PR #3, opened 2016-05-06
Added:  action mode displays count of selected items on History (Google guideline)

EladKeyshawn:


niedzielski:
Replace by patch in [Gerrit](https://gerrit.wikimedia.org/r/#/c/287210/).

PR #9, opened 2017-03-29
i18n: Update "Save" to "Publish" for page edits

jdforrester:
Bug: T142609


jdforrester:
I don't have an Android build system so I'm flying blind here. :-)

jdforrester:
Eurgh, development's in gerrit.

PR #10, opened 2017-05-11
change 349862  : Hide the keyboard as long as the user did not presse…

krimokrimo007:
…d on the edit text of langages preferences


LunarTwilight:
Hi! This GutHub repo is a mirror from https://gerrit.wikimedia.org, so you might want to read https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/Android/App_hacking.

PR #16, opened 2017-12-10
Fix Adaptive Icon

brianroadifer:
updated the launcher icon so that it is more compatible for Android Oreo, applied same fix to development channels icons. Legacy icons are still in place and have been updated as well using Android Studio Image asset maker.


reetssydney:
hi @brianroadifer - Thanks for patch but this Github repo is a mirror only, all our pull requests actually go through Gerrit. Please see our [contribution guide](https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/Android/App_hacking) for tips on getting started! Note the ticket for fixing Adaptive icons is on our workboard at <https://phabricator.wikimedia.org/T174301> for linking to in the Gerrit patch if you choose to continue.

dbrant:
This patch has been migrated to Gerrit:
https://gerrit.wikimedia.org/r/399198

PR #21, opened 2018-05-22
Updated de (German) translation

m-rey:


dbrant:
Thanks; if you'd like to contribute to the app, please see our [getting started guide](https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/Android/App_hacking). Note that the strings inside the `strings_no_translate.xml` files do not need to be translated.

PR #31, opened 2019-02-19
Fix: Theme change Flicker

sharvaniharan:
- A flicker appears everytime there is a theme change. This is because the theme color background is not applied in time. Changing only the background color while loading the page leaves out the header and the header retains the color of the previous theme and causes a flicker. The theme class is getting added too late.
- Removed js/theme.js and made those changes ahead of time at the CommunicationBridge.java
-A temp fix was put in during the light mode fix here : https://gerrit.wikimedia.org/r/c/mediawiki/services/mobileapps/+/487908. This is a more root level fix.

Change-Id: Idda135b0596097649cdcf59b817af404b22011a2


app/src/main/java/org/wikipedia/bridge/CommunicationBridge.java
```diff
@@ -68,11 +67,11 @@ public CommunicationBridge(final WebView webView) {
     public void resetHtml(@NonNull Context context, @NonNull String assetFileName, @NonNull String wikiUrl) {
         String html = "";
         try {
-            final int rgbMask = 0xFFFFFF;
             html = FileUtil.readFile(WikipediaApp.getInstance().getAssets().open(assetFileName))
                     .replace("$wikiurl", wikiUrl)
-                    .replace("$bodybackground", String.format("#%06X",
-                            ResourceUtil.getThemedColor(context, R.attr.paper_color) & rgbMask));
+                    .replace("$pageLibThemeClass", Theme.getThemePageLibClass(WikipediaApp.getInstance().getCurrentTheme()))
+                    .replace("$pageLibDimImgClass", (WikipediaApp.getInstance().getCurrentTheme().isDark()) ? "pagelib_dim_images" : "");
```
cooltey: parentheses can be removed `(WikipediaApp.getInstance().getCurrentTheme().isDark())`

app/src/main/java/org/wikipedia/theme/Theme.java
```diff
@@ -68,4 +68,20 @@ public boolean isDark() {
         this.resourceId = resourceId;
         this.nameId = nameId;
     }
+
+    public static String getThemePageLibClass(Theme theme) {
+        switch (theme.getFunnelName()) {
+            case "light":
```
cooltey: Maybe it is okay to remove the `case "light"` since it is a default case?

app/src/main/java/org/wikipedia/bridge/CommunicationBridge.java
sharvaniharan: Done

app/src/main/java/org/wikipedia/theme/Theme.java
sharvaniharan: Done

PR #32, opened 2019-02-19
Split off network/model library.

dbrant:
This officially splits the app off from our new library which will now contain all model classes and most of the networking client code.

https://phabricator.wikimedia.org/T215949


app/build.gradle
```diff
@@ -217,6 +217,7 @@ dependencies {
     implementation "com.android.support:palette-v7:$supportVersion"
     implementation "com.android.support:preference-v14:$supportVersion"
 
+    implementation "com.dmitrybrant:wikimedia-android-data-client:0.0.13"
```
cooltey: Are we going to apply an official Wikimedia Bintray account in the future?

sharvaniharan:
Looks good... it is in blocked so leaving it open until we discuss.

PR #50, opened 2019-02-21
Remove obsolete scripts and update current ones.

dbrant:


dbrant:
@cooltey Nope! I've removed it. We can bring it back if we decide to run instrumented tests on CI.

PR #51, opened 2019-02-21
Get correct WikiSite() when long pressing on links

cooltey:
If an article contains links with different domain names, the app will show an error page when long pressing to open the link.

https://phabricator.wikimedia.org/T214950

### DIFF
--- a/app/src/main/java/org/wikipedia/LongPressHandler.java
+++ b/app/src/main/java/org/wikipedia/LongPressHandler.java
@@ -15,6 +15,7 @@ import org.wikipedia.dataclient.WikiSite;
 import org.wikipedia.history.HistoryEntry;
 import org.wikipedia.page.PageTitle;
 import org.wikipedia.readinglist.AddToReadingListDialog;
+import org.wikipedia.util.log.L;
 
 import static org.wikipedia.util.DeviceUtil.hideSoftKeyboard;
 import static org.wikipedia.util.UriUtil.isValidPageLink;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/LongPressHandler.java
```diff
@@ -50,8 +51,13 @@ public void onCreateContextMenu(ContextMenu menu, View view,
             if (result.getType() == WebView.HitTestResult.SRC_ANCHOR_TYPE) {
                 Uri uri = Uri.parse(result.getExtra());
                 if (isValidPageLink(uri)) {
-                    title = ((WebViewContextMenuListener) contextMenuListener).getWikiSite()
-                            .titleForInternalLink(uri.getPath());
+                    WikiSite wikiSite = new WikiSite(uri);
+                    // the following logic keeps the correct language code if the domain has multiple variants (e.g. zh).
+                    if (wikiSite.dbName().equals(((WebViewContextMenuListener) contextMenuListener).getWikiSite().dbName())
```
dbrant: Doesn't WikiSite already have this logic (in the constructor)? It's not clear why it's necessary here.

cooltey: If in a `zh-hant` article has a link `https://zh.wikipedia.org/wiki/Taiwan`, the default language should be `zh-hant`.

The logic in the WikiSite constructor only get the `zh` as the language code when get the link above (because the link does not contain a subdomain?), which will let the article having mixed variants.

PR #52, opened 2019-02-22
Migrate dialog styles to MaterialComponents theme.

dbrant:
This moves our Dialog styles to use the MaterialComponents theme instead of the older AppCompat. No visual differences are expected. This also sets an explicit background color on the search language selection bar, since this is now required.


sharvaniharan:
Awesome!

PR #57, opened 2019-02-27
Dialog does not show after backing to the feed

cooltey:
It seems to be happened on serval devices (including real devices and emulator with certain profile settings)

Steps:
1. Fresh install
2. Read an article and add it to a reading list
3. press “back button” and back to the feed

https://phabricator.wikimedia.org/T217303


cooltey:
Since it has only happened on my device, I discovered that the recently changed `Theme.MaterialComponents.Light.Dialog.Alert` makes the issue.

cooltey:
> Since both light and dark modes now have the same parent, we should now move this to main styles with just a common style for alertDialogTheme.

Done. Please see changes in 96a5e3a.

PR #64, opened 2019-03-05
Tweaks: enhancing reading lists search

cooltey:
Show lists + list items in same list based on search term
https://phabricator.wikimedia.org/T214505

Contextual actions for list item
https://phabricator.wikimedia.org/T214506


cooltey:
CC: @sharvaniharan
Using the same method of updating the thumbnail from 3fb7ef1

PR #65, opened 2019-03-05
Design review  - T214504

sharvaniharan:
**Updated review comments**

- Translated colors to the other themes according to the Android Theme guidelines.
- Adjusted thumbnail width to 56x56px
- Clipped scrolling to the start of title & description.
- Blue shadow on the left of scrollview removed
- Search within a list (e.g. Saved) works exactly the same as the global “My lists“ search. Shows Chips.

Phab : https://phabricator.wikimedia.org/T214504

### DIFF
--- a/app/src/main/java/org/wikipedia/readinglist/ReadingListFragment.java
+++ b/app/src/main/java/org/wikipedia/readinglist/ReadingListFragment.java
@@ -615,7 +615,6 @@ public class ReadingListFragment extends Fragment implements
                 getView().setTitleEllipsis();
                 getView().setDescriptionMaxLines(2);
                 getView().setDescriptionEllipsis();
-                getView().setListItemImageDimensions(DimenUtil.roundedDpToPx(ARTICLE_ITEM_IMAGE_DIMENSION), DimenUtil.roundedDpToPx(ARTICLE_ITEM_IMAGE_DIMENSION));
                 getView().setUpChipGroup(ReadingListBehaviorsUtil.INSTANCE.getListsContainPage(page));
             } else {
                 getView().hideChipGroup();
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/readinglist/ReadingListFragment.java
```diff
@@ -599,6 +600,18 @@ void bindItem(ReadingListPage page) {
             getView().setSecondaryActionHint(R.string.reading_list_article_make_offline);
             getView().setSearchQuery(currentSearchQuery);
             PageAvailableOfflineHandler.INSTANCE.check(page, available -> getView().setViewsGreyedOut(!available));
+
+            if (!TextUtils.isEmpty(currentSearchQuery)) {
+                getView().setTitleMaxLines(2);
+                getView().setTitleEllipsis();
+                getView().setDescriptionMaxLines(2);
+                getView().setDescriptionEllipsis();
+                getView().setListItemImageDimensions(DimenUtil.roundedDpToPx(ARTICLE_ITEM_IMAGE_DIMENSION), DimenUtil.roundedDpToPx(ARTICLE_ITEM_IMAGE_DIMENSION));
```
cooltey: The thumbnail size should be set to the default size if out of the actionMode. 
Please see code below: (or you can rebase it)
https://github.com/wikimedia/apps-android-wikipedia/blob/cc25e3fa4e4382983ee0dbef3011145eedaab8ca/app/src/main/java/org/wikipedia/readinglist/ReadingListFragment.java#L610-L628

sharvaniharan: Good catch! thank you..done

PR #70, opened 2019-03-07
Visual changes to translate descriptions

sharvaniharan:
**Translate descriptions**
   -Replaced bidirectional with unidirectional icon.
   -Added clear call to action text (“Add German translation“) indicating users to add translation.
   -Added " Translated by you" after user translates.
   
Phab: https://phabricator.wikimedia.org/T217170


app/src/main/java/org/wikipedia/editactionfeed/AddTitleDescriptionsItemFragment.kt
```diff
@@ -132,10 +139,13 @@ class AddTitleDescriptionsItemFragment : Fragment() {
...
 
         if (parent().source == InvokeSource.EDIT_FEED_TRANSLATE_TITLE_DESC) {
-            val spannableDescription = SpannableString(sourceDescription)
-            spannableDescription.setSpan(BackgroundColorSpan(ResourceUtil.getThemedColor(requireContext(), R.attr.text_highlight_color)), 0, sourceDescription.length, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
-            viewArticleSubtitle.text = TextUtils.concat(String.format(getString(R.string.translation_source_description), app.language().getAppLanguageCanonicalName(parent().langFromCode)), spannableDescription)
+            viewArticleSubtitleContainer.visibility = VISIBLE
+            viewArticleSubtitleAddedBy.visibility = GONE
+            viewArticleSubtitleEdit.visibility = GONE
+            viewArticleSubtitle.text = sourceDescription
+            callToActionText.text = String.format(getString(R.string.add_translation), app.language().getAppLanguageLocalizedName(parent().langToCode))
```
cooltey: Should it use `getAppLanguageCanonicalName` instead?

app/src/main/res/values/strings.xml
```diff
@@ -643,6 +644,7 @@
     <string name="editactionfeed_translate_title_dialog_learn_more">Learn more</string>
     <string name="editactionfeed_to_language">To: </string>
     <string name="editactionfeed_added_by_you">Added by you:</string>
+    <string name="editactionfeed_translated_by_you">Translated by you:</string>
```
cooltey: The title of the card screen should be changed to `Translate descriptions` (remove the **title**)

app/src/main/java/org/wikipedia/editactionfeed/AddTitleDescriptionsItemFragment.kt
sharvaniharan: Done. Good catch!

app/src/main/res/values/strings.xml
sharvaniharan: I have changed both the title string and DescriptionEditActivity title both to 'Translate descriptions' . I will change the DescriptionEditActivity title to 'Translate description' if Robin feels it has to be singular.

app/src/main/res/layout/fragment_add_title_descriptions.xml
```diff
@@ -50,16 +50,16 @@
 
 
             <ImageView
-                android:id="@+id/arrows"
-                android:layout_width="40dp"
-                android:layout_height="match_parent"
+                android:id="@+id/arrow"
+                android:layout_width="16dp"
+                android:layout_height="8dp"
```
cooltey: Would it be better if it has a bigger clickable area? (by adding paddings?)

sharvaniharan: Good catch! Done!

sharvaniharan:
> Noticed that in the `DescriptionEdit` screen, the text `article` in the page summary field should be changed to the "`language` + description", for example `English description`.

Done!

PR #71, opened 2019-03-07
Tie together Suggested Edit screens with new backend API.

dbrant:


app/src/main/java/org/wikipedia/notifications/NotificationEditorTasksHandler.java
```diff
@@ -0,0 +1,91 @@
...
+                    R.drawable.ic_mode_edit_white_24dp, R.color.accent50, intent);
+        }
+    }
+
+    private static void maybeShowTranslateDescriptionUnlockNotification(@NonNull Context context) {
+        if (!WikipediaApp.getInstance().isAnyActivityResumed()) {
+            Intent intent = AddTitleDescriptionsActivity.Companion.newIntent(context, Constants.InvokeSource.EDIT_FEED_TRANSLATE_TITLE_DESC);
+            NotificationCompat.Builder builder = NotificationPresenter.getDefaultBuilder(context);
+            PendingIntent pendingIntent = PendingIntent.getActivity(context, 1, intent, PendingIntent.FLAG_UPDATE_CURRENT);
+            builder.addAction(0, context.getString(R.string.title_description_get_started), pendingIntent);
```
cooltey: Noticed that the `builder` should be sent into the `NotificationPresenter.showNotification()`.

dbrant: Whoops! You are correct.

cooltey:
Have noticed that in the suggested edits tasks screen, if users switch to other app and back, it will not be able to go back to MainActivity when clicking on the back button.

dbrant:
@cooltey You're correct; I changed those activities to be `singleInstance`, when they should really be `singleTask`.

app/src/main/java/org/wikipedia/notifications/NotificationEditorTasksHandler.java
```diff
@@ -0,0 +1,91 @@
...
+import static org.wikipedia.Constants.MIN_LANGUAGES_TO_UNLOCK_TRANSLATION;
+
+public final class NotificationEditorTasksHandler {
+
+    public static void dispatchEditorTaskResults(@NonNull Context context, @NonNull EditorTaskCounts results) {
+        int numTargetsPassed = results.getDescriptionEditTargetsPassed().size();
+        Object eventToDispatch = null;
+
+        if (numTargetsPassed > 1) {
+            if (!Prefs.isEditActionAddDescriptionsUnlocked()) {
```
sharvaniharan: When would this happen?

dbrant: After 50 edits (and after 48 hours of no reverts).

dbrant: Oh, if you meant to say "Why would adding descriptions be locked if the user has already passed 1 target?" the answer is "On a fresh install of the app."

sharvaniharan: I'm confused isn't it 'numTargetsPassed=1 is add descs unlocked' after 5 edits 48hr wait  and then 50 edits 48 hrs 'numTargetsPassed=2 add trans unlocked'?
isn't it, if 2 targets are unlocked then AddDescriptions is definitely unlocked long time back? and even if the user has lost previleges due to reverts... AddDescriptions would first be unlocked before numTargetsPassed > 1 ever happens?

sharvaniharan: > Oh, if you meant to say "Why would adding descriptions be locked if the user has already passed 1 target?" the answer is "On a fresh install of the app."

Ahhh! ding! ding! ding! thank you.. so much logic all organized so neatly with this branch ! merging it now!

PR #76, opened 2019-03-08
Design tweak of enhancing reading lists search function

cooltey:
Update the thumbnail color
https://phabricator.wikimedia.org/T214505#5010784

Enable the `X` button
https://phabricator.wikimedia.org/T214505#5010855

Adjust arrow icon position and hintText color in SearchFragment 
https://phabricator.wikimedia.org/T214505#5011757

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -362,7 +362,7 @@
         <item name="android:contentDescription">@null</item>
         <item name="placeholderImage">?attr/material_theme_border_color</item>
         <item name="failureImage">?attr/material_theme_border_color</item>
-        <item name="backgroundImage">?attr/material_theme_border_color</item>
+        <item name="backgroundImage">?attr/paper_color</item>
         <item name="roundedCornerRadius">2dp</item>
         <item name="roundTopLeft">false</item>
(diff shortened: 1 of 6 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -362,7 +362,7 @@
         <item name="android:contentDescription">@null</item>
         <item name="placeholderImage">?attr/material_theme_border_color</item>
         <item name="failureImage">?attr/material_theme_border_color</item>
-        <item name="backgroundImage">?attr/material_theme_border_color</item>
+        <item name="backgroundImage">?attr/paper_color</item>
```
dbrant: It doesn't seem like `paper_color` is the right choice here. Robin seems to be asking for a variation of `paper_color` that is "shaded" with an opacity. He will need to invent a new meta-color for this purpose.

dbrant: OK! This fixes the color of placeholder thumbnails in the composite (2x2) thumbnail, but it also breaks the color of thumbnails of articles that don't have an image. For example:
device-2019-03-12-161225

cooltey: done!

PR #79, opened 2019-03-11
Fix: set proper margin on tablets in landscape mode

sharvaniharan:
**Root cause and fix**

When we load the page, we set a custom horizontal margin on the page, customized on screen size.
Removed all margin setting code, since this is handled by the css upstream.

Before: https://drive.google.com/open?id=1snAS2L0LkknD2LJsTRgF36M1__RsRbuB
Fixed: https://drive.google.com/open?id=1bmvnwfeBD-jxbZb93NvRAXsTqoBDo0KY

Phab: https://phabricator.wikimedia.org/T216856

### DIFF
--- a/app/src/main/assets/bundle.js
+++ b/app/src/main/assets/bundle.js
@@ -230,6 +230,10 @@ bridge.registerListener( "clearContents", function() {
     clearContents();
 });
 
+bridge.registerListener( "setMargins", function( payload ) {
+    document.getElementById( "content" ).style.marginTop = payload.marginTop + "px";
+});
+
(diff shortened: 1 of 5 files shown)


dbrant:
Unfortunately this seems to break the top margin on non-tablet devices.

device-2019-03-12-142610

The solution might be to just keep setting the Top margin as before, but leave the Left and Right margins alone.

sharvaniharan:
> Unfortunately this seems to break the top margin on non-tablet devices.

Does it look different on current master? It is having the same margin for me on master too ..

PR #87, opened 2019-03-13
Help page update

sharvaniharan:
**Help page updated to external page link**

Url: https://www.mediawiki.org/wiki/Wikimedia_Apps/Suggested_edits

Phab: https://phabricator.wikimedia.org/T217562

Todo: remove unused strings


app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.java
```diff
@@ -308,7 +310,8 @@ private void editFailed(@NonNull Throwable caught, boolean logError) {
 
         @Override
         public void onHelpClick() {
-            startActivity(DescriptionEditHelpActivity.newIntent(requireContext()));
+            UriUtil.visitInExternalBrowser(getContext(),
```
dbrant: Hint: `FeedbackUtil.java` contains numerous convenience functions for visiting specific external pages, e.g. the FAQ page, the privacy policy, etc.  We can simply add another function there that goes to this page.

sharvaniharan: Right! That will be cleaner. Will do this!

app/src/main/res/values/strings.xml
```diff
@@ -777,4 +762,5 @@
     <string name="translate_descriptions_maybe_later">Maybe later</string>
     <string name="title_descriptions_onboarding_got_it">Got it</string>
     <string name="translate_descriptions_onboarding_got_it">Got it</string>
+    <string name="android_app_edit_help_url">https://www.mediawiki.org/wiki/Wikimedia_Apps/Suggested_edits</string>
```
dbrant: As I mentioned in Slack, let's make this point to the mobile version of the page. (`m.` instead of `www.`)

sharvaniharan: I left it like that because it was automatically opening the mobile-compatible page with m. . However, just in case it is only chrome that does that, I have changed to m. .

PR #90, opened 2019-03-14
Fix: next article text is visible on randomizer

cooltey:
Happens on Randomizer if the device is on the landscape mode 
https://phabricator.wikimedia.org/T217444


sharvaniharan:
The next card visible at the bottom, is touching the parent bottom, should there be a slight gap 2-4dp? may be a question for Robin?

sharvaniharan:
Also, sometimes the padding cuts the text : https://drive.google.com/open?id=121Ou3DwnhPiOqIBB28r3y8bkh6nsa_tN

cooltey:
>The next card visible at the bottom, is touching the parent bottom, should there be a slight gap 2-4dp? may be a question for Robin?

Maybe I can adjust the view to nearly full height when it's in the landscape mode, and we can see if it works for Robin.

> Also, sometimes the padding cuts the text : https://drive.google.com/open?id=121Ou3DwnhPiOqIBB28r3y8bkh6nsa_tN

The cutting seems to be a tricky one since each card should have a bottom padding, will try to work on it.

sharvaniharan:
> Maybe I can adjust the view to nearly full height when it's in the landscape mode, and we can see if it works for Robin.

I think that should be fine @cooltey

sharvaniharan:
> The cutting seems to be a tricky one since each card should have a bottom padding, will try to work on it.

agree. guess let us ignore for now.

cooltey:
> I think that should be fine @cooltey 

@sharvaniharan Done!

PR #105, opened 2019-03-22
Add custom shadow to ToC scroller for API 19 devices.

cooltey:
The ToC scroller can be barely seen if it does not have a shadow.
It might be good to add a custom "shadow" to the ToC scroller before we officially move up to 21.

Screenshot_1553215289
Screenshot_1553215279


dbrant:
Sure, looks good! Let's just remember to revert this when we drop API 19.

PR #115, opened 2019-03-25
Tweak of visual changes to add description

cooltey:
https://phabricator.wikimedia.org/T217169#5053622
https://phabricator.wikimedia.org/T217173#5052973


cooltey:
> So the comment #10 is no longer true because of the changing heights I guess.
> https://phabricator.wikimedia.org/T217169#5053622
> "10. Space from card to dice should be 48dp (it’s currently ~ 62dp)"

It is still true if the card reaches the maximum height, and it will still need a bottom 48dp margin.

PR #118, opened 2019-03-27
Fix standard ActionMode when editing.

dbrant:
For some reason we were overriding the default ActionMode in the editing activity. This is incorrect, since the overridden actions ("edit here", "define", etc) don't make sense in the editing context.

https://phabricator.wikimedia.org/T219393

### DIFF
--- a/app/src/main/java/org/wikipedia/edit/EditSectionActivity.java
+++ b/app/src/main/java/org/wikipedia/edit/EditSectionActivity.java
@@ -54,7 +54,6 @@ import org.wikipedia.util.StringUtil;
 import org.wikipedia.util.log.L;
 import org.wikipedia.views.PlainPasteEditText;
 import org.wikipedia.views.ViewAnimations;
-import org.wikipedia.views.ViewUtil;
 import org.wikipedia.views.WikiErrorView;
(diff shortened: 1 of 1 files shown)


sharvaniharan:
overflow shows 'Search Wikipedia' twice..
overflow

dbrant:
@sharvaniharan  This is quite expected, since you have the Dev, Alpha, Prod, etc. apps installed on your device. This is the default ActionMode that you should see in all EditTexts on your device.

cooltey:
The reason why I added that was for the devices using API 19. 

Noticed that we don't actually need to inflate another menu resource: 
`mode.getMenuInflater().inflate(R.menu.menu_text_select, menu);`

Screenshot_1553716488

sharvaniharan:
> @sharvaniharan This is quite expected, since you have the Dev, Alpha, Prod, etc. apps installed on your device. This is the default ActionMode that you should see in all EditTexts on your device.

That's right! in that case it checks out! @cooltey feel free to merge if everything looks good to you on 19.

dbrant:
@cooltey  Ahh thanks, I've put back the bits needed for API 19.

PR #123, opened 2019-03-27
T219358

NoamDev:
I have changed NetworkSecurityConfiguration in the manifest to trust user certificates,
because starting in android nougat they are not trusted by default,
which makes the app impossible to use when using a network filterin service (like Netspark).
more details can be found in Phabricator ticket [T219358](https://phabricator.wikimedia.org/T219358).


NoamDev:
@dbrant do you have any idea why the check have failed?

NoamDev:
I think it happened because it's a fork merge.
should continue contributing through my fork? because I have no permission to create a new branch in this repo.

NoamDev:
yes, as noted [here](https://docs.travis-ci.com/user/pull-requests#pull-requests-and-security-restrictions),
travis-ci can't use encrypted files on forks-pull requests. from security reasons.

dbrant:
@NoamDev I don't think I'm aware of any other major apps that are currently including this attribute in their manifest (e.g. Twitter, Facebook, Instagram, Firefox, etc).  Do these apps work on your device? If so... how/why?

dbrant:
As I understand it, by allowing user-added certificates, you're leaving the door wide open to MITM attacks. In the context of Wikipedia, this would mean that "Netspark" would be able to serve you content that isn't genuine, and might not be coming from the real Wikipedia. We'd like to avoid this if possible.

NoamDev:
Firefox doesn't honor certificates that were installed by the user but allows you to import pem version of the certificate to the browser. chrome does. I have inspected chrome's apk and if you look in AndroidManifest.xml and xml/network-security-config.xml you'll see that did exactly what I suggest (you can assure that by using ann app like apk parse, or extracting the apk on your pc).

You're right about the MIDM attacks but the user has to install the crt file. If he installed it, it means he trust netspark.

My main point is that if you won't allow the user to use wikipedia app, he'll probably use chrome which will trust the user installed certificates, and then all you have earned is discomfort.

By the way, note that before android nougat, every app trusted the user installed certificates.

PR #124, opened 2019-03-27
Add RTL support for views

cooltey:
Some views do not respect the RTL language. It can be simple to fix by adding the `setConditionalLayoutDirection`  method.


sharvaniharan:
lgtm... Will let Dmitry also take a look once since it is an rtl change.. @dbrant please do the honors ;)

PR #128, opened 2019-03-28
Add confirmation dialog when logging out.

dbrant:
The "log out" button in our Navigation menu is pretty easy to tap, and is rather abrupt in logging out the user. It would be nice to have a confirmation message that asks if you actually want to log out.

### DIFF
--- a/app/src/main/java/org/wikipedia/main/MainActivity.java
+++ b/app/src/main/java/org/wikipedia/main/MainActivity.java
@@ -10,6 +10,7 @@ import android.support.annotation.Nullable;
 import android.support.v4.view.GravityCompat;
 import android.support.v4.widget.DrawerLayout;
 import android.support.v7.app.ActionBarDrawerToggle;
+import android.support.v7.app.AlertDialog;
 import android.support.v7.view.ActionMode;
 import android.support.v7.widget.Toolbar;
 import android.view.Gravity;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/main/MainActivity.java
```diff
@@ -250,18 +251,23 @@ protected void clearToolbarElevation() {
...
+                        if (AccountUtil.isLoggedIn()) {
+                            WikipediaApp.getInstance().logOut();
+                            FeedbackUtil.showMessage(MainActivity.this, R.string.toast_logout_complete);
+                            if (Prefs.isReadingListSyncEnabled() && !ReadingListDbHelper.instance().isEmpty()) {
+                                ReadingListSyncBehaviorDialogs.removeExistingListsOnLogoutDialog(MainActivity.this);
+                            }
+                            Prefs.setReadingListsLastSyncTime(null);
+                            Prefs.setReadingListSyncEnabled(false);
+                        } else {
+                            getFragment().onLoginRequested();
```
cooltey: The `AccountUtil.isLoggedIn()` should be moved out as the main logic of behavior, otherwise, users will not be able to log in.

dbrant: 🤦‍♀️ doh, that is correct!

PR #132, opened 2019-03-30
Using api for missing description and translation articles

sharvaniharan:
Using api for missing description and translation articles


sharvaniharan:
We can do this after the api is stabilized.

PR #134, opened 2019-04-01
Wire up analytics for Suggested Edits.

dbrant:
https://phabricator.wikimedia.org/T218594


app/src/main/java/org/wikipedia/descriptions/DescriptionEditActivity.java
```diff
@@ -93,13 +95,20 @@ private void copyLink(@NonNull String url) {
 
     @Override
     public DescriptionEditFragment createFragment() {
+        invokeSource = (InvokeSource) getIntent().getSerializableExtra(INTENT_EXTRA_INVOKE_SOURCE);
+        if (invokeSource == InvokeSource.EDIT_FEED_TITLE_DESC) {
+            SuggestedEditsFunnel.get().getAddDescriptionStats().click();
+        } else if (invokeSource == InvokeSource.EDIT_FEED_TRANSLATE_TITLE_DESC) {
+            SuggestedEditsFunnel.get().getTranslateDescriptionStats().click();
```
cooltey: Would it be good to send `InvokeSource` into the `SuggestedEditsFunnel` class and let the method do the logic and action? It might be easier to manage if we add more `InvokeSources` in the future. (e.g. `EDIT_FEED_IMAGE_CAPTION`)

app/src/main/java/org/wikipedia/analytics/SuggestedEditsFunnel.java
```diff
@@ -2,35 +2,112 @@
...
     }
 
-    @Override
-    protected JSONObject preprocessData(@NonNull JSONObject eventData) {
-        preprocessData(eventData, "ui_version", SUGGESTED_EDITS_UI_VERSION);
-        preprocessData(eventData, "api_version", SUGGESTED_EDITS_API_VERSION);
-        return super.preprocessData(eventData);
+    public void contributionsOpened() {
+        contributionsOpenedCount++;
     }
```
sharvaniharan: May be we can use this now that 'My Contributions' is ready..

app/src/main/java/org/wikipedia/descriptions/DescriptionEditActivity.java
sharvaniharan: +1

dbrant: Correct!

app/src/main/java/org/wikipedia/analytics/SuggestedEditsFunnel.java
dbrant: Done

PR #135, opened 2019-04-01
Get source article for translate screen.

sharvaniharan:
**Get source article for translate screen.**
- Changed the MissingDescriptionProvider getNextArticleWithMissingDescription to return the source page summary and target pagetitle , so we can launch the DescriptionEditActivity successfully.

Phab:https://phabricator.wikimedia.org/T217170
Zeplin: https://app.zeplin.io/project/57a120b91998d8977642a238/dashboard?seid=5c77d1b7c8339abca24b4a1a


cooltey:
@sharvaniharan Does this PR relate to any specific ticket?

sharvaniharan:
Oh yes... will link it

sharvaniharan:
Done

cooltey:
Just noticed that in the following screen:
Screenshot_1554243351

Should it be showing [From Language] article content with the text label "Add [Target Language] description"? I am kind of confusing right now.
https://app.zeplin.io/project/57a120b91998d8977642a238/screen/5c77d20a86d8b86292fd9750

sharvaniharan:
Yes @cooltey [From Language] content. 😂 Is the screenshot from the newest patch of this branch ? because I am getting this:
rtl

sharvaniharan:
This is correct, but only the direction here should be of the [From Lang]. Will change that..

sharvaniharan:
I guess the confusion is because Robin has not switched the languages in the zeplin design, he has moved the direction arrow instead ..
https://app.zeplin.io/project/57a120b91998d8977642a238/screen/5c8689adec0dca37faac5ec8

cooltey:
> I guess the confusion is because Robin has not switched the languages in the zeplin design, he has moved the direction arrow instead ..
> https://app.zeplin.io/project/57a120b91998d8977642a238/screen/5c8689adec0dca37faac5ec8

For the Zeplin design, the entire app layout is in RTL because the system language is an RTL language.

sharvaniharan:
> For the Zeplin design, the entire app layout is in RTL because the system language is an RTL language, and that's why the arrow switches.

Ah! right..

app/src/main/java/org/wikipedia/editactionfeed/provider/MissingDescriptionProvider.kt
```diff
@@ -62,8 +62,8 @@ object MissingDescriptionProvider {
                     ServiceFactory.get(WikiSite(Service.WIKIDATA_URL))
                             .getWikidataLabelsAndDescriptions(StringUtils.join(qNumbers, '|'))
                 }
-                .map<Pair<String, PageTitle>> { response ->
-                    var sourceDescriptionAndTargetTitle: Pair<String, PageTitle>? = null
+                .map<Pair<PageTitle, PageTitle>> { response ->
+                    var sourceDescriptionAndTargetTitle: Pair<PageTitle, PageTitle>? = null
```
cooltey: It would be better if also changes the name from `sourceDescriptionAndTargetTitle` to `sourceAndTargetPageTitles`

app/src/main/java/org/wikipedia/editactionfeed/provider/MissingDescriptionProvider.kt
```diff
@@ -97,8 +97,8 @@ object MissingDescriptionProvider {
...
```
cooltey: same as the comment above.

app/src/main/java/org/wikipedia/editactionfeed/provider/MissingDescriptionProvider.kt
cooltey: NVM, I've made a follow-up commit f72fc98 in PR #142

PR #139, opened 2019-04-02
Fix incorrect null check when clicking on description card view

sharvaniharan:
Change to summary so Title Descriptions click doesn't get locked.


dbrant:
Quick nitpick: Can we please have more descriptive titles for these pull requests? 
"Fix for title Descriptions" is nearly devoid of meaning, whereas your summary ("Change to summary so Title Descriptions click doesn't get locked.") is much better, and can easily just be the title.  And even better might be something like "Fix incorrect null check when clicking on description card view".

This will make things much easier for future developers who browse our commit history.

sharvaniharan:
👍

PR #141, opened 2019-04-02
Tweak the logic of login/logout for SuggestedEdits

cooltey:
If a current user has an account that has unlocked the `SuggestedEdits`, and then the user logs out to another account that has not unlocked the `SuggestedEdits`, the preferences and the item in drawer menu should be changed right after the user logged in.

Also, I've renamed some preferences and methods to `SuggestedEdits`

### DIFF
--- a/app/src/main/java/org/wikipedia/notifications/NotificationEditorTasksHandler.java
+++ b/app/src/main/java/org/wikipedia/notifications/NotificationEditorTasksHandler.java
@@ -78,7 +78,7 @@ public final class NotificationEditorTasksHandler {
     }
 
     public static void maybeShowTranslateDescriptionUnlockNotification(@NonNull Context context, boolean forced) {
-        if (!WikipediaApp.getInstance().isAnyActivityResumed() && !Prefs.isSuggestedEditsTranslateDescriptionsMessageShown()  || forced) {
+        if (!WikipediaApp.getInstance().isAnyActivityResumed() && !Prefs.isSuggestedEditsTranslateDescriptionsMessageShown() || forced) {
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/wikipedia/notifications/NotificationEditorTasksHandler.java
```diff
@@ -62,7 +63,8 @@ public static void dispatchEditorTaskResults(@NonNull Context context, @NonNull
     }
 
     public static void maybeShowEditDescriptionUnlockNotification(@NonNull Context context, boolean forced) {
-        if (!WikipediaApp.getInstance().isAnyActivityResumed() || forced) {
+        if (!WikipediaApp.getInstance().isAnyActivityResumed() && !Prefs.isSuggestedEditsAddDescriptionsMessageShown() || forced) {
+            Prefs.setSuggestedEditsAddDescriptionsMessageShown(true);
```
dbrant: I don't think I understand the purpose of this new preference. Do we need it? It looks like it prevents the notification (or dialog) from being shown more than once.  But what if the user logs out, and then another user logs in, not knowing about these notifications? Or is my understanding wrong?

cooltey: The idea is to prevent that if the **same user** logs out and logs in again, it should not see the dialog/notification again, and it is true that other user will not see the dialog/notification after passed the targets.

cooltey: Removed!

PR #147, opened 2019-04-03
Enable caching of suggested edit titles.

dbrant:
This is a huge improvement in the efficiency of our API usage, since this will now cache the PageTitles to be used in subsequent requests to get suggested edits.


app/src/main/java/org/wikipedia/editactionfeed/provider/MissingDescriptionProvider.kt
```diff
@@ -13,27 +13,53 @@ import org.wikipedia.dataclient.restbase.page.RbPageSummary
...
 import org.wikipedia.wikidata.Entities
 import java.util.*
+import kotlin.collections.ArrayList
 
 object MissingDescriptionProvider {
+    val articlesWithMissingDescriptionCache : Stack<String> = Stack()
+    var articlesWithMissingDescriptionCacheLang : String = ""
+    val articlesWithTranslatableDescriptionCache : Stack<Pair<PageTitle, PageTitle>> = Stack()
+    var articlesWithTranslatableDescriptionCacheFromLang : String = ""
+    var articlesWithTranslatableDescriptionCacheToLang : String = ""
```
cooltey: The `var` and `val` should be private

app/src/main/java/org/wikipedia/editactionfeed/provider/MissingDescriptionProvider.kt
```diff
@@ -13,27 +13,53 @@ import org.wikipedia.dataclient.restbase.page.RbPageSummary
...
+            if (articlesWithMissingDescriptionCacheLang != wiki.languageCode()) {
+                // evict the cache if the language has changed.
+                articlesWithMissingDescriptionCache.clear()
+            }
+            if (articlesWithMissingDescriptionCache.size > 0) {
+                cachedTitle = articlesWithMissingDescriptionCache.pop()
+            }
+        }
+
+        return (if (!TextUtils.isEmpty(cachedTitle)) Observable.just(cachedTitle) else
```
cooltey: I am a little bit confusing that the should `cachedTitle`be a type of `RbPageSummary`?

app/src/main/java/org/wikipedia/editactionfeed/provider/MissingDescriptionProvider.kt
```diff
@@ -49,39 +75,57 @@ object MissingDescriptionProvider {
     @Deprecated("Remove when the new API is deployed to production.")
     fun getNextArticleWithMissingDescription(sourceWiki: WikiSite, targetLang: String, sourceLangMustExist: Boolean): Observable<Pair<PageTitle, RbPageSummary>> {
         val targetWiki = WikiSite.forLanguageCode(targetLang)
+        var cachedPair: Pair<PageTitle, PageTitle>? = null
```
cooltey: Should the `cachedPair` be the type of `Pair<PageTitle, RbPageSummary>`?

app/src/main/java/org/wikipedia/editactionfeed/provider/MissingDescriptionProvider.kt
dbrant: Done

app/src/main/java/org/wikipedia/editactionfeed/provider/MissingDescriptionProvider.kt
dbrant: Nope, we want to cache the *seed* titles from which the RbPageSummary will be retrieved. Remember that the initial call gets up to fifty (50) random articles, but only uses one of them for getting the `RbPageSummary` (it throws away the other 49).  This optimization keeps the other 49, and reuses them in subsequent calls.

app/src/main/java/org/wikipedia/editactionfeed/provider/MissingDescriptionProvider.kt
dbrant: Similarly to the above, we're caching the seed titles that will be used in the requests to get `RbPageSummary`. In this case the caching is even more important because we're caching the results of two different requests.

app/src/main/java/org/wikipedia/editactionfeed/provider/MissingDescriptionProvider.kt
cooltey: I see. I thought it is a `return` value of the method and didn't notice that is actually a part of the `return` before calling `flatMap`

PR #153, opened 2019-04-04
Loading DescriptionEditView smoothly

sharvaniharan:
Added a delay on the DescriptionEditView's handler before making it visible, to make it load smoothly.


dbrant:
Can you share an example of what this is fixing and/or why it's bothersome?

sharvaniharan:
I am trying to see if we can do anything about point #3 in https://phabricator.wikimedia.org/T217170#5084694

dbrant:
Surely there's a better way to address it than imposing an arbitrary artificial delay.

The undesirable "jump" occurs because the `DescriptionEditFragment` makes an API call to get the page summary (asynchronously).  So, wouldn't it be better to show the `editView` when that API call is complete?

Or better yet, note that we only show the page summary when invoking the `DescriptionEditFragment` from the suggested edits workflow, where we *already have* the page summary loaded in the card. So, couldn't we pass the page summary through a Bundle, so that we wouldn't even need to make an API call in `DescriptionEditFragment`?

sharvaniharan:
Although, not sure if the delay and late load is enough to the eye to make it look smoother

> The undesirable "jump" occurs because the `DescriptionEditFragment` makes an API call to get the page summary (asynchronously). So, wouldn't it be better to show the `editView` when that API call is complete?

I have tried this, the ui is updated only after we have the summary in hand, and I tried having the view be invisible and making it visible only after it has the text init.. still it is jumpy... It is the difference in time between the full activity being loaded and the summary text being updated. 

But I can try sending the summary when we launch the DescriptionEditView so as to avoid the call... which would fix only in this case, but better than nothing..

cooltey:
> But I can try sending the summary when we launch the DescriptionEditView so as to avoid the call... which would fix only in this case, but better than nothing..

If you'd like to avoid the API call in `DescriptionEditFragment`, you'll need to send also the "target" `pageTitle` as well since in `Translate Description` flow, the result from API call is bening used in the bottom "read" bar.

sharvaniharan:
We are currently sending targetPageTitle for translation @cooltey for the same reason..

sharvaniharan:
> > But I can try sending the summary when we launch the DescriptionEditView so as to avoid the call... which would fix only in this case, but better than nothing..
> 
> If you'd like to avoid the API call in `DescriptionEditFragment`, you'll need to send the "target" `pageTitle` as well since in `Translate Description` flow, the result from API call is being used in the bottom "read" bar.

Ah! I get it ..You meant target summary! because we set the title of that bar from summary!

app/src/main/java/org/wikipedia/descriptions/DescriptionEditActivity.java
```diff
@@ -32,21 +34,22 @@
...
     private ExclusiveBottomSheetPresenter bottomSheetPresenter = new ExclusiveBottomSheetPresenter();
 
     public static Intent newIntent(@NonNull Context context,
                                    @NonNull PageTitle title,
+                                   @NonNull String extract,
                                    @Nullable String highlightText,
-                                   boolean reviewEnabled,
                                    @Nullable CharSequence translationSourceDescription,
                                    @Nullable String translationSourceLanguageCode,
                                    @NonNull InvokeSource invokeSource) {
```
sharvaniharan: Had to reduce a parameter to introduce a new one, since we can derive reviewEnabled, chose that.

sharvaniharan:
better solution here : https://github.com/wikimedia/apps-android-wikipedia/pull/157

PR #162, opened 2019-04-05
Fix image dimming in dark mode(s).

dbrant:
The changes made in #31 seems to have removed too much functionality, and broke image dimming. This restores the explicit call into the page-library to dim the images when necessary.


cooltey:
The image dimming seems working correctly to me on the current master branch. Do you have the steps of reproducing it?

dbrant:
@cooltey Even when you turn it off?

cooltey:
> @cooltey Even when you turn it off?
I think something went wrong when the Android Studio building the APK. Now I see it actually fixes the issue.

PR #165, opened 2019-04-05
Rename related items to SuggestedEdits

cooltey:
According to Niklas's words, we can just update the English string names and they will do the rest of the work.

https://phabricator.wikimedia.org/T218806


sharvaniharan:
Shouldn't all classes, Activities and Fragments under org.wikipedia.suggestededits package be renamed to SuggestedEdits* names? something like SuggestedEditTask, SuggestedEditTasksFragment,etc..

cooltey:
> Shouldn't all classes, Activities and Fragments under org.wikipedia.suggestededits package be renamed to SuggestedEdits* names? something like SuggestedEditTask, SuggestedEditTasksFragment,etc..

Yes, I should follow the patterns in our project.

sharvaniharan:
> > Shouldn't all classes, Activities and Fragments under org.wikipedia.suggestededits package be renamed to SuggestedEdits* names? something like SuggestedEditTask, SuggestedEditTasksFragment,etc..
> 
> Yes, I should follow the patterns in our project.

Awesome! I was wondering if we should or if it is overkill 😅

PR #39, opened 2019-02-19
My contributions page and update dialog/notification texts

cooltey:
**My contributions page**
https://phabricator.wikimedia.org/T207339

**Dialog/notification concept**
https://phabricator.wikimedia.org/T217369

PR #167, opened 2019-04-08
Show local language names in My Contributions.

dbrant:
Also make sure to clear any Disposables when the fragment is destroyed. (don't forget this!)


cooltey:
@sharvaniharan @dbrant 
Not sure should we following the design:
https://github.com/wikimedia/apps-android-wikipedia/pull/39#pullrequestreview-221264823

PR #169, opened 2019-04-08
Migrate to AndroidX.

dbrant:


app/build.gradle
```diff
@@ -194,33 +194,24 @@ dependencies {
...
-        exclude group: 'com.android.support'
-    }
+    implementation "com.google.android.material:material:1.0.0"
+    implementation "androidx.core:core:1.0.1"
+    implementation "androidx.cardview:cardview:1.0.0"
+    implementation 'androidx.constraintlayout:constraintlayout:1.1.3'
+    implementation "androidx.fragment:fragment:1.0.0"
+    implementation "androidx.palette:palette:1.0.0"
+    implementation "androidx.preference:preference:1.0.0"
+    implementation "androidx.recyclerview:recyclerview:1.0.0"
```
sharvaniharan: Lgtm... just one question... should we not have variable for version numbers? is it because they are all different?

dbrant: It actually looks like all of these packages are now decoupled from each other, so certain components might receive updates more often than others.

sharvaniharan: cool! will merge after travis is complete.

PR #177, opened 2019-04-10
Optimization: improve xml layout of OnThisDay card.

dbrant:
The layout of the OnThisDay card was unnecessarily heavy, with multiple nested usages of `RelativeLayout` and incorrect usages of `match_parent`, which was slowing down the layout/rendering pipeline.
(No visual changes are intended.)

### DIFF
--- a/app/src/main/res/layout/view_card_on_this_day.xml
+++ b/app/src/main/res/layout/view_card_on_this_day.xml
@@ -13,18 +13,12 @@
         android:layout_width="match_parent"
         android:layout_height="48dp" />
 
-    <FrameLayout
-        android:id="@+id/view_on_this_day_top_container"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content">
 
-        <View
-            android:id="@+id/gradient_layout"
-            android:layout_width="1dp"
-            android:layout_height="26dp"
-            android:layout_marginStart="27dp"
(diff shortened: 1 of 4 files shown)


app/src/main/res/layout/view_card_on_this_day.xml
```diff
@@ -40,85 +36,94 @@
...
-        android:layout_marginStart="27dp"
-        android:layout_weight="1"
-        android:background="?attr/chart_shade5"
-        android:clipChildren="false"
-        android:clipToPadding="false">
-
-        <RelativeLayout
-            android:id="@+id/view_on_this_day_click_container"
+        <LinearLayout
+            android:id="@+id/view_on_this_day_text_container"
```
cooltey: Feels like the `view_on_this_day_text_container` and `view_on_this_day_click_container` and be combined since the `LinearLayout` contains only one `FrameLayout`

dbrant: The `LinearLayout` is providing the thin dark line on the left (by having a dark background), and then the `FrameLayout` has a light background with a 1dp margin.  If we want to, we can simplify this further in another patch, but this would be too complex for this one.

cooltey: Make senses.

PR #178, opened 2019-04-10
Convert OnThisDay to Rx.

dbrant:
...it was still using the old Retrofit `Call` style.

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
+++ b/app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
@@ -49,9 +49,9 @@ import butterknife.BindView;
 import butterknife.ButterKnife;
 import butterknife.OnClick;
 import butterknife.Unbinder;
-import retrofit2.Call;
-import retrofit2.Callback;
-import retrofit2.Response;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
```diff
@@ -152,42 +152,25 @@ private void requestEvents(int month, int date) {
...
-        errorView.setError(t);
-        errorView.setVisibility(View.VISIBLE);
-        progressBar.setVisibility(View.GONE);
-        eventsRecycler.setVisibility(View.GONE);
+        disposables.add(ServiceFactory.getRest(wiki).getOnThisDay(month + 1, date)
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe(response -> {
+                    onThisDay = response;
+                    progressBar.setVisibility(View.GONE);
```
cooltey: Move `progressBar.setVisibility(View.GONE);` to `doFinally()`?

dbrant: Sure!

PR #187, opened 2019-04-11
Move undo/redo edit buttons to keyboard overlay view.

dbrant:
https://phabricator.wikimedia.org/T219692


sharvaniharan:
Looks good... I think the ticket also calls for a change to "OOUI icons for Template and Reference"

dbrant:
> Looks good... I think the ticket also calls for a change to "OOUI icons for Template and Reference"

This can be done in a follow-up PR. (I'm still back-and-forth with Robin on the specific icons.)

PR #191, opened 2019-04-13
Re-used translations audit

sharvaniharan:
Created local strings where the app was using android.R.string strings, so we can have our own translations for the same string in different contexts.
Strings changes : 
android.R.string.ok
android.R.string.cancel
android.R.string.yes
android.R.string.no
Total occurrences: 29

Phab: https://phabricator.wikimedia.org/T215255


app/src/main/res/values/strings.xml
```diff
@@ -778,4 +778,31 @@
...
+    <string name="customize_lang_selection_dialog_cancel_button_text">Cancel</string>
+    <string name="log_out_dialog_cancel_button_text">Cancel</string>
+    <string name="editing_login_cancel_button_text">Cancel</string>
+    <string name="remove_from_list_dialog_cancel_button_text">Cancel</string>
+    <string name="remove_from_offline_cancel_button_text">Cancel</string>
+    <string name="download_using_mobile_data_cancel_button_text">Cancel</string>
+    <string name="remove_language_dialog_cancel_button_text">Cancel</string>
+    <string name="custom_date_picker_dialog_cancel_button_text">Cancel</string>
+    <string name="delete_list_dialog_cancel_button_text">Cancel</string>
+    <string name="cancel">Cancel</string>
```
cooltey: Is it possible that have these "OK" items having the same naming logic for the related components?

For example, the `delete_list_dialog_ok_button_text` should be named as `reading_list_delete_confirm_ok_button_text`, and put it under the `reading_list_delete_confirm` item.

```
    <string name="reading_list_added_articles_list_titled">Added %1$d articles to %2$s</string>
    <string name="reading_list_delete_confirm">Are you sure you want to delete %s?</string>
    <string name="reading_list_delete_confirm_ok_button_text">OK</string>    <=== Like this
```

Thank you.

sharvaniharan: Done.

app/src/main/res/values/strings.xml
```diff
@@ -780,4 +804,10 @@
     <string name="main_drawer_close">Close</string>
     <string name="main_drawer_help">Help</string>
     <string name="main_drawer_login">Log in / join Wikipedia</string>
+    <string name="custom_date_picker_dialog_ok_button_text">OK</string>
+    <string name="text_input_dialog_ok_button_text">OK</string>
+    <string name="ok">OK</string>
```
cooltey: Not sure why don't move the strings that relate to developer preferences to `dev_settings_strings.xml`?

app/src/main/res/values/strings.xml
```diff
@@ -780,4 +804,10 @@
...
+    <string name="text_input_dialog_cancel_button_text">Cancel</string>
+    <string name="custom_date_picker_dialog_cancel_button_text">Cancel</string>
+    <string name="cancel">Cancel</string>
```
cooltey: Not sure why don't move the strings that relate to developer preferences to `dev_settings_strings.xml`?

sharvaniharan: Done.

app/src/main/res/values/strings.xml
sharvaniharan: Done

PR #198, opened 2019-04-15
Upgrade TranslationTests and update/remove string items

cooltey:
The new `TranslationTests` contains three test cases:

**testAllTranslations**:
 - Test all translations that contain parameters such as `%d`, `%s`, and compare with the base `strings.xml`.

**testPluralsDeclaration**:
 - Test the plurals items in non `en` languages to avoid build failures.

**testUnsupportedTexts**:
 - Test whether the string items contain any Wikitext or Markdown formats.

https://phabricator.wikimedia.org/T198474
https://phabricator.wikimedia.org/T158807


dbrant:
@cooltey This all looks perfectly good to me!
However, what do you think of the following:  Let's *keep* the old tests, in addition to the new ones that you made. (i.e. rename the old tests to `TranslationTestsOld.java`)
This can be just for a little while, until we're completely confident that the new tests catch at least all the errors that the old tests also caught.

cooltey:
> @cooltey This all looks perfectly good to me!
> However, what do you think of the following: Let's _keep_ the old tests, in addition to the new ones that you made. (i.e. rename the old tests to `TranslationTestsOld.java`)
> This can be just for a little while, until we're completely confident that the new tests catch at least all the errors that the old tests also caught.

Sounds good to me! I've added the old TransationTests and renamed it.

PR #200, opened 2019-04-17
Use Constants.InvokeSource instead of SearchInvokeSource

sharvaniharan:
Use Constants.InvokeSource instead of SearchInvokeSource


app/src/main/java/org/wikipedia/Constants.java
```diff
@@ -77,7 +77,13 @@
...
         FEED,
         NOTIFICATION,
-        APP_SHORTCUTS
+        APP_SHORTCUTS,
+        TOOLBAR,
+        WIDGET,
+        INTENT_SHARE,
+        INTENT_PROCESS_TEXT,
+        FEED_BAR,
+        VOICE
```
cooltey: Not sure but should these `Search` related invoke source can have names like `SEARCH_VOICE`?

dbrant: I believe the point of these patches is to abstract away the various invoke sources into general-purpose constants, usable from multiple contexts. Therefore in this case it might actually be beneficial to have more general names than specific ones.

PR #207, opened 2019-04-18
Update gradle build tool version to 3.4.0

cooltey:


cooltey:
Need to investigate what's new here: https://developer.android.com/studio/releases/gradle-plugin

cooltey:
Found that in Gradle build tool `3.4.0`, it has a task call `app:packageAlphaReleaseUnitTestForUnitTest` when running `testAlphaRelease`, which does not exist in the previous version.

cooltey:
Found that the issue might also relate to the `RobolectricTestRunner`
https://github.com/robolectric/robolectric/issues/4921

PR #209, opened 2019-04-19
Update checkstyle ImportOrder rule with the new 'androidx' imports

cooltey:
After we migrated to AndroidX, the **CheckStyle** rules of import order should be updated, too.

The main change in this PR is to update the `ImportOrder` rules in `gradle/src/checkstyle.xml`, and the rest of the job is to re-run the `Optimize imports` in Android Studio.

### DIFF
--- a/gradle/src/checkstyle.xml
+++ b/gradle/src/checkstyle.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <!DOCTYPE module PUBLIC
-    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
-    "https://checkstyle.org/dtds/configuration_1_3.dtd">
+    "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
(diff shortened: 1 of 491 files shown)


gradle/src/checkstyle.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <!DOCTYPE module PUBLIC
-    "-//Puppy Crawl//DTD Check Configuration 1.2//EN"
-    "http://www.puppycrawl.com/dtds/configuration_1_2.dtd">
+    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+    "https://checkstyle.org/dtds/configuration_1_3.dtd">
```
dbrant: Is it necessary to change this DTD spec? This causes the checkstyle task to fail on my Windows workstation, since it doesn't seem to honor the SSL certificate of the `checkstyle.org` domain.

cooltey: That's weird. The original DTD URL seems to be a `404 not found` URL and that's why I changed it. I can change it back.

PR #214, opened 2019-04-20
Use Constants.InvokeSource in RandomActivity

sharvaniharan:
Use Constants.InvokeSource instead of individual source constants in RandomActivity


app/src/main/java/org/wikipedia/Constants.java
```diff
@@ -75,6 +75,7 @@
         FEED_CARD_SUGGESTED_EDITS_TRANSLATE_DESC,
         SUGGESTED_EDITS_ONBOARDING,
         ONBOARDING_DIALOG,
+        APP_SHORTCUTS,
```
dbrant: Can we add new values at the end of the list, so that the ordinal numbers wouldn't be reshuffled every time we refactor these parts?

sharvaniharan: Done. Will adjust https://github.com/wikimedia/apps-android-wikipedia/pull/200/files after this is merged.

PR #215, opened 2019-04-23
Use correct resources for the espresso test class.

cooltey:


sharvaniharan:
Should we even invest more time into this or just remove all Espresso related code?

PR #217, opened 2019-04-24
ToC onboarding tweaks: scroller and onboarding text visibility

cooltey:
https://phabricator.wikimedia.org/T208054


app/src/main/java/org/wikipedia/page/ToCHandler.java
```diff
@@ -134,18 +137,15 @@ public ToCHandler(final PageFragment fragment, ViewGroup tocContainer, PageScrol
     void setupToC(@NonNull Page page, @NonNull WikiSite wiki, boolean firstPage) {
         adapter.setPage(page);
         rtl = L10nUtil.isLangRTL(wiki.languageCode());
+        showOnboading = Prefs.isTocTutorialEnabled() && !page.isMainPage() && !firstPage && !Prefs.showActionFeedIndicator();
         tocList.setRtl(rtl);
         setConditionalLayoutDirection(tocContainer, wiki.languageCode());
         FrameLayout.LayoutParams params = (FrameLayout.LayoutParams)tocContainer.getLayoutParams();
-        params.gravity = rtl ? Gravity.LEFT : Gravity.RIGHT;
+        params.gravity = rtl ? Gravity.START : Gravity.END;
```
dbrant: There is a reason we were using explicit `LEFT` and `RIGHT` for the gravity. When the system language is an RTL language, "start" will mean "right", and the position of the ToC will be broken now.

Please restrict this PR to what it says it does: "scroller and onboarding text visibility".

cooltey: I see. Have also removed the unrelated changes and will have other PR to handle it.

cooltey: #221

PR #218, opened 2019-04-25
Using PopupMenu for the LongPress actions on links

cooltey:
Use trhe `PopupMenu` for the long press actions on links or `ListView` items.

The only concern is that we cannot set header text to `PopupMenu` 

https://phabricator.wikimedia.org/T220218


app/src/main/java/org/wikipedia/LongPressHandler.java
```diff
@@ -55,67 +50,64 @@ public void onCreateContextMenu(ContextMenu menu, View view,
...
-                return true;
-            case R.id.menu_long_press_share_page:
-                contextMenuListener.onShareLink(title);
-                return true;
-            case R.id.menu_long_press_add_to_list:
-                contextMenuListener.onAddToList(title, CONTEXT_MENU);
-                return true;
-            default:
-            return false;
+            AlertDialog.Builder builder = new AlertDialog.Builder(view.getContext());
```
dbrant: Can we use a `PopupMenu` instead of `AlertDialog`? That way, we can directly inflate the menu resource, and not require another array of strings.

cooltey: Sure, it would be much easier. Only concern is that we are no longer able to set the header for the menu.

PR #222, opened 2019-04-25
Follow-up to ToC tweaks.

dbrant:
I'm not sure why we needed that specific condition. It doesn't seem to be necessary.


cooltey:
Found this was added on this commit: 
https://github.com/wikimedia/apps-android-wikipedia/commit/86957e269cc5b0068cd0f8a36e68567bf0a26371#diff-2a6071302496297f2817a75c39a77d40

cooltey:
> Found this was added on this commit:
> [86957e2#diff-2a6071302496297f2817a75c39a77d40](https://github.com/wikimedia/apps-android-wikipedia/commit/86957e269cc5b0068cd0f8a36e68567bf0a26371#diff-2a6071302496297f2817a75c39a77d40)

Since we've changed the logic of showing the unlock dialog, I think it is fine to be removed.

PR #223, opened 2019-04-25
Update string values "Editor tasks" to "Suggested edits"

cooltey:
Not sure why we have not noticed some of the strings are still using the old name.


app/src/main/res/values/strings.xml
```diff
@@ -650,6 +650,7 @@
     <!-- /Description editing tutorial -->
 
     <!-- Suggested edits -->
+    <string name="suggested_edits_activity_title">Suggested edits</string>
```
dbrant: Couldn't we just change the text of the existing string resource instead of creating a new one?

cooltey: Done!

PR #233, opened 2019-04-29
Call the logout API with token when logging out.

dbrant:


cooltey:
The code looks good to me. 
Will we handle the behaviors of checking the login token status in other PR?

dbrant:
> Will we handle the behaviors of checking the login token status in other PR?

Correct, this PR doesn't need to care about the validity of the current token status.

PR #239, opened 2019-04-30
Developer Settings for testing suggested cards

sharvaniharan:
-Turning on the developer setting for descriptions/translations dialog will not help load cards as it gets unset with the next poll, so created an exclusive switch for suggested edits feed cards.
- User still needs to be logged in, the cards should be enabled and still under the preBeta flag.

**Phab**: https://phabricator.wikimedia.org/T217373


cooltey:
Does this PR contain other changes rather than only the developer settings changes?

sharvaniharan:
oh god... i might have branched from my other wip branch.. will remove unnecessary code..

sharvaniharan:
created a new PR

PR #242, opened 2019-04-30
Integrate with References endpoint

cooltey:
Remaining works:
1. Rewrite the JS code to send only the key of the reference: **#cite_note-`keyofreferencesmap**
2. Send `ReferenceCollection` with only the keys instead of everything.

https://phabricator.wikimedia.org/T221665


dbrant:
@cooltey  Would you mind retargeting this onto the `mobile_html` branch, and hooking it up to the new `reference_clicked` event?

cooltey:
Sure, will do it.

cooltey:
@dbrant Done.

app/src/main/java/org/wikipedia/page/references/References.java
```diff
@@ -0,0 +1,64 @@
...
+import androidx.annotation.NonNull;
+
+import com.google.gson.annotations.SerializedName;
+
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.List;
+import java.util.Map;
+
+public class References {
```
dbrant: If you put the `@SuppressWarnings("unused,NullableProblems")` annotation at the top of this class itself, you won't need to put it for every field.

app/src/main/java/org/wikipedia/page/references/References.java
```diff
@@ -0,0 +1,64 @@
...
+    @SuppressWarnings("unused,NullableProblems")  @NonNull private String revision;
+    @SuppressWarnings("unused,NullableProblems") @NonNull private String tid;
+    @SuppressWarnings("unused,NullableProblems") @NonNull @SerializedName("references_by_id") private Map<String, Reference> referencesMap;
```
dbrant: Marking things as `@NonNull` does not guarantee that they will actually be NonNull when it's deserialized. Everything should be `@Nullable`, and then checked/guarded in the getter methods.

app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -917,17 +924,45 @@ private void setupMessageHandlers() {
...
+                L.d("Detached from activity, so stopping reference click.");
+                return;
+            }
 
-                if (!isAdded()) {
-                    L.d("Detached from activity, so stopping reference click.");
-                    return;
-                }
+            disposables.add(ServiceFactory.getRest(getTitle().getWikiSite()).getReferences(getTitle().getConvertedText())
+                    .subscribeOn(Schedulers.io())
```
dbrant: Could we actually store/cache the References response in the fragment, so that we don't have to make the network call every time?

app/src/main/java/org/wikipedia/page/references/References.java
cooltey: Thanks! You're correct!

app/src/main/java/org/wikipedia/page/references/References.java
cooltey: Done!

app/src/main/java/org/wikipedia/page/PageFragment.java
cooltey: Sure.

cooltey:
Note: this PR is ready for review now.

app/src/main/java/org/wikipedia/page/references/References.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public static class ReferenceBackLink  {
+        @Nullable private String href;
+
+        @NonNull
+        public String getHref() {
+            return StringUtils.defaultString(href);
+        }
+    }
```
sharvaniharan: Do they have to be classes? looks like we are just passing back the value with a defaultString fallback.

sharvaniharan: Looks good to me otherwise..

cooltey: It refers to the reference endpoint. Each object contains `back_links` and `content`.
You can see more detail in: 
https://en.wikipedia.org/api/rest_v1/page/references/Barack_Obama

PR #245, opened 2019-05-02
Improve behavior with zero tabs, and improve Back behavior.

dbrant:
...Forgot to handle the case of zero tabs, which was showing a "0" in the tabs button!.


cooltey:
The code looks good to me. 

Just wondering that if we press the back button to go back to explore feed and click the **same** article again, the tab will show multiple **same** articles, is this expected?

dbrant:
> Just wondering that if we press the back button to go back to explore feed and click the **same** article again, the tab will show multiple **same** articles, is this expected?

Good idea, and... done.

PR #248, opened 2019-05-03
Separated layout view for SuggestedEdits Feed card

sharvaniharan:
-Previously the feed card was designed to share the same layout as the suggested edits feed card. However, as the designs are evolving and differing in many aspects, it is becoming harder to manage them. This separates them out by creating its own layout for the feed card.


app/src/main/res/layout-land/view_suggested_edit_card.xml
```diff
@@ -0,0 +1,127 @@
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
cooltey: Maybe I am wrong, but I think we don't need the landscape view for the card view since the card view will have almost the same ratio.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsCardView.kt
```diff
@@ -29,23 +29,26 @@ class SuggestedEditsCardView(context: Context) : DefaultFeedCardView<SuggestedEd
...
-        prepareViews()
-
         isTranslation = card.isTranslation
         sourceSummary = card.sourceSummary
         targetSummary = card.targetSummary
 
         setLayoutDirectionByWikiSite(WikiSite.forLanguageCode(sourceSummary!!.lang), this)
 
+        cardView.setOnClickListener {
+            if (callback != null && card != null) {
```
cooltey: the `card` is always `NonNull`, so there's no need to check the value

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsCardView.kt
```diff
@@ -29,23 +29,26 @@ class SuggestedEditsCardView(context: Context) : DefaultFeedCardView<SuggestedEd
     var isTranslation: Boolean = false
     var sourceSummary: RbPageSummary? = null
     var targetSummary: RbPageSummary? = null
-    var addedDescription: String? = null
 
     init {
-        inflate(getContext(), R.layout.fragment_suggested_edits_add_descriptions_item, this)
+        inflate(getContext(), R.layout.view_suggested_edit_card, this)
```
cooltey: You may need to update the import from the `
import kotlinx.android.synthetic.main.fragment_suggested_edits_add_descriptions_item.view.*` to the `
import kotlinx.android.synthetic.main.view_suggested_edit_card.view.*`

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsCardView.kt
```diff
@@ -29,23 +29,26 @@ class SuggestedEditsCardView(context: Context) : DefaultFeedCardView<SuggestedEd
...
         targetSummary = card.targetSummary
 
         setLayoutDirectionByWikiSite(WikiSite.forLanguageCode(sourceSummary!!.lang), this)
 
+        cardView.setOnClickListener {
+            if (callback != null && card != null) {
+                callback!!.onSuggestedEditsCardClick(this)
+            }
+        }
+        viewArticleSubtitle.visibility = View.GONE
```
cooltey: Would it be possible to set the default `android:visibility="gone"` in the XML?

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsCardView.kt
sharvaniharan: Done

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsCardView.kt
sharvaniharan: Done

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsCardView.kt
sharvaniharan: It poses a problem during recycling... it starts showing previous card's subtitle

app/src/main/res/layout-land/view_suggested_edit_card.xml
sharvaniharan: Awesome catch! My eye had just gotten used to seeing two layouts..

sharvaniharan: Done

PR #249, opened 2019-05-03
ToC tweaks: remove peek behavior of the scroller

cooltey:
The PR will make the scroller hide completely and will show only with the ToC list.
https://phabricator.wikimedia.org/T213133


cooltey:
Note: I've removed the `peek()` method in `ToCInteractionFunnel`. 

Looks like the schema should be updated, too. ( the `hide()` should be removed as well, but will leave until we have the new schema).

PR #257, opened 2019-05-08
Hygiene: Formatting code in res folder

sharvaniharan:
No functionality changes - Only ran a 'Reformat code' on res folder.
When we are making changes to xmls, if we want to quickly format the code with Ctrl+Alt+L, it re-arranges the whole file. This will start us off with a clean slate for future changes.


dbrant:
Not sure I'm seeing the point. Also, rearranging all the `strings.xml` files is futile because they will be overwritten by the next TWN sync.

sharvaniharan:
ok we will ignore the strings files going forward or i can reset now.. but if it doesn't make much difference either way...can we please do it anyway... it would make it much easier to work on xmls[for me at least], and it is the best way the IDE feels the xml attributes should be arranged within components so if anything it is an improvement.. I will maintain them periodically on my own if commits happen without format.

sharvaniharan:
No changes needed on strings fiels

PR #259, opened 2019-05-09
Monitor whether the user has been logged out in the background.

dbrant:
https://phabricator.wikimedia.org/T221889


cooltey:
I didn't see the dialog shown after logged out from another device, the steps I've tried were:

1. Logged in on both devices with the same account.
2. Logged out on device `A`
3. Go to device `B`, and click on `Suggested edits`, and only saw the Snackbar showing the message of `you've been logged out`, which I think it might come from the API response. 


The other issue I've seen was:

After the step `3.` above, if I go to `reading lists` page and swipe down to refresh the page, my account will be connected again, and I will no long see any error messages when entering the `Suggested edits` page.

dbrant:
@cooltey  All of that is expected.  This is to handle the case if you _change your credentials_ on another device, so that we're no longer able to log in automatically within the app.

cooltey:
Thanks, @dbrant
I think I confused by the dialog text when hitting the log out button.

PR #37, opened 2019-02-19
Network Interceptors to Kotlin

sharvaniharan:
CommonHeaderRequestInterceptor.java -> CommonHeaderRequestInterceptor.kt
StatusResponseInterceptor.java -> StatusResponseInterceptor.kt
StripMustRevalidateResponseInterceptor.java ->StripMustRevalidateResponseInterceptor.kt
TestStubInterceptor.java -> TestStubInterceptor.kt
UnsuccessfulResponseInterceptor.java -> UnsuccessfulResponseInterceptor.kt
WikipediaZeroResponseInterceptor.java -> WikipediaZeroResponseInterceptor.kt

Change-Id: Ibf6d527e477bdd31fe1d3536e8ae7870f49063d7

PR #264, opened 2019-05-12
Convert ImageUrlUtil to Kotlin

veyndan:


dbrant:
Sure why not. Ideally, however, a pull request should contain a meaningful improvement in functionality or a bug fix (and a conversion to Kotlin can be a side effect of the improvement).

veyndan:
I was trying to deduce whether such a change was acceptable and found that someone made a similar such change in #37.

> Ideally, however, a pull request should contain a meaningful improvement in functionality or a bug fix (and a conversion to Kotlin can be a side effect of the improvement).

Could I get an example for when conversion to Kotlin is the side effect of some improvement?

Also, are improvements in the conciseness and readability of code welcomed? Such changes are neither an improvement in functionality or bug fixes, though personally I would say they can reduce the likelihood of bugs to occur in the future.

PR #265, opened 2019-05-13
Replace Kotlin occurrences of TextUtils.isEmpty(String) with String#isNullOrEmpty()

veyndan:
`String?#isNullOrEmpty()` uses contracts such that the string is implicitly non-nullable if the check passes. For example, [here](https://github.com/wikimedia/apps-android-wikipedia/compare/master...veyndan:kt-TextUtils?expand=1#diff-eb25cdd07ccca1c70f96371dd907b1b6R112) the not-null assertion operator is no longer required as the [surrounding conditional](https://github.com/wikimedia/apps-android-wikipedia/compare/master...veyndan:kt-TextUtils?expand=1#diff-eb25cdd07ccca1c70f96371dd907b1b6R102) states the use of the string in that instance is not-null.

As an additional bonus, the linter tells us that `String?#isNullOrEmpty()` can be replaced with `String#isEmpty()` when the string is non-null, removing a redundant non-null check, and more importantly, making the code more concise.

### DIFF
--- a/app/src/main/java/org/wikipedia/descriptions/DescriptionEditReviewView.kt
+++ b/app/src/main/java/org/wikipedia/descriptions/DescriptionEditReviewView.kt
@@ -39,7 +39,7 @@ class DescriptionEditReviewView @JvmOverloads constructor(
         articleSubtitle!!.text = StringUtils.capitalize(description)
         articleExtract!!.text = StringUtil.fromHtml(pageSummary.extractHtml)
 
-        if (pageSummary.thumbnailUrl.isNullOrEmpty()) {
+        if (pageSummary.thumbnailUrl.isNullOrBlank()) {
             articleImage.visibility = View.GONE
             articleExtract.maxLines = ARTICLE_EXTRACT_MAX_LINE_WITHOUT_IMAGE
         } else {
--- a/app/src/main/java/org/wikipedia/descriptions/DescriptionEditReviewView.kt
+++ b/app/src/main/java/org/wikipedia/descriptions/DescriptionEditReviewView.kt
@@ -2,7 +2,6 @@ package org.wikipedia.descriptions
 
 import android.content.Context
 import android.net.Uri
-import android.text.TextUtils
 import android.util.AttributeSet
 import android.view.View
 import androidx.constraintlayout.widget.ConstraintLayout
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/wikipedia/descriptions/DescriptionEditReviewView.kt
```diff
@@ -40,7 +39,7 @@ class DescriptionEditReviewView @JvmOverloads constructor(
         articleSubtitle!!.text = StringUtils.capitalize(description)
         articleExtract!!.text = StringUtil.fromHtml(pageSummary.extractHtml)
 
-        if (TextUtils.isEmpty(pageSummary.thumbnailUrl)) {
+        if (pageSummary.thumbnailUrl.isNullOrEmpty()) {
```
cooltey: For the thumbnails URLs, I think it would be better if we use `isNullOrBlank()` here.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsCardView.kt
```diff
@@ -70,7 +69,7 @@ class SuggestedEditsCardView(context: Context) : DefaultFeedCardView<SuggestedEd
     }
 
     private fun showImageOrExtract() {
-        if (TextUtils.isEmpty(sourceSummary!!.thumbnailUrl)) {
+        if (sourceSummary!!.thumbnailUrl.isNullOrEmpty()) {
```
cooltey: I think it would be better if we use `isNullOrBlank()` here.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsAddDescriptionsItemFragment.kt
```diff
@@ -140,7 +139,7 @@ class SuggestedEditsAddDescriptionsItemFragment : Fragment() {
         }
 
         viewArticleExtract.text = StringUtil.fromHtml(sourceSummary!!.extractHtml)
-        if (TextUtils.isEmpty(sourceSummary!!.thumbnailUrl)) {
+        if (sourceSummary!!.thumbnailUrl.isNullOrEmpty()) {
```
cooltey: For the thumbnails URLs, I think it would be better if we use `isNullOrBlank()` here.

veyndan:
Updated with `isNullOrBlank()`

PR #267, opened 2019-05-13
Remove unused CompositeDisposable

veyndan:
`disposable` is never added to, so only every clearing an empty `CompositeDisposable` is rather pointless.


cooltey:
That's a nice catch!

PR #268, opened 2019-05-13
Make properties of SuggestedEditsCard read-only

veyndan:


app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsCard.kt
```diff
@@ -8,24 +8,17 @@ import org.wikipedia.feed.model.CardType
...
-    var age: Int? = null
-    var isTranslation: Boolean = false
-    var sourceSummary: RbPageSummary? = null
-    var targetSummary: RbPageSummary? = null
+class SuggestedEditsCard(
+        wiki: WikiSite,
+        val isTranslation: Boolean,
+        val sourceSummary: RbPageSummary?,
+        val targetSummary: RbPageSummary?
+) : WikiSiteCard(wiki) {
```
cooltey: Please have the format like the following code:

```
SuggestedEditsCard(wiki: WikiSite,
                   val isTranslation: Boolean,
                   val sourceSummary: RbPageSummary?,
                   val targetSummary: RbPageSummary?) : WikiSiteCard(wiki) {
...
```

veyndan: Is there any style guide that you're using for Kotlin? My variation is identical to the one [recommended by Kotlin](https://kotlinlang.org/docs/reference/coding-conventions.html#class-header-formatting).

cooltey: No, we are not using the style guide for Kotlin yet, thank you for providing the reference.
But it would be better if it formats as the style of how we are following/doing now.

veyndan: Maybe a Checkstyle like tool for Kotlin could be used at some point in the future.

I've made the changes.

cooltey: Yes, it would make sense.

PR #269, opened 2019-05-13
Fix: Invisible app bar on article page with image

cooltey:
Instead of continuously updating the opacity of toolbar and its icons, we can use the height of the header view and update opacity when `ScrollY` reaches/over the height.

Also, update methods and constructor to the `package-private` level.

https://phabricator.wikimedia.org/T223121


dbrant:
Great! That's all it needs to be.

PR #271, opened 2019-05-14
Update build script to generate Bundles in addition to APKs.

dbrant:
Alright! This updates our build script to do all of the following:

- It will generate a universal `.APK`, like before.
- It will also generate a Bundle, for uploading to the Play Store, which is an `.AAB` file. When creating a new release in the Play Store (for both Beta and Production apps), we shall now use these bundles instead of the APK.

These files are placed in the `/releases` directory, as before.

### DIFF
--- a/scripts/make-release.py
+++ b/scripts/make-release.py
@@ -77,7 +77,6 @@ def git_push_tag(target, version_name):
 
 def make_release(flavors, custom_channel):
     sh.cd(PATH_PREFIX)
-    # ./gradlew -q assembleDevDebug
     args = [GRADLEW,
             '-q',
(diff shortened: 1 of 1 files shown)


cooltey:
I tested the `.aab` file with the official `bundletool`, and it looks great!
https://codelabs.developers.google.com/codelabs/your-first-dynamic-app/index.html#4

The PR looks good to me, and feel free to test it and merge it if it looks good to you as well @sharvaniharan

sharvaniharan:
@cooltey having too many problems with my ide today... getting build errors without reasons... so if it looks good to you feel free to merge!

PR #274, opened 2019-05-14
Fix: first list item of BottomContentView is not clickable

cooltey:
The first item of "Read more" in the `BottomContentView` is not clickable, and you can also notice that the ripple animation is gone.

https://phabricator.wikimedia.org/T223398


app/src/main/java/org/wikipedia/page/bottomcontent/BottomContentView.java
```diff
@@ -400,11 +400,7 @@ public long getItemId(int position) {
 
         @Override
         public View getView(int position, View convView, ViewGroup parent) {
-            PageItemView<RbPageSummary> itemView = (PageItemView<RbPageSummary>) convView;
-            if (itemView == null) {
-                itemView = new PageItemView<>(getContext());
-            }
-            itemView.setLayoutParams(new ListView.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT));
+            PageItemView<RbPageSummary> itemView = new PageItemView<>(getContext());
```
dbrant: Is it actually necessary to remove the logic for reusing the `convView`?  I believe the solution might just be to remove the `setLayoutParams()` call.

cooltey: Yes, remove the `setLayoutParams()` can resolve the issue, but it will reproduce the bug in #252 again.

dbrant: I see; in that case, what about just putting the `setLayoutParams()` call into the `if` statement that checks if `itemView == null`?  (Unless I'm missing what the issue was in #252 )

cooltey: In #252, the `setLayoutParams()` was moved out from the `if (itemView == null)` logic.

dbrant: lol  so them what was the issue in #252? I don't think I'm seeing it.

PR #282, opened 2019-05-17
Fix: incorrect toolbar opacity in PageActivity

cooltey:
The issue can be reproduced by the following steps:

1. Go to article __Perfect season__
2. Click on the second paragraph link: __Exhibition game__

This PR also brings the `tabsButton.setTabCount` back to constructor because we don't need to update the value every time after scrolled.

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageToolbarHideHandler.java
+++ b/app/src/main/java/org/wikipedia/page/PageToolbarHideHandler.java
@@ -23,7 +23,6 @@ import static org.wikipedia.util.ResourceUtil.getThemedColor;
 
 public class PageToolbarHideHandler extends ViewHideHandler {
     private static final int FULL_OPACITY = 255;
-    private static final int DELAY_UPDATE_MS = 100;
 
     private ArgbEvaluator argbEvaluator = new ArgbEvaluator();
     private boolean fadeEnabled;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/page/PageToolbarHideHandler.java
```diff
@@ -56,12 +56,13 @@
      */
     void setFadeEnabled(boolean enabled) {
         fadeEnabled = enabled;
-        update();
+
+        // to avoid the toolbar applies the opacity before image loaded
+        pageFragment.getWebView().postDelayed(this::update, DELAY_UPDATE_MS);
```
dbrant: When I see code that uses `postDelayed` to work around a problem, it usually means that there's an incomplete understanding of the underlying issue. (see below)

app/src/main/java/org/wikipedia/page/PageToolbarHideHandler.java
```diff
@@ -56,12 +56,13 @@
...
     }
 
     @Override
     protected void onScrolled(int oldScrollY, int scrollY) {
-        tabsButton.setTabCount(WikipediaApp.getInstance().getTabCount());
 
         int opacity = fadeEnabled && scrollY < (pageFragment.getHeaderView().getHeight() - toolbarHeight) ? 0 : FULL_OPACITY;
```
dbrant: Instead of using `getHeaderView().getHeight()`, can we just use `DimenUtil.leadImageHeightForDevice()`? I believe that should fix it.

cooltey: Thanks, you're right!

PR #286, opened 2019-05-18
Using Coroutine in ReadingListBehaviorsUtil.kt

cooltey:
Replace to the Kotlin `coroutine` on a few places that we've already use `RxJava` in the Kotlin file.


sharvaniharan:
Great first step towards Coroutines!! :)

PR #295, opened 2019-05-20
Fix positioning/style of Add Description button in Suggested Edits.

dbrant:
There are a number of problems with the updated "Add Description" button:
* The "elevation" is specified in the xml, but doesn't actually work. The proper parameter to use is `android:elevation` instead of `app:elevation`.
* The vertical alignment of the "arrow" buttons (to the left and right of "Add Description") isn't properly centered.
* The width of the "Add Description" button is constant (hard-coded), which means that it disregards other languages where the length of the text in the button might be different, which will make the button look weird, or will cut off the text.

Let's please pay closer attention to the details, and keep other languages in mind when building layouts.


dbrant:
@cooltey  I noticed that, but there's actually no easy way to do it. If we let it fill the button, then the highlight goes outside the boundaries of the blue shape, which looks weird. (Unless you know of a better way.)

cooltey:
> @cooltey I noticed that, but there's actually no easy way to do it. If we let it fill the button, then the highlight goes outside the boundaries of the blue shape, which looks weird. (Unless you know of a better way.)

I see, let's keep tracking it to see if we have a solution for that.

PR #296, opened 2019-05-20
Fix multi-select and scroll behavior in Reading List fragment.

dbrant:
There are a few serious issues in the ReadingListFragment where the scrolling becomes really weird (when you scroll down to collapse the toolbar, then try scrolling back up), and also when trying to multi-select items:
* The blue multi-select checkbox appears "on top" of the thumbnail image, which looks weird.
* Sometimes the multi-select state doesn't get updated unless the list is scrolled.
* Scrolling the list up and down produces strange behavior and artifacts in the collapsing toolbar.
* The system status bar is not translucent, as it should be.

This PR fixes all of the above issues.


app/src/main/java/org/wikipedia/readinglist/ReadingListFragment.java
```diff
@@ -522,14 +520,9 @@ public void onOffsetChanged(AppBarLayout appBarLayout, int verticalOffset) {
...
-            recyclerView.post(() -> {
-                if (isAdded()) {
-                    DeviceUtil.updateStatusBarTheme(requireActivity(), toolbar, toolbarExpanded && transparentStatusBarEnabled);
-                    requireActivity().getWindow().clearFlags(WindowManager.LayoutParams.FLAG_TRANSLUCENT_STATUS);
-                    requireActivity().getWindow().setStatusBarColor(transparentStatusBarEnabled
-                            ? Color.TRANSPARENT : ResourceUtil.getThemedColor(requireActivity(), R.attr.main_status_bar_color));
-                }
-            });
+            DeviceUtil.updateStatusBarTheme(requireActivity(), toolbar,
+                    (appBarLayout.getTotalScrollRange() + verticalOffset) > appBarLayout.getTotalScrollRange() / 2);
```
cooltey: Should we add `actionMode == null &&` into the logic since the list is still scrollable when multi-selecting items?

Screenshot_1558380118

dbrant: Sure!

PR #301, opened 2019-05-21
Fix the issue of media playback overlaps the navigation bar.

cooltey:
It is reproducible on my Pixel 3 XL running Android 9.0.

The full-screen mode in the `GalleryActivity` will also hide the navigation bar.

https://phabricator.wikimedia.org/T223948


sharvaniharan:
Great first step towards Coroutines!! :)

dbrant:
This is kind of unfortunate, since we're adding code to work around a problem on a _single model_ of device (I can't reproduce this on any other devices). Let's make sure to remember to check occasionally if this code is still relevant or necessary.

PR #305, opened 2019-05-23
Add API underpinnings for getting Image Captions.

dbrant:


app/src/main/java/org/wikipedia/dataclient/mwapi/media/MediaHelper.kt
```diff
@@ -0,0 +1,28 @@
...
+import org.wikipedia.dataclient.Service
+import org.wikipedia.dataclient.ServiceFactory
+import org.wikipedia.dataclient.WikiSite
+
+import java.util.HashMap
+
+import io.reactivex.Observable
+
+object MediaHelper {
+    private const val COMMONS_DB_NAME = "commonswiki";
```
cooltey: No `;` needed in a Kotlin file

app/src/main/java/org/wikipedia/dataclient/Service.java
```diff
@@ -344,4 +348,8 @@
                                              @Nullable @Field("summary") String summary,
                                              @NonNull @Field("token") String token,
                                              @Nullable @Field("assert") String user);
+
+
+
+
```
cooltey: Adding multiple empty lines by accident?

app/src/main/java/org/wikipedia/dataclient/mwapi/media/MediaHelper.kt
dbrant: done

app/src/main/java/org/wikipedia/dataclient/Service.java
dbrant: whoops

PR #309, opened 2019-05-23
Tech Debt: Add html attribute to AppTextView

veghtomi:
Add html attribute to AppTextView.

https://phabricator.wikimedia.org/T156496


veghtomi:
@dbrant I cannot add reviewers so I indicate it here.

PR #310, opened 2019-05-23
Bottom sheet popup for adding/translating/editing image captions

sharvaniharan:
-Added ImagePreviewDialog
-Updated all details ui for the image
-Updated the toolbar.
-Updated the Bottom Read bar to accommodate both the article and image caption under different conditions

NOTE: ignore all the comments with the string  : 
"//Todo: only for testing - change after image captions logic is solidified"
All logic around it will need to be tweaked just a little after the image captions logic is solidified.

Have created a reminder task for it: https://phabricator.wikimedia.org/T224709
Phab: https://phabricator.wikimedia.org/T223130


app/src/main/java/org/wikipedia/descriptions/DescriptionEditReadArticleBarView.kt
```diff
@@ -36,7 +41,29 @@ class DescriptionEditReadArticleBarView @JvmOverloads constructor(
             viewArticleImage!!.visibility = VISIBLE
             viewArticleImage!!.loadImage(Uri.parse(pageSummary.thumbnailUrl))
         }
+        viewImageThumbnail.visibility = View.GONE
+        viewReadButton.visibility = View.VISIBLE
+        show()
+    }
+
+    fun setImageDetails(thumbUrl: String, @NonNull fileName: String) {
```
cooltey: No `@NonNull` will be needed

app/src/main/java/org/wikipedia/descriptions/DescriptionEditReadArticleBarView.kt
```diff
@@ -36,7 +41,29 @@ class DescriptionEditReadArticleBarView @JvmOverloads constructor(
...
+        setConditionalLayoutDirection(this, WikipediaApp.getInstance().language().appLanguageCode)
+        viewArticleTitle!!.text = StringUtil.fromHtml(fileName)
+
+        val padding: Int = DimenUtil.dpToPx(8.0f).toInt()
+        viewArticleImage!!.visibility = VISIBLE
+        viewArticleImage!!.setPadding(padding, padding, padding, padding)
+        viewArticleImage!!.setColorFilter(ResourceUtil.getThemedColor(context, R.attr.main_toolbar_icon_color))
+        viewArticleImage.setImageResource(R.drawable.ic_keyboard_arrow_up_black_24dp)
```
cooltey: Please use `setActualImageResource` instead of `setImageResource`

app/src/main/java/org/wikipedia/views/ImageDetailView.kt
```diff
@@ -0,0 +1,34 @@
...
+    constructor(context: Context, attrs: AttributeSet?) : this(context, attrs, 0)
+
+    constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(context, attrs, defStyleAttr) {
+        init(context)
+    }
+
+    private fun init(context: Context) {
+        View.inflate(context, R.layout.view_image_detail, this)
+        layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
+
```
cooltey: To reduce `constructors` and to avoid creating new `init()` method, you can see `SwipeableListView.kt` for more example.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -0,0 +1,143 @@
...
+
+
+class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.OnDismissListener {
+
+
+    private var fileName: String? = null
+    private lateinit var imageInfo: ImageInfo
+
+    private val menuListener = PopupMenu.OnMenuItemClickListener { item ->
+        false
```
cooltey: Empty menu actions?

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -0,0 +1,143 @@
...
+
+    private fun setImageDetails() {
+        loadImage(imageInfo.originalUrl)
+        titleText!!.text = StringUtil.fromHtml(fileName!!.removePrefix("File:"))
+        addDetailPortion(getString(R.string.suggested_edits_image_caption_title), StringUtil.fromHtml(imageInfo.getMetadata()!!.imageDescription()!!.value()).toString())
+        addDetailPortion(getString(R.string.suggested_edits_image_artist), StringUtil.fromHtml(imageInfo.getMetadata()!!.artist()!!.value()).toString())
+        addDetailPortion(getString(R.string.suggested_edits_image_date), imageInfo.getMetadata()!!.dateTime()!!.value())
+        addDetailPortion(getString(R.string.suggested_edits_image_source), imageInfo.getMetadata()!!.imageDescription()!!.source())
+        addDetailPortion(getString(R.string.suggested_edits_image_licensing), imageInfo.getMetadata()!!.licenseShortName()!!.value())
+        detailsHolder.requestLayout()
```
cooltey: Instead of using `imageInfo.getMetadata()`, you can directly use `imageInfo.metadata`

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -0,0 +1,143 @@
...
+
+    companion object {
+
+        fun newInstance(imageInfo: ImageInfo, fileName: String?): ImagePreviewDialog {
+            val dialog = ImagePreviewDialog()
+            val args = Bundle()
+            args.putSerializable("imageInfo", imageInfo)
+            if (fileName != null) {
+                args.putString("filename", fileName)
+            }
```
cooltey: It would be better if create `const` value for "imageInfo" and "filename"

app/src/main/java/org/wikipedia/descriptions/DescriptionEditReadArticleBarView.kt
sharvaniharan: Done

app/src/main/java/org/wikipedia/descriptions/DescriptionEditReadArticleBarView.kt
sharvaniharan: Thank you!! I forgot to look for an alternate for the deprecated api..

app/src/main/java/org/wikipedia/views/ImageDetailView.kt
sharvaniharan: Done.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
sharvaniharan: oh I have a question for Robin on the ticket... I couldn't see what options need to be there..he has answered but I still have questions as that option is already on the screen... so will be done in a separate pr.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
sharvaniharan: Done

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
sharvaniharan: Done

app/src/main/java/org/wikipedia/descriptions/DescriptionEditReadArticleBarView.kt
sharvaniharan: it is not rendering the resource for both setActualImageResource and loadImage(int id). May be something to discuss

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
sharvaniharan: Robin confirmed that there shouldn't be any menu option. So removed.

app/src/main/java/org/wikipedia/views/ImageDetailView.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.wikipedia.R
+
+class ImageDetailView @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null, defStyle: Int = 0) : LinearLayout(context, attrs, defStyle) {
+
+    init {
+        View.inflate(context, R.layout.view_image_detail, this)
+        layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
+    }
+
+    fun setTitle(@NonNull title: String) {
```
cooltey: No `@NonNull` or `@Nullable` annotation will be needed in Kotlin

app/src/main/java/org/wikipedia/views/ImageDetailView.kt
```diff
@@ -0,0 +1,26 @@
...
+        titleTextView.setText(title)
```
cooltey: Please use proper syntax for setting text. 

`titleTextView.text = title`

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -0,0 +1,117 @@
...
+import org.wikipedia.gallery.ImageInfo
+import org.wikipedia.page.ExtendedBottomSheetDialogFragment
+import org.wikipedia.util.L10nUtil.setConditionalLayoutDirection
+import org.wikipedia.util.StringUtil
+import org.wikipedia.util.UriUtil
+import org.wikipedia.util.log.L
+
+
+class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.OnDismissListener {
+
```
cooltey: Unnecessary empty line

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -0,0 +1,117 @@
...
+
+    private var fileName: String? = null
+    private lateinit var imageInfo: ImageInfo
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        val app = WikipediaApp.getInstance()
```
cooltey: Unused variable

app/src/main/java/org/wikipedia/descriptions/DescriptionEditReadArticleBarView.kt
cooltey: I looked at the Zeplin design and compared with the actual implementation, the "arrow" image does not have the same position as the image thumbnail position. It should be an individual `ImageView`. 

https://app.zeplin.io/project/57a120b91998d8977642a238/screen/5cdac7fc25662e679ce616a2

app/src/main/java/org/wikipedia/views/ImageDetailView.kt
sharvaniharan: Done

app/src/main/java/org/wikipedia/views/ImageDetailView.kt
sharvaniharan: Done

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
sharvaniharan: Done

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
sharvaniharan: Done

app/src/main/java/org/wikipedia/descriptions/DescriptionEditReadArticleBarView.kt
sharvaniharan: for the article it is thumnail-title-'READ' button, and for image caption it is arrow-title-thumbnail where the button is.. so Why make an additional imageView?

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -0,0 +1,114 @@
...
+import org.wikipedia.util.UriUtil
+import org.wikipedia.util.log.L
+
+
+class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.OnDismissListener {
+
+    private var fileName: String? = null
+    private lateinit var imageInfo: ImageInfo
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
```
cooltey: `View?` can be `View` since it should not be `null`.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    override fun onDestroyView() {
+        toolbarView!!.setOnClickListener(null)
+        super.onDestroyView()
+    }
+
+    private fun setImageDetails() {
+        loadImage(imageInfo.originalUrl)
+        titleText!!.text = StringUtil.fromHtml(fileName!!.removePrefix("File:"))
```
cooltey: You can use `StringUtil.removeNamespace()` to remove namespace.

app/src/main/java/org/wikipedia/views/ImageDetailView.kt
```diff
@@ -0,0 +1,25 @@
...
+        layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
+    }
+
+    fun setTitle(title: String) {
+        titleTextView.text = title
+    }
+
+    fun setDetail(detail: String) {
+        detailTextView.text = detail
+    }
```
cooltey: I don't think the `setTitle()` and `setDetail()` methods are necessary. We can just directly use the well-named attributes to set up texts.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -0,0 +1,114 @@
...
+        addDetailPortion(getString(R.string.suggested_edits_image_source), imageInfo.metadata!!.imageDescription()!!.source())
+        addDetailPortion(getString(R.string.suggested_edits_image_licensing), imageInfo.metadata!!.licenseShortName()!!.value())
+        detailsHolder.requestLayout()
+    }
+
+    private fun addDetailPortion(@NonNull title: String, @Nullable detail: String?) {
+        if (!detail.isNullOrEmpty()) {
+            val view = ImageDetailView(requireContext())
+            view.setTitle(title)
+            view.setDetail(detail)
```
cooltey: We can just use: 
```
view.titleText.text = title
view.detailTextView.text = detail
```

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -0,0 +1,114 @@
...
+                    override fun onFailure(id: String?, throwable: Throwable?) {
+                        L.d(throwable)
+                    }
+                })
+                .build()
+    }
+
+    companion object {
+        const val IMAGE_INFO = "imageInfo"
+        const val FILE_NAME = "filename"
```
cooltey: These should be `private` constant variables.

app/src/main/res/layout/dialog_image_preview.xml
```diff
@@ -0,0 +1,143 @@
...
+                    android:fontFamily="serif"
+                    android:lineSpacingMultiplier="0.9"
+                    android:maxLines="3"
+                    android:paddingBottom="4dp"
+                    android:textColor="?attr/primary_text_color"
+                    android:textSize="20sp"
+                    tools:text="Lorem ipsum" />
+            </LinearLayout>
+
+            <org.wikipedia.views.ZoomableDraweeViewWithBackground
```
cooltey: I think just use the `FaceAndColorDetectImageView` with `app:actualImageScaleType="centerCrop"` should be enough.

app/src/main/res/layout/dialog_image_preview.xml
```diff
@@ -0,0 +1,143 @@
...
+                        android:textSize="16sp"
+                        android:textStyle="normal" />
+
+                    <ImageView
+                        android:layout_width="@dimen/bottom_toolbar_icon_size"
+                        android:layout_height="@dimen/bottom_toolbar_icon_size"
+                        android:layout_gravity="center_vertical"
+                        android:contentDescription="@null"
+                        android:padding="12dp"
+                        android:tint="?attr/main_toolbar_icon_color"
```
cooltey: Should the `tint` be `?attr/colorAccent`?

app/src/main/java/org/wikipedia/descriptions/DescriptionEditReadArticleBarView.kt
cooltey: Not actually related to this thread, but I am seeing the "arrow" icon has too many "left" margin than the design. The arrow should remain in the same position as the arrow in `Image Preview Dialog`

app/src/main/java/org/wikipedia/descriptions/DescriptionEditReadArticleBarView.kt
```diff
@@ -34,9 +39,31 @@ class DescriptionEditReadArticleBarView @JvmOverloads constructor(
...
-            viewArticleImage!!.loadImage(Uri.parse(pageSummary.thumbnailUrl))
+            ViewUtil.loadImageUrlInto(viewArticleImage!!, pageSummary.thumbnailUrl)
+        }
+        viewImageThumbnail.visibility = View.GONE
+        viewReadButton.visibility = View.VISIBLE
+        show()
+    }
+
+    fun setImageDetails(thumbUrl: String, fileName: String) {
+        setConditionalLayoutDirection(this, WikipediaApp.getInstance().language().appLanguageCode)
```
cooltey: Just note that the layout direction should rely on the language code from ImageDetails data.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -0,0 +1,114 @@
...
+
+class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.OnDismissListener {
+
+    private var fileName: String? = null
+    private lateinit var imageInfo: ImageInfo
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        val rootView = inflater.inflate(R.layout.dialog_image_preview, container)
+
+        setConditionalLayoutDirection(rootView, WikipediaApp.getInstance().language().appLanguageCode)
```
cooltey: Just note that the layout direction should rely on the language code from ImageDetails data.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
sharvaniharan: Done

app/src/main/res/layout/dialog_image_preview.xml
sharvaniharan: With ZoomableDraweeViewWithBackground you can pinch and zoom.. so since this is centered around the image... could we leave it, since it doesn't harm?

app/src/main/res/layout/dialog_image_preview.xml
sharvaniharan: Great catch!

app/src/main/java/org/wikipedia/descriptions/DescriptionEditReadArticleBarView.kt
sharvaniharan: I will just make a note for now..and change it with other changes

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
sharvaniharan: I will just make a note for now..and change it with other changes

app/src/main/java/org/wikipedia/descriptions/DescriptionEditActivity.java
```diff
@@ -67,12 +68,21 @@ public void onDescriptionEditSuccess() {
...
 
     @Override
-    public void onPageSummaryContainerClicked(@NonNull PageTitle pageTitle) {
-        bottomSheetPresenter.show(getSupportFragmentManager(),
-                LinkPreviewDialog.newInstance(new HistoryEntry(pageTitle,
-                        getIntent().hasExtra(EXTRA_INVOKE_SOURCE) && getIntent().getSerializableExtra(EXTRA_INVOKE_SOURCE) == PAGE_ACTIVITY
-                                ? HistoryEntry.SOURCE_EDIT_DESCRIPTION : HistoryEntry.SOURCE_SUGGESTED_EDITS),
-                        null));
+    public void onBottomBarContainerClicked(@NonNull Object object) {
+        if (object instanceof PageTitle) {
```
cooltey: The `SuggestedEditsSummary` has the `PageTitle`, and if you passing `InvokeSource` in the `onBottomBarContainerClicked` would be much better than sending an `Object` value.

app/src/main/java/org/wikipedia/descriptions/DescriptionEditBottomBarView.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    fun show() {
+        visibility = VISIBLE
+    }
+
+    fun hide() {
+        visibility = GONE
+    }
+
+    fun setSummary(summary: SuggestedEditsSummary) {
```
cooltey: I think it would be nice if we keep the same method and add an `InvokeSource` to do future logic of setting views.

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.java
```diff
@@ -68,7 +68,7 @@
 
     public interface Callback {
         void onDescriptionEditSuccess();
-        void onPageSummaryContainerClicked(@NonNull PageTitle pageTitle);
+        void onBottomBarContainerClicked(@NonNull Object object);
```
cooltey: Please see comment above.

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.java
```diff
@@ -372,8 +372,17 @@ public void onCancelClick() {
         }
 
         @Override
-        public void onReadArticleClick() {
-            callback().onPageSummaryContainerClicked(pageTitle);
+        public void onBottomBarClick() {
```
cooltey: Please see comment above.

app/src/main/java/org/wikipedia/descriptions/DescriptionEditView.java
```diff
@@ -197,8 +197,19 @@ public void setSummaries(@NonNull Activity activity, @NonNull SuggestedEditsSumm
...
+    }
+
+    private void setUpBottomBar() {
+        switch (invokeSource) {
+            case SUGGESTED_EDITS_ADD_CAPTION:
+            case SUGGESTED_EDITS_TRANSLATE_CAPTION:
+                bottomBarContainer.setImageDetails(suggestedEditsSummary);
+                break;
+            default:
+                bottomBarContainer.setSummary(suggestedEditsSummary);
```
cooltey: Please see comment above.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -0,0 +1,111 @@
...
+
+                    override fun onFailure(id: String?, throwable: Throwable?) {
+                        L.d(throwable)
+                    }
+                })
+                .build()
+    }
+
+    companion object {
+        private const val SUMMARY = "summary"
```
cooltey: Can we have a descriptive naming like `ARG_SUMMARY`

app/src/main/res/layout/dialog_image_preview.xml
```diff
@@ -0,0 +1,143 @@
...
+                    android:textSize="14sp"
+                    tools:text="Lorem ipsum" />
+            </LinearLayout>
+
+            <org.wikipedia.views.ZoomableDraweeViewWithBackground
+                android:id="@+id/galleryImage"
+                android:layout_width="match_parent"
+                android:layout_height="446dp"
+                android:contentDescription="@null"
+                app:actualImageScaleType="fitXY" />
```
cooltey: Can we use `fitCenter` for the `scaleType`?

app/src/main/res/layout/dialog_image_preview.xml
cooltey: While testing the "zoomable" feature in a preview dialog, the function does not work properly for me. I am still thinking use the general `FaceAndColorDetectImageView` would be better.

app/src/main/res/layout/dialog_image_preview.xml
cooltey: Has it been fixed? I am still seeing the incorrect tint color.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
sharvaniharan: Done

app/src/main/res/layout/dialog_image_preview.xml
sharvaniharan: Yes I tried that since it doesnt distort the image... but I am guessing that the height is fixed in the design and with fitCenter it will not be. So will re-check with Robin once and then change

app/src/main/java/org/wikipedia/descriptions/DescriptionEditBottomBarView.kt
sharvaniharan: Makes sense

PR #319, opened 2019-05-28
Show structured captions in Gallery, if available!

dbrant:
This makes an extra API call when looking at images in the Gallery, and shows a structured caption for the current image if it's available in the user's current language.

https://phabricator.wikimedia.org/T224492

### DIFF
--- a/app/src/main/java/org/wikipedia/gallery/GalleryActivity.java
+++ b/app/src/main/java/org/wikipedia/gallery/GalleryActivity.java
@@ -51,6 +51,7 @@ import org.wikipedia.theme.Theme;
 import org.wikipedia.util.ClipboardUtil;
 import org.wikipedia.util.FeedbackUtil;
 import org.wikipedia.util.GradientUtil;
+import org.wikipedia.util.ReleaseUtil;
 import org.wikipedia.util.ShareUtil;
 import org.wikipedia.util.StringUtil;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/gallery/GalleryActivity.java
```diff
@@ -542,11 +557,60 @@ public void layOutGalleryDescription() {
...
+        if (captions != null && captions.containsKey(app.getAppOrSystemLanguageCode())) {
+            descriptionStr = captions.get(app.getAppOrSystemLanguageCode());
+            captionEditButton.setVisibility(View.VISIBLE);
+
+            // and if we have another language in which the caption doesn't exist, then offer
+            // it to be translatable.
+            boolean allowTranslate = false;
+            for (String lang : app.language().getAppLanguageCodes()) {
+                if (!captions.containsKey(lang)) {
+                    allowTranslate = true;
```
cooltey: Should we add a break if it is true?

cooltey: Or, maybe we can set `allowTranslate = true` as an initial value, and then add the `break` in the `if (captions.containsKey(lang))` logic.

dbrant: Sure!

PR #320, opened 2019-05-28
Allow voice input for description/caption editing.

dbrant:
https://phabricator.wikimedia.org/T224501

### DIFF
--- a/app/src/main/res/layout/view_description_edit.xml
+++ b/app/src/main/res/layout/view_description_edit.xml
@@ -63,12 +63,24 @@
             </LinearLayout>
 
             <ImageView
+                android:id="@+id/view_description_edit_help_button"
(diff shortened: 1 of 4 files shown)


app/src/main/res/layout/view_description_edit.xml
```diff
@@ -62,13 +62,25 @@
 
             </LinearLayout>
 
+            <ImageView
+                android:id="@+id/view_description_edit_help_button"
```
cooltey: Do we need to add a logic to hide the help button if in a __Suggested edits__ flows?
https://app.zeplin.io/project/57a120b91998d8977642a238/screen/5ce57c59ff418a1e46d91e2d

dbrant: Correct!

PR #326, opened 2019-05-30
Fix: update unlock suggested edits preferences correctly.

cooltey:
To reproduce:
1. Have an account that has 50+ verified description edits.
2. Keep only "English" app language and log out.
3. Log in the same account and go to the Suggested edits
4. You will see the "Multilingual task" card with incorrect messages. It should be the one with "add languages" button


dbrant:
nice one.

PR #329, opened 2019-05-31
Update Gallery to use updated Media endpoint with structured data.

dbrant:


app/src/main/java/org/wikipedia/wikidata/Entities.java
```diff
@@ -25,12 +26,20 @@
         return entities.values().iterator().next();
     }
 
+    @Override
+    public void postProcess() {
+        if (getFirst() != null && getFirst().isMissing()) {
+            throw new RuntimeException("The requested entity was not found.");
```
cooltey: Looks like it breaks the logic of `getNextImageWithMissingCaption()`, and now it returns error everytime. Do you have any idea of that?

PR #332, opened 2019-06-01
Fix: deep link action will open a new tab

cooltey:
https://phabricator.wikimedia.org/T224789


dbrant:
This is expected behavior, and should not be modified until Design approves the change.

PR #333, opened 2019-06-02
Fix error icon incorrect rendering on API 21 and 23

veghtomi:
https://phabricator.wikimedia.org/T163230

A [great article](https://medium.com/androiddevelopers/using-vector-assets-in-android-apps-4318fd662eb9) detailing why `ContextCompat` cannot be used for vector assets.


veghtomi:
@dbrant it can be reviewed.

PR #345, opened 2019-06-06
Fix: for image touching scrollview issue in reading list

sharvaniharan:


sharvaniharan:
Not sure how to solve the problem of zero gap between the image and scrollview ... but I guess it is very edge case so can we ignore?
image

cooltey:
> Not sure how to solve the problem of zero gap between the image and scrollview ... but I guess it is very edge case so can we ignore?

If someone has a very long description on a reading list, it will happen. So I don't think that is an edge case.

sharvaniharan:
Agree

PR #354, opened 2019-06-10
Show correct total contribution count, taking captions into account.

dbrant:


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.java
```diff
@@ -142,6 +142,9 @@ private void fetchUserContributions() {
                     for (int count : editorTaskCounts.getDescriptionEditsPerLanguage().values()) {
                         totalEdits += count;
                     }
+                    for (int count : editorTaskCounts.getCaptionEditsPerLanguage().values()) {
+                        totalEdits += count;
```
sharvaniharan: We will have an empty map if the user has no caption edits, which will result in the error when we try to access its values:
error

dbrant: Whoops, good catch. This should be fixed now.

PR #355, opened 2019-06-10
Gradient bottom on the extract/image info area in Suggested edits

cooltey:
It would make sense if the view has a fading edge and also scrollable at the same time.

https://phabricator.wikimedia.org/T224500


cooltey:
> 
> 
> Looks good but is merge pending the discussion you are having on the ticket with Robin?

Yes, that's waiting for the response from Robin.

cooltey:
Hi @sharvaniharan, this PR is ready for review now.

PR #361, opened 2019-06-11
Follow up: tweaking UI and logics for the bottom bar in Suggested edits

cooltey:
- Use `FrameLayout` for the left icons and the right icons. Not sure about dynamically sets up the margins and paddings but I think to add a `ViewGroup` does not hurt the efficiency (also good for maintenance) 
- Rename the `suggested_edits_image_*` to `suggested_edits_image_preview_dialog_*` since we already have the similar string resources.
- Update some logic of using the `InvokeSource`


app/src/main/java/org/wikipedia/descriptions/DescriptionEditActivity.java
```diff
@@ -69,31 +72,22 @@ public void onDescriptionEditSuccess() {
 
     @Override
     public void onBottomBarContainerClicked(@NonNull InvokeSource invokeSource) {
```
sharvaniharan: This change is not needed, as the feed card makes this a switch statement. Could we hold this off till that gets merged. So if you can remove this change here, rest of it looks good to go

cooltey: The idea is to reduce the unnecessary value been created and also reduce the repeated functions get called.

I think it is easy that just add `FEED_CARD_SUGGESTED_EDITS_TRANSLATE_DESC`, `FEED_CARD_SUGGESTED_EDITS_TRANSLATE_IMAGE_CAPTION` to line `76`

and add `FEED_CARD_SUGGESTED_EDITS_IMAGE_CAPTION` and `FEED_CARD_SUGGESTED_EDITS_TRANSLATE_IMAGE_CAPTION` to line `82`

sharvaniharan: Yes @cooltey  makes sense for now... the switch was to facilitate readability if we end up having more types in the future... but this is more efficient for now... will merge this!

PR #363, opened 2019-06-11
Extract description from html

sharvaniharan:
To avoid :
desc


sharvaniharan:
This is the file...https://commons.wikimedia.org/wiki/File:Vertical_(31096669255).jpg so it must just be a cse of someone mistakenly entering a link for description.. Will close this

PR #368, opened 2019-06-13
Add caption CTA on article pages

sharvaniharan:
Made ui changes to add call-to-action button.
Todo: conditionally show the right text

Phab: https://phabricator.wikimedia.org/T224728


app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -168,11 +188,54 @@ private void loadLeadImage(@Nullable String url) {
...
 
+    private void updateCallToAction() {
+        if ((!Prefs.isSuggestedEditsAddCaptionsUnlocked() && !Prefs.isSuggestedEditsTranslateCaptionsUnlocked()) || getPage() == null) {
+            return;
+        }
+        WikipediaApp app = WikipediaApp.getInstance();
+        String imageName = getPage().getPageProperties().getLeadImageName();
+        String filename = "File:" + imageName;
+
+        imageCaptionDisposable = MediaHelper.INSTANCE.getImageCaptions(filename)
```
cooltey: Not sure about the logic because a local image file should not be able to edit the caption. (e.g.: https://en.wikipedia.org/wiki/Toy_Story_4)

Also, I've noticed that the source description/caption is empty in the "add image caption" screen.

Please see the logic of `updateGalleryDescription` in `GalleryActivity`.

app/src/main/res/layout/view_page_header.xml
```diff
@@ -26,6 +26,39 @@
             android:layout_width="match_parent"
             android:layout_height="match_parent" />
 
+        <LinearLayout
+            android:id="@+id/call_to_action_container"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_gravity="bottom"
+            android:paddingStart="12dp"
+            android:paddingBottom="12dp"
```
cooltey: Should use `layout_margin` for the container, otherwise, the text inside the container will not have correct gravity.
(You can see the issue of ripple animation when long pressing on the button)

sharvaniharan: Fixed the xml issue. Couldn't reproduce the crash, but I found an issue which could be affecting this as well. So please check if it is fixed now.

app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.java
```diff
@@ -71,6 +75,15 @@ public void setCallback(@Nullable Callback callback) {
         this.callback = callback;
     }
 
+    public void setUpCallToAction(String callToActionText) {
+        if (callToActionText == null) {
+            callToActionContainer.setVisibility(GONE);
+        } else {
+            callToActionContainer.setVisibility(VISIBLE);
+            callToActionTextView.setText(callToActionText);
```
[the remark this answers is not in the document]
cooltey: Has this been resolved?

app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.java
```diff
@@ -98,6 +111,12 @@ public void onScrollChanged(int oldScrollY, int scrollY, boolean isHumanScroll)
         }
     }
 
+    @OnClick(R.id.call_to_action_container) void onCallToActionContainerClicked() {
```
cooltey: Would it be better if just call it as `onCallToActionClicked()`?

app/src/main/res/layout/view_page_header.xml
```diff
@@ -26,6 +26,39 @@
             android:layout_width="match_parent"
             android:layout_height="match_parent" />
 
+        <LinearLayout
```
cooltey: The clickable area is quite small. Would it be better if adds `padding` to the `LinearLayout`?

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -168,11 +194,65 @@ private void loadLeadImage(@Nullable String url) {
...
+                                                    sourceSummary = new SuggestedEditsSummary(sourceTitle.getPrefixedText(), sourceTitle.getWikiSite().languageCode(), sourceTitle,
+                                                            sourceTitle.getDisplayText(), sourceTitle.getDisplayText(), null, getLeadImageUrl(), getLeadImageUrl(),
+                                                            null, null, null, null);
+
+                                                    targetSummary = new SuggestedEditsSummary(targetTitle.getPrefixedText(), targetTitle.getWikiSite().languageCode(), targetTitle,
+                                                            targetTitle.getDisplayText(), targetTitle.getDisplayText(), null, getLeadImageUrl(), getLeadImageUrl(),
+                                                            null, null, null, null);
+                                                    pageHeaderView.setUpCallToAction(String.format(app.getResources().getString(R.string.suggested_edits_article_cta_translate_image_caption), lang));
```
cooltey: Should use `getAppLanguageLocalizedName()` for the `lang`

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -168,11 +194,65 @@ private void loadLeadImage(@Nullable String url) {
...
     }
 
+    private void updateCallToAction() {
+        if ((!Prefs.isSuggestedEditsAddCaptionsUnlocked() && !Prefs.isSuggestedEditsTranslateCaptionsUnlocked()) || !getLeadImageUrl().contains(COMMONS_IMAGE_URL) || getPage() == null) {
+            pageHeaderView.setUpCallToAction(null);
+            return;
+        }
+        WikipediaApp app = WikipediaApp.getInstance();
+        String imageName = getPage().getPageProperties().getLeadImageName();
+        String filename = "File:" + imageName;
```
cooltey: This might cause a "non-exist" issue of a file page. Please see #437

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -44,6 +66,10 @@
     @NonNull private final PageHeaderView pageHeaderView;
 
     private int displayHeightDp;
+    private Disposable imageCaptionDisposable;
+    private SuggestedEditsSummary sourceSummary, targetSummary;
+    private boolean isTranslation;
+    private CompositeDisposable disposables = new CompositeDisposable();
```
cooltey: Not sure why we need two `disposables` in a class...

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -168,11 +194,65 @@ private void loadLeadImage(@Nullable String url) {
...
+                                sourceSummary = new SuggestedEditsSummary(getTitle().getPrefixedText(), app.getAppOrSystemLanguageCode(), getTitle(),
+                                        getTitle().getDisplayText(), getTitle().getDisplayText(), StringUtil.fromHtml(item.getDescription().getHtml()).toString(), item.getThumbnailUrl(), item.getPreferredSizedImageUrl(),
+                                        null, null, null, null);
+
+                                return;
+                            }
+                        }
+                    }
+                    if (app.language().getAppLanguageCodes().size() > 1 && Prefs.isSuggestedEditsTranslateCaptionsUnlocked()) {
+                        imageCaptionDisposable = MediaHelper.INSTANCE.getImageCaptions(filename)
```
cooltey: Can we use `map` or `flapMap` for the logic? I don't think to have another `disposable` inside a `disposable` is a good idea.

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
sharvaniharan: awesome catch!

sharvaniharan:
https://github.com/wikimedia/apps-android-wikipedia/pull/368#discussion_r300170996
This is not changed because, when we have the situation where :
- you open an image which is not in commons [ add caption not shown]
- move to the next image which happens to be in commons [ add caption shown]
- now you move back, we need a way to hide it , or else the button is still showing when the image is not editable.

app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.java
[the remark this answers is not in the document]
sharvaniharan: This is not changed because, when we have the situation where :

you open an image which is not in commons [ add caption not shown]
move to the next image which happens to be in commons [ add caption shown]
now you move back, we need a way to hide it , or else the button is still showing when the image is not editable.

app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.java
```diff
@@ -32,10 +33,13 @@
 public class PageHeaderView extends LinearLayoutOverWebView implements ObservableWebView.OnScrollChangeListener {
     @BindView(R.id.view_page_header_image) FaceAndColorDetectImageView image;
     @BindView(R.id.view_page_header_image_gradient) View gradientView;
+    @BindView(R.id.call_to_action_container) View callToActionContainer;
+    @BindView(R.id.call_to_action_text) TextView callToActionTextView;
     @Nullable private Callback callback;
 
     public interface Callback {
         void onImageClicked();
+        void onArticleCTAClicked();
```
cooltey: Would it be better if just call it as `onAddCaptionClicked()`

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -237,4 +334,14 @@ private boolean isFragmentAdded() {
     private FragmentActivity getActivity() {
         return parentFragment.getActivity();
     }
+
+    public void dispose() {
+        if (disposables != null && !disposables.isDisposed()) {
```
dbrant: These checks to not need to be performed.

app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.java
```diff
@@ -32,10 +33,13 @@
 public class PageHeaderView extends LinearLayoutOverWebView implements ObservableWebView.OnScrollChangeListener {
     @BindView(R.id.view_page_header_image) FaceAndColorDetectImageView image;
     @BindView(R.id.view_page_header_image_gradient) View gradientView;
+    @BindView(R.id.call_to_action_container) View callToActionContainer;
+    @BindView(R.id.call_to_action_text) TextView callToActionTextView;
     @Nullable private Callback callback;
 
     public interface Callback {
         void onImageClicked();
+        void onAddCaptionClicked();
```
dbrant: I like the fact that all of your other call-to-action fields and variables have purpose-agnostic names, like `callToActionTextView` and `setupCallToAction()`.  So then, shouldn't this callback function also have an agnostic name, such as `onCallToActionClicked()`?

app/src/main/res/layout/view_page_header.xml
```diff
@@ -26,6 +26,40 @@
...
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginStart="8dp"
+                android:fontFamily="sans-serif-medium"
+                android:letterSpacing="0.04"
+                android:lineSpacingExtra="6sp"
+                android:textAllCaps="true"
+                android:textColor="?attr/paper_color"
+                android:textSize="14sp"
+                android:textStyle="normal"
```
dbrant: This is the default, and does not need to be specified explicitly.

app/src/main/res/layout/view_page_header.xml
```diff
@@ -26,6 +26,40 @@
...
+                android:id="@+id/call_to_action_text"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginStart="8dp"
+                android:fontFamily="sans-serif-medium"
+                android:letterSpacing="0.04"
+                android:lineSpacingExtra="6sp"
+                android:textAllCaps="true"
+                android:textColor="?attr/paper_color"
+                android:textSize="14sp"
```
dbrant: This is the default, and does not need to be specified explicitly.

app/src/main/res/layout/view_page_header.xml
```diff
@@ -26,6 +26,40 @@
...
+                app:srcCompat="@drawable/ic_add_gray_white_24dp" />
+
+            <TextView
+                android:id="@+id/call_to_action_text"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginStart="8dp"
+                android:fontFamily="sans-serif-medium"
+                android:letterSpacing="0.04"
+                android:lineSpacingExtra="6sp"
```
dbrant: This is not necessary, since there will not be more than one line.

app/src/main/res/layout/view_page_header.xml
```diff
@@ -26,6 +26,40 @@
...
+                android:tint="?attr/paper_color"
+                app:srcCompat="@drawable/ic_add_gray_white_24dp" />
+
+            <TextView
+                android:id="@+id/call_to_action_text"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_marginStart="8dp"
+                android:fontFamily="sans-serif-medium"
+                android:letterSpacing="0.04"
```
dbrant: This is not necessary, since it will not make a difference to the human eye.

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -33,6 +55,7 @@
      * the page title.
      */
     private static final int MIN_SCREEN_HEIGHT_DP = 480;
+    private static final String COMMONS_IMAGE_URL = "/wikipedia/commons/";
```
dbrant: This is not a URL, and probably shouldn't be called a "URL".

app/src/main/java/org/wikipedia/descriptions/DescriptionEditView.java
```diff
@@ -101,7 +101,9 @@ public void setPageTitle(@NonNull PageTitle pageTitle) {
         this.pageTitle = pageTitle;
         originalDescription = pageTitle.getDescription();
         setHintText();
-        setDescription(originalDescription);
+        if (invokeSource != SUGGESTED_EDITS_ADD_CAPTION && invokeSource != SUGGESTED_EDITS_TRANSLATE_CAPTION) {
```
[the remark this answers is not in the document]
cooltey: Did this be resolved? I am still seeing the same change.

app/src/main/java/org/wikipedia/page/PageActivity.java
```diff
@@ -683,6 +684,11 @@ public void onActivityResult(int requestCode, int resultCode, final Intent data)
             } else if (resultCode == TabActivity.RESULT_LOAD_FROM_BACKSTACK) {
                 pageFragment.reloadFromBackstack();
             }
+        } else if (requestCode == Constants.ACTIVITY_REQUEST_IMAGE_CAPTION_EDIT
+                && resultCode == RESULT_OK) {
+            pageFragment.refreshPage();
+            FeedbackUtil.showMessage(this, getString(R.string.description_edit_success_saved_image_caption_in_lang_snackbar,
+                    app.language().getAppLanguageLocalizedName(StringUtils.defaultString(pageFragment.getLeadingImageEditLang()))));
```
dbrant: Since the return value of `getLeadingImageEditLang()` can be null/empty, it won't make sense to feed it into a string that always expects a language parameter. I suggest checking if the language is empty, and using the non-language-specific string `description_edit_success_saved_image_caption_snackbar` in that case.

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -44,6 +67,9 @@
     @NonNull private final PageHeaderView pageHeaderView;
 
     private int displayHeightDp;
+    private SuggestedEditsSummary sourceSummary, targetSummary;
```
dbrant: Since either of these fields can be `null`, they should be annotated as `@Nullable`, so that the IDE will give you hints that might help you catch possible NPE cases, of which I see at least one already (see below).

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -237,4 +334,14 @@ private boolean isFragmentAdded() {
...
+
+    public void dispose() {
+        if (disposables != null && !disposables.isDisposed()) {
+            disposables.clear();
+        }
+    }
+
+    public String getLeadingImageEditLang() {
+        return isTranslation ? targetSummary.getPageTitle().getWikiSite().languageCode() : sourceSummary.getPageTitle().getWikiSite().languageCode();
+    }
```
dbrant: This will produce a NPE and crash the app if `sourceSummary` or `targetSummary` is null.  This can happen if the activity is destroyed (e.g. "don't keep activities) and then recreated when coming back from editing.

app/src/main/java/org/wikipedia/descriptions/DescriptionEditView.java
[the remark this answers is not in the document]
sharvaniharan: yes

PR #375, opened 2019-06-17
Improve caching layer for suggested edits.

dbrant:
This substantially improves the efficiency of serving up suggested edits by caching both suggested Descriptions and Captions (across multiple threads).


dbrant:
@cooltey  I don't really think the slowness of the results has to do with this PR.  It's very slow for me on `master` too, which is because there are very few images with existing captions in other languages.  This will need to be solved in a separate PR.

cooltey:
> 
> 
> @cooltey I don't really think the slowness of the results has to do with this PR. It's very slow for me on `master` too, which is because there are very few images with existing captions in other languages. This will need to be solved in a separate PR.

I see, let's keep tracking it.

PR #382, opened 2019-06-19
Remove placeholder for the image view in review screen

cooltey:
`#17` in https://phabricator.wikimedia.org/T225635


sharvaniharan:
I think Robin wants color-group-21 according to https://www.mediawiki.org/wiki/Wikimedia_Apps/Android_theme_guidelines. setting it @null will not do it. Even for failure image we need to set this attribute color. because, our settings 'show images' will default to failure image styles.

cooltey:
> 
> 
> I think Robin wants color-group-21 according to https://www.mediawiki.org/wiki/Wikimedia_Apps/Android_theme_guidelines. setting it @null will not do it. Even for failure image we need to set this attribute color. because, our settings 'show images' will default to failure image styles.

Make sense. PR is ready for review now.

PR #392, opened 2019-06-22
Update image ratio in ViewPager cards to 16:9 and adjust thumbnail resolution to 800dp

cooltey:
`#27` in https://phabricator.wikimedia.org/T225635

This PR includes the changes of Suggested edits and Randomizer cards. Both have similar UIs and they should be updated at the same time.


app/src/main/java/org/wikipedia/random/RandomItemFragment.java
```diff
@@ -139,8 +142,11 @@ public void onGlobalLayout() {
                 extractView.getViewTreeObserver().removeOnGlobalLayoutListener(this);
             }
         });
+
         imageView.loadImage(TextUtils.isEmpty(summary.getThumbnailUrl()) ? null
-                : Uri.parse(summary.getThumbnailUrl()));
+                : Uri.parse((WikipediaApp.getInstance().isTablet()
+                ? ImageUrlUtil.getUrlForPreferredSize(summary.getThumbnailUrl(), PREFERRED_TABLET_THUMBNAIL_SIZE)
```
dbrant: Any reason we can't just use 800dp in all cases?  The image looks kind of pixelated on phone displays, too...

cooltey: Yes, we can!

PR #404, opened 2019-06-25
Add a method of enabling full width preview dialog

cooltey:
`#28` in https://phabricator.wikimedia.org/T225635


cooltey:
> 
> 
> I dont see why the LinkPreviewDialog needs to have the fullwidth as just an optional arguement. We can make it fullwidth always, like in ImagePreviewDialog. Here are a couple places I feel it wouldn't hurt to have it fullwidth

I can ping Robin about this concern, but I think your request can be an independent task for doing it because maybe not everywhere needs a fullwidth dialog. The reason why he liked to make it full width maybe because of the "bottom bar" in suggested edit is full width.

sharvaniharan:
Agree. Can we hold off on merging till we discuss with Robin... just to avoid reverting

sharvaniharan:
Rest of it looks good, so will merge this so that other preview dialog dependent work doesnt have to wait, as we discussed offline.

PR #411, opened 2019-06-26
Fix padding around suggested edit card.

dbrant:
If you look closely, the top border of the suggested edit card is cut off (overlapped by the toolbar).


cooltey:
Great fix!

PR #415, opened 2019-06-27
Fix: prevent showing mixed Chinese variants articles when opening /zh/ or /wiki/ URLs

cooltey:
URLs for testing
https://zh.wikipedia.org/wiki/核武器
https://zh.wikipedia.org/zh/核武器

### DIFF
--- a/app/src/main/java/org/wikipedia/language/LanguageUtil.java
+++ b/app/src/main/java/org/wikipedia/language/LanguageUtil.java
@@ -11,6 +11,7 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.core.os.LocaleListCompat;
 
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/language/LanguageUtil.java
```diff
@@ -141,5 +143,19 @@ private static boolean isTraditionalChinesePredominantInCountry(@Nullable String
         return TRADITIONAL_CHINESE_COUNTRY_CODES.contains(country);
     }
 
+    @NonNull
+    public static String getFirstSelectedChineseVariant() {
+        if (FIRST_SELECTED_ZH_LANG_CODE == null) {
```
dbrant: What if the user changes the selected Chinese variant while the app is running?
Is this really so expensive that it needs to be cached?

cooltey: You're correct! and done.

PR #422, opened 2019-06-28
Show links and make them clickable

sharvaniharan:
Phab: https://phabricator.wikimedia.org/T226118


app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -137,6 +134,14 @@ class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.
         ViewUtil.loadImageUrlInto(galleryImage, url)
     }
 
+    private fun getClickListenerFor(url: String?): View.OnClickListener {
+        return View.OnClickListener {
```
cooltey: Should it have a `url` null check here? I am seeing some links will open a blank page.

app/src/main/res/layout/view_image_detail.xml
```diff
@@ -27,16 +28,33 @@
...
         android:layout_height="wrap_content"
-        android:layout_marginTop="11dp"
-        android:fontFamily="sans-serif"
-        android:lineSpacingExtra="8sp"
-        android:textColor="?attr/primary_text_color"
-        android:textSize="16sp"
-        android:textStyle="normal"
-        tools:text="Watercolor portrait of Ada King, Countess of Lovelace (Ada Lovelace)" />
+        android:orientation="horizontal"
+        android:paddingTop="12dp">
```
cooltey: It would be better if add `android:background="?android:selectableItemBackground"` to this view

and please don't forget to add `paddingBottom` to make the text align to the middle. (and will need to adjust the parent's bottom padding)

app/src/main/java/org/wikipedia/util/StringUtil.java
```diff
@@ -261,4 +264,16 @@ public static String listToJSONString(@Nullable List<Integer> list) {
 
     private StringUtil() {
     }
+
+    public static ArrayList<String> extractUrlsFromhtmlString(String html) {
+        ArrayList<String> result = new ArrayList<>();
+        String regex = "(https)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]";
+        Pattern pattern = Pattern.compile(regex);
+        Matcher matcher = pattern.matcher(html);
```
cooltey: Is it necessary to create a new `Pattern`? It would be better if does the following: 
`Matcher matcher = Pattern.compile(regex).matcher(html);`

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
sharvaniharan: I don't think it is null urls. sometimes the response itself doesn't have well-formed ones. However, I have put in a null check.

app/src/main/res/layout/view_image_detail.xml
sharvaniharan: I did not want to align the external link icon to the center, because sometimes the details have a lot of new lines and that makes the icon look out of place.

app/src/main/java/org/wikipedia/util/StringUtil.java
sharvaniharan: I feel things like this make code easily readable, but I have changed it anyway.

sharvaniharan:
Done with the link changes.

app/src/main/java/org/wikipedia/gallery/ExtMetadata.java
```diff
@@ -57,13 +57,32 @@
     }
 
     @NonNull public String dateTime() {
-        return StringUtils.defaultString(dateTimeOriginal == null ? null : dateTimeOriginal.value());
+        if (dateTimeOriginal == null || dateTimeOriginal.value() == null) {
+            return "";
```
cooltey: Any reason why to change the logic to display an empty string?

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -104,39 +103,56 @@ class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.
...
+        addDetailPortion(getString(R.string.suggested_edits_image_preview_dialog_date), StringUtil.fromHtml(suggestedEditsSummary.metadata!!.dateTime()), null)
+        addDetailPortion(getString(R.string.suggested_edits_image_preview_dialog_source), StringUtil.fromHtml(suggestedEditsSummary.metadata!!.creditUrl()), null)
+        addDetailPortion(getString(R.string.suggested_edits_image_preview_dialog_licensing), SpannableString(suggestedEditsSummary.metadata!!.licenseShortName()), getClickListenerFor(suggestedEditsSummary.metadata!!.licenseUrl()))
+        addDetailPortion(getString(R.string.suggested_edits_image_preview_dialog_more_info), SpannableString(getString(R.string.suggested_edits_image_preview_dialog_file_page_link_text)), getClickListenerFor(getString(R.string.suggested_edits_image_file_page_commons_link, suggestedEditsSummary.title)))
         detailsHolder.requestLayout()
     }
 
-    private fun addDetailPortion(titleString: String, @Nullable detail: String?, shouldAddAccentTint: Boolean) {
-        if (!detail.isNullOrEmpty()) {
+    private fun addDetailPortion(titleString: String, @Nullable detail: Spanned, linkClickListener: View.OnClickListener?) {
```
cooltey: Please don't use `@Nullable` annotation for Kotlin because you can use `?` for the type.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -104,39 +103,56 @@ class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.
...
+        if (!detail.isEmpty()) {
```
cooltey: Not sure why don't follow the original logic.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -104,39 +103,56 @@ class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.
...
             val view = ImageDetailView(requireContext())
             view.titleTextView.text = titleString
-            if (shouldAddAccentTint) {
+            view.detailTextView.setMovementMethod(movementMethod)
```
cooltey: Please use the `view.detailTextView.movementMethod = movementMethod`

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -104,39 +103,56 @@ class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.
...
+            if (linkClickListener != null) {
                 view.detailTextView.setTextColor(ResourceUtil.getThemedColor(context!!, R.attr.colorAccent))
+                view.externalLinkView.visibility = VISIBLE
+                view.detailsContainer.setOnClickListener(linkClickListener)
             }
-            view.detailTextView.text = StringUtil.strip(StringUtil.removeHTMLTags(detail))
+            view.detailTextView.text = detail
```
cooltey: Why don't just use the following:
```
view.detailTextView.text = StringUtil.strip(StringUtil.fromHtml(detail))
```

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -104,39 +103,56 @@ class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.
             // Show the image description when a structured caption does not exist.
             addDetailPortion(getString(R.string.suggested_edits_image_preview_dialog_description_in_language_title,
                     WikipediaApp.getInstance().language().getAppLanguageLocalizedName(suggestedEditsSummary.lang)),
-                    suggestedEditsSummary.description, false)
+                    SpannableString(suggestedEditsSummary.description), null)
```
cooltey: I think we can just use `view.detailTextView.text = StringUtil.strip(StringUtil.fromHtml(detail))`. Please see the following comment.

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -104,39 +103,56 @@ class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.
...
         } else {
             addDetailPortion(getString(R.string.suggested_edits_image_preview_dialog_caption_in_language_title,
                     WikipediaApp.getInstance().language().getAppLanguageLocalizedName(suggestedEditsSummary.lang)),
-                    if (suggestedEditsSummary.pageTitle.description.isNullOrEmpty()) suggestedEditsSummary.description
-                    else suggestedEditsSummary.pageTitle.description, false)
+                    SpannableString(if (suggestedEditsSummary.pageTitle.description.isNullOrEmpty()) suggestedEditsSummary.description
```
cooltey: same

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
```diff
@@ -104,39 +103,56 @@ class ImagePreviewDialog : ExtendedBottomSheetDialogFragment(), DialogInterface.
...
+                view.externalLinkView.visibility = VISIBLE
+                view.detailsContainer.setOnClickListener(linkClickListener)
             }
-            view.detailTextView.text = StringUtil.strip(StringUtil.removeHTMLTags(detail))
+            view.detailTextView.text = detail
             detailsHolder.addView(view)
         }
     }
 
+    private val movementMethod = LinkMovementMethodExt { url: String -> handleExternalLink(context, Uri.parse(url)) }
```
cooltey: Should we verify the link to make sure it starts with `https` or `http`?

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
cooltey: Screenshot_1562622962

app/src/main/java/org/wikipedia/views/ImagePreviewDialog.kt
sharvaniharan: because i have started sending a spanned instead of string... so it wont be null ...only empty

sharvaniharan:
the descriptions with html has been automatically resolved:
description

PR #430, opened 2019-07-01
Make suggested edit cards scrollable and clickable and rippleable.

dbrant:


app/src/main/res/layout/fragment_suggested_edits_cards_item.xml
```diff
@@ -40,143 +36,157 @@
...
                 android:layout_height="wrap_content"
-                android:minHeight="231dp"
-                android:orientation="vertical"
-                android:paddingTop="11dp">
-
-                <TextView
-                    android:id="@+id/viewArticleTitle"
+                android:clickable="true"
+                android:focusable="true"
+                android:background="?attr/selectableItemBackground"
```
cooltey: It would be better if change this to `android:foreground="?attr/selectableItemBackground"`

dbrant: Why is that?

cooltey: The `ImageView` and the `LinearLayout` that contains a blue left label with a description will not be covered by ripple animation.

dbrant: Ah ok. I was testing on API 27, where the issue wasn't as evident.
Done.

PR #432, opened 2019-07-01
Fix: add "no description" string for the suggested edits explore feed card

cooltey:
...to prevent empty description field on the edit screen.


app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
```diff
@@ -33,6 +35,7 @@ class SuggestedEditsFeedClient(private var invokeSource: Constants.InvokeSource)
 
     override fun request(context: Context, wiki: WikiSite, age: Int, cb: FeedClient.Callback) {
         this.age = age
+        this.context = context
```
dbrant: Can we not store the instance of `Context`, and instead use `WikipediaApp.getInstance()`?

cooltey: Yes, done.

PR #437, opened 2019-07-02
Fix: use correct the namespace for submitting image captions

cooltey:
`#30` in https://phabricator.wikimedia.org/T225635#5299612

When reading an article in German, the output from the REST endpoint will have a title with a German file namespace.

https://de.wikipedia.org/wiki/Lucio_Amelio
https://de.wikipedia.org/api/rest_v1/page/media/Lucio_Amelio


sharvaniharan:
The caption edit is going through fine...but there is a tiny bug with the success toast
I only edited it, not translation.
Screenshot_1562274642

cooltey:
In that case, we can have other PR to fix it.

sharvaniharan:
@cooltey i did not get a chance to verify if it is just the toast or if there is a bigger problem with the target pagetitle or something. So please look into it and @dbrant can take that call tomorrow.

PR #443, opened 2019-07-05
Fix: use landscape specific layout for proper imageView aspect ratio in Randomizer

cooltey:
The current `app:viewAspectRatio` in `fragment_random_item.xml` is only suitable for portrait mode.


cooltey:
> 
> 
> It is just a one logic change. do we really need its own landscape layout for this? Can we just handle this dynamically to avoid maintenance headaches?

Makes sense, and done!

PR #444, opened 2019-07-05
Tweak: keep "deep link" users in the app when pressing the back button

cooltey:
https://phabricator.wikimedia.org/T224789#5306112

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageActivity.java
+++ b/app/src/main/java/org/wikipedia/page/PageActivity.java
@@ -243,7 +243,11 @@ public class PageActivity extends BaseActivity implements PageFragment.Callback,
     public boolean onOptionsItemSelected(MenuItem item) {
         switch (item.getItemId()) {
             case android.R.id.home:
-                onBackPressed();
+                if (app.haveMainActivity()) {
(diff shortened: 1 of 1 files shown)


sharvaniharan:
just checking app.haveMainActivity() might not be enough here... if we have stuff on the stack it will not work as expected:

https://youtu.be/yXTG0AUqpR4

cooltey:
@sharvaniharan This PR fixes the issue when users open a deep link from a web browser, the "back" button show go back to the explore feed. It does not relate to the issue you've mentioned.

sharvaniharan:
but the ticket mentions that the system back should take them back to chrome and stay in the app. That is what is not happening as seen in the video..

sharvaniharan:
Sorry the link was wrong . Here is the correct one
https://youtu.be/y78VDTH043g

Sorry for the confusion.

sharvaniharan:
@cooltey as discussed offline, I will post the video on the ticket since the system back button only works on a fresh launch as expected. I agree that it will be too much code to maintain, in order to detect a deep-link launch and behave differently. Will post the video on the ticket too, so Robin is aware of this decision.

PR #448, opened 2019-07-09
Fix: Chinese variants issues that title in the lists does not match the article itself

cooltey:
Since we use the `PageTitle.getDisplayText()` in many places, the Chinese users will notice that the article's title does not match the title they just saw.

This PR fixes the following parts that make the title matches the title in the article: 
 - Articles' title in the History page 
 - Articles' title in the Reading list 
 - Articles' title in the TabsActivity 
 - Use the `PageTitle.getRequestUrlText()` when requesting data from the endpoint

Ticket:
https://phabricator.wikimedia.org/T178201


cooltey:
Found a possible issue in reading list endpoint since it is not able to save language variant when adding an entry. 

For example, the Internet of things in Chinese is: `https://zh.wikipedia.org/wiki/物联网`, but the display name is `IoT技術` in Traditional Chinese and `物联网` in Simplified Chinese. When saving these two articles into the same list, the endpoint will consider they are the same entry. 

Not sure if it is possible to have the RI team fix that by adding a new column for adding language variant.

cooltey:
TODO: will fix the conflicts after `mobile-html` merged.

cooltey:
Closed since #623 has a better implementation.

PR #454, opened 2019-07-09
Correctly preserve instanceState of Suggested Edit cards.

dbrant:
This fixes the issue of not being able to click the Suggested Edit card when returning from editing the caption. (with "don't keep activities" enabled)


cooltey:
Great fix!

PR #460, opened 2019-07-10
Adjusting bottom sheet peekHeight

sharvaniharan:
On tablets, the default peekheight of bottomsheets is very small. so, we need to set it dynamically to a fraction of the screen height.
**Phab**: https://phabricator.wikimedia.org/T223130

### DIFF
--- a/app/src/main/java/org/wikipedia/page/ExtendedBottomSheetDialogFragment.java
+++ b/app/src/main/java/org/wikipedia/page/ExtendedBottomSheetDialogFragment.java
@@ -1,9 +1,14 @@
 package org.wikipedia.page;
 
 import android.content.res.Configuration;
+import android.content.res.Resources;
 import android.os.Bundle;
+import android.view.View;
 import android.view.ViewGroup;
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/wikipedia/page/ExtendedBottomSheetDialogFragment.java
```diff
@@ -26,6 +31,23 @@ public void onCreate(Bundle savedInstanceState) {
     public void onStart() {
         super.onStart();
         setWindowLayout();
+        setPeekHeightPerDevice();
+    }
+
+    private void setPeekHeightPerDevice() {
```
dbrant: If you look in `AddToReadingListDialog` line 120, it shows you how to do this using a single line of code. Also let's do this in `ImagePreviewDialog`, instead of the base class.

sharvaniharan:
I have picked heights which i felt were best for phones, 7" and 10" tablets respectively since Robin is ooo. Please let me know if you feel any of them need to be changed.

PR #462, opened 2019-07-10
Fix: don't create a default list every time when upgrading the database

cooltey:
The `currentLists` is empty, so it will let the `createDefaultList()` create a default list every time when upgrading the database.


dbrant:
Sorry, what issue is this fixing?  Where is the default list being recreated?

cooltey:
> 
> 
> Sorry, what issue is this fixing? Where is the default list being recreated?

This relates to this PR: #448 since the db version will be updated to `19`.

dbrant:
...But that code won't run, will it? It should only run when the DB is upgraded to 18.

cooltey:
> 
> 
> ...But that code won't run, will it? It should only run when the DB is upgraded to 18.

I think you're right. I checked my code and found that is not necessary to run in `19`. Will close this PR.

PR #477, opened 2019-07-16
Article cta design tweak

sharvaniharan:
-Fixed bottom gradient
-Changes text to "Add Image Caption (Language)" for both adding and translating image captions.

**Phab**: https://phabricator.wikimedia.org/T224728


app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -234,7 +234,7 @@ private void updateCallToAction() {
                                 captionSourcePageTitle = new PageTitle(title[0], new WikiSite(Service.COMMONS_URL, getTitle().getWikiSite().languageCode()));
 
                                 if (!captions.containsKey(getTitle().getWikiSite().languageCode())) {
-                                    pageHeaderView.setUpCallToAction(app.getResources().getString(R.string.suggested_edits_article_cta_add_image_caption));
+                                    pageHeaderView.setUpCallToAction(String.format(app.getResources().getString(R.string.suggested_edits_article_cta_image_caption), app.language().getAppLanguageLocalizedName(getTitle().getWikiSite().languageCode())));
```
cooltey: Please use `getString(resId, arg)` instead of `String.format()`

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -255,7 +255,7 @@ private void updateCallToAction() {
                                             callToActionTargetSummary = new SuggestedEditsSummary(captionTargetPageTitle.getPrefixedText(), captionTargetPageTitle.getWikiSite().languageCode(), captionTargetPageTitle,
                                                     captionTargetPageTitle.getDisplayText(), captionTargetPageTitle.getDisplayText(), null, getLeadImageUrl(), getLeadImageUrl(),
                                                     null, null, null, null);
-                                            pageHeaderView.setUpCallToAction(String.format(app.getResources().getString(R.string.suggested_edits_article_cta_translate_image_caption), app.language().getAppLanguageLocalizedName(lang)));
+                                            pageHeaderView.setUpCallToAction(String.format(app.getResources().getString(R.string.suggested_edits_article_cta_image_caption), app.language().getAppLanguageLocalizedName(lang)));
```
cooltey: Please use `getString(resId, arg)` instead of `String.format()`

app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.java
```diff
@@ -138,6 +138,7 @@ private void init() {
         ButterKnife.bind(this);
         ViewCompat.setTransitionName(this, getContext().getString(R.string.transition_floating_queue));
         gradientView.setBackground(getPowerGradient(R.color.black38, Gravity.TOP));
+        gradientView.setBackground(getPowerGradient(R.color.black38, Gravity.BOTTOM));
```
cooltey: Not related to this, but I found that in dark mode, the text is almost unseeable. 
Should we change the text and the icon color to be just white? (or ask Robin for the suggestion)

dbrant:
Was this merged by accident, or is there a follow-up coming?
- The bottom gradient should not be shown if the call-to-action button is not visible.
- The text of the button should always be white.

cooltey:
@dbrant There's a follow-up PR for it (since Sharavni should be off now). 

Will update the PR with the "The bottom gradient should not be shown if the call-to-action button is not visible."

PR #479, opened 2019-07-16
Follow up: get rid of String.format() and use proper text and icon color for dark theme

cooltey:


dbrant:
Excellent; one more bit of feedback from the previous PR:
- The bottom gradient should not be shown if the call-to-action button is not visible.

PR #487, opened 2019-07-18
Replace ProgressDialog with ProgressBar: LoginActivity

cooltey:
https://phabricator.wikimedia.org/T228464

### DIFF
--- a/app/src/main/java/org/wikipedia/login/LoginActivity.java
+++ b/app/src/main/java/org/wikipedia/login/LoginActivity.java
@@ -216,8 +216,7 @@ public class LoginActivity extends BaseActivity {
             loginClient = new LoginClient();
         }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/login/LoginActivity.java
```diff
@@ -233,11 +231,7 @@ private void doLogin() {
         return new LoginClient.LoginCallback() {
             @Override
             public void success(@NonNull LoginResult result) {
-                if (!progressDialog.isShowing()) {
-                    // no longer attached to activity!
-                    return;
-                }
-                progressDialog.dismiss();
+                showProgressBar(false);
```
dbrant: Note: this no longer checks if the activity is still alive, which might lead to unexpected behavior, including crashing. See my follow-up.

PR #494, opened 2019-07-19
Beginnings of integration with page-library Footer.

dbrant:


app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -23,4 +27,43 @@ object JavaScriptActionHandler {
...
+                "    '${model.page?.displayTitle}'," +
+                "    [" +
+                (if (showLanguagesLink) "pagelib.c1.Footer.MenuItemType.languages, " else "") +
+                (if (showEditHistoryLink) "pagelib.c1.Footer.MenuItemType.lastEdited, " else "") +
+                (if (showTalkLink) "pagelib.c1.Footer.MenuItemType.talkPage, " else "") +
+                (if (showMapLink) "pagelib.c1.Footer.MenuItemType.coordinate " else "") +
+                "],   {" +
+                "        'readMoreHeading': '" + context.getString(R.string.read_more_section) + "'," +
+                "        'menuDisambiguationTitle': '" + context.getString(R.string.page_similar_titles) + "'," +
+                "        'menuLanguagesTitle': '" + context.getString(R.string.language_count_link_text, model.page?.pageProperties?.languageCount) + "'," +
```
cooltey: If the current article is available in Chinese, it will need to be updated since we automatically add either Simplified or Traditional Chinese in the "Other languages" list.

app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -31,10 +32,12 @@ object JavaScriptActionHandler {
     @JvmStatic
     fun setFooter(context: Context, model: PageViewModel): String {
         if (model.page == null) {
-            return "";
+            return ""
         }
 
-        val showLanguagesLink = model.page!!.pageProperties.languageCount > 0
+        val languageCount = L10nUtil.getUpdatedLanguageCountIfNeeded(model.page!!.title.wikiSite.languageCode(),
```
cooltey: Should take this `languageCount` to the `menuLanguagesTitle` as the correct value of language count?

dbrant: Doh!

PR #498, opened 2019-07-22
add alertdialog in base activity and implemented in activity class

ashishpujari204:


dbrant:
Thanks for the PR.  I understand the motivation for making convenience methods, but `AlertDialog.Builder` is kind of already a convenience method, and doesn't need to be simplified further.  From what I can see, this pull request adds more than 60 lines of code, with no real impact on readability or functionality.

PR #499, opened 2019-07-22
Fix: prevent possible crash after left PageFragment

cooltey:
Steps to reproduce:

1. Read any article
2. Turn on airplane mode
3. Back to explore feed


dbrant:
Note: this was already fixed in my Page Footer PR.

PR #503, opened 2019-07-24
Clean up junk data by verifying the last modified date

cooltey:
Another implementation of this ticket: https://phabricator.wikimedia.org/T209115

A simpler way of cleaning up the "old" files inside the cache folder.


dbrant:
This will be superseded by #924

PR #523, opened 2019-07-31
Remove isPreBetaRelease from Suggested edits FeedContentType

cooltey:


app/src/main/java/org/wikipedia/feed/FeedContentType.java
```diff
@@ -126,7 +125,6 @@ public FeedClient newClient(AggregatedFeedContentClient aggregatedClient, int ag
     private boolean enabled = true;
 
     private boolean perLanguage;
-    private boolean showInConfig = true;
```
sharvaniharan: We should not remove the architecture to keep/hide options. Let us only remove the prebeta flag.

sharvaniharan:
https://github.com/wikimedia/apps-android-wikipedia/pull/527

PR #525, opened 2019-08-01
Simplify and speed up the call-to-action on top of lead image.

dbrant:
The logic for showing the CTA was using the `media` endpoint, which will be deprecated soon, and is not really necessary for this purpose anyway.

### DIFF
--- a/app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
+++ b/app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
@@ -3,6 +3,7 @@ package org.wikipedia.page.leadimages;
 import android.graphics.Bitmap;
 import android.net.Uri;
 import android.text.TextUtils;
+import android.util.Pair;
 import android.view.View;
 
 import androidx.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -208,53 +208,40 @@ private void updateCallToAction() {
...
-                    for (GalleryItem item : list) {
-                        if (getPage() != null && item.getFilePage().contains(StringUtil.addUnderscores(getPage().getPageProperties().getLeadImageName()))) {
-                            galleryItem[0] = item;
-                            title[0] = item.getFilePage().equals(Service.COMMONS_URL) ? item.getTitles().getCanonical() : UriUtil.getTitleFromUrl(item.getFilePage());
-                            return MediaHelper.INSTANCE.getImageCaptions(title[0]);
-                        }
-                    }
-                    return null;
-                })
+        String imageTitle = "File:" + getPage().getPageProperties().getLeadImageName();
```
sharvaniharan: @dbrant Are we sure this will not create a problem given that some files are "File:" and some are File: "?

dbrant: Not sure what the second case means. Can you give an example?

sharvaniharan: Sorry. Not an issue. got confused with something else.

sharvaniharan:
@cooltey since you are also reviewing currently please merge if it looks good to u ...

PR #528, opened 2019-08-01
Fix: Feedback message after edit from Suggested Edits cards

sharvaniharan:
Fixed correct feedback message on caption/description published through feed cards

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/FeedFragment.java
+++ b/app/src/main/java/org/wikipedia/feed/FeedFragment.java
@@ -235,14 +235,16 @@ public class FeedFragment extends Fragment implements BackPressedHandler {
                 || requestCode == ACTIVITY_REQUEST_ADD_A_LANGUAGE) {
             refresh();
         } else if (requestCode == ACTIVITY_REQUEST_DESCRIPTION_EDIT && resultCode == RESULT_OK) {
-            boolean isTranslation = false;
+            boolean isTranslation;
             if (suggestedEditsCardView != null) {
                 suggestedEditsCardView.refreshCardContent();
                 isTranslation = suggestedEditsCardView.isTranslation();
+                if (suggestedEditsCardView.getCard() != null) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/feed/FeedFragment.java
```diff
@@ -235,14 +235,16 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
             refresh();
         } else if (requestCode == ACTIVITY_REQUEST_DESCRIPTION_EDIT && resultCode == RESULT_OK) {
-            boolean isTranslation = false;
+            boolean isTranslation;
             if (suggestedEditsCardView != null) {
                 suggestedEditsCardView.refreshCardContent();
                 isTranslation = suggestedEditsCardView.isTranslation();
+                if (suggestedEditsCardView.getCard() != null) {
+                    FeedbackUtil.showMessage(this, isTranslation && app.language().getAppLanguageCodes().size() > 1
+                            ? (suggestedEditsCardView.getCard().getInvokeSource() == FEED_CARD_SUGGESTED_EDITS_TRANSLATE_DESC) ? getString(R.string.description_edit_success_saved_in_lang_snackbar, app.language().getAppLanguageLocalizedName(app.language().getAppLanguageCodes().get(1))) : getString(R.string.description_edit_success_saved_image_caption_in_lang_snackbar, app.language().getAppLanguageLocalizedName(app.language().getAppLanguageCodes().get(1)))
```
cooltey: Maybe we can have the `getString()` contains the logic of checking `InvokeSource` to reduce the repeated `app.language().getAppLanguageLocalizedName(app.language().getAppLanguageCodes().get(1)`

sharvaniharan: oh i tried that ... osme have arguement and some don't ... go ttricky

PR #529, opened 2019-08-01
Switch over to new MediaList endpoint, and stop using Media endpoint.

dbrant:


app/src/main/java/org/wikipedia/gallery/GalleryActivity.java
```diff
@@ -534,25 +533,24 @@ private void loadGalleryContent() {
...
         galleryPager.setPageTransformer(false, null);
         // first, verify that the collection contains the item that the user
         // initially requested, if we have one...
         int initialImagePos = -1;
         if (initialFilename != null) {
-            for (GalleryItem item : list) {
+            for (MediaListItem item : list) {
                 // sometimes the namespace of a file would be in different languages rather than English.
-                String title = StringUtil.removeNamespace(item.getTitles().getCanonical());
-                String titleFromFilePageUrl = StringUtil.removeNamespace(addUnderscores(UriUtil.getTitleFromUrl(item.getFilePage())));
```
cooltey: Not sure how to fix the issue we resolved recently in #517 since there's no `file_page` from the response. The issue will be reproducible now.

app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.java
```diff
@@ -71,46 +71,42 @@
...
+        mediaListItem = (MediaListItem) getArguments().getSerializable(ARG_GALLERY_ITEM);
         pageTitle = getArguments().getParcelable(ARG_PAGETITLE);
         imageTitle = new PageTitle(Namespace.FILE.toLegacyString(),
-                StringUtil.removeNamespace(galleryItem.getTitles().getCanonical()),
-                new WikiSite(galleryItem.getFilePage()));
-
-        if (getArguments().getBoolean(ARG_FEED_FEATURED_IMAGE)) {
-            age = getArguments().getInt(ARG_AGE);
-        }
+                StringUtil.removeNamespace(mediaListItem.getTitle()),
```
cooltey: Gallery will crash if open the featured image from the explore feed.

app/src/main/java/org/wikipedia/gallery/GalleryActivity.java
dbrant: I'm not able to reproduce the issue. Can you?

cooltey: Although the description and other image information can be displayed successfully, the response from `https://en.wikipedia.org/api/rest_v1/page/media/Badminton_at_the_1992_Summer_Olympics` contains only one file.

If you swipe to the next image, you will see exactly the same image as the previous one.

dbrant: In that case, the `media-list` endpoint should be updated to resolve redirects, and no longer give us redirected titles. We should not be doing the work of matching up redirects.
If this is the only issue remaining, then we can live with it for now.

cooltey: I see. Will create a ticket for that issue.

PR #536, opened 2019-08-02
Update button view type from TextView to Button

cooltey:
https://phabricator.wikimedia.org/T229707


app/src/main/java/org/wikipedia/captcha/CaptchaHandler.java
```diff
@@ -58,7 +59,7 @@ public CaptchaHandler(final Activity activity, final WikiSite wiki, final View p
         captchaImage = activity.findViewById(R.id.captcha_image);
         captchaText = ((TextInputLayout) activity.findViewById(R.id.captcha_text)).getEditText();
         captchaProgress = activity.findViewById(R.id.captcha_image_progress);
-        TextView submitButton = activity.findViewById(R.id.captcha_submit_button);
+        Button submitButton = activity.findViewById(R.id.captcha_submit_button);
```
dbrant: Did you know:  `Button` is just a subclass of `TextView`, so technically in our Java code we don't even need to change our types to `Button` (we can leave them as `TextView`), since we're just using them to set the text and/or visibility.

cooltey: Yes, I know, and I just want to make it consistent. If you have a strong feeling that we should leave as they are, I can revert the changes.

PR #539, opened 2019-08-06
Mobile-html: integrate with Table of Contents.

dbrant:


cooltey:
Found some issues:
1. 
https://en.wikipedia.org/wiki/Siberian_accentor

If clicks on "Notes", the lead image will overlap the article content
Screenshot_1565138412


2. Click on "About this article" does not reach to the bottom.

PR #542, opened 2019-08-06
Mobile html: Text selection

sharvaniharan:
Phab: https://phabricator.wikimedia.org/T219995

### DIFF
--- a/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
+++ b/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
@@ -28,6 +28,11 @@ object JavaScriptActionHandler {
     }
 
     @JvmStatic
+    fun getTextSelection(): String {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -28,6 +28,10 @@ object JavaScriptActionHandler {
     }
 
     @JvmStatic
+    fun getTextSelection(): String {
+        return String.format("pagelib.c1.InteractionHandling.getSelectionInfo()")
+    }
+
     fun getOffsets(): String {
```
cooltey: Looks like you removed the `@JvmStatic` by accident.

dbrant: I have updated the PR.

cooltey: Thanks!

PR #544, opened 2019-08-07
Mobile html: Attempt to Render page  better

sharvaniharan:
- When we switch tabs, the previous article used to be seen until the new content loaded. This is taken care of by hiding-unhiding the webview
- Webview unhiding is after the margins are set, hence removes the flicker


sharvaniharan:
Updated description @cooltey

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -269,8 +269,10 @@ private void updateCallToAction() {
     }
 
     private void initArticleHeaderView() {
-        pageHeaderView.addOnLayoutChangeListener((View v, int left, int top, int right, int bottom,
-                                                  int oldLeft, int oldTop, int oldRight, int oldBottom) -> setWebViewPaddingTop());
+        pageHeaderView.addOnLayoutChangeListener((v, left, top, right, bottom, oldLeft, oldTop, oldRight, oldBottom) -> {
+            LeadImagesHandler.this.setWebViewPaddingTop();
+            parentFragment.getWebView().setVisibility(VISIBLE);
```
dbrant: Would it be more logical to put this in `onPageFinished()` in PageFragment?

app/src/main/java/org/wikipedia/page/PageActivity.java
```diff
@@ -681,9 +681,11 @@ public void onActivityResult(int requestCode, int resultCode, final Intent data)
                 return;
             }
             if (resultCode == TabActivity.RESULT_NEW_TAB) {
+                pageFragment.getWebView().setVisibility(View.GONE);
                 loadMainPage(TabPosition.NEW_TAB_FOREGROUND);
                 animateTabsButton();
             } else if (resultCode == TabActivity.RESULT_LOAD_FROM_BACKSTACK) {
+                pageFragment.getWebView().setVisibility(View.GONE);
```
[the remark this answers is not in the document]
sharvaniharan: Done. I just thought we need to hide only during atab switch and had put it there, but this makes sense for any unforeseen case in general..have made this change.

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
sharvaniharan: Changes done @dbrant

PR #552, opened 2019-08-09
New performance measure

sharvaniharan:
Comparison of load time

Old measurements taken from PR by @joewalsh : https://github.com/wikimedia/apps-android-wikipedia/commit/77861cf6f8148c906ec46b926c75387b8d3eee14

|Detail|Old Performance|mobile-html Performance|
|-------|-----------------|-------------|     
Page with image|~200ms|~600ms|
Page without image|~140ms|~400ms|
Re-open same page|~200ms|~150ms|
From existing tab|~160ms|~200ms|
 
**Note**: The page is painted a split second later during old measurements and mobile-html is fully painted

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageFragment.java
+++ b/app/src/main/java/org/wikipedia/page/PageFragment.java
@@ -10,6 +10,7 @@ import android.graphics.Bitmap;
 import android.net.Uri;
 import android.os.Bundle;
 import android.text.TextUtils;
+import android.util.Log;
 import android.view.ActionMode;
 import android.view.LayoutInflater;
 import android.view.Menu;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -422,6 +422,7 @@ private void initWebViewListeners() {
             @Override
             public void onPageFinished(WebView view, String url) {
                 super.onPageFinished(view, url);
+                L.d("ttfp: " + activeTimer.getElapsedMillis());
```
dbrant: At this stage, the page is _nowhere near_ finished loading. Right after this line, we call the extremely heavy `JavaScriptActionHandler.setUp()` function, which rejiggers the entire DOM and causes multiple re-renders.  There is also the `setMargin()` call, which gets called even later, and causes the whole page to jump down below the lead image. We have to somehow stop the timer when _all of that_ is complete.

joewalsh:
> Note: The page is painted a split second later during old measurements and mobile-html is fully painted 

I noticed this as well - feel free to adjust the measurement on my branch to reflect when the actual paint occurs.

PR #558, opened 2019-08-10
Show the ErrorView if received a bad response from the endpoint

cooltey:
Send a new `WebResourceResponse` with a bad code `404` in the `catch` in `shouldInterceptRequest`, and then use the response to touch the `onReceivedError()`

This PR also add the logic of making `webView.setVisibility(VISIBLE)` since it remains `INVISIBLE` when returning from an `error View`.


app/src/main/java/org/wikipedia/dataclient/okhttp/OkHttpWebViewClient.java
```diff
@@ -45,24 +45,27 @@
...
+                response =  new WebResourceResponse(rsp.body().contentType().type() + "/" + rsp.body().contentType().subtype(),
+                            rsp.body().contentType().charset(Charset.defaultCharset()).name(),
+                            rsp.code(),
+                            StringUtils.defaultIfBlank(rsp.message(), "Unknown error"),
+                            toMap(addResponseHeaders(rsp.headers())),
+                            getInputStream(rsp));
             }
         } catch (Exception e) {
+            // TODO: we can send actual error message by handling the exception message.
+            response = new WebResourceResponse(null, null, 404, "Unknown error", null, null);
```
sharvaniharan: @cooltey Why not use the same error code and error message to create this response, as we have above?

PR #573, opened 2019-08-17
Optimize and improve vector drawables

mahozad:
<strong>launcher_foreground.xml:</strong>
- Remove bad paths and fix distortions in icon (shown in red color)
Screenshot (588)

- Use clip-path instead of a custom white circle (so the icon correctly shows the gradient in `launcher_background.xml`) (changes highlighted in green color below)
Layer 0

Final result:
Layer 0

<hr>

<strong>feed_header_wordmark.xml:</strong> Fix bottom of the letters cropped, Remove unneeded (duplicate) paths

Before:
before
After:
after

<hr>

<strong>wp_wordmark.svg:</strong> Remove a redundant path (shown in red color below):

Screenshot

<hr>

Make all drawable files [end with a newline](https://stackoverflow.com/q/729692/8583692)

<hr>

Use reference  colors in drawables where possible


cooltey:
Thanks, this PR looks good to me. 

One thing I have concern is that uses the reference colors in drawable, what if we change the reference colors in the future? Although it might be a low possibility to change the color.

mahozad:
> One thing I have concern is that uses the reference colors in drawable, what if we change the reference colors in the future? Although it might be a low possibility to change the color.

I don't know how to revert that commit in PR but I think it has no harm in it. If very minor changes are introduced to the colors in the future, the icons adapt automatically :)

dbrant:
Trimmed away stuff that doesn't matter, and now ready to merge.

PR #575, opened 2019-08-19
Prevent possible crash on videoView in GalleryItemFragment

cooltey:
https://appcenter.ms/orgs/ci-apps-hockeyapp-f5mf/apps/Wikipedia/crashes/errors/4256860024u/reports

### DIFF
--- a/app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.java
+++ b/app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.java
@@ -209,6 +209,9 @@ public class GalleryItemFragment extends Fragment {
      * @param pagerPosition    The pager's current position that is displayed to the user.
      */
     public void onUpdatePosition(int fragmentPosition, int pagerPosition) {
+        if (!isAdded()) {
+            return;
+        }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.java
```diff
@@ -209,6 +209,9 @@ public boolean onOptionsItemSelected(MenuItem item) {
      * @param pagerPosition    The pager's current position that is displayed to the user.
      */
     public void onUpdatePosition(int fragmentPosition, int pagerPosition) {
+        if (!isAdded()) {
```
dbrant: This might fix this particular crash, but it doesn't address the root of the problem, which is that in `GalleryActivity`, the `imageCaptionDisposable` never gets disposed.  We should probably dispose it in `onDestroy`.

cooltey: Makes sense and done.

I added the `imageCaptionDisposable.dispose();`, please let me know if you think this logic of checking  `isAdded()` should be removed.

PR #576, opened 2019-08-20
Hugely optimize page loading (learnings from mobile-html profiling).

dbrant:
This PR includes numerous significant optimizations for loading page content in our WebView, especially when navigating from one page to another.

- This removes *all* flicker when loading a new page on top of the current one.
- This also removes the flicker of loading the new lead image over the current one.
- And it also optimizes the internals of the Javascript that loads and transforms the actual section content when sending it to the WebView.

This makes our page loading logic truly optimal.


dbrant:
Whoops, will fix in a bit.

dbrant:
@cooltey  Should be fixed now.

cooltey:
@dbrant 
Found other issues:
1. In any Chinese article, scroll to the bottom and tap any non-cached article, the page will become blank (or, the bottom view was removed) when loading.
2. The offline mode does not work. 
    - Read any article -> turn on Airplane mode -> refresh the page.

dbrant:
@cooltey  I'm not able to reproduce either of those. I'll need specific instructions on which link(s) to click.

cooltey:
@dbrant 
Please check this link: https://zh.wikipedia.org/wiki/%E8%9C%9C%E8%9C%82?wprov=sfla1
and click on the second or third item in the "read more" section.

> The offline mode does not work. 
For this issue, I cleared the cache and it is no longer reproducible.

dbrant:
@cooltey  It's working for me. What device / API version are you using? What do you see in logcat?

cooltey:
@dbrant 
Please see this video: https://www.youtube.com/watch?v=KQP78SGwxME
(start from 00:20)

Looks like if the page has not been visited recently, it will need extra time to process, and that's how I reproduce the blank issue.

dbrant:
@cooltey  So the page finishes loading eventually? Is this actually specific to this PR? I was able to see this in production (I think).

cooltey:
@dbrant 
Yes, you're correct, but they show in different ways.
(production -> go to the top and load)
(this -> stay in the same position -> hide bottom view and load)

This PR looks good to me now and maybe we can do some optimization later

dbrant:
Right, when I look at the actual network traffic, the requests to `zhwiki` are actually taking as long as 5 seconds.
Feel free to merge when ready.

PR #577, opened 2019-08-21
Fix: incorrect button background in text selection toolbar

cooltey:
https://phabricator.wikimedia.org/T230606

This PR also removed unused tool links and tool tag.


app/src/main/res/values/styles.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools">
 
-    <style name="BaseAppTheme" parent="Theme.MaterialComponents.Light.NoActionBar">
+    <style name="BaseAppTheme" parent="Theme.MaterialComponents.Light.NoActionBar.Bridge">
```
dbrant: This seems like a very blunt way of fixing the issue. The `Bridge` themes inherit from the old `AppCompat` themes, which we should be trying to get away from, instead of going back to.

Can we just override the button style itself, with something like this:
```
<item name="materialButtonStyle">@style/Widget.MaterialComponents.Button.TextButton</item>`
```
(put it in the `BottomSheetDialogLight` and `BottomSheetDialogDark` styles)

cooltey: Hi @dbrant, thanks for providing the suggestion and example.

I tried to put `materialButtonStyle` item to `BottomSheetDialogLight` and `BottomSheetDialogDark`, but nothing happened. I tested it on an API 23 emulator.

dbrant: @cooltey Try also adding the same `materialButtonStyle` item to the top-level `BaseAppTheme`.  (and verify that no other buttons are affected by accident)

cooltey: Will do, thanks!

cooltey: Done by adding `materialButtonStyle` to `BaseAppTheme` and tested it on API 23, API 24 and API 28, every button looks good.

sharvaniharan:
Lgtm!

PR #582, opened 2019-08-21
Fix: show ErrorView if click on an SVG file with no other available images in Gallery

cooltey:
Steps to reproduce:
1. https://en.wikipedia.org/wiki/Playoffs
2. Click on the book SVG icon
3. Open Gallery and see an empty view

### DIFF
--- a/app/src/main/java/org/wikipedia/gallery/GalleryActivity.java
+++ b/app/src/main/java/org/wikipedia/gallery/GalleryActivity.java
@@ -546,6 +546,11 @@ public class GalleryActivity extends BaseActivity implements LinkPreviewDialog.C
     }
 
     private void applyGalleryList(@NonNull List<MediaListItem> list) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/gallery/GalleryActivity.java
```diff
@@ -546,6 +546,11 @@ private void loadGalleryContent() {
     }
 
     private void applyGalleryList(@NonNull List<MediaListItem> list) {
+        if (list.isEmpty()) {
```
dbrant: This doesn't really fix the root of the problem, which is that we're failing to load an image if it's not listed in the article's media-list endpoint.

See my other fix in #587, which should fix this issue indirectly.

cooltey: Yes, your PR works correctly!

PR #598, opened 2019-08-24
Update deprecated methods and minor code clean up

cooltey:
`getLoaderManager()` and `getSupportLoaderManager()` is going to be deprecated, and should use `LoaderManager.getInstance()` now.

Also, removed some `null` variables and added annotations.


app/src/main/java/org/wikipedia/search/RecentSearchesFragment.java
```diff
@@ -19,6 +19,7 @@
 import androidx.loader.content.CursorLoader;
 import androidx.loader.content.Loader;
 
+import org.jetbrains.annotations.NotNull;
```
dbrant: We prefer to use the `@NonNull` annotation from `androidx.annotation`, instead of the one from `org.jetbrains.annotations`.

PR #611, opened 2019-08-28
Utilize the displaytitle returned by /page/summary - SearchResult

cooltey:
https://phabricator.wikimedia.org/T227092

This PR relates to #448  since it would be better if sends requests using the `converetedTitle()`

TODO: add `displaytitle` to a new table?


cooltey:
#621 has a better solution.

PR #619, opened 2019-08-30
Minor unwanted code removal

sharvaniharan:


app/src/main/java/org/wikipedia/settings/Prefs.java
```diff
@@ -531,21 +464,6 @@ public static void setAskedForPermissionOnce(@NonNull String permission) {
         setBoolean(R.string.preference_key_permission_asked + permission, true);
     }
 
-    public static void setReadingListsCurrentUser(@Nullable String userName) {
-        setString(R.string.preference_key_reading_lists_current_user_hash,
```
cooltey: Should we remove this in `preference_keys.xml`?

sharvaniharan: you are right @cooltey ... removed..

PR #632, opened 2019-09-10
Onboard new users with snackbar and pulse animation on SE nav item

cooltey:
https://phabricator.wikimedia.org/T231444


sharvaniharan:
@cooltey should the pulse animation timed? right now, it stays indefinitely now.

cooltey:
> 
> 
> @cooltey should the pulse animation be timed to disappear with the snackbar? right now, it stays indefinitely now.

I think the pulsing animation should keep animating but the Snackbar will disappear after 10 seconds.

cooltey:
> view_nav_tab_overlay_fill

Done!

PR #640, opened 2019-09-14
Exclude languages while creating app bundles

sharvaniharan:
The app bundle splitting process splits apk according different screen densities, cpu architectures and languages. While this has not been a problem in the past,Playstore threw an error while trying to split the apk according to languages as it doesn't recognize th 'qq' flavor. This patch requests languages to be excluded during splitting, hence avoiding the problem.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -135,6 +135,12 @@ android {
             returnDefaultValues = true
         }
     }
+
+    bundle {
+        language {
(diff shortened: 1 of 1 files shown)


sharvaniharan:
@joewalsh I will self-merge this so we can roll to beta right away... but we can push again on Monday if there are any changes you feel need to be done.

PR #646, opened 2019-09-20
Fix "open new tab" does not work properly

cooltey:
https://phabricator.wikimedia.org/T232275


cooltey:
> I'm afraid this is unnecessarily complex, and doesn't fix the root of the issue. Please see #651

You're right. Closed.

PR #662, opened 2019-10-01
Differentiate timeout errors from connection errors.

dbrant:
https://phabricator.wikimedia.org/T232847


app/src/main/java/org/wikipedia/views/WikiErrorView.java
```diff
@@ -89,8 +93,9 @@ public void setError(@Nullable Throwable caught) {
     ErrorType getErrorType(@Nullable Throwable caught) {
         if (caught != null && is404(caught)) {
             return ErrorType.PAGE_MISSING;
-        }
-        if (caught != null && isOffline(caught)) {
+        } else if (caught != null && isTimeout(caught)) {
+            return ErrorType.TIMEOUT;
+        } else if (caught != null && isOffline(caught)) {
```
cooltey: Do these `caught != null` null check conditions necessary?

dbrant: nope!

PR #680, opened 2019-10-04
Handle page revision to get correct resposes from endpoints

cooltey:
https://phabricator.wikimedia.org/T228814

TODO: add `revision` when requesting ToC endpoint (if we have one)


app/src/main/java/org/wikipedia/dataclient/page/PageSummary.java
```diff
@@ -31,6 +31,7 @@
     @Nullable private Thumbnail thumbnail;
     @Nullable @SerializedName("originalimage") private Thumbnail originalImage;
     @Nullable private String lang;
+    @Nullable private String revision;
```
dbrant: Shouldn't this be a `long` instead of a `String`? (and everywhere else?)

cooltey: make sense, done!

PR #709, opened 2019-10-15
Using RxJava on announcement client and simplify code

cooltey:


app/src/main/java/org/wikipedia/feed/announcement/AnnouncementClient.java
```diff
@@ -24,72 +23,41 @@
...
     public void request(@NonNull Context context, @NonNull WikiSite wiki, int age, @NonNull Callback cb) {
         cancel();
-        call = request(ServiceFactory.getRest(wiki));
-        call.enqueue(new CallbackAdapter(cb, true));
+        disposables.add(ServiceFactory.getRest(wiki).getAnnouncements()
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe(list -> {
+                    List<Card> cards = new ArrayList<>();
+                    if (list != null) {
```
dbrant: The null check is not necessary. If the observable is successful, the result is guaranteed to be nonnull. Otherwise it will fall through to the error block.
These five lines can be replaced with a single expression-lambda.

cooltey: Thanks, you're correct!

cooltey: @dbrant Done (in case you missed this).

PR #710, opened 2019-10-16
Fix the long press issue on find in page menu item

cooltey:
Long pressing on the menu item of the find in page feature in the `EditSectionActivity` will cause a crash.


cooltey:
> 
> 
> This does not fix the _root cause_ of the crash; it simply adds more logic without understanding why the crash happens. The solution should be one line of code at the most.

Thanks for your tips, @dbrant! It actually fixes the root of the issue.

Will go to implement the long-press function in other PR.

PR #715, opened 2019-10-20
Add missing rtl languages: azb, lrc, nqo

amire80:


dbrant:
Thanks for the update!

PR #716, opened 2019-10-20
Add N'Ko and Balinese languages

amire80:


dbrant:
@amire80 Thanks for this, too.  However, a couple of notes about this one:
- XML requires apostrophes to be escaped, which is why the build is failing on this PR.
- The `languages_list.xml` file is actually auto-generated by a script that we run periodically (`scripts/generate_wiki_languages.py`).

I will close this PR (and run the script), but thanks for reminding me that we haven't run the script in a while. We should really think about adding it as a build step.

amire80:
> I will close this PR (and run the script), but thanks for reminding me that we haven't run the script in a while. We should really think about adding it as a build step.

I appreciate this reply! This was all just a naïve guess :)

Can you please update the page https://wikitech.wikimedia.org/wiki/Add_a_wiki in a way that will ensure that Wikipedias in new languages are immediately usable? I did this: https://wikitech.wikimedia.org/w/index.php?title=Add_a_wiki&type=revision&diff=1841598&oldid=1841595 . But now I see that it's not totally correct.

Thank you!

dbrant:
> Can you please update the page

Done!

PR #720, opened 2019-10-23
Rework "Updated UI for Suggested Edits screen"

dbrant:


app/src/main/java/org/wikipedia/dataclient/mwapi/UserContributions.kt
```diff
@@ -1,11 +1,12 @@
...
-    @SerializedName("title")
-    var title: String? = null
+    val user: String = ""
+        get() { return field.orEmpty() }
+    val title: String = ""
+        get() { return field.orEmpty() }
+    val timestamp: String = ""
+        get() { return field.orEmpty() }
+    val comment: String = ""
+        get() { return field.orEmpty() }
```
cooltey: It should be fine to remove the `get() { return field.orEmpty() }` since the `val` values cannot be reassigned and the default value is `empty`.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTaskView.kt
```diff
@@ -3,42 +3,34 @@ package org.wikipedia.suggestededits
...
 
     fun setUpViews(task: SuggestedEditsTask, callback: Callback?) {
-        setResourcesByDeviceSize()
         updateTranslateActionUI()
-        taskImageDetailView.setTitle(task.title!!)
-        taskImageDetailView.setDescription(task.description!!)
-        taskImageDetailView.setImageDrawable(task.imageDrawable!!)
-        taskImageDetailView.setImageTint(ResourceUtil.getThemedAttributeId(context!!, R.attr.themed_icon_color))
+        taskTitle.text = task.title!!
+        taskDescription.text = task.description!!
```
cooltey: The `!!` can be removed.

app/src/main/res/layout/view_image_title_description.xml
```diff
@@ -1,43 +1,43 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
-    android:id="@+id/imageTitleDescriptionContainer"
+    xmlns:tools="http://schemas.android.com/tools"
```
cooltey: `xmlns:app="http://schemas.android.com/apk/res-auto"` can be removed.

app/src/main/java/org/wikipedia/dataclient/mwapi/UserContributions.kt
dbrant: Your statement is not exactly right: these values can definitely be reassigned, when they are deserialized from JSON.  And if the JSON response does not contain one of these fields, how sure are you that it will receive the default `empty` value?

app/src/main/res/layout/view_image_title_description.xml
dbrant: OK, but it's very annoying the next time I want to add an `app:` attribute, and I have to hunt down the correct `xmlns` statement to put at the top of the file. Can we think about having these declarations in each XML file by default?

dbrant:
@cooltey ^ all ready!

app/src/main/res/layout/view_image_title_description.xml
cooltey: Make sense to me to have the statement by default.

PR #725, opened 2019-10-28
Integrate app with Edit Streak API.

dbrant:
Also rip out the old Targets numbers.


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -217,7 +216,9 @@ class SuggestedEditsTasksFragment : Fragment() {
                     } else if (!maybeSetDisabledStatus(80)) { // TODO: use edit quality metric from API response
 
                         editQualityStatsView.setGoodnessState((editQualityState++) % 7) // TODO: use edit quality metric from API response
-                        // TODO: apply edit streak from API response
+
+                        editStreakStatsView.setTitle(resources.getQuantityString(R.plurals.suggested_edits_edit_streak_detail_text,
+                                editorTaskCounts.editStreak, editorTaskCounts.editStreak))
```
cooltey: Just curious, when will it show `editorTaskCounts.getLastEditDate()`?

dbrant: Good call, and... done.

PR #727, opened 2019-10-29
Handle NavTab layouts visibility properly

cooltey:
We have an `EventBus` in `BaseActivity` for listening to the logout status in the background, and we should handle the logout event properly for the `NavTab` layouts in `MainFragment`

And, we also need to take care of the "log in" dialog button to update the `NavTab` layouts.


dbrant:
Excellent; you beat me to it.

PR #729, opened 2019-10-30
Show fundraising dialog on PageFragment

cooltey:
https://phabricator.wikimedia.org/T235538


cooltey:
The PR is ready for review, and please use checkout this patch on Gerrit to test: https://gerrit.wikimedia.org/r/#/c/mediawiki/services/wikifeeds/+/545071/

Note: will remove the `getRestLocal()` method after the wikifeed patch got merged.

app/src/main/java/org/wikipedia/settings/Prefs.java
```diff
@@ -802,6 +802,22 @@ public static void setShouldShowSuggestedEditsTooltip(boolean enabled) {
...
+        setBoolean(R.string.preference_key_visited_article_page, visited);
+    }
+
+    public static int fundraisingDialogShownInYear() {
+        return getInt(R.string.preference_key_fundraising_dialog_shown_in_year, 0);
+    }
+
+    public static void setFundraisingDialogShownInYear(int year) {
+        setInt(R.string.preference_key_fundraising_dialog_shown_in_year, year);
+    }
```
dbrant: This needs to be much more "general".  Calling it "fundraisingDialog" is too specific. Announcements could be made for all kinds of reasons, not just fundraising, and not just once per year.

If the goal is to remember which announcement(s) have already been shown, we need to store a list of announcement `id`s (taken from the `Announcement` object), and seeing if any of them match the requested announcement.

app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -1239,4 +1247,21 @@ void openImageInGallery(@NonNull String language) {
...
+    private void maybeShowFundraisingDialog() {
+        if (Prefs.hasVisitedArticlePage() && Prefs.fundraisingDialogShownInYear() < Calendar.getInstance().get(Calendar.YEAR)) {
+            disposables.add(ServiceFactory.getRest(getTitle().getWikiSite()).getAnnouncements()
+                    .subscribeOn(Schedulers.io())
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .subscribe(list -> {
+                        String country = GeoUtil.getGeoIPCountry();
+                        Date now = new Date();
+                        for (Announcement announcement : list.items()) {
+                            if (shouldShow(announcement, country, now) && announcement.placement().equals(FUNDRAISING_PLACEMENT_ARTICLE)) {
```
dbrant: The `FUNDRAISING_PLACEMENT_` constants should probably be renamed to `ANNOUNCEMENT_PLACEMENT_`. I can easily imagine wanting to make an announcement that doesn't have to do with fundraising.

app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -1239,4 +1247,21 @@ void openImageInGallery(@NonNull String language) {
...
+                                new AnnouncementDialog(requireActivity(), new AnnouncementCard(announcement)).show();
```
dbrant: We should probably add `setCancelable(false)` so that the user must choose one of the options in the dialog, and won't dismiss it with an accidental tap outside the dialog area.

app/src/main/java/org/wikipedia/settings/Prefs.java
cooltey: Thanks! Done.

app/src/main/java/org/wikipedia/page/PageFragment.java
cooltey: Makes sense. Done

app/src/main/java/org/wikipedia/page/PageFragment.java
cooltey: Sure!

app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -1239,4 +1245,25 @@ void openImageInGallery(@NonNull String language) {
         leadImagesHandler.openImageInGallery(language);
     }
 
+    private void maybeShowFundraisingDialog() {
```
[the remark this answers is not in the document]
cooltey: Make sense. Done.

app/src/main/java/org/wikipedia/settings/Prefs.java
```diff
@@ -802,6 +802,37 @@ public static void setShouldShowSuggestedEditsTooltip(boolean enabled) {
...
+            return emptySet;
+        }
+        //noinspection unchecked
+        Set<String> cards = GsonUnmarshaller.unmarshal(emptySet.getClass(),
+                getString(R.string.preference_key_announcement_shown_dialogs, null));
+        return cards != null ? cards : emptySet;
+    }
+
+    public static void setAnnouncementShownDialogs(@NonNull Set<String> announcementIds) {
+        setString(R.string.preference_key_announcement_shown_dialogs, GsonMarshaller.marshal(announcementIds));
```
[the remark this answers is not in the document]
cooltey: You are correct! Done.

dbrant:
👍 👍

PR #731, opened 2019-10-30
Extra-final design updates in Suggested Edits.

dbrant:


app/src/main/java/org/wikipedia/navtab/NavTabLayout.kt
```diff
@@ -37,11 +36,12 @@ class NavTabLayout constructor(context: Context, attrs: AttributeSet) : BottomNa
...
-                        labelView.setSingleLine(false)
-                        labelView.maxLines = 2
-                        labelView.ellipsize = TextUtils.TruncateAt.END
-                        labelView.textAlignment = View.TEXT_ALIGNMENT_CENTER
+                        val largeLabel = menuItemView.findViewById<TextView>(R.id.largeLabel)
+                        largeLabel.ellipsize = TextUtils.TruncateAt.END
+                        largeLabel.setPadding(0, paddingTop, 0, paddingBottom)
+                        val smallLabel = menuItemView.findViewById<TextView>(R.id.smallLabel)
+                        smallLabel.ellipsize = TextUtils.TruncateAt.END
+                        smallLabel.setPadding(0, paddingTop, 0, paddingBottom)
```
cooltey: This works fine but somehow it makes the pulsing icon mismatch the edit pencil icon.

dbrant: Done!

PR #735, opened 2019-11-04
Final (for real) design tweaks for Suggested Edits.

dbrant:
https://phabricator.wikimedia.org/T231445


sharvaniharan:
Would it be better to dismiss tooltips on click of tasks as well... when we click a few times on different tooltips it will get queued up and keep showing on SuggestedEditsCardsActivity too..

dbrant:
@sharvaniharan  Good point, and... done.

PR #744, opened 2019-11-05
[SPIKE] Investigate optimal offline format for saving articles

cooltey:
https://phabricator.wikimedia.org/T234965


cooltey:
Not a good approach for our project.

PR #747, opened 2019-11-06
Make Suggested Edit feed cards dependent on paused/disabled state.

dbrant:
This will make the Suggested Edit feed cards no longer appear if the feature is paused or disabled. Note: it may be necessary to refresh the feed for the updating to take effect.

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
+++ b/app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
@@ -15,7 +15,7 @@ import org.wikipedia.feed.model.Card
 import org.wikipedia.page.Namespace
 import org.wikipedia.page.PageTitle
 import org.wikipedia.suggestededits.SuggestedEditsSummary
-import org.wikipedia.suggestededits.SuggestedEditsTasksFragment
+import org.wikipedia.suggestededits.SuggestedEditsUserStats
 import org.wikipedia.suggestededits.provider.MissingDescriptionProvider
 import org.wikipedia.util.StringUtil
 import java.util.*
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
```diff
@@ -34,6 +36,18 @@ class SuggestedEditsFeedClient(private var invokeSource: Constants.InvokeSource)
...
         this.age = age
         cancel()
+
+        if (age == 0) {
+            // In the background, fetch the user's latest contribution stats, so that we can update whether the
+            // Suggested Edits feature is paused or disabled, the next time the feed is refreshed.
+            SuggestedEditsTasksFragment.updateStatsInBackground()
+        }
+
+        if (SuggestedEditsTasksFragment.isDisabled() || SuggestedEditsTasksFragment.maybePauseAndGetEndDate() != null) {
```
sharvaniharan: I feel this knowledge of edit quality should be kept with a utility class available from all parts of the app, like AccountUtil class, rather than the Fragment being responsible for this. Or, maybe we can have the EditorTaskCounts class do all the calculations and offer isPausedForUser() and isDisabledForUser()? Not sure if Fragment is the best place for it. Come to think of it, a AccountUtil should probably calculate all stats info?

dbrant: Sure, good point. I went ahead and created a new utility class that holds this information.  I didn't put it into `EditorTaskCounts` since that's a model class, and neither into `AccountUtil` since I feel like that class shouldn't contain any state information other than user credentials.
lmk what you think.

PR #754, opened 2019-11-08
Support the long press action for find-in-text in edit screen

cooltey:
Also, removed the `lastOccurrenceSearchFlag` since it breaks the `onFindNextLongClicked` function in edit screen. We can simply add `fragment.getWebView().findNext(false);` in `FindInWebPageActionProvider.java` to keep it works as it is.


cooltey:
> 
> 
> This breaks at least one thing when finding text in the WebView:
> 
>     * open an article, e.g. "Korean War"
> 
>     * click "find in page", and type "ko"
> 
>     * long-press the "down" arrow.
> 
>     * notice that the interface is now broken -- neither the "up" or "down" arrows are working.

Thanks and done. I've added the original logic back and added an extra variable to it.

PR #756, opened 2019-11-09
Migrate from HockeySDK to App Center SDK.

dbrant:


app/src/main/res/values-af/strings.xml
```diff
@@ -249,11 +249,7 @@
   <string name="error_page_does_not_exist">Hierdie blad bestaan nie</string>
   <string name="view_wiki_error_message_offline">Kan nie aan die internet koppel nie</string>
   <string name="reference_title">Verwysing %s</string>
-  <string name="crash_report_dialog_title">Jammer, toep het vorige keer gebom</string>
-  <string name="crash_report_dialog_dont_send_button">Slaan oor</string>
-  <string name="crash_report_dialog_always_send_button">Stuur altyd</string>
-  <string name="crash_report_dialog_send_button">Stuur</string>
-  <string name="crash_report_activity_title">Toepassingfout</string>
+    <string name="crash_report_activity_title">Toepassingfout</string>
```
sharvaniharan: IDE adds this tab space when we delete string resources in bulk... @cooltey @dbrant should we remove it manually or should we just let it be..

dbrant: This should be corrected automatically with the next TWN sync.

PR #760, opened 2019-11-13
Add proper header views to HistoryFragment

cooltey:
https://phabricator.wikimedia.org/T186758


sharvaniharan:
Lgtm!

app/src/main/java/org/wikipedia/history/HistoryFragment.java
```diff
@@ -337,137 +330,169 @@ private void restartLoader() {
 
         @Override
         public void onLoadFinished(@NonNull Loader<Cursor> cursorLoader, Cursor cursor) {
-            adapter.setCursor(cursor);
+            List<Object> list = new ArrayList<>();
+            while (cursor.moveToNext()) {
+                IndexedHistoryEntry indexedEntry = new IndexedHistoryEntry(cursor);
+                // Check the previous item, see if the times differ enough
+                // If they do, display the section header.
+                // Always do it this is the first item.
```
sharvaniharan: Very minor typo... missing "if"

cooltey: I copied it from the old logic, and...done.

cooltey:
> The "Clear browsing history" button no longer works. (click the trash button with nothing selected, select "Yes", and nothing happens.)

Thanks and done.

> The header text doesn't seem to match the design -- the font style and color are different.  However, I personally like this new style, especially because it's consistent with other places where we use header text, like language selection.  Let's ask Robin for a definitive statement.

Sure! https://phabricator.wikimedia.org/T186758#5661295

PR #778, opened 2019-11-19
Use default action mode backpress behavior on searching

cooltey:
https://phabricator.wikimedia.org/T238349

The removed overriding method was made because of this ticket: 
https://phabricator.wikimedia.org/T188569

It would be better if we follow the default behavior.


dbrant:
I wonder why we were doing it this way...?

cooltey:
> 
> 
> I wonder why we were doing it this way...?
Referring to the demo in this ticket: https://phabricator.wikimedia.org/T188569
I guess it was more solving the issue on the old design and now it's not suitable for the current design.

PR #779, opened 2019-11-19
Hide the bottom navigation tab layouts correctly

cooltey:
Since we have two layouts (nav tabs and nav tabs overlay) on the bottom, we should hide them properly. 

Using a simple animation on hiding might be a good idea.


dbrant:
This doesn't seem to work as intended. It leaves a space where the bottom navigation was:
device-2019-11-21-105833

And we probably don't need an animation when hiding it; the animation looks a little weird in combination with the keyboard popping up.

cooltey:
> 
> 
> This doesn't seem to work as intended. It leaves a space where the bottom navigation was:
> device-2019-11-21-105833
> 
> And we probably don't need an animation when hiding it; the animation looks a little weird in combination with the keyboard popping up.

Make sense, and looks like the animation breaks the expected behavior. 

Done!

PR #800, opened 2019-11-27
Handle read more links properly on mobile-html

cooltey:
Known issue: maybe not related to this PR, but if keeps clicking read more item, the pages will be broken after 2nd time clicked.


cooltey:
> 
> 
> Can we reduce this PR to fixing only the broken issues, i.e. the `convertedTitle` and link format? We can worry about proper eventlogging of the read-more items at a later time.


The current read more item URLs contain the `?event-logging-label` parameter and it will break the action when requesting the URL in `LinkPreviewDialog`. 

Screenshot_1575580353

dbrant:
Right; keep those changes, but not the new `onReadMoreLinkClicked()` logic. We can deal with eventlogging (analytics) of read-more items later.

cooltey:
> 
> 
> Right; keep those changes, but not the new `onReadMoreLinkClicked()` logic. We can deal with eventlogging (analytics) of read-more items later.

I created the `onReadMoreLinkClicked` not for the analytics, it was because that the current production app does *not* open the `LinkPreviewDialog` when clicking on read more links. If I remove the `onReadMoreLinkClicked()` and use `onInternalLinkClicked()`, it will show the `LinkPreviewDialog` everytime when clicking on the read more items.

dbrant:
> it will show the LinkPreviewDialog everytime when clicking on the read more items.
Isn't that what we want? (i.e. isn't it a bug if we don't show the link preview dialog for these items?)

cooltey:
> 
> 
> > it will show the LinkPreviewDialog everytime when clicking on the read more items.
> 
> Isn't that what we want? (i.e. isn't it a bug if we don't show the link preview dialog for these items?)

Make sense. Done!

PR #801, opened 2019-11-27
Use AndroidAppV3 to get latest announcement API content

cooltey:
The release version `302` and `303` contains the new announcement logic to not show card on feed, which should not happen.

This PR associate with the patch on Gerrit and hopefully, it can solve the issue: https://gerrit.wikimedia.org/r/#/c/mediawiki/services/wikifeeds/+/553391/


cooltey:
Close it due to a non-issue.

PR #802, opened 2019-11-28
Page load without PageLead

sharvaniharan:


app/src/main/java/org/wikipedia/dataclient/page/Protection.java
```diff
@@ -4,20 +4,23 @@
...
 
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
 
 /** Protection settings for a page */
 public class Protection {
-    @SuppressWarnings("MismatchedReadAndWriteOfArray") @NonNull private Set<String> edit = Collections.emptySet();
+    @Nullable private String edit;
+    @Nullable private String move;
```
cooltey: Is it being used?

app/src/main/java/org/wikipedia/dataclient/page/Protection.java
```diff
@@ -4,20 +4,23 @@
...
 
-    // TODO should send them all, but callers need to be updated, too, (future patch)
     @Nullable
     public String getFirstAllowedEditorRole() {
-        return edit.isEmpty() ? null : edit.iterator().next();
+        return edit == null || edit.isEmpty() ? "" : edit;
```
cooltey: Would it be better if we use `StringUtils.defaultString(edit)`?

app/src/main/java/org/wikipedia/dataclient/page/Protection.java
```diff
@@ -4,20 +4,23 @@
...
     }
 
     @NonNull
     public Set<String> getEditRoles() {
-        return Collections.unmodifiableSet(edit);
+        Set<String> roleSet = new HashSet<>();
+        roleSet.add(edit);
+        return edit != null ? Collections.unmodifiableSet(roleSet) : Collections.emptySet();
```
cooltey: The logic does not make sense to me. 

The `edit != null` logic should happen before creating any object, otherwise, it will be a redundant object.

app/src/main/java/org/wikipedia/page/Page.java
```diff
@@ -10,7 +9,7 @@
  */
 public class Page {
     @NonNull private final PageTitle title;
-    @NonNull private final List<Section> sections;
+    @NonNull private  List<Section> sections;
```
cooltey: Redundent empty space.

app/src/main/java/org/wikipedia/page/Page.java
```diff
@@ -29,6 +33,9 @@ public Page(@NonNull PageTitle title, @NonNull List<Section> sections,
         return sections;
     }
 
+    @NonNull public void setSections(List<Section> sections) {
```
cooltey: A `void` function should not have annotation, and please add `@NonNull` to the section parameter.

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
```diff
@@ -228,35 +228,29 @@ private void pageLoadFromNetwork(final ErrorCallback errorCallback) {
...
+
+                            if ((pair.first.raw().cacheResponse() != null && pair.first.raw().networkResponse() == null)
+                                    || OfflineCacheInterceptor.SAVE_HEADER_SAVE.equals(pair.first.headers().get(OfflineCacheInterceptor.SAVE_HEADER))) {
+                                showPageOfflineMessage(Objects.requireNonNull(pair.first.raw().header("date", "")));
+                            }
+                        },
+                        throwable -> {
+                            L.e("Page details network response error: ", throwable);
+                            commonSectionFetchOnCatch(throwable);
+                        }));
```
cooltey: Looks like the page will break after clicking on an in-article link.

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
```diff
@@ -272,19 +266,19 @@ private void showPageOfflineMessage(@NonNull String dateHeader) {
         }
     }
 
-    private void pageLoadLeadSectionComplete(PageLead pageLead) {
+    private void createPage(PageSummary pageSummary, MediaList mediaList) {
```
cooltey: Can this have annotations on the parameters?

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
```diff
@@ -272,19 +266,19 @@ private void showPageOfflineMessage(@NonNull String dateHeader) {
...
         if (!fragment.isAdded()) {
             return;
         }
+        List<MediaListItem> items = mediaList.getItems("image");
 
-        Page page = pageLead.toPage(model.getTitle());
-        bridge.execute(JavaScriptActionHandler.setUpEditButtons(true, !page.getPageProperties().canEdit()));
+        String leadImageUrl = "https:" + items.get(0).getImageUrl(DimenUtil.calculateLeadImageWidth());
 
+        String leadImageName = items.get(0).getTitle().replace("File:", "").trim();
```
cooltey: Would it be better if it has a null check on the `items.get(0)`?

app/src/main/java/org/wikipedia/page/PageProperties.java
```diff
@@ -27,34 +30,60 @@
...
      * FIXME: This is not a true page property, since it depends on current user.
      */
-    private final boolean canEdit;
+    private boolean canEdit;
 
     /**
      * Side note: Should later be moved out of this class but I like the similarities with
      * PageProperties(JSONObject).
      */
+    public PageProperties(PageSummary pageSummary, String leadImageName, String leandImageUrl) {
```
cooltey: Can this have annotations on the parameters?

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsItemFragment.kt
```diff
@@ -205,10 +205,11 @@ class SuggestedEditsCardsItemFragment : Fragment() {
             }
 
             else -> {
-                disposables.add(MissingDescriptionProvider.getNextArticleWithMissingDescription(WikiSite.forLanguageCode(parent().langFromCode))
-                        .subscribeOn(Schedulers.io())
-                        .observeOn(AndroidSchedulers.mainThread())
-                        .subscribe({ pageSummary ->
+                disposables.add(MissingDescriptionProvider.getNextArticleWithMissingDescription(WikiSite.forLanguageCode(parent().langFromCode))!!
+                        .subscribeOn(Schedulers.io())!!
+                        .observeOn(AndroidSchedulers.mainThread())!!
```
cooltey: Please remove the `!!`.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsItemFragment.kt
```diff
@@ -220,7 +221,7 @@ class SuggestedEditsCardsItemFragment : Fragment() {
                                     null, null, null
                             )
                             updateContents()
-                        }, { this.setErrorState(it) }))
+                        }, { this.setErrorState(it) })!!)
```
cooltey: Please remove the `!!`.

app/src/main/java/org/wikipedia/suggestededits/provider/MissingDescriptionProvider.kt
```diff
@@ -27,7 +28,7 @@ object MissingDescriptionProvider {
 
     // TODO: add a maximum-retry limit -- it's currently infinite, or until disposed.
 
-    fun getNextArticleWithMissingDescription(wiki: WikiSite): Observable<PageSummary> {
+    fun getNextArticleWithMissingDescription(wiki: WikiSite): Observable<Response<PageSummary>>? {
```
cooltey: Not sure why to put `?` on this method?

app/src/main/java/org/wikipedia/dataclient/page/Protection.java
sharvaniharan: I think we can remove the concept of checking for 'Protection' object, since that logic is now handled by the service... so will remove this in the next commit

app/src/main/java/org/wikipedia/dataclient/page/PageSummary.java
```diff
@@ -24,6 +30,8 @@
 
     @Nullable private String type;
     @Nullable private Titles titles;
+    @SuppressWarnings("unused,NullableProblems") @Required
+    @NonNull private String title;
```
dbrant: All fields should be `@Nullable` (and use `defaultString()` in getter methods), because there's no guarantee that deserialized structures will respect nullability.  The `Required` annotation doesn't work exactly like you think it does, and should eventually be removed.

dbrant: Also the `unused` suppression is already applied at the class level.

app/src/main/java/org/wikipedia/dataclient/page/PageSummary.java
```diff
@@ -32,6 +40,30 @@
     @Nullable @SerializedName("originalimage") private Thumbnail originalImage;
     @Nullable private String lang;
     private int pageid;
+    @SuppressWarnings("unused,NullableProblems") @NonNull private String displaytitle;
```
dbrant: `displayTitle` should not be used; instead we already have `titles.display`.

app/src/main/java/org/wikipedia/dataclient/page/PageSummary.java
```diff
@@ -32,6 +40,30 @@
...
+    @SuppressWarnings("unused") @Nullable private String redirected;
```
dbrant: Does the `redirected` field exist in the Summary response?

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
```diff
@@ -272,19 +266,22 @@ private void showPageOfflineMessage(@NonNull String dateHeader) {
...
             return;
         }
+        List<MediaListItem> items = mediaList.getItems("image");
 
-        Page page = pageLead.toPage(model.getTitle());
-        bridge.execute(JavaScriptActionHandler.setUpEditButtons(true, !page.getPageProperties().canEdit()));
-
+        if (items.get(0) != null) {
+            leadImageUrl = "https:" + items.get(0).getImageUrl(DimenUtil.calculateLeadImageWidth());
+            leadImageName = items.get(0).getTitle().replace("File:", "").trim();
```
dbrant: The lead image comes from `summary->thumbnail` or `summary->originalimage`.  The "first" image in the media list will almost certainly not be the lead image.

app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -440,6 +445,27 @@ public void onReceivedError(WebView view, int errorCode, String description, Str
         });
     }
 
+    private void setPageProtection() {
+        bridge.evaluate(JavaScriptActionHandler.getProtection(), value -> {
+            Protection protection = GsonUtil.getDefaultGson().fromJson(value, Protection.class);
+            if (model.getPage() != null) {
+                model.getPage().getPageProperties().setProtection(protection);
+                bridge.execute(JavaScriptActionHandler.setUpEditButtons(true, !model.getPage().getPageProperties().canEdit()));
```
dbrant: Do we actually need to execute `setUpEditButtons()` again? Since we already send the user's edit roles to the page in the `setup` callback, the page should be already aware if the user can edit the page.

sharvaniharan: Correct!

app/src/main/java/org/wikipedia/dataclient/page/PageSummary.java
sharvaniharan: No @dbrant Summary response doesn't have redirected. However, we can get rid of the logic of checking for that string since we uniformly create the PageTitle object with the canonical title.

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
sharvaniharan: @cooltey this should be fixed now... can you please confirm?

app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -792,6 +820,7 @@ public void scrollToSection(@NonNull String sectionAnchor) {
     }
 
     public void onPageLoadComplete() {
+        editHandler.setPage(model.getPage());
```
cooltey: Crash when opening internal links:

```
2019-12-20 10:43:29.789 16941-16941/org.wikipedia.dev E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.wikipedia.dev, PID: 16941
    java.lang.NullPointerException: Attempt to invoke interface method 'java.util.Iterator java.util.List.iterator()' on a null object reference
        at org.wikipedia.page.ToCHandler$ToCAdapter.setPage(ToCHandler.java:243)
        at org.wikipedia.page.ToCHandler.setupToC(ToCHandler.java:134)
        at org.wikipedia.page.PageFragment.setupToC(PageFragment.java:911)
        at org.wikipedia.page.PageFragment.onPageLoadComplete(PageFragment.java:843)
        at org.wikipedia.page.PageFragment.lambda$updateSections$5$PageFragment(PageFragment.java:467)
        at org.wikipedia.page.-$$Lambda$PageFragment$FGcCUBkp65M8A1pyqvlfx8vSEd4.onReceiveValue(Unknown Source:4)
        at uf.onResult(Unknown Source:2)
        at org.chromium.android_webview.AwContents.a(PG:906)
        at aeG.run(Unknown Source:4)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6669)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
cooltey: Please see my other comment in `PageFragment.java`

app/src/main/java/org/wikipedia/page/PageFragment.java
cooltey: Found out it is the purged issue.

cooltey:
Please fix the conflict and also the build failed. (The `SavedPageSyncService`

app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -440,6 +445,29 @@ public void onReceivedError(WebView view, int errorCode, String description, Str
...
+    private void updateSections() {
+        bridge.evaluate(JavaScriptActionHandler.getSections(), value -> {
+            Section[] secArray = GsonUtil.getDefaultGson().fromJson(value, Section[].class);
+            if (secArray != null) {
+                sections = Arrays.asList(secArray);
+                if (model.getPage() != null) {
+                    model.getPage().setSections(sections);
+                }
+            }
+            onPageLoadComplete();
```
cooltey: Looks like it will need to add two more items into the ToC: the article title and the "About this article".

sharvaniharan:
> Please fix the conflict and also the build failed. (The `SavedPageSyncService`

Done.

sharvaniharan:
@cooltey - There is some jankiness in the toc scroller... i think it is because of y offsets. I feel that should be handled in a separate PR.

app/src/main/java/org/wikipedia/dataclient/page/PageSummary.java
```diff
@@ -118,4 +145,32 @@ public int id() {
...
+    }
+
+    @Nullable
+    public Location getGeo() {
+        return coordinates;
+    }
+
+    @Nullable
+    public String getTimestamp() {
+        return timestamp;
```
cooltey: Would it be better if use `StringUtils.defaultString()` on the string methods?

app/src/main/java/org/wikipedia/dataclient/page/PageSummary.java
```diff
@@ -118,4 +145,32 @@ public int id() {
...
+    }
+
+    @Nullable
+    public String getWikiBaseItem() {
+        return wikiBaseItem;
```
cooltey: same

app/src/main/java/org/wikipedia/dataclient/page/Protection.java
```diff
@@ -3,21 +3,27 @@
...
+import java.util.HashSet;
 import java.util.Set;
 
 /** Protection settings for a page */
 public class Protection {
-    @SuppressWarnings("MismatchedReadAndWriteOfArray") @NonNull private Set<String> edit = Collections.emptySet();
+    @Nullable private String edit;
 
-    // TODO should send them all, but callers need to be updated, too, (future patch)
     @Nullable
```
cooltey: It should be `@NonNull` if you're using the `StringUtils.defaultString()`.

dbrant:
Looks good, but will have a follow-up patch shortly.

app/src/main/java/org/wikipedia/page/PageProperties.java
```diff
@@ -27,34 +30,60 @@
...
      * Side note: Should later be moved out of this class but I like the similarities with
      * PageProperties(JSONObject).
      */
+    public PageProperties(@NonNull PageSummary pageSummary) {
+        pageId = pageSummary.getPageId();
+        namespace = pageSummary.getNamespace();
+        revisionId = pageSummary.getRevision();
+        displayTitleText = defaultString(pageSummary.getDisplayTitle());
+        geo = pageSummary.getGeo();
+        languageCount = 1;
```
dbrant: @sharvaniharan  Were you going to update this to receive the language count from somewhere else?
(cc @joewalsh )

PR #805, opened 2019-12-03
Pre settings

sharvaniharan:


app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -38,20 +39,20 @@ object JavaScriptActionHandler {
...
+        return String.format("{" +
+                "\"platform\": \"pcs.c1.Platforms.ANDROID\"," +
+                "\"clientVersion\": \"%s\"," +
+                "\"theme\": \"%s\"," +
+                "\"dimImages\": \"%b\"," +
+                "\"margins\": { \"top\": \"%dpx\", \"right\": \"%dpx\", \"bottom\": \"%dpx\", \"left\": \"%dpx\" }," +
+                "\"areTablesInitiallyExpanded\": \"%b\"," +
+                "\"textSizeAdjustmentPercentage\": \"100%%\"," +
+                "\"loadImages\": \"%b\"" +
+                "}", BuildConfig.VERSION_NAME, app.currentTheme.funnelName,
```
cooltey: Not sure why to change the string with escaped slashes. Can we use the previous format since it looks better?

sharvaniharan: the pcsClient expects the string that is to be returned by getSetupSettings() in that format @cooltey . Earlier it was like theme:dark, now it needs to be like "theme":"dark". Hence the escaping.

app/src/main/java/org/wikipedia/bridge/CommunicationBridge.java
```diff
@@ -49,7 +49,7 @@ public CommunicationBridge(final WebView webView) {
         webView.getSettings().setAllowUniversalAccessFromFileURLs(true);
         webView.getSettings().setMediaPlaybackRequiresUserGesture(false);
         webView.setWebChromeClient(new CommunicatingChrome());
-        webView.addJavascriptInterface(new BridgeMarshaller(), "marshaller");
```
cooltey: Now the in-article link does not work as expected.

Received `Uncaught ReferenceError: marshaller is not defined`.

app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
cooltey: I see. Is it possible to use a single quote to them?

app/src/main/java/org/wikipedia/bridge/CommunicationBridge.java
sharvaniharan: Good catch! fixed.

app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
sharvaniharan: tried... but no.. only in double quotes " "

app/src/main/java/org/wikipedia/bridge/CommunicationBridge.java
```diff
@@ -49,6 +49,7 @@ public CommunicationBridge(final WebView webView) {
         webView.getSettings().setAllowUniversalAccessFromFileURLs(true);
         webView.getSettings().setMediaPlaybackRequiresUserGesture(false);
         webView.setWebChromeClient(new CommunicatingChrome());
+        webView.addJavascriptInterface(new CommunicatingChrome(), "pcsClient");
```
dbrant: Why are we creating another instance of `CommunicatingChrome` here? That's not what that class is intended for.  Couldn't we just add the new setup method to `BridgeMarshaller` and rename it from `marshaller` to `pcsClient`?

app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -430,7 +430,7 @@ public void onPageFinished(WebView view, String url) {
                 pageFragmentLoadState.onPageFinished();
                 updateProgressBar(false, true, 0);
                 webView.setVisibility(View.VISIBLE);
-                bridge.execute(JavaScriptActionHandler.setUp(leadImagesHandler.getPaddingTop()));
+                bridge.execute(JavaScriptActionHandler.setTopMargin(leadImagesHandler.getPaddingTop()));
```
dbrant: We shouldn't need to issue another call to `setTopMargin` here.  We should instead be setting it correctly in the initial `setup` call.

app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -38,20 +39,20 @@ object JavaScriptActionHandler {
...
-                "margins: { top: '%dpx', right: '%dpx', bottom: '%dpx', left: '%dpx' }," +
-                "areTablesInitiallyExpanded: %b," +
-                "textSizeAdjustmentPercentage: '100%%'," +
-                "loadImages: %b" +
-                "})", BuildConfig.VERSION_NAME, app.currentTheme.funnelName.toUpperCase(),
+        return String.format("{" +
+                "\"platform\": \"pcs.c1.Platforms.ANDROID\"," +
+                "\"clientVersion\": \"%s\"," +
+                "\"theme\": \"%s\"," +
+                "\"dimImages\": \"%b\"," +
```
dbrant: Boolean parameters do not need quotes around them. In fact, this causes them to break.

app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -38,20 +39,20 @@ object JavaScriptActionHandler {
...
+                "\"margins\": { \"top\": \"%dpx\", \"right\": \"%dpx\", \"bottom\": \"%dpx\", \"left\": \"%dpx\" }," +
+                "\"areTablesInitiallyExpanded\": \"%b\"," +
+                "\"textSizeAdjustmentPercentage\": \"100%%\"," +
+                "\"loadImages\": \"%b\"" +
```
dbrant: Same.

app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -8,13 +8,15 @@ import org.wikipedia.dataclient.RestService
...
+import org.wikipedia.util.DimenUtil.getDensityScalar
+import org.wikipedia.util.DimenUtil.leadImageHeightForDevice
 import org.wikipedia.util.L10nUtil
 import org.wikipedia.util.L10nUtil.formatDateRelative
 
 object JavaScriptActionHandler {
     @JvmStatic
     fun setHandler(): String {
-        return "pcs.c1.InteractionHandling.setInteractionHandler((interaction) => { marshaller.onReceiveMessage(JSON.stringify(interaction))})"
+        return "pcs.c1.InteractionHandling.setInteractionHandler((interaction) => { pcsClient.onReceiveMessage(JSON.stringify(interaction))})"
```
joewalsh: @sharvaniharan it should be possible to remove `setHandler()` - PCS should call `onRecieveMessage` on `pcsClient` directly without having to call `setInteractionHandler`

sharvaniharan: Done. Thanks @joewalsh

PR #806, opened 2019-12-03
Rip out MobileView, and remove RestBase differentiation.

dbrant:
This removes our remaining dependency on the `mobileview` API, and therefore also removes the necessity for the `RbSwitch` class that was used to decide whether to use `mobileview` or RestBase.

This also renames the `Rb`-named classes to regular names, e.g. `RbPageSummary` -> `PageSummary`.


app/src/main/java/org/wikipedia/dataclient/page/PageLead.java
```diff
@@ -5,22 +5,256 @@
...
 
 /**
- * Gson POJI for loading the first stage of page content.
+ * Gson POJO for loading the first stage of page content.
  */
-public interface PageLead {
-    /** Note: before using this check that #hasError is false */
-    Page toPage(PageTitle title);
+public class PageLead implements PageLeadProperties {
+    @SuppressWarnings("unused") private int ns;
```
cooltey: Would it be better if move the `@SuppressWarnings("unused")` to the top of this class?

dbrant: This class will go away when we switch over to mobile-html.

cooltey: I see.

PR #817, opened 2019-12-06
Fix table-of-contents link to "About this article" footer.

dbrant:


dbrant:
> Found that the "about this article" item on ToC does not change to a blue color if sit on the section, would that be an issue?

This might need to be fixed later, but definitely not in this PR.

PR #821, opened 2019-12-06
Implement ToC using PCS getTableOfContents() method

cooltey:
The current way of building ToC has Chinese variant issues (please see "Taiwan" in Traditional Chinese"), and it would be better if we use the PCS `getTableOfContents()` to build the ToC.

https://github.com/wikimedia/mediawiki-services-mobileapps/blob/1001bb39c42ac83ca4095f9921a3e419a3cded02/pagelib/docs/pcs/pcs.md#gettableofcontents

### DIFF
--- a/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
+++ b/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
@@ -36,6 +36,11 @@ object JavaScriptActionHandler {
     }
 
     @JvmStatic
+    fun getTableOfContent(): String {
+        return "pcs.c1.Page.getTableOfContents()"
+    }
+
+    @JvmStatic
     fun getOffsets(): String {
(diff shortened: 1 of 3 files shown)


sharvaniharan:
@cooltey I feel there are a few very nice things in this PR which I have missed on mine. I will incorporate the TOP_SECTION_ID, and updating the section heading without checking for lead. However, I would still like to have the Section Class updated, so we can have them as objects rather than parse the json object. Secondly, after the switch to PageSummary, since the pageProperties will be updated asynchronously, I am not sure that it would be a great idea to setUoToc with it here. So is it ok to close this and I will make those improvements in mine?

cooltey:
@sharvaniharan Sounds good. Just curious that does update the `Section` class affect anything on the editing? I'm seeing some classes are using the `Section` class as well.

sharvaniharan:
@cooltey EditHandler uses the same Section object ... so should be fine.

cooltey:
I see. Closed.

PR #546, opened 2019-08-08
Fetching and caching infrastructure for mobile-html

cooltey:
https://phabricator.wikimedia.org/T219994

TODO:
 - Need to re-download the articles in reading lists and clear all the existing catches.

PR #828, opened 2019-12-10
Add srcset data class for media-list endpoint

cooltey:
The `page/media-list` has the `srcset`, which provides the image URLs and their scales. It will be used in the #546 when downloading images for offline use.

### DIFF
--- a/app/src/main/java/org/wikipedia/gallery/MediaListItem.java
+++ b/app/src/main/java/org/wikipedia/gallery/MediaListItem.java
@@ -55,14 +55,14 @@ public class MediaListItem implements Serializable {
         @Nullable private String src;
         @Nullable private String scale;
 
-        @Nullable
+        @NonNull
         public String getSrc() {
-            return src;
+            return StringUtils.defaultString(src);
         }
 
-        @Nullable
+        @NonNull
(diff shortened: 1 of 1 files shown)


sharvaniharan:
@cooltey this is also there in my PR for removing PageLead. I have also tied it to the device size, please take a look and lmk which one to keep

cooltey:
> 
> 
> @cooltey this is also there in my PR for removing PageLead. I have also tied it to the device size, please take a look and lmk which one to keep

Is it possible to merge this PR and then you can add the logic you've done in your PR? If so, I will be very appreciated!

 I'd like to continue the work on #546

sharvaniharan:
Sure @cooltey Can we add a provision to get url when scale is specified? and then we can merge

cooltey:
> 
> 
> Sure @cooltey Can we add a provision to get url when scale is specified? and then we can merge

Yes, I write a logic to pick the best scale of the image from a preferred size since there are so many different image sizes.

Please see: 
https://en.wikipedia.org/api/rest_v1/page/media-list/Juice_Wrld 
and 
https://en.wikipedia.org/api/rest_v1/page/media-list/Barack_Obama

PR #846, opened 2019-12-18
Fix: fade-in gallery images from Black instead of default (white).

dbrant:
If you look closely when entering the gallery, the image fades in from a flash of white color, which looks a little jarring. This patch makes the fade look perfectly smooth from the background black color.


cooltey:
Excellent!

PR #847, opened 2019-12-18
Improve link parsing in wikitext editing previews.

dbrant:
https://phabricator.wikimedia.org/T240769


app/src/main/java/org/wikipedia/views/WikiTextKeyboardView.java
```diff
@@ -117,12 +117,20 @@ public void setEditText(@NonNull PlainPasteEditText editText) {
         }
         @Nullable String title = null;
         CharSequence selection = editText.getInputConnection().getSelectedText(0);
-        if (selection != null && selection.length() > 0) {
-            title = trimPunctuation(selection.toString()).replace("[[", "").replace("]]", "");
+        String before;
+        String after;
```
cooltey: Just a minor thing that these objects can be moved into line 125.

dbrant: ok!

PR #850, opened 2019-12-19
Sending proper title object to the resetHTML()

cooltey:
**Steps to reproduce**

1. Search "Obama" and you will see the redirected result
2. The page will break after it reaching the `resetHTML()`.

### DIFF
--- a/app/src/main/java/org/wikipedia/bridge/CommunicationBridge.java
+++ b/app/src/main/java/org/wikipedia/bridge/CommunicationBridge.java
@@ -18,8 +18,7 @@ import com.google.gson.JsonObject;
 import org.apache.commons.lang3.StringUtils;
 import org.wikipedia.dataclient.RestService;
 import org.wikipedia.json.GsonUtil;
-import org.wikipedia.util.StringUtil;
-import org.wikipedia.util.UriUtil;
+import org.wikipedia.page.PageTitle;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/bridge/CommunicationBridge.java
```diff
@@ -70,7 +71,7 @@ public void onPageFinished() {
     public void resetHtml(@NonNull String wikiUrl, String title) {
         isDOMReady = false;
         pendingJSMessages.clear();
-        communicationBridgeListener.getWebView().loadUrl(wikiUrl + RestService.REST_API_PREFIX + RestService.PAGE_HTML_ENDPOINT + UriUtil.encodeURL(title));
+        communicationBridgeListener.getWebView().loadUrl(wikiUrl + RestService.REST_API_PREFIX + RestService.PAGE_HTML_ENDPOINT + UriUtil.encodeURL(StringUtil.addUnderscores(title)));
```
dbrant: Could we change the `title` parameter to be a `PageTitle`, and then use `title.getPrefixedText()` here?

cooltey: Yes! It makes sense. But I would use `title.getConvertedText()` to prevent the variant issue

PR #851, opened 2019-12-19
Add LeakCanary to credits.

Johnett:


dbrant:
I'm not sure this is necessary, since LeakCanary doesn't actually get shipped with the Release APK (it becomes a no-op).

PR #853, opened 2019-12-19
Add .ogg url support for mobile-html

cooltey:
The Barack Obama article has horn icon, which is clickable, and should open the Gallery for it.


dbrant:
The decision was to bounce the user to an external browser for this action.

PR #856, opened 2019-12-20
Kotlin/create account to kotlin

m-rahimy:
Converting 2 classes to kotlin, reduced 90 lines of code


app/build.gradle
```diff
@@ -30,6 +30,10 @@ android {
         targetCompatibility = JAVA_VERSION
     }
 
+    androidExtensions {
+        experimental = true
```
cooltey: We usually don't use an experimental extension on our app.

m-rahimy: Can you check if replacing `experimental = true` with `features = ["parcelize"]` makes errors in build or not? `@Parcelize` is not an experimental feature anymore.

cooltey: By adding the feature flag will cause a building error. 

I think it would be fine to remove the entire `androidExtensions` if it is not an experimental feature anymore.
```
androidExtensions {
     experimental = true
}
```

m-rahimy: It seems the error is caused by missing kotlin synthetic when we add `androidExtensions` and remove `experimental = true`. 

removing the entire `androidExtensions {    experimental = true }`  does not lead to failing build and `Parcelize` annotaion works now.

PR #858, opened 2019-12-20
Ensure that Suggested Edits tooltip is shown only once.

dbrant:


cooltey:
Since there was no specific note in the ticket (https://phabricator.wikimedia.org/T231444) that talks about showing the tooltip only once, I am happy to merge this PR.

PR #863, opened 2020-01-02
Fix showing of proper chooser intent when sharing link to article.

dbrant:
https://phabricator.wikimedia.org/T241792


dbrant:
OK, this is ready for review.  This seems to be the most reasonable way to handle the "sharing" chooser, without messing with the "external browser" chooser.

PR #874, opened 2020-01-06
Add pop-up list of references.

dbrant:


app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -38,6 +38,12 @@ object JavaScriptActionHandler {
         return "window.scrollTo(0, document.getElementById('pcs-footer-container-menu').offsetTop - ${DimenUtil.getNavigationBarHeight(context)});"
     }
 
+    @JvmStatic
+    fun scrollToAnchor(anchorLink: String): String {
+        val anchor = if (anchorLink.contains("#")) anchorLink.substring(anchorLink.indexOf("#") + 1) else anchorLink
+        return "var el = document.getElementById('$anchor'); window.scrollTo(0, el.offsetTop - (screen.height / 2)); setTimeout(function(){ el.style.backgroundColor='#ee0'; setTimeout(function(){ el.style.backgroundColor=null; }, 500); }, 250);"
```
cooltey: Would it be better if it has multiple lines to increase the readability?

app/src/main/java/org/wikipedia/page/references/ReferenceListDialog.kt
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    override fun onStart() {
+        super.onStart()
+        BottomSheetBehavior.from(view!!.parent as View).peekHeight = DimenUtil
+                .roundedDpToPx(DimenUtil.getDimension(R.dimen.readingListSheetPeekHeight))
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        L10nUtil.setConditionalLayoutDirection(view, callback()!!.linkHandler.getWikiSite().languageCode())
```
cooltey: `getWikiSite()` can be changed to `wikiSite`

dbrant: ok!

app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
dbrant: ok!

PR #878, opened 2020-01-07
No longer load Main Page in empty tabs.

dbrant:
The new tab will now be empty. (and I mean _really_ empty.)

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageActivity.java
+++ b/app/src/main/java/org/wikipedia/page/PageActivity.java
@@ -449,7 +449,7 @@ public class PageActivity extends BaseActivity implements PageFragment.Callback,
 
     public void loadEmptyPage(TabPosition position) {
         PageTitle title = new PageTitle(Constants.EMPTY_PAGE_TITLE, app.getWikiSite());
-        HistoryEntry historyEntry = new HistoryEntry(title, HistoryEntry.SOURCE_MAIN_PAGE);
+        HistoryEntry historyEntry = new HistoryEntry(title, HistoryEntry.SOURCE_INTERNAL_LINK);
         loadPage(title, historyEntry, position);
     }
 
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/wikipedia/page/PageActivity.java
```diff
@@ -452,12 +447,8 @@ public void loadPage(@NonNull final PageTitle title,
...
 
-    /**
-     * Go directly to the Main Page of the current Wiki, optionally allowing state loss of the
-     * fragment manager. Useful for when this function is called from an AsyncTask result.
-     */
-    public void loadMainPage(TabPosition position) {
-        PageTitle title = MainPageClient.getMainPageTitle();
+    public void loadEmptyPage(TabPosition position) {
+        PageTitle title = new PageTitle(Constants.EMPTY_PAGE_TITLE, app.getWikiSite());
         HistoryEntry historyEntry = new HistoryEntry(title, HistoryEntry.SOURCE_MAIN_PAGE);
```
sharvaniharan: We should probably have an empty source?

app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -639,28 +641,48 @@ public void loadPage(@NonNull PageTitle title, @NonNull HistoryEntry entry,
...
 
-        closePageScrollFunnel();
-        pageFragmentLoadState.load(pushBackStack);
-        scrollTriggerListener.setStagedScrollY(stagedScrollY);
-        updateBookmarkAndMenuOptions();
+
+
+
+
+        } else {
```
sharvaniharan: Minor... remove extra lines.

dbrant: whoops!

app/src/main/java/org/wikipedia/page/PageActivity.java
dbrant: I changed the source to `SOURCE_INTERNAL_LINK` so that it's slightly more consistent with what this is.  I don't know if we need a new constant for an "empty" source, since it won't be used anywhere else except here, and no other part of the app will care about the source of this entry.

PR #881, opened 2020-01-08
Fix: suggested Edits analytics not working

cooltey:
Not sure why we didn't put the `log()` inside the `successRunnable`, but it works for me.

https://phabricator.wikimedia.org/T242137


cooltey:
> 
> 
> This is not actually how this funnel is supposed to work. Hint: If you recall, the SuggestedEdits fragment used to be inside a separate activity, but then we moved it directly into MainActivity. If you look at the history of the original `SuggestedEditsTasksActivity`, you should be able to see where the `.log()` function was originally used, and where it should be used now.

Thanks for the hint. 
Not sure if we want to put the `SuggestedEditsFunnel.get().log()` in the `onDestory()` of `SuggestedEditsTasksFragment`, but please let me know where to put them to make this work correctly.

dbrant:
I think that's absolutely right.

PR #894, opened 2020-01-14
Prevent crashes when loading page

cooltey:
Sometimes the `model.getPage()` will be a `null` object, and it will cause crashes.

It is often reproducible on the Chinese articles on trending.


dbrant:
@cooltey   Hang on, this seems to have broken the Table of Contents entirely (it only shows the title and "about this article" now).  Would you mind taking a look and fixing?

cooltey:
@dbrant No problem, I'll take care of it.

PR #897, opened 2020-01-15
Reset HTML after the loading process was done

cooltey:
This PR fixes: https://github.com/wikimedia/apps-android-wikipedia/pull/894#issuecomment-574755120


app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
```diff
@@ -214,6 +214,7 @@ private void pageLoadFromNetwork(final ErrorCallback errorCallback) {
                         model.shouldSaveOffline() ? OfflineCacheInterceptor.SAVE_HEADER_SAVE : null, null, model.getTitle().getPrefixedText())
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
+                .doAfterTerminate(() -> bridge.resetHtml(model.getTitle().getWikiSite().url(), model.getTitle()))
```
dbrant: This means that we won't start loading the page until we're finished receiving the Summary response.  The whole idea was to load the HTML _in parallel_ with fetching the Summary response.  Can we still do that?

cooltey: I see. I added a callback method to update the `Section` list, please have a look.

Also, I just found that the current section ID does not match the list size, and I fixed that as well.

dbrant:
Sorry, this is now superseded by #902

PR #903, opened 2020-01-16
Introduce task for Image Tags to suggested edits screen.

dbrant:
This now contains the entire workflow, including browsing, selecting, and publishing the tags.


sharvaniharan:
Awesome patch @dbrant! just a couple minor things. 
1. There is a top padding in the design on the image. Looks like long descriptions are flowing on top of the image. Looks ok to me, but might need to be verified with Robin.
top_padding

2. The publish button should have a different design before selection : https://app.zeplin.io/project/57a120b91998d8977642a238/screen/5dfc9d816049c216fb5d7bc7 and change after to: https://app.zeplin.io/project/57a120b91998d8977642a238/screen/5dfc9d6cd78bad98eea117bc
publish_button

sharvaniharan:
3. Orientation change needs to be handled. Currently we are losing selection, and sometimes a totally fresh card is loading.

dbrant:
@sharvaniharan 

> There is a top padding in the design on the image. Looks like long descriptions are flowing on top of the image. Looks ok to me, but might need to be verified with Robin.

This is expected. See the following mock: https://zpl.io/VkzNNKq

> The publish button should have a different design before selection

Done!

> Orientation change needs to be handled. Currently we are losing selection, and sometimes a totally fresh card is loading.

Done!

dbrant:
Self-merging to make available for design review, and since remaining comments were minor, can be reviewed in subsequent patches.

PR #907, opened 2020-01-21
Pcs-to-mobilehtml converter

sharvaniharan:
Ready for review...
~~-Serving up the required js properly while offline so we do not need to do anything extra~~
~~-Special titles: have left a comment in code~~
~~-Should add logic to delete the converted files as and when the reading list pages get visited while online~~


sharvaniharan:
@dbrant if you have any time, I need help with this,  so i can focus on the other 4 changes:
 1. The execution of the conversion method is getting interrupted
and 
2. The converted html is getting displayed as a string 
converted

app/src/main/java/org/wikipedia/dataclient/okhttp/OfflineCacheInterceptor.java
```diff
@@ -50,6 +61,22 @@ public Response intercept(@NonNull Chain chain) throws IOException {
...
+            File file = new File(WikipediaApp.getInstance().getFilesDir() + "/" + CONVERTED_FILES_DIRECTORY_NAME + "/" + request.url().toString().substring(indexOfTitleDelimiter + 1));
+            if (file.exists()) {
+                FileInputStream fileInputStream = new FileInputStream(file);
+                String contentJSON = FileUtil.readFile(fileInputStream);
+                cacheCandidate = new Response.Builder()
+                        .code(STATUS_CODE_SUCCESS)
+                        .request(request)
+                        .protocol(HTTP_1_1)
+                        .message("")
+                        .body(ResponseBody.create(MEDIA_JSON, contentJSON))
```
dbrant: Hint: you're saying that the mime-type of this response is JSON. But... is it actually JSON?

sharvaniharan: yes! i was changing the 'content-type' header but i guess that was getting overwritten... thank you 😅

app/src/main/java/org/wikipedia/util/FileUtil.java
```diff
@@ -97,5 +98,30 @@ public static String bytesToUserVisibleUnit(Context context, long bytes) {
         return context.getString(R.string.size_mb, bytesToMB(bytes));
     }
 
+    static void writeToFileInDirectory(String data, String directory, String filename) {
+        File file = new File(directory, filename);
+
+        FileOutputStream outputStream = null;
+        try {
+            file.createNewFile();
+            outputStream = new FileOutputStream(file, true);
```
dbrant: The `true` parameter causes the contents to be appended to the end of the file, instead of re-creating the file with the new data.  This can simply be `new FileOutputStream(file, true)`, which defaults to `false`.

sharvaniharan: Done.

PR #909, opened 2020-01-21
Introduce snackbar for Image Tagging task.

dbrant:


dbrant:
Self-merging, because approved.

PR #914, opened 2020-01-23
Fix getting of language link count when loading pages.

dbrant:
The conversion from PageLead to PageSummary omitted the `languageCount` parameter, and set it to a hard-coded constant.


dbrant:
superseded by #916

PR #920, opened 2020-02-04
Save open tabs to list

sharvaniharan:
Save articles from tabs list to a reading list

**Phab**: https://phabricator.wikimedia.org/T224339


app/src/main/java/org/wikipedia/page/tabs/TabActivity.java
```diff
@@ -278,6 +285,16 @@ public boolean onOptionsItemSelected(MenuItem item) {
         return super.onOptionsItemSelected(item);
     }
 
+    private void saveTabsToList() {
```
cooltey: Would it be better to add the logic of checking "empty" tab when saving to list?

sharvaniharan: good one will add..

sharvaniharan: Done @cooltey

cooltey: Nice! Can we have a message if the "Save all tabs" action actually does nothing?

sharvaniharan: have asked for a copy on ticket @cooltey .. if everything else is good we can merge and then add the copy as follow-up?

cooltey: Make sense!

PR #954, opened 2020-02-18
Basic push notification support

mdholloway:
Adds initial support for push notifications via Firebase Cloud Messaging.
On startup, the app will retrieve an identifier token from Firebase, and
register or update its registration with the configured push notification
service, if any.

Setup steps:

1) Pull this patch and rebuild the project to pull in new required
   dependencies.

2) Retrieve the file at https://phabricator.wikimedia.org/F31621302 and
   save it as app/google-services.json.

3) Launch the app, navigate to the developer settings, and do the
   following:

     a) set the push service base URL to "https://pushd.wmflabs.org"

     b) toggle "enable push notifications" to true

   Then close and relaunch the app.

To send a test push notification, post a message payload to the
prototype service. Example:

  curl -X POST -d "title=Foo&msg=Bar" https://pushd.wmflabs.org/event/broadcast

NOTES:

* The google-services.json Firebase credentials assume that the app
  ID will be org.wikipedia.dev (i.e., the Dev flavor). Other build
  flavors will not likely work with these credentials.

* A README describing the full push service API is available at
  https://github.com/mdholloway/pushd/blob/master/README.md. Please note
  that this is currently only a prototype, and implementation details are
  very likely to change.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,4 +1,5 @@
 apply plugin: 'com.android.application'
+apply plugin: 'com.google.gms.google-services'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
 
@@ -174,6 +175,8 @@ dependencies {
     implementation "androidx.preference:preference:1.1.0"
     implementation "androidx.recyclerview:recyclerview:1.1.0"
     implementation 'com.google.android:flexbox:2.0.1'
+    implementation 'com.google.firebase:firebase-messaging:20.1.0'
+    implementation 'com.google.firebase:firebase-analytics:17.2.2'
 
     implementation ('com.github.michael-rapp:chrome-like-tab-switcher:0.4.6') {
         exclude group: 'org.jetbrains'
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -341,6 +341,14 @@
             android:name=".savedpages.SavedPageSyncService"
             android:permission="android.permission.BIND_JOB_SERVICE" />
 
+        <service
+            android:name=".firebase.WikipediaAppFirebaseMessagingService"
+            android:exported="false">
+            <intent-filter>
+                <action android:name="com.google.firebase.MESSAGING_EVENT" />
+            </intent-filter>
+        </service>
+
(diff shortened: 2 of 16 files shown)


mdholloway:
Looks like google-services.json is needed to build the project, which is causing CI to fail.  It's not technically a secret but also not really meant to be committed to version control (since third party users, project forkers, etc. will want their own creds).  I'll leave it failing for now.

dbrant:
Very, very interesting!
How will this work on devices that don't have Google services installed?

mdholloway:
That's an excellent question.  It looks like it's technically possible to run a service to receive push notification messages from an arbitrary source, but [that's been made difficult to do well](https://hub.packtpub.com/android-8-forces-foss-apps-to-use-firebase-for-push-notifications-or-label-them-as-using-too-much-battery/) in recent Android releases.  (Hmm, I wonder how the system is determining that a message is coming from Firebase...)

For now, I wonder whether we could handle it as simply as saying that Google services are required in order to receive push notifications, or if the app would crash on attempt to create the background service.

Otherwise, I don't know that we have any good options in this regard.  What do you think?

mdholloway:
FWIW, the app doesn't crash on an emulator without Google Play services.

berndsi:
Michael, I changed the order of applying developer settings in the above instructions because the app crashed if the push service base URL was unset.

mdholloway:
@berndsi Thanks! The instructions before your fix didn't crash my Nexus 6P, but did crash an emulator for me just now.

dbrant:
I'm catching up on the infrastructure of how push notifications actually work... But I'm wondering in what sense this is actually using Firebase. (i.e. if we're using our own server for the notifications, then what is it that Firebase is doing for us?)

mdholloway:
It's actually Firebase (or, in the case of iOS, Apple Push Notification Service (APNs)) that pushes notification messages to app users.  The service that the app is communicating with here is for registering push subscribers, managing subscribers' subscriptions to topics, and forwarding messages as approprate to FCM or APNs to be pushed on to clients.  It doesn't push notifications to clients directly.

mdholloway:
Closing this early integration prototype in favor of the demo client to be created: https://phabricator.wikimedia.org/T252690

PR #992, opened 2020-03-02
Feed card for Image Tags

sharvaniharan:
https://phabricator.wikimedia.org/T244866


app/src/main/AndroidManifest.xml
```diff
@@ -240,6 +240,10 @@
         <activity android:name=".onboarding.SuggestedEditsOnboardingActivity" />
 
         <activity android:name=".suggestededits.SuggestedEditsImageTagsOnboardingActivity" />
+        <activity
+            android:name=".suggestededits.SuggestedEditsFeedCardImageTagActivity"
+            android:launchMode="singleTask"
```
dbrant: What's the reasoning for making this `singleTask`?

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
```diff
@@ -32,6 +33,7 @@ class SuggestedEditsFeedClient(private var action: DescriptionEditActivity.Actio
     private var targetSummary: SuggestedEditsSummary? = null
     private val langFromCode: String = app.language().appLanguageCode
     private val langToCode: String = if (app.language().appLanguageCodes.size == 1) "" else app.language().appLanguageCodes[1]
+    private var page: MwQueryPage? = null
```
dbrant: Does `page` need to be a member variable here? Couldn't it just be a parameter of `toSuggestedEditsCard()`?

app/src/main/AndroidManifest.xml
sharvaniharan: Done

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
sharvaniharan: Done

sharvaniharan:
Changes done @dbrant

PR #995, opened 2020-03-03
Hide HTML tags from title in Share-a-fact

berndsi:
Some display titles include HTML tags. This patch ensures that
(1) the visible title does not show HTML tags anymore and
(2) the file name used for sharing an image does not include HTML tags.

Bug: https://phabricator.wikimedia.org/T246847

### DIFF
--- a/app/src/main/java/org/wikipedia/page/shareafact/ShareHandler.java
+++ b/app/src/main/java/org/wikipedia/page/shareafact/ShareHandler.java
@@ -111,7 +111,7 @@ public class ShareHandler {
                 .subscribe(imageLicense -> {
                     final Bitmap snippetBitmap = SnippetImage.getSnippetImage(fragment.requireContext(),
                             fragment.getLeadImageBitmap(),
-                            title.getDisplayText(),
+                            StringUtil.fromHtml(title.getDisplayText()),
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/page/shareafact/ShareHandler.java
```diff
@@ -250,10 +250,10 @@ public boolean onMenuItemClick(MenuItem item) {
             rootView.findViewById(R.id.share_as_image_button)
                     .setOnClickListener((v) -> {
                         String introText = context.getString(R.string.snippet_share_intro,
-                                title.getDisplayText(),
+                                StringUtil.fromHtml(title.getDisplayText()),
                                 UriUtil.getUrlWithProvenance(context, title, R.string.prov_share_image));
-                        ShareUtil.shareImage(context, resultBitmap, title.getDisplayText(),
-                                title.getDisplayText(), introText);
+                        ShareUtil.shareImage(context, resultBitmap, title.getText(),
```
berndsi: Using `getText()` instead of `getDisplayText()` fixes the "interpreted" issue shown in the bug report.

dbrant:
Nice! thanks.

PR #1001, opened 2020-03-04
Fix issue(s) with sharing URLs to external apps.

dbrant:
https://phabricator.wikimedia.org/T246514

This should fix the "share" issues that users have been reporting, at a slight cost: the "Wikipedia" app will be available as one of the choices for sharing, even though we want to share a URL with an external app.


sharvaniharan:
Seems like it is adding duplicates @dbrant :

now
before

dbrant:
@sharvaniharan Shoot, how about now??

PR #1007, opened 2020-03-07
Follow-up: Create functionality for adding arbitrary tags

sharvaniharan:
-Add string resource
-Make dialog cancelable


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageTagsFragment.kt
```diff
@@ -186,7 +186,7 @@ class SuggestedEditsImageTagsFragment : SuggestedEditsItemFragment(), CompoundBu
 
     private fun addChip(label: MwQueryPage.ImageLabel?, typeface: Typeface) {
         val chip = Chip(requireContext())
-        chip.text = label?.label ?: "Add tag..."
+        chip.text = label?.label ?: getString(R.string.suggested_edits_image_tags_add_tag)
```
dbrant: whoops! good catch.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageTagDialog.kt
```diff
@@ -180,6 +180,7 @@ class SuggestedEditsImageTagDialog : DialogFragment() {
             args.putBoolean("useClipboardText", useClipboardText)
             args.putString("lastText", lastText)
             dialog.arguments = args
+            dialog.isCancelable = true
```
dbrant: isn't this the default?

sharvaniharan: i think it isnt ... i was not able to cancel the dialog just by clicking outside of it..

dbrant: Not sure how that's possible (it's not the case on any of my devices). Cancelable is initialized to `true` in the source code of `Dialog`. And the [documentation](https://developer.android.com/reference/android/app/AlertDialog.Builder#setCancelable(boolean)) says that the default is `true`.

sharvaniharan: ah! i thought 'false' is the default... but makes sense... i guess the click outside the dialog was not working only at some spots: https://youtu.be/v3ink7P8vy4 but setting it to true does not resolve it.. but we can ignore as it is minor.

PR #1008, opened 2020-03-09
Localisation updates from https://translatewiki.net.

sharvaniharan:


app/src/main/res/values-diq/strings.xml
```diff
@@ -422,7 +422,6 @@
   <string name="description_edit_text_hint">Şınasiya meqaley</string>
   <string name="description_edit_article_description_label">Meqale</string>
   <string name="description_edit_add_caption_label">Şınastiya resımi</string>
-  <string name="description_edit_add_caption_hint" fuzzy="true">Bınnuştey resimi (%s)</string>
```
dbrant: From a previous review:
@sharvaniharan It's not useful to delete a resource from localized translations, because it will simply come back when the next sync happens. Instead, log on to TWN and actually edit the translation and fix the offending content. (In this case it's clearly the extraneous "(%s)" parameter, which you can simply edit out.)

After you edit it on TWN, paste the edited content here, and it will be in sync again.

sharvaniharan: @dbrant I remembered that but for looks like I do not have permission...

Screen Shot 2020-03-09 at 2 23 54 PM

https://translatewiki.net/wiki/Project:Users says staff members can, but not sure what happened there. I will delete %s here for now, and will see how I can resolve my account issues..

PR #1025, opened 2020-03-16
Using 'Prepare to scroll' for reference click

sharvaniharan:
https://gerrit.wikimedia.org/r/#/c/mediawiki/services/mobileapps/+/579537/4/pagelib/docs/pcs/pcs.md

Use https://apps.wmflabs.org/en.wikipedia.org/v1/page/mobile-html/Admiral_Akbar for testing.

Note: We will not need to use the /page/references endpoint anymore. Will remove the usage of it in the next PR.


app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -965,25 +963,6 @@ private int getForegroundTabPosition() {
     private void setupMessageHandlers() {
         linkHandler = new LinkHandler(requireActivity()) {
             @Override public void onPageLinkClicked(@NonNull String anchor, @NonNull String linkText) {
-                dismissBottomSheet();
-                if (anchor.startsWith("cite_note")) {
-                    // It's a link to another reference within the page.
```
dbrant: The purpose of this was to handle clicking on references *within* the text of a reference. How will this be handled now?  For an example, go to the [[Earth]] article, scroll to the "Name and etymology" section, tap the [n 7] reference in the first sentence, and tap the [23] reference within the reference.

sharvaniharan: This is done @dbrant . Since references endpoint will be deprecated, for references withing reference content, it only scrolls to the place where the reference is.

PR #1032, opened 2020-03-17
'Prepare to scroll' for toc

sharvaniharan:


app/src/main/java/org/wikipedia/page/ToCHandler.java
```diff
@@ -447,6 +450,7 @@ private void scrollToListSectionByOffset(int yOffset) {
             adapter.setHighlightedSection(itemToSelect);
             currentItemSelected = itemToSelect;
         }
+        JavaScriptActionHandler.prepareToScrollTo(adapter.getItem(currentItemSelected).getAnchor(), "{ highlight: false }");
```
dbrant: Do we really need to do this here at all? The smooth thumb-scroller works just fine without it.

app/src/main/java/org/wikipedia/page/ToCHandler.java
```diff
@@ -152,6 +152,7 @@ void log() {
     }
 
     void scrollToSection(@NonNull String sectionAnchor) {
+        bridge.execute(JavaScriptActionHandler.prepareToScrollTo(sectionAnchor, "{ highlight: false }"));
         for (Section section : adapter.sections) {
             if (section.getAnchor().equals(sectionAnchor)) {
                 scrollToSection(section);
```
dbrant: Do we still need this other `scrollToSection()` call, then? It sounds like we're making it scroll twice.

app/src/main/java/org/wikipedia/page/ToCHandler.java
```diff
@@ -164,6 +165,8 @@ private void scrollToSection(@Nullable Section section) {
         if (section == null) {
             return;
         }
+        bridge.execute(JavaScriptActionHandler.prepareToScrollTo(section.getAnchor(), "{ highlight: false }"));
+
         if (section.getId() == ABOUT_SECTION_ID) {
             bridge.execute(JavaScriptActionHandler.scrollToFooter(webView.getContext()));
         } else {
```
dbrant: Do we still need this other scrolling logic after this? It sounds like we're making it scroll twice.

app/src/main/java/org/wikipedia/page/ToCHandler.java
sharvaniharan: This is to expand if the section scrolled to is collapsed.... so we need it here... I agree not having to scroll twice, so have made the scroll_to_anchor listener on PageFragment respond with a scroll only to references.

app/src/main/java/org/wikipedia/page/ToCHandler.java
sharvaniharan: same

app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -1001,8 +1001,10 @@ private void setupMessageHandlers() {
             }
         });
         bridge.addListener("scroll_to_anchor", (String messageType, JsonObject payload) -> {
-            int diffY = DimenUtil.roundedDpToPx(payload.getAsJsonObject("rect").get("y").getAsFloat());
-            webView.setScrollY(webView.getScrollY() + diffY - webView.getHeight() / 2);
+            if (payload.get("anchor").getAsString().contains("cite_")) {
```
dbrant: Why are we now restricting this for only references? Why not allow this handler to scroll to sections, as well?

app/src/main/java/org/wikipedia/page/ToCHandler.java
dbrant: To clarify: the "scroll_to_anchor" handler will automatically scroll to the desired section anyway. So why do we need to do `scrollToSection()` here a second time?

app/src/main/java/org/wikipedia/page/ToCHandler.java
dbrant: To clarify: the "scroll_to_anchor" handler will automatically scroll to the desired section anyway. So why do we need to do `scrollToSection()` here a second time?

app/src/main/java/org/wikipedia/page/ToCHandler.java
```diff
@@ -446,6 +449,7 @@ private void scrollToListSectionByOffset(int yOffset) {
         if (itemToSelect != currentItemSelected) {
             adapter.setHighlightedSection(itemToSelect);
             currentItemSelected = itemToSelect;
+            bridge.execute(JavaScriptActionHandler.prepareToScrollTo(adapter.getItem(currentItemSelected).getAnchor(), "{ highlight: false }"));
```
dbrant: To clarify: is this actually necessary? This means that the "References" section will expand itself while the user is _in the middle_ of scrolling using the thumb-circle.  That sounds jarring -- is it really desirable behavior?

sharvaniharan: So... the whole reason we are even trying to use 'prepareToScrollTo' in the toc at all, was to expand sections that we were about to scroll to... but  I do see the point in it being jarring if you are just scrolling by, and the sections are just expanding unnecessarily. 

The decision to restrict "scroll_to_anchor" handler on PageFragment was somewhat tied to this...  the handler would asynchronously scroll to sections when thumb-scrolling, which was glitchy.. so in order to 'only expand but not scroll' when 'scroll_to_anchor' requests came from toc, I had to restrict the handler...

I will remove this prepareToScrollTo and remove constraints on the 'scroll_to_anchor' handler and remove extra scrollToSection calls..

sharvaniharan:
Done

dbrant:
much better, and so much simpler.

PR #1033, opened 2020-03-18
Develop/download notification

m-rahimy:
Since showing the downloading file's progress is known to be a good UX approach, we get notification visibility of the downloading image as input for the `performDownloadRequest` method.

### DIFF
--- a/app/src/main/java/org/wikipedia/gallery/MediaDownloadReceiver.java
+++ b/app/src/main/java/org/wikipedia/gallery/MediaDownloadReceiver.java
@@ -60,7 +60,8 @@ public class MediaDownloadReceiver extends BroadcastReceiver {
 
     private void performDownloadRequest(@NonNull Context context, @NonNull Uri uri,
                                         @NonNull String targetDirectoryType,
-                                        @NonNull String targetFileName, @Nullable String mimeType) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/gallery/MediaDownloadReceiver.java
```diff
@@ -38,7 +38,7 @@ public void setCallback(@Nullable Callback callback) {
     public void download(@NonNull Context context, @NonNull FeaturedImage featuredImage) {
         String filename = FileUtil.sanitizeFileName(featuredImage.title());
         String targetDirectory = Environment.DIRECTORY_PICTURES;
-        performDownloadRequest(context, Uri.parse(featuredImage.getOriginal().getSource()), targetDirectory, filename, null);
+        performDownloadRequest(context, Uri.parse(featuredImage.getOriginal().getSource()), targetDirectory, filename, null, DownloadManager.Request.VISIBILITY_VISIBLE);
```
dbrant: I'm not sure I see the benefit of this. Why wouldn't we always want to use `VISIBILITY_VISIBLE_NOTIFY_COMPLETED`, which shows the notification during downloading as well as when it's completed?  This change will no longer show the notification when it's complete.

PR #1034, opened 2020-03-18
The dark background in the splash layout implies that there would be …

m-rahimy:
Changing the dark background in the splash layout to white, implicitly showing a white-themed application at start.


dbrant:
This is by design. Our app allows the user to choose between a light and dark theme, which is persisted across launches, but the splash screen cannot detect the user's theme preference before the app is launched. Therefore we decided it would be better to have a dark splash screen in case the user has the dark theme selected, rather than having an unexpected flash of whiteness.

PR #1045, opened 2020-03-23
Remove unused override methods

cooltey:
Looks like we have `SuggestedEditsImageTagsFragment.Callback` implemented in `SuggestedEditsCardsFragment`.


sharvaniharan:
oh this is for feed cards @cooltey we have a separate activity for feed cards as they differ slightly in the ui components

cooltey:
> 
> 
> oh this is for feed cards @cooltey we have a separate activity for feed cards as they differ slightly in the ui components

Ah! right. Closed

PR #1046, opened 2020-03-24
Automatically go to the next image only if the user stays at the same image

cooltey:
This PR will check if the user has already gone to the next page during the processing or not. If yes, then automatically "next page" will be skipped.


cooltey:
> This is a good catch, but it can be done in a much simpler way.
> Hint: modify the `nextPage()` function into something like `nextPage(sourceFragment: Fragment?)` and pass the calling fragment into it. And check if the sourceFragment is the topBaseChild.

You are right! That's definitely a better way of doing it.

Done.

dbrant:
That's it!

PR #1056, opened 2020-03-25
Make Suggested edits "Next Page" button works properly

cooltey:
Looks like the previous fix #1046 breaks the "next page" button.

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
@@ -106,7 +106,7 @@ class SuggestedEditsCardsFragment : Fragment(), SuggestedEditsImageTagsFragment.
             if (nextButton.drawable is Animatable) {
                 (nextButton.drawable as Animatable).start()
             }
-            nextPage(this)
+            nextPage()
         }
         updateBackButton(0)
         addContributionButton.setOnClickListener { onSelectPage() }
@@ -156,6 +156,12 @@ class SuggestedEditsCardsFragment : Fragment(), SuggestedEditsImageTagsFragment.
 
     override fun getSinglePage(): MwQueryPage? {
         return null
+    }
+
+    override fun nextPageIfNeeded(sourceFragment: Fragment?) {
+        if (sourceFragment == topBaseChild()) {
+            nextPage()
+        }
     }
 
     override fun updateActionButton() {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
```diff
@@ -244,12 +250,10 @@ class SuggestedEditsCardsFragment : Fragment(), SuggestedEditsImageTagsFragment.
         updateActionButton()
     }
 
-    override fun nextPage(sourceFragment: Fragment?) {
-        if (sourceFragment == topBaseChild()) {
```
dbrant: add `|| sourceFragment == null` here.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
```diff
@@ -106,7 +106,7 @@ class SuggestedEditsCardsFragment : Fragment(), SuggestedEditsImageTagsFragment.
             if (nextButton.drawable is Animatable) {
                 (nextButton.drawable as Animatable).start()
             }
-            nextPage(this)
```
dbrant: pass `null` here

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
```diff
@@ -232,7 +238,7 @@ class SuggestedEditsCardsFragment : Fragment(), SuggestedEditsImageTagsFragment.
                         else -> getString(R.string.description_edit_success_saved_snackbar)
                     }
             )
-            nextPage(this)
```
dbrant: pass `null` here

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
cooltey: done

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
cooltey: done

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
cooltey: done

PR #1058, opened 2020-03-25
Show the disambiguation description in the article language

cooltey:
This PR also removes redundant if the statement of checking `pageSummary` since it is always to be the instance of `PageSummary `


sharvaniharan:
remove  "createLegacyExtractText(...)"

cooltey:
> 
> 
> remove "createLegacyExtractText(...)"

You are right! Done.

PR #1079, opened 2020-03-30
Switch over to Glide.

dbrant:
https://phabricator.wikimedia.org/T238886


app/src/main/java/org/wikipedia/views/FaceAndColorDetectImageView.java
```diff
@@ -1,29 +1,45 @@
...
 import static org.wikipedia.settings.Prefs.isImageDownloadEnabled;
 
-public class FaceAndColorDetectImageView extends SimpleDraweeView {
-    private RetainingDataSourceSupplier<CloseableReference<CloseableImage>> supplier;
+public class FaceAndColorDetectImageView extends AppCompatImageView {
+    public static final int PAINT_FLAGS = Paint.DITHER_FLAG | Paint.FILTER_BITMAP_FLAG;
+    private static final Paint DEFAULT_PAINT = new Paint(PAINT_FLAGS);
+    private static final int BITMAP_COPY_WIDTH = 200;
 
     public interface OnImageLoadListener {
```
cooltey: It is not showing here but should we remove the unused `onImageFailed()` method from the interface?

app/src/main/res/values/styles.xml
```diff
@@ -337,35 +337,13 @@
     </style>
 
     <style name="SimpleDraweeViewPlaceholder">
```
cooltey: Should we rename the `SimpleDrawee` to another proper name since it has been removed?

app/src/main/java/org/wikipedia/views/FaceAndColorDetectImageView.java
dbrant: ok!

app/src/main/res/values/styles.xml
dbrant: sure!

sharvaniharan:
Will review this before i finish my day today..

PR #1080, opened 2020-03-30
Design tweaks : T247248

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T247248

### DIFF
--- a/app/src/main/res/layout/group_find_references_in_page.xml
+++ b/app/src/main/res/layout/group_find_references_in_page.xml
@@ -10,9 +10,10 @@
     <TextView
         android:id="@+id/reference_label"
         android:layout_width="wrap_content"
-        android:layout_height="48dp"
-        android:gravity="center_vertical"
+        android:layout_height="match_parent"
         android:layout_marginStart="16dp"
+        android:background="?android:attr/selectableItemBackground"
+        android:gravity="center_vertical"
         android:maxLines="1"
         android:textColor="?attr/colorAccent"
         android:textSize="16sp"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/group_find_references_in_page.xml
```diff
@@ -10,8 +10,8 @@
     <TextView
         android:id="@+id/reference_label"
         android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:layout_gravity="center_vertical"
+        android:layout_height="48dp"
+        android:gravity="center_vertical"
         android:layout_marginStart="16dp"
```
cooltey: Looks like the text does not align with the count text.

Would it be better if we do something like this:
```
        android:layout_gravity="center_vertical"
        android:padding="16dp"
        android:background="?android:attr/selectableItemBackground"
```
Set padding, change height to `wrap_content`, change to the previous gravity setting and add background of clicking ripple animation.

sharvaniharan:
Thats a good catch @cooltey will add the ripple and arrange it differently

sharvaniharan:
@cooltey its done.. although i had to center it in a different way because Robin wants the click area to be large..

cooltey:
> @cooltey its done.. although i had to center it in a different way because Robin wants the click area to be large..

I think add padding to the textView should be enough to enlarge the click area, isn't it?

sharvaniharan:
the height of that parent layout itself is 48dp... so we can either just expand it to match parent and align center_vertical, what i have done... or have it wrap_content and give it a padding, and have the parent assign gravity... Is there any additional value to the second way...if there is, i will change it...

cooltey:
> 
> 
> the height of that parent layout itself is 48dp... so we can either just expand it to match parent and align center_vertical, what i have done... or have it wrap_content and give it a padding, and have the parent assign gravity... Is there any additional value to the second way...if there is, i will change it...

It makes sense. But If you look closely, you'll find that the ripple animation of the reference text actually touches to the edge of the text, which looks weird to me and that's why I would recommend adding padding to the text.

sharvaniharan:
There! @cooltey 😁

PR #1092, opened 2020-04-01
Properly integrate with Install Referrer API.

dbrant:
https://phabricator.wikimedia.org/T240378


app/src/main/java/org/wikipedia/analytics/InstallReferrerListener.java
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * To test the Google Play Install Referrer functionality:
+ *
+ * - Make sure the app is uninstalled.
+ * - Launch the Play Store via the usual specially crafted campaign link, such as:
+ *   https://play.google.com/store/apps/details?id=org.wikipedia&referrer=utm_source%3Dtest_source%26utm_medium%3Dtest_medium%26utm_term%3Dtest-term%26utm_content%3Dtest_content%26utm_campaign%3Dtest_name
+ *
+ * - ...But do NOT click the "install" button in the Play Store to install the app.
+ * - Launch or debug the app in the usual way.
+ * - The Install Referrer service should work, and will pass through the correct referrer.
```
cooltey, line 28: I've only seen once when launching the dev app. Does it happen every time after clicked the link and then launch the app?

dbrant: Ideally you should test this with the Production flavor, since that's what the Play Store knows about.

cooltey: I see! You're right!

PR #1094, opened 2020-04-01
SE V4 analytics

sharvaniharan:


app/src/main/java/org/wikipedia/analytics/SuggestedEditsFunnel.java
```diff
@@ -141,40 +148,75 @@ public void failure(Action action) {
...
 
-    public void log() {
-        log(
-                "edit_tasks", GsonUtil.getDefaultGson().toJson(statsCollection),
-                "help_opened", helpOpenedCount,
-                "scorecard_opened", contributionsOpenedCount,
-                "source", (invokeSource.getName())
-        );
+    private void removeStatsWithNoValues(JSONObject editTasksJSONObject) {
+        ArrayList<String> editTaskKeysToRemove = new ArrayList<>();
```
dbrant, line 175: I'm afraid this is way too complex for what it needs to be.
Remember that `Gson` is a very powerful and versatile library.  It would have been worthwhile to check whether it provides built-in functionality to do what we need here.  I would estimate that this is achievable with maybe 3 or 4 lines of code, plus a little boilerplate.

sharvaniharan: The TypeAdapter way is exponentially better!!!🙌

PR #1102, opened 2020-04-03
Follow-up: Switch over to Glide. 

sharvaniharan:
Looks like we dont need this package


dbrant:
We do need it.
It handles the `@GlideModule` annotation that integrates `OkHttpGlideModule` into the pipeline.  (Without it, the app builds successfully (for some reason), but the module doesn't get integrated.)

PR #1107, opened 2020-04-03
Fix: make sure that goToTab() function works properly

cooltey:
The `goToTab()` function is broken because of the change of setting up the `View.generateViewId()` for each `NavTab` menu.


cooltey:
> This is a good catch, but it's achievable with one line of code.

Not sure if the new commit matches your finding. Please let me know if it needs another change.

dbrant:
correct!

PR #1110, opened 2020-04-03
Prevent showing the PopupMenu when long pressing on reference links

cooltey:
When long-pressing on a reference link, it will trigger the `PopupMenu`, and somehow it is not quite right and will produce some weird actions such as jump to the top of the article or open the main page after clicking on the "Open in new tab" action.


cooltey:
> This is a good catch, but it's fixable with one line of code, instead of 23.
Done!

dbrant:
correct!

PR #1118, opened 2020-04-06
Reformat string files - helpful for future resource deletions

sharvaniharan:
When we delete string resources the ide always adds an extra space. This means that we have to go into each file and delete that space... If we have a resources folder that is formatted according to the ide, it will either automatically resolve that, or we can easily run a 'reformat'.


sharvaniharan:
oh! i thought they will only manipulate the most current version of those files... so I thought if not the next one, at least the next-to-next one should set it right...

PR #1119, opened 2020-04-06
Set up article Uri correctly for Chinese variants

cooltey:
When sharing page links on Chinese articles, we should give the user a correct link with proper language variant.

For example: 
https://zh.wikipedia.org/zh-hant/2019冠状病毒病西班牙疫情 for Traditional Chinese
instead of 
https://zh.wikipedia.org/wiki/2019冠状病毒病西班牙疫情 (default page that has mixed variants)


dbrant:
Looks good. How about a couple of quick unit tests in `PageTitleTests` to test these variants?

cooltey:
> Looks good. How about a couple of quick unit tests in `PageTitleTests` to test these variants?

Makes sense. Will do it!

cooltey:
> Looks good. How about a couple of quick unit tests in `PageTitleTests` to test these variants?

Done!

dbrant:
very nice

PR #1120, opened 2020-04-07
Don't override system-provided ActionMode items.

sharvaniharan:
**Phab:**https://phabricator.wikimedia.org/T248717


app/src/main/java/org/wikipedia/page/PageActivity.java
```diff
@@ -699,15 +700,32 @@ public void onDestroy() {
...
         }
         currentActionModes.add(mode);
     }
 
+    private void modifyMenu(ActionMode mode) {
+        Menu menu = mode.getMenu();
+        ArrayList<MenuItem> menuItemsList = new ArrayList<>();
+
+        for (int i = 0; i < menu.size(); i++) {
+            String title = menu.getItem(i).getTitle().toString();
```
dbrant, line 715: We actually _do_ want to remove the "Search Wikipedia" item, because we're already in the Wikipedia app.  And we can reliably check for `R.string.search_hint` because it will match in all languages.

app/src/main/java/org/wikipedia/page/PageActivity.java
```diff
@@ -699,15 +700,32 @@ public void onDestroy() {
...
+            if (!title.contains(getString(R.string.menu_text_select_share))) {
+                menuItemsList.add(menu.getItem(i));
+            }
+        }
+
+        menu.clear();
+        mode.getMenuInflater().inflate(R.menu.menu_text_select, menu);
+
+        for (MenuItem menuItem : menuItemsList) {
+            menu.add(Menu.NONE, menuItem.getItemId(), Menu.NONE, menuItem.getTitle()).setIntent((menuItem.getIntent() == null) ? null : menuItem.getIntent().addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)).setIcon(menuItem.getIcon());
```
dbrant, line 725: I figured it out for this to work completely properly:
* Set the `group` parameter to `menuItem.getGroupId()` instead of `NONE`.
* No longer need the `NEW_TASK` flag

sharvaniharan: Done. great find!

app/src/main/java/org/wikipedia/page/PageActivity.java
sharvaniharan: Done

sharvaniharan:
Charlotte has approved removal of share-a-fact

PR #1122, opened 2020-04-08
Remove swipe to hide the cards

sharvaniharan:
-Remove swipe to hide cards


dbrant:
@sharvaniharan  If you make this PR restricted to just removing the swipe-to-dismiss of the cards, we can merge it and proceed from there.

sharvaniharan:
@dbrant good idea... shall i keep the drawerlayout gradle inclusion? it will give the out-of-the-box peek behavior for android 10, which we are sure we want..

dbrant:
> @dbrant good idea... shall i keep the drawerlayout gradle inclusion? it will give the out-of-the-box peek behavior for android 10, which we are sure we want..

OK, I'm a bit nervous about depending on a non production-ready (`beta01`) component, but we can try it since it's easily interchangeable if we need to revert.

sharvaniharan:
ok. updating now..

PR #1125, opened 2020-04-08
Set up edit description/image caption input field max length to 250 characters

cooltey:
https://phabricator.wikimedia.org/T231863


dbrant:
Do we want to apply the 250-character limit to both captions and descriptions?  It seems a bit odd to have a different limit for captions than descriptions.

cooltey:
> 
> 
> Do we want to apply the 250-character limit to both captions and descriptions? It seems a bit odd to have a different limit for captions than descriptions.

Totally agree with you. Will let Charlotte know about this.

PR #1126, opened 2020-04-09
Remove singleTask from MainActivity's launchMode

cooltey:
https://phabricator.wikimedia.org/T126725

Looks like the `singleTop` is a recommended way of the `launchMode`, but some of the existing behaviors are changed because of this change. e.g.: 

feed -> article -> home button -> shortcut to search -> article -> press back -> back to feed -> back to previous article.

@dbrant @sharvaniharan  any suggestion?


dbrant:
> Looks like the `singleTop` is a recommended way of the `launchMode`

Recommended by whom?

> some of the existing behaviors are changed because of this change

The actual root of the issue is that our `MainActivity` is set to be `singleTask`, so when it's invoked from the home launcher, it's forced to relaunch the activity and clear the stack.  Can we get away with just removing `singleTask` from MainActivity, and nothing else?  What (if any) changes in behavior will result from that?

cooltey:
> > Looks like the `singleTop` is a recommended way of the `launchMode`
> 
> Recommended by whom?

My bad. It says that `singleTask` is not recommended for general use in here: https://developer.android.com/guide/topics/manifest/activity-element

> > some of the existing behaviors are changed because of this change
> 
> The actual root of the issue is that our `MainActivity` is set to be `singleTask`, so when it's invoked from the home launcher, it's forced to relaunch the activity and clear the stack. Can we get away with just removing `singleTask` from MainActivity, and nothing else? What (if any) changes in behavior will result from that?
You are correct! I don't see changes in behavior if we do that. Thanks!

cooltey:
@dbrant I cannot reproduce the unexpected behaviors I've found anymore. The patch is ready for review now. @sharvaniharan please review to see if anything goes wrong.

PR #1131, opened 2020-04-10
Follow up: further refine onboarding layouts.

dbrant:
The views were looking too lopsided towards the bottom. This centers everything in the middle of the screen.


cooltey:
I have the same feeling by looking at the non-centered images and texts, but I also concern about how the design would think. Let's merge it and see how it goes. :D

PR #1138, opened 2020-04-13
Don't allow cleartext external links from intent.

dbrant:


app/src/main/java/org/wikipedia/page/PageActivity.java
```diff
@@ -306,8 +307,13 @@ protected void onNewIntent(Intent intent) {
 
     private void handleIntent(@NonNull Intent intent) {
         if (Intent.ACTION_VIEW.equals(intent.getAction()) && intent.getData() != null) {
-            WikiSite wiki = new WikiSite(intent.getData());
-            PageTitle title = wiki.titleForUri(intent.getData());
+            Uri uri = intent.getData();
+            if (ReleaseUtil.isProdRelease() && uri.getScheme() != null && uri.getScheme().equals("http")) {
```
cooltey, line 311: Just curious that would it be possible if it starts sending `https` in the future? Should we use `startWith("http")`?

dbrant: If the URL already starts with `https`, then we don't need to change anything about it.  This is just for transforming `http` into `https`.

PR #1139, opened 2020-04-13
Use PhotoView in the Gallery for zooming images.

dbrant:
After the switchover to Glide, the zooming functionality in the Gallery was relying on our custom pinch-to-zoom function, which snaps the image back when the pinch is released. This is not desirable, since we want the image to remain zoomed after pinching.

This PR adds the `PhotoView` component, which is a very popular component precisely for zooming ImageViews..  We were actually going to use this component a long time ago, but it's not compatible with Fresco, which is why we couldn't use it.

P.S. After including this library, our method count is still about 42000, so we're still good for a while.


cooltey:
Works as expected!

PR #1143, opened 2020-04-14
Correctly pass back and handle errors from the WebView.

dbrant:


app/src/main/java/org/wikipedia/dataclient/okhttp/HttpStatusException.java
```diff
@@ -29,10 +32,10 @@ public HttpStatusException(@NonNull Response rsp) {
         }
     }
 
-    public HttpStatusException(@Nullable ServiceError error) {
-        serviceError = error;
-        code = 0;
-        url = "";
+    public HttpStatusException(int code, String url, String message) {
```
cooltey, line 35: Could we have annotations on these parameters?

dbrant: Sure!

PR #1150, opened 2020-04-15
Create full-screen swipe gesture for Nav Drawer and Table of Contents.

dbrant:
https://phabricator.wikimedia.org/T235673

* Allows you to swipe anywhere in MainActivity to bring out the drawer.
* Independent of any hacks involving Android 10 gesture exclusion zones, etc.
* Refactored the Table of Contents to be a proper DrawerLayout.
* Allows you to swipe anywhere in PageActivity to bring out the ToC.


app/src/main/java/org/wikipedia/page/ToCHandler.java
```diff
@@ -100,15 +90,20 @@ public void onReceiveValue(String value) {
...
         this.bridge = bridge;
-        this.tocContainer = tocContainer;
+        this.drawerLayout = drawerLayout;
         this.scrollerView = scrollerView;
         scrollerViewParams = new FrameLayout.LayoutParams(DimenUtil.roundedDpToPx(SCROLLER_BUTTON_SIZE), DimenUtil.roundedDpToPx(SCROLLER_BUTTON_SIZE));
 
-        tocList = tocContainer.findViewById(R.id.toc_list);
+        containerView = drawerLayout.findViewById(R.id.toc_container);
+        final int transparency = 0xe0000000;
+        final int colorMask = 0xffffff;
```
cooltey, line 103: Looks like the ToC background has a more transparent level than before?

dbrant: > 1. Swipe to dismiss feed card seems to be disabled?

This was already done last week (on purpose), based on updated designs.

> 2. When the `ActionMode` is on, which are the "search" on the reading list page, history page and "find in page" on PageFragment, the "swipe to open drawer" should be disabled?

Good observation, and... done!

> Looks like the ToC background has a more transparent level than before?

Also by design, but can be tweaked upon review.

cooltey: I see. The PR looks good to me now! Will let @sharvaniharan review the PR as well.

PR #1153, opened 2020-04-15
Add tags into image bottomsheet

cooltey:
https://phabricator.wikimedia.org/T250199

Since it is difficult to get images that have tags, and the example image below shows how it actually looks like.

Screenshot_20200415-133554_Wikipedia Dev


cooltey:
Please note that the base branch is #1154

cooltey:
Note: should still put this into HOLD since the SE image tags feature is still unclear.

PR #1173, opened 2020-04-21
Improved support for keyboard navigation, Phase 1.

dbrant:
This PR literally just adds the `focusable` flag to components that were missing it.
This is a win-win for keyboard navigation, as well as keeping the code lint-free.


cooltey:
Great! I believe it also optimizes accessibility that has mentioned here: https://phabricator.wikimedia.org/T226401

PR #1176, opened 2020-04-21
Store wikitext section into SharedPreferences temporarily to prevent a crash

cooltey:
This may fix the issue here: https://phabricator.wikimedia.org/T242449#6073882

For a really long wikitext section, the app will crash when using the app switcher and going back to the app. If we want to keep the edit states, saving them into the `SharedPreferences` might be a way to solve the `TransactionTooLargeException` error.

### DIFF
--- a/app/src/main/java/org/wikipedia/edit/EditSectionActivity.java
+++ b/app/src/main/java/org/wikipedia/edit/EditSectionActivity.java
@@ -233,12 +233,6 @@ public class EditSectionActivity extends BaseActivity {
     }
 
     @Override
-    public void onPause() {
-        super.onPause();
-        Prefs.storeTemporaryWikitext(sectionWikitext);
-    }
-
-    @Override
     public void onDestroy() {
         disposables.clear();
         captchaHandler.dispose();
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/wikipedia/edit/EditSectionActivity.java
```diff
@@ -180,8 +180,8 @@ public void onCreate(Bundle savedInstanceState) {
             funnel.logStart();
         }
 
-        if (savedInstanceState != null && savedInstanceState.containsKey("sectionWikitext")) {
-            sectionWikitext = savedInstanceState.getString("sectionWikitext");
+        if (savedInstanceState != null && savedInstanceState.containsKey("hasTemporaryWikitextStored")) {
```
banansb, line 183: Duplicate of #

banansb, line 183:
```suggestion
        if (savedInstanceState != null && savedInstanceState.containsKey("hasTemporaryWikitextStored")) {
```

app/src/main/java/org/wikipedia/edit/EditSectionActivity.java
```diff
@@ -610,7 +616,7 @@ public void finishActionMode() {
     @Override
     protected void onSaveInstanceState(Bundle outState) {
         super.onSaveInstanceState(outState);
```
dbrant, line 618: Can we save the temporary wikitext here, instead of in `onPause`?
By saving it in `onPause`, it will get saved unnecessarily, e.g. when the user simply backs out of editing.

cooltey: Sure and done!

PR #1184, opened 2020-04-30
Consolidate app theme colors and styles

cooltey:
Since we are using `paper_color` as the toolbar color, status bar color, and navigation bar color on the main page, there's no reason for us to change the rest of the screens to keep the consistency. 

This PR also combines `main_toolbar_*` and `page_toolbar_*` into `toolbar_*` attributes to reduce confusion while implementing colors to new screens.


cooltey:
The designer approved the changes, and PR is ready for review and merge.

PR #1186, opened 2020-04-30
Fix: move PCS calls to onPageMetadataLoaded to prevent incomplete page loading

cooltey:
When loading Traditional Chinese trending articles for the first, sometimes the page will not a complete status of it. You will see the ToC and footer are gone, and the protection status is also incorrect.

This PR moves the PCS calls into one function to reduce multiple bridge calls.

### DIFF
--- a/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
+++ b/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
@@ -46,7 +46,20 @@ object JavaScriptActionHandler {
 
     @JvmStatic
     fun getRevision(): String {
-        return "pcs.c1.Page.getRevision();"
+        return "pcs.c1.Page.getRevision()"
+    }
+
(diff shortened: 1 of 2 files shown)


cooltey:
> 
> 
> This may "work", but it's highly unlikely to be a totally correct solution. PCS sends us the `setup` and `final_setup` events for a reason -- it's trying to tell us when it's safe to start making additional calls to perform additional page operations. If we ignore those events, we risk corrupting the state of the page in different ways.
> 

Make sense.

> I don't see the connection of _why_ this fixes the issue. If you are seeing problems with loading articles, the issue should probably be fixed on the `mobile-html` side, not on our side.

I combined the PCS calls based on the previous comments about too many `bridge` calls, and I agree that this might be fixed on the PCS side. Will talk with Joe.

cooltey:
Found the issue might come from the summary call in `pageLoadFromNetwork()`. Will update the PR about it.

PR #1192, opened 2020-05-02
Migrate to ThreeTenABP from java.util.Date and its associated classes.

Isira-Seneviratne:
[ThreeTenABP ](https://github.com/JakeWharton/ThreeTenABP) is a port of [ThreeTenBP](https://www.threeten.org/threetenbp/apidocs/index.html), a backport of Java 8's Time API (JSR-310) for Java 6 and 7, for Android.

Using this library allows the Java 8 Time API to be used on Android without having to check the API level, and also makes migrating to the standard Java 8 Time API easier as it will be made available on versions of Android older than API level 26 [starting with Android Gradle plugin 4.0](https://developer.android.com/studio/preview/features#j8-desugar) (currently at beta 5).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -205,6 +205,7 @@ dependencies {
     implementation 'org.apache.commons:commons-lang3:3.9'
     implementation 'org.jsoup:jsoup:1.12.1'
     implementation 'com.github.chrisbanes:PhotoView:2.3.0'
+    implementation 'com.jakewharton.threetenabp:threetenabp:1.2.4'
 
     annotationProcessor "com.jakewharton:butterknife-compiler:$butterknifeVersion"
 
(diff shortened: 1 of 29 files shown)


dbrant:
This is very interesting, but the ThreeTenABP library has a README that specifically says:

> **Attention**: Development on this library is winding down. Please consider switching to Android Gradle plugin 4.0, java.time.*, and its core library desugaring feature in the coming months.

Since we'd rather not depend on libraries that are explicitly deprecated, can we just switch over to `java.time.*` when it's ready?

Isira-Seneviratne:
On Sat, May 2, 2020, 5:55 PM Dmitry Brant <notifications@github.com> wrote:

> This is very interesting, but the ThreeTenABP library has a README that
> specifically says:
>
> *Attention*: Development on this library is winding down. Please consider
> switching to Android Gradle plugin 4.0, java.time.*, and its core library
> desugaring feature in the coming months.
>
> Since we'd rather not depend on libraries that are explicitly deprecated,
> can we just switch over to java.time.* when it's ready?
>

Sure, I guess.

Earlier I made the changes using the beta plugin.

PR #1193, opened 2020-05-02
Rewrite ReadingListBehaviorsUtil.getListsContainPage() using standard Kotlin functions.

Isira-Seneviratne:
Use Kotlin's `filter` and `any` functions to simplify the filtering operation in getListsContainPage().

This makes the code more idiomatic and allows the function to be declared as a single expression function.


cooltey:
That's a great optimization, thanks for your contribution!

PR #1194, opened 2020-05-03
Use Apache Commons Collections utility methods

Isira-Seneviratne:
The Apache Commons Collections library includes utility classes that perform standard collection operations in a null-safe way and also provide methods that return an empty `Collection` if the value passed in was null.

### DIFF
--- a/app/src/main/java/org/wikipedia/readinglist/ReadingListBookmarkMenu.java
+++ b/app/src/main/java/org/wikipedia/readinglist/ReadingListBookmarkMenu.java
@@ -24,6 +24,10 @@ import io.reactivex.Completable;
 import io.reactivex.android.schedulers.AndroidSchedulers;
 import io.reactivex.schedulers.Schedulers;
 
+import static org.apache.commons.collections4.CollectionUtils.isEmpty;
+import static org.apache.commons.collections4.CollectionUtils.isNotEmpty;
+import static org.apache.commons.collections4.CollectionUtils.size;
+
 public class ReadingListBookmarkMenu {
     public interface Callback {
         void onAddRequest(@Nullable ReadingListPage page);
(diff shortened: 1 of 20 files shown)


app/src/main/java/org/wikipedia/readinglist/ReadingListBookmarkMenu.java
```diff
@@ -73,7 +77,7 @@ private void showMenu() {
         menu.getMenuInflater().inflate(menuRes, menu.getMenu());
         menu.setOnMenuItemClickListener(new PageSaveMenuClickListener());
 
-        if (listsContainingPage.size() == 1) {
+        if (size(listsContainingPage) == 1) {
```
dbrant, line 80: Why do we need a third-party library to get the size of a list?

app/src/main/java/org/wikipedia/readinglist/ReadingListBookmarkMenu.java
```diff
@@ -82,16 +86,16 @@ private void showMenu() {
             menu.setGravity(Gravity.END);
 
             MenuItem addToOtherItem = menu.getMenu().findItem(R.id.menu_add_to_other_list);
-            addToOtherItem.setVisible(listsContainingPage.size() > 0);
-            addToOtherItem.setEnabled(listsContainingPage.size() > 0);
+            addToOtherItem.setVisible(isNotEmpty(listsContainingPage));
```
dbrant, line 89: I could see perhaps changing this to `!list.isEmpty()`, but what is the benefit of using a new third-party dependency for this?

app/src/main/java/org/wikipedia/readinglist/ReadingListBookmarkMenu.java
Isira-Seneviratne: That particular list is nullable, so calling its size() method directly might throw a `NullPointerException`.

app/src/main/java/org/wikipedia/readinglist/ReadingListBookmarkMenu.java
Isira-Seneviratne: Same as earlier: that list is nullable, so calling its `size()` method might throw a `NullPointerException`.

PR #1204, opened 2020-05-05
Use totally random query for image tags.

dbrant:


sharvaniharan:
This is so much better than the suggested images! It would have been great if this was shipped! 🌟

dbrant:
There's every possibility we'll ship this as a follow-up afterwards.  The one minor problem with this is that it doesn't detect if the random image _already_ has tag(s).  That will require a few more rather complex queries.

sharvaniharan:
oh nice! makes sense..

cooltey:
This is definitely a better implementation than the current one. Nice job 👍

PR #1206, opened 2020-05-07
Move getPageViews() to SuggestedEditsUserStats

cooltey:
This PR benefits #1191 which can share the `getPageViews()` method.


sharvaniharan:
@Cooltey can we also make the 'pageTitle->pageviews' map available? it will be helpful for the contributions page too...

PR #1207, opened 2020-05-07
Correctly encode path segments when requesting from RESTBase.

dbrant:


cooltey:
New commit looks good to me. I'll just merge it. @sharvaniharan Please reply to this ticket if you see anything works incorrectly.

PR #1209, opened 2020-05-07
Combine API calls for getting image tags

cooltey:
This returns the image tags in the `HashMap` format.

```
{en=[Exotic Shorthair, house cat]}
```


cooltey:
Pending discussion: the `pageid` will be empty if we get `imageinfo` from `*.wikipedai.org`.

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.java
```diff
@@ -0,0 +1,63 @@
...
+import org.wikipedia.dataclient.mwapi.MwResponse;
+import org.wikipedia.json.PostProcessingTypeAdapter;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+@SuppressWarnings("unused")
+public class Claims extends MwResponse implements PostProcessingTypeAdapter.PostProcessable {
+    @Nullable private Map<String, List<Claim>> claims;
```
dbrant, line 18: Don't need to implement `PostProcessable` here.

app/src/main/java/org/wikipedia/commons/ImageTagsProvider.kt
```diff
@@ -0,0 +1,38 @@
...
+    fun getImageTagsObservable(pageId: String, langCode: String): Observable<Map<String, List<String>>> {
+        return ServiceFactory.get(WikiSite(Service.COMMONS_URL)).getClaims("M$pageId")
+                .subscribeOn(Schedulers.io())
+                .flatMap { claims ->
+                    val depicts = claims.claims()["P180"]
+                    val ids = mutableListOf<String?>()
+                    depicts?.forEach {
+                        ids.add(it.mainSnak?.dataValue?.value?.id)
+                    }
+                    ServiceFactory.get(WikiSite(Service.WIKIDATA_URL)).getWikidataLabels(ids.joinToString(separator = "|"), langCode)
```
dbrant, line 20: This will send a request to Wikidata even if the `depicts` list is empty. Can we return a dummy observable from here if the list is empty?

app/src/main/java/org/wikipedia/commons/ImageTagsProvider.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import io.reactivex.Observable
+import io.reactivex.schedulers.Schedulers
+import org.wikipedia.dataclient.Service
+import org.wikipedia.dataclient.ServiceFactory
+import org.wikipedia.dataclient.WikiSite
+import java.util.*
+
+object ImageTagsProvider {
+    fun getImageTagsObservable(pageId: String, langCode: String): Observable<Map<String, List<String>>> {
```
dbrant, line 11: `pageId` should be a `Long`.

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.java
cooltey: Done

app/src/main/java/org/wikipedia/commons/ImageTagsProvider.kt
cooltey: I check the `pageId` in `MwQueryPage`, it is a `int` type. Can we use that?

app/src/main/java/org/wikipedia/commons/ImageTagsProvider.kt
cooltey: Sure!

PR #1210, opened 2020-05-07
Return entire result of pageViews 

sharvaniharan:
If we return the entire result, we can process it however we want when we use this.
Will help : https://app.zeplin.io/project/57a120b91998d8977642a238/screen/5eb0309f4913f81a5fb9b31f

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsUserStats.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsUserStats.kt
@@ -41,7 +41,7 @@ object SuggestedEditsUserStats {
                 }
     }
 
-    fun getPageViewsObservable(): Observable<Long> {
+    fun getPageViewsObservable(): Observable<Array<Any?>> {
         val qLangMap = HashMap<String, HashSet<String>>()
         return ServiceFactory.get(WikiSite(Service.WIKIDATA_URL)).getUserContributions(AccountUtil.getUserName()!!, 10)
                 .subscribeOn(Schedulers.io())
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsUserStats.kt
```diff
@@ -41,7 +41,7 @@ object SuggestedEditsUserStats {
                 }
     }
 
-    fun getPageViewsObservable(): Observable<Long> {
+    fun getPageViewsObservable(): Observable<Array<Any?>> {
```
cooltey, line 44: Maybe we can make this method to `private` and rename it to `getPageViewsObservableList`, and let it returns `ArrayList<Observable<MwQueryResponse>>`

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsUserStats.kt
```diff
@@ -115,7 +115,7 @@ object SuggestedEditsUserStats {
                                 }
                             }
                         }
-                        totalPageViews
+                        resultList
```
cooltey, line 118: Since the logic above will be used somewhere, it would be better if we can create two public methods:

`getPageViewsObservable` which keeps our existing logic.
`getPageViewsListObservable` which can be use in the contribution screen.

sharvaniharan:
@cooltey I think i will close this for now, because it makes sense for me to make the pageview call myself on the contributions screen than call this, because I make all the apis in the chain anyway, and this will be a repetition.

PR #1220, opened 2020-05-12
Indicate that Commons pages/images are protected

cooltey:
https://phabricator.wikimedia.org/T235028

The newly created accounts (less than 4 days old) will not be able to edit pages in Commons if the pages are semi-protected and we should not disable all the semi-protected pages in Commons.


cooltey:
> 
> 
> Code looks good @cooltey but I can still see the image caption shown in gallery mode has editable pencil next to it... the tas shows that it should be a 'lock+pencil'. Also, I was able to update the caption

Thanks. May I know which article reflects the issue you've mentioned? Ideally, the edit pencil should be hidden if the image is not editable.

sharvaniharan:
**Dont see the CTA which is correct**
Screenshot_20200513-151627 2
**But the pencil is there**
Screenshot_20200513-151600

sharvaniharan:
Oh sorry ... one minute too late... i just rebuilt and i don't see it anymore.. so may be i was testing a wrong build.

cooltey:
> 
> 
> Oh sorry ... one minute too late... i just rebuilt and i don't see it anymore.. so may be i was testing a wrong build.

No problem. I noticed that it is needed to show the locked edit pencil if the file is protected. Code updated.

cooltey:
The PR is ready for review again.

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -131,41 +132,52 @@ private void updateCallToAction() {
...
-                                        captionSourcePageTitle.setDescription(currentCaption);
-                                        callToActionSourceSummary = new SuggestedEditsSummary(captionSourcePageTitle.getPrefixedText(), captionSourcePageTitle.getWikiSite().languageCode(), captionSourcePageTitle,
-                                                captionSourcePageTitle.getDisplayText(), currentCaption, getLeadImageUrl());
-
-                                        callToActionTargetSummary = new SuggestedEditsSummary(captionTargetPageTitle.getPrefixedText(), captionTargetPageTitle.getWikiSite().languageCode(), captionTargetPageTitle,
-                                                captionTargetPageTitle.getDisplayText(), null, getLeadImageUrl());
-                                        pageHeaderView.setUpCallToAction(app.getResources().getString(R.string.suggested_edits_article_cta_image_caption_in_language, app.language().getAppLanguageLocalizedName(lang)));
-                                        break;
-                                    }
+                    if (pair != null) {
```
dbrant, line 149: The `pair` parameter will never be null, so this check is not necessary.

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -131,41 +132,52 @@ private void updateCallToAction() {
...
+                ServiceFactory.get(getTitle().getWikiSite()).getImageInfo(imageTitle, WikipediaApp.getInstance().getAppOrSystemLanguageCode()),
+                ServiceFactory.get(new WikiSite(Service.COMMONS_URL)).getProtectionInfo(imageTitle), (captions, imageInfoRsp, protectionInfoRsp) -> {
+                    boolean protectedFile = false;
+                    for (Protection protection : protectionInfoRsp.query().firstPage().protection()) {
+                        if (protection.getType().equals("edit") && !protectionInfoRsp.query().userInfo().getGroups().contains(protection.getLevel())) {
+                            protectedFile = true;
+                            break;
+                        }
+                    }
+                    return protectedFile ? null : new Pair<>(captions, imageInfoRsp);
```
dbrant, line 144: You must never return `null` for the result of an `Observable`. This will cause an internal exception in RxJava, and will be passed on to onError, and will never make it into the `subscribe` block.

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
cooltey: Done!

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
cooltey: You're right. Done!

PR #1221, opened 2020-05-13
Add option to move articles between reading lists

cooltey:
https://phabricator.wikimedia.org/T237152


app/src/main/java/org/wikipedia/readinglist/AddToReadingListDialog.java
```diff
@@ -45,34 +45,38 @@
     private View listsContainer;
     private View onboardingContainer;
     private View onboardingButton;
-    private InvokeSource invokeSource;
     private CreateButtonClickListener createClickListener = new CreateButtonClickListener();
-    private CompositeDisposable disposables = new CompositeDisposable();
+    List<ReadingList> readingLists = new ArrayList<>();
+    InvokeSource invokeSource;
```
dbrant, line 50: Can these be `private`?

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
```diff
@@ -240,12 +256,44 @@ void addPagesToList(@NonNull SQLiteDatabase db, @NonNull ReadingList list, @NonN
...
+                    continue;
+                }
+                movePageToList(db, sourceReadingListId, list, title);
+                movedTitles.add(title.getDisplayText());
+            }
+            db.setTransactionSuccessful();
+        } finally {
+            db.endTransaction();
+        }
+        if (movedTitles.isEmpty()) {
```
dbrant, line 275: Should this be `!movedTitles.isEmpty()`?

app/src/main/java/org/wikipedia/readinglist/AddToReadingListDialog.java
cooltey: Yes!

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
cooltey: Oops, you're right.

cooltey:
> 
> 
> After I move an article to another reading list, the syncing of the list doesn't seem to work. (When I refresh my lists on another device, the moved page is not synced.)

I have changed the logic to "add" to the new list and "remove" from the old list, and the sync works. Please let me know if that does not work for you.

app/src/main/java/org/wikipedia/readinglist/AddToReadingListDialog.java
```diff
@@ -201,14 +205,33 @@ private void showCreateListDialog() {
...
+
+        if (titles.size() == 1) {
+            addAndDismiss(readingList, titles.get(0));
+            return;
+        }
+
+        execute(readingList, titles);
+    }
+
+    void execute(final ReadingList readingList, final PageTitle title) {
```
dbrant, line 234: Naming: what does this execute?

app/src/main/java/org/wikipedia/readinglist/AddToReadingListDialog.java
```diff
@@ -201,14 +205,33 @@ private void showCreateListDialog() {
...
             String message = getString(R.string.reading_list_article_limit_message, readingList.title(), SiteInfoClient.getMaxPagesPerReadingList());
             FeedbackUtil.makeSnackbar(getActivity(), message, FeedbackUtil.LENGTH_DEFAULT).show();
             dismiss();
             return;
         }
 
+        execute(readingList, title);
+    }
+
+    private void addAndDismiss(final ReadingList readingList, final List<PageTitle> titles) {
```
dbrant, line 218: Do we need a separate function for this? Can we have just a single `addAndDismiss` function that takes a list of Titles, and give it a singleton list in the case of a single title?

app/src/main/java/org/wikipedia/readinglist/AddToReadingListDialog.java
```diff
@@ -220,29 +243,14 @@ private void addAndDismiss(final ReadingList readingList, final PageTitle title)
...
-            dismiss();
-            return;
-        }
-
-        if (titles.size() == 1) {
-            addAndDismiss(readingList, titles.get(0));
-            return;
-        }
-
+    void execute(final ReadingList readingList, final List<PageTitle> titles) {
```
dbrant, line 253: Similarly, can we have just a single `execute` function that takes a list of Titles, and give it a singleton list in the case of a single title?

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
```diff
@@ -74,6 +75,16 @@ public static ReadingListDbHelper instance() {
         return lists;
     }
 
+    private ReadingList getListWithoutContentsById(SQLiteDatabase db, long readingListId) {
```
dbrant, line 78: Return value should probably be annotated `@Nullable`.

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
```diff
@@ -240,12 +266,52 @@ void addPagesToList(@NonNull SQLiteDatabase db, @NonNull ReadingList list, @NonN
...
+        ReadingListPage sourceReadingListPage = getPageByTitle(db, sourceReadingListId, title);
+        ReadingList sourceReadingList = getListWithoutContentsById(db, sourceReadingListId);
+        if (sourceReadingListPage != null && sourceReadingList != null) {
+            addPageToList(db, list, title);
+            markPageForDeletion(db, sourceReadingList, sourceReadingListPage);
+            WikipediaApp.getInstance().getBus().post(new ReadingListSyncEvent());
+        }
+    }
+
+    private void markPageForDeletion(SQLiteDatabase db, @NonNull ReadingList list, @NonNull ReadingListPage page) {
```
dbrant, line 308: Can we just use the existing `markPagesForDeletion` function?

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
```diff
@@ -240,12 +266,52 @@ void addPagesToList(@NonNull SQLiteDatabase db, @NonNull ReadingList list, @NonN
...
 
+    public List<String> movePagesToListIfNotExist(long sourceReadingListId, @NonNull ReadingList list, @NonNull List<PageTitle> titles) {
+        SQLiteDatabase db = getWritableDatabase();
+        db.beginTransaction();
+        List<String> movedTitles = new ArrayList<>();
+        try {
+            for (PageTitle title : titles) {
+                if (getPageByTitle(db, list.id(), title) != null) {
+                    continue;
+                }
```
dbrant, line 277: This will behave incorrectly.  If you move a title from a reading list to another reading list that already has that title, the title won't be removed from the first reading list.

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
```diff
@@ -171,12 +182,27 @@ public void addPageToList(@NonNull ReadingList list, @NonNull PageTitle title, b
         }
     }
 
+    public void movePageToList(long sourceReadingListId, @NonNull ReadingList list, @NonNull PageTitle title, boolean queueForSync) {
```
dbrant, line 185: Can you just use `movePagesToListIfNotExist` instead?

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
```diff
@@ -682,7 +748,7 @@ public ReadingListPage getPageById(long id) {
     }
 
     @Nullable
-    private ReadingListPage getPageByTitle(SQLiteDatabase db, @NonNull ReadingList list, @NonNull PageTitle title) {
```
dbrant, line 685: What's the rationale for changing this to use the `readingListId`?  It would be much cleaner to use structured objects where possible.

app/src/main/java/org/wikipedia/readinglist/AddToReadingListDialog.java
cooltey: I changed it to `run` and not sure if it is a proper name. Do you have any suggestions for this method?

app/src/main/java/org/wikipedia/readinglist/AddToReadingListDialog.java
cooltey: Make sense, done!

app/src/main/java/org/wikipedia/readinglist/AddToReadingListDialog.java
cooltey: Yes!

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
cooltey: Done!

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
cooltey: Done!

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
cooltey: Yes!

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
cooltey: Make sense. Done!

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
cooltey: Good catch! I noticed that the message `All good! %1$s already contains %2$s.` will no longer be shown with the new logic because the `movedTitles` will not be empty. 

If we still want to show the message above, we'll need to call `getPageByTitle()` in the `MoveToReadingListDialog` for the message. Should we do that?

app/src/main/java/org/wikipedia/readinglist/AddToReadingListDialog.java
dbrant: Ideally the name of a function should convey a sense of _what it does_.  Names like `run()`, `doIt()`, `go()`, etc. are pretty much meaningless.  I can see how it's hard to give this function a name because it can do different things when it's overridden. See my follow-up for the best I could come up with.

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
dbrant: I think this is fine as is.

PR #1236, opened 2020-05-18
Add null check for processBitmapForSharing() method

cooltey:


sharvaniharan:
The annotation is changed and check added? isnt it the same?

cooltey:
> 
> 
> The annotation is changed and check added? isnt it the same?

It is because of the original code shows warning that `processBitmapForSharing()` is possible to return `null`

PR #1240, opened 2020-05-19
Consolidate colors for app theme consistency

cooltey:
Since the major app background is changed to `paper_color`, this PR updates the remaining Activities and Fragments to make the app theme consistent.


cooltey:
Note: have confirmed with Robin.

PR #1242, opened 2020-05-19
Priority: fix incorrect target language when translating descriptions/captions.

dbrant:
When adding a caption in a different language from the "Add caption" prompt in the lead image, the captions are sometimes going to the incorrect language.

### DIFF
--- a/app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.java
+++ b/app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.java
@@ -312,7 +312,7 @@ public class DescriptionEditFragment extends Fragment {
         @SuppressWarnings("checkstyle:magicnumber")
         private void postDescription(@NonNull String editToken) {
 
-            disposables.add(ServiceFactory.get(pageTitle.getWikiSite()).getSiteInfo()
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.java
```diff
@@ -312,7 +312,7 @@ public void twoFactorPrompt() {
         @SuppressWarnings("checkstyle:magicnumber")
         private void postDescription(@NonNull String editToken) {
 
-            disposables.add(ServiceFactory.get(pageTitle.getWikiSite()).getSiteInfo()
+            disposables.add(ServiceFactory.get(WikiSite.forLanguageCode(pageTitle.getWikiSite().languageCode())).getSiteInfo()
```
sharvaniharan, line 315: Wouldn't this just result in pageTitle.getWikiSite() ?

dbrant: Nope! That's the key. When editing image captions, the value you get from `pageTitle.getWikiSite()` is commons.wikimedia.org, but the language code of the site is the true language, e.g. "en", "ru", etc.  The language is obviously not "commons".  So, when requesting `siteInfo` from commonswiki, we always get a language code of `en`, when in fact we want the `siteInfo` for the wiki that corresponds to the language code.

PR #1249, opened 2020-05-20
getting article and image details with .zip

sharvaniharan:
Just for comparison of response times

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
@@ -11,6 +11,7 @@ import androidx.fragment.app.Fragment
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import io.reactivex.Observable
+import io.reactivex.ObservableSource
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
```diff
@@ -179,28 +180,36 @@ class SuggestedEditsContributionsFragment : Fragment(), SuggestedEditsTypeItem.C
     }
 
     private fun getArticleContributionDetails() {
-        disposables.add(Observable.fromIterable(continuedArticlesContributions)
-                .subscribeOn(Schedulers.io())
-                .doFinally {
+        val summaryObservables = ArrayList<ObservableSource<PageSummary>?>()
+        for (contribution in continuedArticlesContributions) {
+            summaryObservables.add(ServiceFactory.getRest(contribution.wikiSite).getSummary(null, contribution.title).subscribeOn(Schedulers.io()).onErrorReturnItem(PageSummary()))
```
dbrant, line 185: Hint:  If you look in logcat, you will see that these `summaryObservables` are correctly working in parallel, but your `imageInfoObservables` below are working in series.  What's different about them?

PR #1265, opened 2020-05-26
Constraint views correctly to prevent views overlap

cooltey:
In Chromebook, the feed cards are smaller and the views may overlap with each other. For example, the feed card that shows Suggested edits add image caption with `zh-hant` language code.

### DIFF
--- a/app/src/main/res/layout/view_card_header.xml
+++ b/app/src/main/res/layout/view_card_header.xml
@@ -24,7 +24,7 @@
 
     <TextView
         android:id="@+id/view_card_header_title"
-        android:layout_width="wrap_content"
+        android:layout_width="0dp"
         android:layout_height="wrap_content"
         android:layout_gravity="center_vertical"
         android:layout_marginStart="8dp"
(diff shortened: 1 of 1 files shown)


app/src/main/res/layout/view_card_header.xml
```diff
@@ -35,7 +35,9 @@
         android:textSize="14sp"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintStart_toEndOf="@+id/view_card_header_image"
+        app:layout_constraintEnd_toStartOf="@+id/view_card_header_subtitle"
         app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintHorizontal_weight="1"
```
dbrant, line 40: Is this actually necessary, if we already specify `toStartOf` and `toEndOf`?

cooltey: It looks fine if we remove it, but I still think it would be better to set this because the `android:layout_width` is `0dp`.

dbrant: We have many other cases where the width or height is `0dp` because we have `toStartOf` and `toEndOf` constraints, all without needing `weight`.

cooltey: You convinced me. 👍

PR #1268, opened 2020-05-27
Resurrect the Main Page.

dbrant:
https://phabricator.wikimedia.org/T253397


dbrant:
> Found a bug:
> Go to the English main page -> change the language to `Traditional Chinese` -> then change the language to `Simplified Chinese` -> see error message.
> 
> Looks like the namespace has been added two `Wikipedia:`.

Done.

dbrant:
> Found another issue when loading an article with a lead image.
> 
> Find a Chinese article that has a lead image in trending -> you will notice that the content of the article will be loaded at first and the lead image will show up later -> lead image overlaps the top section.

Done!

PR #1273, opened 2020-05-28
Add right-click event support to trigger the long-click event

cooltey:
> Unable to right click on text using trackpad/mouse


app/src/main/java/org/wikipedia/util/DeviceUtil.java
```diff
@@ -111,6 +111,12 @@ public static void updateStatusBarTheme(@NonNull Activity activity, @Nullable To
         }
     }
 
+    public static void setRightClickEvent(@NonNull View view) {
```
dbrant, line 114: This name could be much more specific. Something like `setContextClickAsLongClick`.

cooltey: Done!

PR #1274, opened 2020-05-29
Upgrade to RxJava 3.0.

dbrant:
This upgrades our code to the newest RxJava 3.0 package, along with RxAndroid and Retrofit.
Most of the changes are package-name restructuring, with no expected impact to functionality. It will be worth reading the documentation to see what new features or conveniences we can use going forward.


cooltey:
Looks good to me!

PR #1275, opened 2020-05-29
Add an option to move article in ReadingListBookmarkMenu 

cooltey:
https://phabricator.wikimedia.org/T237152#6177262

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/news/NewsFragment.java
+++ b/app/src/main/java/org/wikipedia/feed/news/NewsFragment.java
@@ -30,6 +30,7 @@ import org.wikipedia.json.GsonUnmarshaller;
 import org.wikipedia.page.ExclusiveBottomSheetPresenter;
 import org.wikipedia.page.PageActivity;
 import org.wikipedia.readinglist.AddToReadingListDialog;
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/wikipedia/feed/news/NewsFragment.java
```diff
@@ -166,6 +168,12 @@ public void onAddPageToList(@NonNull HistoryEntry entry) {
                     AddToReadingListDialog.newInstance(entry.getTitle(), NEWS_ACTIVITY));
         }
 
+        @Override
+        public void onMovePageToList(long sourceReadingListId, @NonNull HistoryEntry entry) {
+            bottomSheetPresenter.show(getChildFragmentManager(),
+                    MoveToReadingListDialog.newInstance(sourceReadingListId, entry.getTitle(), MOST_READ_ACTIVITY));
```
dbrant, line 174: Is `MOST_READ_ACTIVITY` the correct constant here?

cooltey: It is not. Updated and done!

PR #1276, opened 2020-06-01
Use comma to spearate image tags in FilePage

cooltey:
https://phabricator.wikimedia.org/T250199#6181270

### DIFF
--- a/app/src/main/java/org/wikipedia/commons/FilePageView.kt
+++ b/app/src/main/java/org/wikipedia/commons/FilePageView.kt
@@ -79,7 +79,7 @@ class FilePageView constructor(context: Context, attrs: AttributeSet? = null) : 
     }
 
     private fun getImageTags(imageTags: Map<String, List<String>>, languageCode: String) : String? {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/commons/FilePageView.kt
```diff
@@ -79,7 +79,7 @@ class FilePageView constructor(context: Context, attrs: AttributeSet? = null) :
     }
 
     private fun getImageTags(imageTags: Map<String, List<String>>, languageCode: String) : String? {
-        return imageTags[languageCode]?.joinToString(separator = "\n")
+        return imageTags[languageCode]?.joinToString(separator = ", ")
```
dbrant, line 82: Do all languages use an ASCII comma to separate items in a list? What about Chinese? Or Hebrew?

cooltey: We can use ASCII comma for Chinese, but I think it does not work for Hebrew. I'll let Robin know about this.

cooltey:
proper solution: #1283

PR #1277, opened 2020-06-01
Load Main Page(s) via Mobile Web instead of via API.

dbrant:
Thanks to some ideas from @joewalsh, we shall now load the Main Page (of any language wiki) by literally loading the mobile web URL into the WebView. (this is in fact what the iOS app does.)

This has a couple of drawbacks:

* It will only appear in the Light theme (i.e. will ignore the user's theme selection).

* The Main Page of many wikis often has links to various Special and Portal pages that may also render weirdly in the app.

* The Main Page won't have the other affordances that other pages have, such as the table of contents, collapsible tables, etc.

...But at least this will allow us to present the Main Page to users who need it, making sure it's formatted correctly because we're leveraging mobile web, instead of the still-evolving parsoid `page/html` API.

### DIFF
--- a/app/src/main/java/org/wikipedia/bridge/CommunicationBridge.java
+++ b/app/src/main/java/org/wikipedia/bridge/CommunicationBridge.java
@@ -20,6 +20,7 @@ import org.wikipedia.dataclient.RestService;
 import org.wikipedia.dataclient.ServiceFactory;
 import org.wikipedia.json.GsonUtil;
 import org.wikipedia.page.PageTitle;
+import org.wikipedia.page.PageViewModel;
 import org.wikipedia.util.UriUtil;
 import org.wikipedia.util.log.L;
 
@@ -51,7 +52,7 @@ public class CommunicationBridge {
 
     public interface CommunicationBridgeListener {
         WebView getWebView();
-        PageTitle getPageTitle();
+        PageViewModel getModel();
         boolean isPreview();
     }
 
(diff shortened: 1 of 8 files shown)


dbrant:
@cooltey   The issue you mentioned yesterday should now be fixed.

cooltey:
> 
> 
> @cooltey The issue you mentioned yesterday should now be fixed.

Thanks! The issue has been fixed correctly. I noticed that if we click on the images on the Main page, it will open its own image viewer and it is difficult to close it. Should we fix that?

dbrant:
> > @cooltey The issue you mentioned yesterday should now be fixed.
> 
> Thanks! The issue has been fixed correctly. I noticed that if we click on the images on the Main page, it will open its own image viewer and it is difficult to close it. Should we fix that?

We should not.

cooltey:
> 
> 
> > > @cooltey The issue you mentioned yesterday should now be fixed.
> > 
> > 
> > Thanks! The issue has been fixed correctly. I noticed that if we click on the images on the Main page, it will open its own image viewer and it is difficult to close it. Should we fix that?
> 
> We should not. That would become prohibitively complex to dig into the event structure of the elements in the main page.

Makes sense. @sharvaniharan Please have a review and merge it if it looks good to you too.

sharvaniharan:
> > > @cooltey The issue you mentioned yesterday should now be fixed.
> > 
> > 
> > Thanks! The issue has been fixed correctly. I noticed that if we click on the images on the Main page, it will open its own image viewer and it is difficult to close it. Should we fix that?
> 
> We should not. That would become prohibitively complex to dig into the event structure of the elements in the main page.

I had the same question... if we should handle such actions... but will merge now..

PR #1279, opened 2020-06-01
Support webview right-click event when using mouse

cooltey:


cooltey:
Close this and replace it with a better solution.

PR #1280, opened 2020-06-02
Add mouse click supports on WebView

cooltey:
>  Unable to right click on text using trackpad/mouse

Because I cannot either find a proper way of triggering the system default `ActionMode` when selecting text nor call the system `ActionModeCallback` for the Webview `ActionMode` when calling `startActionMode()`, The simplest way of the solution is to show a custom `PopupMenu` for the right-click event.

Here are the screenshots:
Screenshot 2020-06-03 at 4 20 38 PM
Screenshot 2020-06-03 at 4 21 14 PM]


cooltey:
Better implementation: #1408

PR #1295, opened 2020-06-09
Improve app navigation

sharvaniharan:
**Phab:**https://phabricator.wikimedia.org/T253238

attr?windowBackground should be used instead of color_group_60


app/src/main/res/layout/fragment_main.xml
```diff
@@ -21,33 +19,69 @@
...
+                android:id="@+id/fragment_main_nav_tab_overlay_layout"
+                android:layout_width="match_parent"
+                android:layout_height="match_parent"
+                android:elevation="8dp" />
+
+        </FrameLayout>
+
+        <LinearLayout
+            android:id="@+id/menu_layout"
+            android:layout_width="60dp"
```
cooltey, line 61: I tested it with the Samsung S9 and looks like the width is not correct because I am seeing the background color of the parent view. Also, this button should have a click ripple on this.

app/src/main/java/org/wikipedia/main/MainFragment.java
```diff
@@ -520,4 +532,42 @@ public void accept(Object event) {
...
+                        .setNegativeButton(R.string.logout_dialog_cancel_button_text, null)
+                        .setPositiveButton(R.string.preference_title_logout, (dialog, which) -> {
+                            WikipediaApp.getInstance().logOut();
+                            FeedbackUtil.showMessage(requireActivity(), R.string.toast_logout_complete);
+                            Prefs.setReadingListsLastSyncTime(null);
+                            Prefs.setReadingListSyncEnabled(false);
+                            resetNavTabLayouts();
+                        }).show();
+            } else {
+                onLoginRequested();
```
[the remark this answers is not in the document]
sharvaniharan: Not sure... I still need to discuss that with Robin.

app/src/main/res/layout/fragment_main.xml
```diff
@@ -21,33 +19,69 @@
...
+            android:orientation="vertical">
 
-    </FrameLayout>
+            <ImageView
+                android:id="@+id/menu_icon"
+                android:layout_width="60dp"
+                android:layout_height="wrap_content"
+                android:contentDescription="@null"
+                android:paddingTop="10dp"
+                android:tint="?attr/nav_tab_item_color_state"
```
cooltey, line 74: The icon color looks being activated in dark mode, is it expected?

app/src/main/res/layout/fragment_main.xml
```diff
@@ -21,33 +19,69 @@
...
+                android:id="@+id/fragment_main_nav_tab_overlay_layout"
+                android:layout_width="match_parent"
+                android:layout_height="match_parent"
+                android:elevation="8dp" />
+
+        </FrameLayout>
+
+        <LinearLayout
+            android:id="@+id/nav_more_container"
+            android:layout_width="80dp"
```
dbrant, line 60: If we want this button to appear the same size as the other buttons, then we should probably make the `weight` of the other buttons be `0.8`, and the weight of this button be `0.2`.

app/src/main/res/layout/fragment_main.xml
```diff
@@ -21,33 +19,69 @@
...
-        android:id="@+id/fragment_main_nav_tab_container"
+    <LinearLayout
         android:layout_width="match_parent"
-        android:layout_height="?attr/actionBarSize"
-        android:layout_gravity="bottom"
-        android:background="?attr/nav_tab_background_color"
-        android:elevation="6dp">
+        android:layout_height="wrap_content"
+        android:baselineAligned="false"
+        android:orientation="horizontal">
```
dbrant, line 26: The button bar no longer has a shadow below it. This view should probably have an `elevation` of 6dp, instead of its child views.

app/src/main/res/layout/fragment_main.xml
```diff
@@ -21,33 +19,69 @@
...
-            android:id="@+id/fragment_main_nav_tab_layout"
-            android:layout_width="match_parent"
-            android:layout_height="match_parent"
+        <FrameLayout
+            android:id="@+id/fragment_main_nav_tab_container"
+            android:layout_width="0dp"
+            android:layout_height="?attr/actionBarSize"
+            android:layout_gravity="bottom"
+            android:layout_weight="1"
             android:background="?attr/nav_tab_background_color"
```
dbrant, line 34: This background should now be applied to the parent view, and removed from here.

app/src/main/res/layout/fragment_main.xml
```diff
@@ -21,33 +19,69 @@
...
+
+        </FrameLayout>
+
+        <LinearLayout
+            android:id="@+id/nav_more_container"
+            android:layout_width="80dp"
             android:layout_height="match_parent"
-            android:elevation="8dp"/>
+            android:background="?attr/nav_tab_background_color"
+            android:elevation="6dp"
```
dbrant, line 63: This does not need an elevation, since it should be applied to the parent.

app/src/main/res/layout/fragment_main.xml
```diff
@@ -21,33 +19,69 @@
...
+                android:elevation="8dp" />
+
+        </FrameLayout>
+
+        <LinearLayout
+            android:id="@+id/nav_more_container"
+            android:layout_width="80dp"
             android:layout_height="match_parent"
-            android:elevation="8dp"/>
+            android:background="?attr/nav_tab_background_color"
```
dbrant, line 62: This does not need a background, since the parent will have the same background.

app/src/main/res/layout/fragment_main.xml
```diff
@@ -21,33 +19,69 @@
...
+            android:elevation="6dp"
+            android:foreground="?attr/selectableItemBackground"
```
dbrant, line 64: This `foreground` can now become a `background`.

app/src/main/res/layout/fragment_main.xml
sharvaniharan: Awesome!

sharvaniharan: the NavTabLayout still needs it, but other than that it worked!

PR #1297, opened 2020-06-09
Encourage account creation

cooltey:
https://phabricator.wikimedia.org/T252550

This PR uses the existing logic that leads user to the create account page in `LoginActivity`.


cooltey:
PR is basically for review and we can review/update the analytics later.
One thing remains: https://phabricator.wikimedia.org/T252550#6219919, eventhough I've added a `v2` version of the string.

sharvaniharan:
Fresh install -> 'My Lists' tab -> login -> create an account -> when you get back to 'My Lists' screen, enable sync -> when the lists load, the background of the item is wrong. It is happening only on my emulator, not on real device.. so please check on your device, but if you cannot reproduce, it can be ignored.
bg

cooltey:
> 
> 
> Fresh install -> 'My Lists' tab -> login -> create an account -> when you get back to 'My Lists' screen, enable sync -> when the lists load, the background of the item is wrong. It is happening only on my emulator, not on real device.. so please check on your device, but if you cannot reproduce, it can be ignored.

I cannot reproduce it but I am guessing this is the same issue as #1253.

cooltey:
> 
> 
> I think we would still need a newLine between "Suggested edits is a new way to edit Wikipedia on Android. It helps you make small but vital contributions to Wikipedia" and "Our goal is to make editing easier and more accessible for everyone! Log in or join Wikipedia to get started" here:
> https://phab.wmfusercontent.org/file/data/asdlxxd2rlwxp2suxhda/PHID-FILE-pnmvbbhrtvi2kn54etmu/sev5-01-no-account-a.png

Done!

sharvaniharan:
Launch app -> not logged in -> SE tab shows option for logging in -> login -> SE options update and load correctly -> logout -> SE does not load back the login option view -> Now any activity on the SE tab will crash the app.
Same flow with account creation.

cooltey:
> Launch app -> not logged in -> SE tab shows option for logging in -> login -> SE options update and load correctly -> logout -> SE does not load back the login option view -> Now any activity on the SE tab will crash the app.
Same flow with account creation.

I am not seeing the crash but I think that this should be handled by creating other PR to fix it since every page needs to refresh once the logout is executed.

sharvaniharan:
@cooltey  it might have to go in this one... I am getting a crash : https://youtu.be/a9DCqazWciE
Screen Shot 2020-06-24 at 9 46 48 AM

I am also guessing that even though you don't get a crash on the tasks, we will end up with one when we try to submit a change, since the user is not logged.

cooltey:
> I am also guessing that even though you don't get a crash on the tasks, we will end up with one when we try to submit a change, since the user is not logged.

Make sense. I've created a PR that handles this issue. #1333  (and I merged that into this PR). It should be fixed.

sharvaniharan:
Lgtm... just one last thing...We might need to refresh all fragments @cooltey 🙈  now the SE cards are still accessible after logout...

app/src/main/res/layout/view_suggested_edits_disabled_states.xml
```diff
@@ -11,65 +11,40 @@
...
+        android:textSize="16sp"
+        android:fontFamily="sans-serif"
         android:textColor="?attr/material_theme_secondary_color"
-        android:textSize="16sp" />
+        android:lineSpacingExtra="8sp"
+        android:layout_marginBottom="12dp"/>
 
-    <LinearLayout
-        android:id="@+id/learnMoreContainer"
+    <com.google.android.material.button.MaterialButton
```
dbrant, line 41: This is a good find.  If you've got some spare cycles, perhaps consider a follow-up PR that updates our other usages of these buttons (e.g. in `view_suggested_edits_task_item.xml`) to use `MaterialButton`.

PR #1303, opened 2020-06-11
Follow-up to user contributions.

dbrant:


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
```diff
@@ -28,32 +30,27 @@ import org.wikipedia.suggestededits.Contribution.Companion.EDIT_TYPE_ARTICLE_DES
...
-class SuggestedEditsContributionsFragment : Fragment(), SuggestedEditsTypeItem.Callback {
+class SuggestedEditsContributionsFragment : Fragment(), SuggestedEditsContributionsHeaderView.Callback {
     private val adapter: ContributionsEntryItemAdapter = ContributionsEntryItemAdapter()
-    private var articleContributions = ArrayList<Contribution>()
-    private var imageContributions = HashSet<Contribution>()
-    private var articleAndImageContributions = ArrayList<Contribution>()
-    private var consolidatedContributionsWithDates: MutableList<Any> = ArrayList()
-    private var continuedArticlesContributions = ArrayList<Contribution>()
-    private val continuedImageContributions = HashSet<Contribution>()
+
```
sharvaniharan, line 44: @dbrant Removing the hashsets has resulted in multiple entries one  for each image tag added... also resulting in multiple network calls to image info..

app/src/main/res/layout/fragment_contributions_suggested_edits.xml
```diff
@@ -1,39 +1,39 @@
...
-            android:layout_height="0dp"
-            android:layout_weight="1"
-            android:clipToPadding="false"
-            android:paddingBottom="16dp"
+            android:layout_height="match_parent"
             android:scrollbars="vertical" />
 
         <ProgressBar
-            android:id="@+id/loadMoreProgressView"
+            android:id="@+id/progressBar"
```
sharvaniharan, line 20: I remember Robin wanting a progress indicator at the bottom, to indicate lazy loading, like the explore feed. This will probably be ok, but might have to check with him.

app/src/main/java/org/wikipedia/suggestededits/Contribution.kt
```diff
@@ -3,8 +3,8 @@ package org.wikipedia.suggestededits
 import org.wikipedia.dataclient.WikiSite
 import java.util.*
 
-class Contribution internal constructor(val qNumber: String, var title: String, var description: String, val editType: Int, var imageUrl: String,
-                                        val date: Date, val wikiSite: WikiSite, var pageViews: Long, var tagCount: Int) {
+class Contribution internal constructor(val qNumber: String, var title: String, var description: String, val editType: Int, var imageUrl: String?,
+                                        val date: Date, val wikiSite: WikiSite, var pageViews: Long) {
```
sharvaniharan, line 7: tagCount was added to assist in the dependent PR for displaying each contribution details

sharvaniharan:
Seeing duplicates on article descriptions and image captions too
Screenshot_20200611-122546-1
Screenshot_20200611-122814

dbrant:
@sharvaniharan   Well... according to your edit history on Wikidata, you made two edits to "poisoning of Sergei and Yulia Skripal" and four edits to "Sankethi dialect", so why wouldn't we show an entry in our list for each of those edits?

image

app/src/main/java/org/wikipedia/suggestededits/Contribution.kt
dbrant: Since each edit can add only one tag at most, how would this be helpful?

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
dbrant: Don't we *want* multiple entries, one for each image tag?  Attempting to "coalesce" multiple edits into one item will not work well, and is bound to explode in complexity.  We're building a list of edits, not a list of articles that you have edited.

sharvaniharan:
my bad! thought they were repeating on our end..

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
sharvaniharan: I agree with that school of thought, but clarified during review and during discussions with Robin that these had to be one entry ...yes.. that did take me some time and complexity to coalesce and group the tags while coding this. Might need to check in with Charlotte and Robin

app/src/main/java/org/wikipedia/suggestededits/Contribution.kt
sharvaniharan: This details design also makes more sense when grouped... if it is agreed to not group then its ok.. we don't need to 
tags

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsHeaderView.kt
```diff
@@ -0,0 +1,77 @@
...
+        fun onTypeItemClick(editType: Int)
+    }
+
+    var callback: Callback? = null
+    private var filterViews = ArrayList<SuggestedEditsTypeItem>()
+
+    init {
+        View.inflate(context, R.layout.view_contributions_header, this)
+        orientation = VERTICAL
+        layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
```
cooltey, line 24: Looks like the `R.layout.view_contributions_header` is a `LinearLayout`, should we change the view to use `merge`?

dbrant: Yep, forgot to finish that one.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
```diff
@@ -144,120 +160,111 @@ class SuggestedEditsContributionsFragment : Fragment(), SuggestedEditsTypeItem.C
...
                             }
-                        }
-                    }
-                }, { t ->
-                    L.e(t)
-                    showError(t)
-                }))
-    }
+                },
+                if (allContributions.isNotEmpty() && imageContributionsContinuation.isNullOrEmpty()) Observable.just(Collections.emptyList<Contribution>()) else
```
cooltey, line 183: Explicit type argument can be removed.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
```diff
@@ -63,68 +64,83 @@ class SuggestedEditsContributionsFragment : Fragment(), SuggestedEditsTypeItem.C
...
+        disposables.clear()
+
+        if (allContributions.isEmpty()) {
+            disposables.add(SuggestedEditsUserStats.getPageViewsObservable().subscribe {
+                totalPageViews = it
+                adapter.notifyDataSetChanged()
+            })
+        }
+
+        disposables.add(Observable.zip(if (allContributions.isNotEmpty() && articleContributionsContinuation.isNullOrEmpty()) Observable.just(Collections.emptyList<Contribution>())
```
cooltey, line 137: Explicit type argument can be removed.

dbrant: I don't think it can. Try it for yourself and see if the build succeeds.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
dbrant: I don't think it can. Try it for yourself and see if the build succeeds.

cooltey: You are right. How about just use `Observable.empty()`?

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
cooltey: You are right. How about just use Observable.empty()?

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
dbrant: `Observable.empty()` cannot be used because it causes the subscriber to wait forever, because it's empty.  It needs to be non-empty so that the subscriber receives something.

cooltey: I see. Thanks.

cooltey:
One question:

Looks like the image caption item shows the "image description" instead of "image caption", which seems to be incorrect to me.

The latest image caption I've added:
Screenshot_1592334454

The actual caption I've actually added:
註解 2020-06-16 120842

dbrant:
@cooltey  I have radically improved the way this logic parses the edit comment, so that it should show the correct description, caption, or image tag. (i.e. the description or caption at the time of editing, instead of the current one)

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
```diff
@@ -186,26 +194,41 @@ class SuggestedEditsContributionsFragment : Fragment(), SuggestedEditsContributi
...
                                 var qNumber = ""
 
-                                for (str in strArr) {
-                                    if (str.contains("wbsetlabel")) {
-                                        val descArr = str.split("|")
+                                val matches = commentRegex.findAll(contribution.comment)
+                                if (matches.any()) {
+                                    val metaComment = deCommentString(matches.first().value)
+
+                                    if (metaComment.contains("wbsetlabel")) {
```
cooltey, line 207: How about replacing the `if` with `when`?

dbrant: ok!

dbrant:
> Found a bug when scrolling down to the bottom in the image captions contribution.
> 
> I have 86 contributions in image captions. When reaching to the bottom of it, the progress bar will remain in loading status, and then the error view will show up with the following error message in logcat:
> 
> Also, the error view "go back" button does not properly and it will not disappear after swipe-to-refresh.

Got it, try again with the latest.

PR #1307, opened 2020-06-16
Refresh the feed to dismiss the sync-readinglist login prompt

cooltey:
If the user logs in the account during the onboarding, the Explore feed should be refreshed.


app/src/main/java/org/wikipedia/main/MainActivity.java
```diff
@@ -161,6 +158,12 @@ public boolean onPrepareOptionsMenu(Menu menu) {
         return true;
     }
 
+    @Override
+    public void onActivityResult(int requestCode, int resultCode, final Intent data) {
+        getFragment().onActivityResult(requestCode, resultCode, data);
```
dbrant, line 163: This seems strange: Fragments should automatically get their own `onActivityResult` call, and we shouldn't need to dispatch it ourselves in this way, should we?

cooltey: If we can use `getFragment().startActivityForResult()` then we don't need to dispatch it. But in this case, the `getFragment().startActivityForResult()` will cause `java.lang.NullPointerException`.

For the `startActivityForResult(TabActivity.newIntent(MainActivity.this), Constants.ACTIVITY_REQUEST_BROWSE_TABS)`, the `onActivityResult` in `MainFragment` is not receiving anything.

PR #1309, opened 2020-06-16
Kotlin: convert DialogTitleWithImage from Java to Kotlin

cooltey:


dbrant:
Is this class actually used anymore?

cooltey:
> Is this class actually used anymore?

I don't think so. I am fine with deleting it, but still not sure if we'll need this custom view in the future.

dbrant:
> > Is this class actually used anymore?
> 
> I don't think so. I am fine with deleting it, but still not sure if we'll need this custom view in the future.

Eh, I'm ok leaving it, too. It might come in handy at some point, and it's easier to leave it than dig through the commit history.

PR #1310, opened 2020-06-17
Allow adding multiple image tags in one edit!

dbrant:


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageTagsFragment.kt
```diff
@@ -331,40 +328,39 @@ class SuggestedEditsImageTagsFragment : SuggestedEditsItemFragment(), CompoundBu
...
+                var claimStr = "{\"claims\":["
+                var commentStr = "/* add-depicts: "
                 for (label in acceptedLabels) {
-                    val claimTemplate = "{\"mainsnak\":" +
+                    if (claimStr.length > 16) {
+                        claimStr += ","
+                    }
+                    if (commentStr.length > 20) {
+                        commentStr += ","
+                    }
```
cooltey, line 340: Could you please add comments here to descrip the meaning of the numbers?

dbrant: You're right, that wasn't very clean. I've refactored it a bit to remove the numbers entirely.

PR #1317, opened 2020-06-18
Create a preference for the QA purpose of reward interstitial

cooltey:
https://phabricator.wikimedia.org/T241847#6236091

It would be too painful to see all the cards without a special preference when doing the QA.


sharvaniharan:
Lgtm... I see Todos @cooltey Do we really need to remove it after QA or just leave it for regression tests and such?

cooltey:
> 
> 
> Lgtm... I see Todos @cooltey Do we really need to remove it after QA or just leave it for regression tests and such?

That's a good point. I can remove the TODO from the code.

sharvaniharan:
ok @cooltey thanks.. will be ready to merge after..

PR #1318, opened 2020-06-19
Don't ask if all tabs should be closed if there are none

thiemowmde:
[Bug: T255853](https://phabricator.wikimedia.org/T255853)


dbrant:
Makes sense, thanks!

PR #1323, opened 2020-06-22
Make Contributions screen show total contributions, not just suggested.

dbrant:


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -115,48 +132,61 @@ class SuggestedEditsTasksFragment : Fragment() {
...
-        disposables.add(SuggestedEditsUserStats.getEditCountsObservable()
-                .map { response ->
-                    var shouldLoadPageViews = false
-                    if (response.query()!!.userInfo()!!.isBlocked) {
 
-                        setIPBlockedStatus()
+        disposables.add(Observable.zip(ServiceFactory.get(WikiSite(Service.COMMONS_URL)).getUserContributions(AccountUtil.getUserName()!!, 10, null).subscribeOn(Schedulers.io()),
+                ServiceFactory.get(WikiSite(Service.WIKIDATA_URL)).getUserContributions(AccountUtil.getUserName()!!, 10, null).subscribeOn(Schedulers.io()),
+                SuggestedEditsUserStats.getEditCountsObservable(),
+                Function3<MwQueryResponse, MwQueryResponse, MwQueryResponse, MwQueryResponse> { commonsResponse, wikidataResponse, suggestedStatsResponse ->
```
cooltey, line 145: Looks like the `suggestedStatsResponse` is never used.

dbrant: Yep, I changed it to an underscore. (The call still needs to be made, because it performs some logic implicitly.)

cooltey: Makes sense!

PR #1332, opened 2020-06-24
Design updates for Contribution Details

sharvaniharan:
Removes revision number in the details screen when it is not the current version

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionDetailsFragment.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionDetailsFragment.kt
@@ -60,7 +60,7 @@ class SuggestedEditsContributionDetailsFragment : Fragment() {
         updateTopGradient()
         contributionContainer.setOnClickListener { startTypeSpecificActivity() }
         contributionDetailText.text = contribution.description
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionDetailsFragment.kt
```diff
@@ -60,7 +60,7 @@ class SuggestedEditsContributionDetailsFragment : Fragment() {
         updateTopGradient()
         contributionContainer.setOnClickListener { startTypeSpecificActivity() }
         contributionDetailText.text = contribution.description
-        revisionText.text = if (contribution.top) getString(R.string.suggested_edits_contribution_current_revision) else contribution.revisionId.toString()
+        if (contribution.top) revisionText.text = getString(R.string.suggested_edits_contribution_current_revision) else revisionLayout.visibility = GONE
```
cooltey, line 63: Should we set the text in XML and use this logic to make the view `VISIBLE` or `GONE`?

sharvaniharan: correct! done.

PR #1335, opened 2020-06-24
Design updates for Navigation changes

sharvaniharan:
**Phab:**https://phabricator.wikimedia.org/T253238
Removed globe for login icon


dbrant:
> > Off canvas menu is overlapping the tab bar, this is suboptimal for the user’s orientation
> 
> Looks like the login dialog should move up to not overlap the nav menu.

I'm not aware of any way to do this using the standard bottom sheet component, and I'm pretty sure we mentioned to Robin that this won't be possible.

PR #1340, opened 2020-06-25
Update "diff < 0" gradient color in dark mode

cooltey:
The status bar icons are almost not visible in dark mode if the `diff < 0`.

If this color set does not work out, maybe we should create another attribute for this diff view gradient.


sharvaniharan:
Yes @Cooltey I think there needs to be a new color group... using colorError might be too strong.. will ask Robin

cooltey:
> 
> 
> Yes @cooltey I think there needs to be a new color group... using colorError might be too strong.. will ask Robin

Makes sense! Will update this PR after we received the new color group from Robin.

PR #1343, opened 2020-06-25
Design tweaks [Final?] for contributions

sharvaniharan:
-Changed pageviews label to accent color
-Image loading done differently to reduce flicker
-Removed 'File:' from the image titles. used 'replaceFirst' just in case of edge case of having files named that way
-Could not reproduce or see in code how text could be cut-off as mentioned in design review. Open to suggestions.
**Phab:*https://phabricator.wikimedia.org/T250190

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
@@ -38,6 +38,7 @@ import org.wikipedia.suggestededits.SuggestedEditsContributionsItemView.Callback
 import org.wikipedia.util.DateUtil
 import org.wikipedia.util.DimenUtil
 import org.wikipedia.util.ResourceUtil
+import org.wikipedia.util.StringUtil
 import org.wikipedia.util.log.L
 import org.wikipedia.views.DefaultViewHolder
 import java.util.*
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsContributionsFragment.kt
```diff
@@ -370,7 +370,8 @@ class SuggestedEditsContributionsFragment : Fragment(), SuggestedEditsContributi
             view.contribution = contribution
             view.setTitle(contribution.description)
             view.setDiffCountText(contribution)
-            view.setDescription(contribution.title)
+            view.setDescription(if (contribution.editType == EDIT_TYPE_IMAGE_TAG || contribution.editType == EDIT_TYPE_IMAGE_CAPTION)
+                contribution.title.replaceFirst("File:", "") else contribution.title)
```
cooltey, line 374: I think we can just remove the namespace without an extra logic.

Also, you can use `StringUtil.removeNamespace(contribution.title)` to remove the namespace.

cooltey: I noticed that the `File:` should be removed in the detail page as well.

sharvaniharan: Correct @cooltey done!

PR #1344, opened 2020-06-25
Encode title passed to Footer.add()

berndsi:
This title is used by PCS to call /page/related as is.

Bug: https://phabricator.wikimedia.org/T255579


dbrant:
The problem is that when we pass data through the Javascript Bridge, it expects the data to be already urlencoded, so when it reaches the WebView, it gets decoded.  Therefore, if we want the title to remain encoded even after it gets to the WebView, we should _double_-urlencode it.

Or we could make the change in mobile-html to encode the title when requesting related titles.

berndsi:
Ok, I have another proposal, to avoid having the client send the title at all, see the associated task. Let me know if you see any issues with that approach.

PR #1349, opened 2020-06-26
Navigation icon tint change

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T253238


app/src/main/res/drawable/ic_baseline_person_24.xml
```diff
@@ -1,5 +1,5 @@
 <vector android:autoMirrored="true" android:height="24dp"
-    android:tint="?attr/colorControlNormal" android:viewportHeight="24"
+    android:tint="?attr/material_theme_secondary_color" android:viewportHeight="24"
```
dbrant, line 2: If this drawable will be tinted explicitly wherever it's used, does it need its own tint here?

sharvaniharan: Correct. removed @dbrant

PR #1350, opened 2020-06-26
Replace custom icon+text button with MaterialButton

cooltey:
https://github.com/wikimedia/apps-android-wikipedia/pull/1297#pullrequestreview-438355114


dbrant:
The buttons look a bit larger than the designs, and also with larger margins around them.
Is it possible to make the button height exactly 32dp, and the spacing around them just like before? (or convince Robin to accept this new appearance)

cooltey:
> 
> 
> The buttons look a bit larger than the designs, and also with larger margins around them.
> Is it possible to make the button height exactly 32dp, and the spacing around them just like before? (or convince Robin to accept this new appearance)

Sure! I created a `FlatButton.Small` for a smaller size button, please check the latest version of the PR.

dbrant:
That's it!

PR #1353, opened 2020-06-27
Tweak: set up "more on this day" background correctly

cooltey:
If you watch closely, the "more on this day" bottom action layout has `paper_color` background and makes the rounded border has a rectangle background. This PR fixes it but still has a minor issue: the ripple animation is still in a square shape. 

Please advise if any suggestions.


dbrant:
@cooltey  We have the go-ahead to disable the overflowing of child cards (for now).  Therefore, we can just remove the usages of `setAllowOverflow()`, and everything should fix itself.

cooltey:
> @cooltey We have the go-ahead to disable the overflowing of child cards (for now). Therefore, we can just remove the usages of `setAllowOverflow()`, and everything should fix itself.

Makes sense. Done!

PR #1360, opened 2020-06-29
Tweak: encourage account creation

cooltey:
https://phabricator.wikimedia.org/T252550#6265292


app/src/main/java/org/wikipedia/notifications/NotificationPollBroadcastReceiver.java
```diff
@@ -190,21 +190,23 @@ private void onNotificationsComplete(@NonNull final Context context, @NonNull Li
...
-                        || (n.category().equals(Notification.CATEGORY_EDIT_THANK) && Prefs.notificationThanksEnabled())
-                        || (n.category().equals(Notification.CATEGORY_THANK_YOU_EDIT) && Prefs.notificationMilestoneEnabled())
-                        || (n.category().equals(Notification.CATEGORY_REVERTED) && Prefs.notificationRevertEnabled())
-                        || (n.category().equals(Notification.CATEGORY_EDIT_USER_TALK) && Prefs.notificationUserTalkEnabled())
-                        || (n.category().equals(Notification.CATEGORY_LOGIN_FAIL) && Prefs.notificationLoginFailEnabled())
-                        || (n.category().startsWith(Notification.CATEGORY_MENTION) && Prefs.notificationMentionEnabled())
-                        || Prefs.showAllNotifications()) {
-
-                    NotificationPresenter.showNotification(context, n, dbNameWikiNameMap.containsKey(n.wiki()) ? dbNameWikiNameMap.get(n.wiki()) : n.wiki());
+        if (!Prefs.isSuggestedEditsHighestPriorityEnabled()) {
```
dbrant, line 193: With this logic, it will just "forget" the notifications, and never show them again, which is not good.   Can we just `return` from this function, and delay showing the notifications until the next poll cycle?

cooltey: Makes sense and done!

PR #1361, opened 2020-06-29
Remove AppBar shadows

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T256388


app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.java
```diff
@@ -124,8 +125,8 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
         searchEmptyView.setEmptyText(R.string.search_reading_lists_no_results);
         readingListView.setLayoutManager(new LinearLayoutManager(getContext()));
         readingListView.setAdapter(adapter);
-        readingListView.addItemDecoration(new DrawableItemDecoration(requireContext(), R.attr.list_separator_drawable, true, false));
-
+        readingListView.addItemDecoration(new DrawableItemDecoration(requireContext(), R.attr.list_separator_drawable, false, true));
```
cooltey, line 128: The top divider will still show up if there's no other reading list than "Saved".

sharvaniharan: Good one @cooltey .. it needed a general fix to all recyclerviews.. addressed here #1369

app/src/main/java/org/wikipedia/history/HistoryFragment.java
```diff
@@ -110,9 +111,20 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
...
+        setUpScrollListener();
         return view;
     }
 
+    private void setUpScrollListener() {
+        historyList.addOnScrollListener(new RecyclerView.OnScrollListener() {
+            @Override
+            public void onScrolled(@NonNull RecyclerView recyclerView, int dx, int dy) {
+                super.onScrolled(recyclerView, dx, dy);
+                ((MainActivity) requireActivity()).updateToolbarElevation(historyList.canScrollVertically(SCROLL_DIRECTION_UP));
```
dbrant, line 123: Can't we just use `historyList.computeVerticalScrollOffset() != 0`, and not require a custom constant?

app/src/main/java/org/wikipedia/history/HistoryFragment.java
```diff
@@ -110,9 +111,20 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
...
+            }
+        });
```
dbrant, line 125: For symmetry, and to avoid leaks, we should call `historyList.clearOnScrollListeners()` when destroying the fragment.

sharvaniharan:
Done @dbrant

app/src/main/java/org/wikipedia/Constants.java
```diff
@@ -60,6 +60,7 @@
     public static final int MAX_READING_LISTS_LIMIT = 100;
 
     public static final int MIN_LANGUAGES_TO_UNLOCK_TRANSLATION = 2;
+    public static final int SCROLL_DIRECTION_UP = -1;
```
dbrant, line 63: This is no longer needed, is it?

sharvaniharan: Done

PR #1366, opened 2020-06-30
Don't show SE cards in feed customisation when logged out

cooltey:
https://phabricator.wikimedia.org/T256773

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/FeedContentType.java
+++ b/app/src/main/java/org/wikipedia/feed/FeedContentType.java
@@ -88,7 +88,7 @@ public enum FeedContentType implements EnumCode {
             return isEnabled() ? new BecauseYouReadClient() : null;
         }
     },
-    SUGGESTED_EDITS(9, R.string.suggested_edits_feed_card_title, R.string.feed_item_type_suggested_edits, false) {
+    SUGGESTED_EDITS(9, R.string.suggested_edits_feed_card_title, R.string.feed_item_type_suggested_edits, false, AccountUtil.isLoggedIn()) {
         @Nullable
         @Override
         public FeedClient newClient(AggregatedFeedContentClient aggregatedClient, int age) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/feed/configure/ConfigureFragment.java
```diff
@@ -176,6 +177,10 @@ private void prepareContentTypeList() {
                 i.remove();
                 continue;
             }
+            if (!AccountUtil.isLoggedIn() && feedContentType == FeedContentType.SUGGESTED_EDITS) {
```
sharvaniharan, line 180: @cooltey This should actually be a change to the constructor in FeedContentType.  
`   SUGGESTED_EDITS(9, R.string.suggested_edits_feed_card_title, R.string.feed_item_type_suggested_edits, false,AccountUtil.isLoggedIn()) {
`
It will automatically be taken care of.

cooltey: Thanks! Done!

app/src/main/java/org/wikipedia/feed/FeedContentType.java
```diff
@@ -88,7 +88,7 @@ public FeedClient newClient(AggregatedFeedContentClient aggregatedClient, int ag
             return isEnabled() ? new BecauseYouReadClient() : null;
         }
     },
-    SUGGESTED_EDITS(9, R.string.suggested_edits_feed_card_title, R.string.feed_item_type_suggested_edits, false) {
+    SUGGESTED_EDITS(9, R.string.suggested_edits_feed_card_title, R.string.feed_item_type_suggested_edits, false, AccountUtil.isLoggedIn()) {
```
dbrant, line 91: Hang on, sorry, @sharvaniharan's previous comment is not correct for this case.  This doesn't actually work because it will make the state of this item permanent. (If you log in afterwards, the item will not appear, because the constructor was already called.)  Your original method was just fine.

cooltey: Done!

PR #1368, opened 2020-06-30
Branding / Terminology changing to 'Edits'

sharvaniharan:

**Note:**  I see a couple more places which have "Suggested edits in the copy" . I have asked Robin for a new copy, but meanwhile we can merge this. I will follow-up with those couple string.


app/src/main/java/org/wikipedia/navtab/NavTab.kt
```diff
@@ -25,7 +25,7 @@ enum class NavTab constructor(private val text: Int, private val icon: Int) : En
             return HistoryFragment.newInstance()
         }
     },
-    SUGGESTED_EDITS(R.string.nav_item_suggested_edits, R.drawable.ic_mode_edit_themed_24dp) {
+    SUGGESTED_EDITS(R.string.nav_item_edits, R.drawable.ic_mode_edit_themed_24dp) {
```
cooltey, line 28: `SUGGESTED_EDITS` -> `EDITS`

dbrant:
Have you confirmed that renaming all the string resources (in all the localized files) will automatically rename the corresponding wiki entries on TranslateWiki?

sharvaniharan:
Splitting this into manageable PRs, without renaming the strings.

PR #1369, opened 2020-07-01
Remove unnecessary decor divider in recyclerviews

sharvaniharan:
Looks like this decor line drew an extra divider on top of the first item. [ In red ]
Screenshot_20200630-171005


dbrant:
What about the divider that is now missing from the second-to-last item?

sharvaniharan:
Thank you @dbrant had not seen that.. I see that when there is only one child top was getting drawn on that one line. Corrected now!

PR #1372, opened 2020-07-01
Add static graph

sharvaniharan:


app/src/main/res/layout/view_contributions_header.xml
```diff
@@ -1,36 +1,49 @@
...
+        <ImageView
+            android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:layout_marginTop="4dp"
-            android:lineSpacingExtra="8sp"
-            android:textColor="?attr/secondary_text_color"
-            android:textSize="16sp"
-            android:textStyle="normal" />
+            android:contentDescription="@null"
+            app:srcCompat="@drawable/ic_edits_graph" />
```
cooltey, line 46: The margin to the left should be `16dp`, otherwise it will too close to the text.

sharvaniharan:
Fixed the margin @cooltey ... not sure if we can help the jump in text... since it just makes space dynamically... I would say let us see what robin wants to suggest?

cooltey:
> 
> 
> Fixed the margin @cooltey ... not sure if we can help the jump in text... since it just makes space dynamically... I would say let us see what robin wants to suggest?

Makes sense to me. Thanks!

PR #1373, opened 2020-07-01
New behaviour/screen for My Lists

cooltey:
https://phabricator.wikimedia.org/T256775


app/src/main/java/org/wikipedia/views/MessageCardView.kt
```diff
@@ -4,17 +4,43 @@ import android.content.Context
...
+
+    fun setMessageTitle(title: String) {
+        messageTitleView.text = title
+    }
+
+    fun setMessageText(text: String) {
+        messageTextView.text = text
+    }
+
+    fun setImageSource(imageSource: Int) {
```
dbrant, line 29: This should probably be named `setImageResource`, and annotate the parameter with `@DrawableRes`.

app/src/main/java/org/wikipedia/views/MessageCardView.kt
```diff
@@ -4,17 +4,43 @@ import android.content.Context
...
+        if (imageSource < 0) {
```
dbrant, line 30: It's not very obvious to implicitly show/hide the image based on the resource parameter.  Perhaps we can have a second (boolean) parameter to show/hide the image.

app/src/main/java/org/wikipedia/views/MessageCardView.kt
```diff
@@ -4,17 +4,43 @@ import android.content.Context
...
+            imageView.visibility = View.GONE
+        } else {
+            imageView.visibility = View.VISIBLE
+            imageView.setImageResource(imageSource)
+        }
+    }
+
+    fun setButton(text: String, listener: OnClickListener, applyListenerToContainer: Boolean) {
```
dbrant, line 38: Can the `text` be an Int, annotated with `@StringRes`?

app/src/main/java/org/wikipedia/views/MessageCardView.kt
cooltey: Makes sense. Done!

app/src/main/java/org/wikipedia/views/MessageCardView.kt
cooltey: You're correct. Done!

app/src/main/java/org/wikipedia/views/MessageCardView.kt
cooltey: Sure! Done.

cooltey:
Pending discussion with Robin: https://phabricator.wikimedia.org/T256775#6283622

PR #1378, opened 2020-07-02
Rename code from 'SuggestedEdits' to 'Edits'

sharvaniharan:

**Note:**
-Anything in feed or notification is still 'Suggested Edits'
-I have not changes analytics or preferences


app/src/main/java/org/wikipedia/edits/EditsSummary.kt
```diff
@@ -1,11 +1,11 @@
-package org.wikipedia.suggestededits
+package org.wikipedia.edits
 
 import org.wikipedia.Constants
 import org.wikipedia.gallery.ExtMetadata
 import org.wikipedia.page.PageTitle
 import org.wikipedia.util.ImageUrlUtil
 
-data class SuggestedEditsSummary(
+data class EditsSummary(
```
dbrant, line 8: `EditsSummary` sounds very odd. Can we come up with a more meaningful name? If someone looks at this code for the first time, would they know what this is for?  Perhaps something like `PageSummaryForEdit`?

app/src/main/java/org/wikipedia/dataclient/RestService.java
```diff
@@ -181,22 +181,22 @@
     @Headers("Cache-Control: no-cache")
     @GET("data/recommendation/caption/addition/{lang}")
     @NonNull
-    Observable<List<SuggestedEditItem>> getImagesWithoutCaptions(@NonNull @Path("lang") String lang);
+    Observable<List<EditItem>> getImagesWithoutCaptions(@NonNull @Path("lang") String lang);
```
dbrant, line 184: These functions are still returning editing suggestions.  Isn't `EditItem` more confusing now?

app/src/main/java/org/wikipedia/edits/EditsTask.kt
```diff
@@ -1,12 +1,13 @@
-package org.wikipedia.suggestededits
+package org.wikipedia.edits
 
 import androidx.annotation.DrawableRes
 
-class SuggestedEditsTask {
+class EditsTask {
```
dbrant, line 5: Generally speaking, classes that have to do with "suggesting" edits to the user should probably keep the terminology.

app/src/main/java/org/wikipedia/edits/EditsTypeItem.kt
```diff
@@ -10,10 +10,10 @@ import androidx.core.content.ContextCompat
 import androidx.core.widget.ImageViewCompat
 import kotlinx.android.synthetic.main.item_suggested_edits_type.view.*
 import org.wikipedia.R
-import org.wikipedia.suggestededits.Contribution.Companion.EDIT_TYPE_GENERIC
+import org.wikipedia.edits.Contribution.Companion.EDIT_TYPE_GENERIC
 import org.wikipedia.util.ResourceUtil
 
-class SuggestedEditsTypeItem constructor(context: Context, attrs: AttributeSet? = null) : LinearLayout(context, attrs) {
+class EditsTypeItem constructor(context: Context, attrs: AttributeSet? = null) : LinearLayout(context, attrs) {
```
dbrant, line 16: Since this class is a View, the name of the class should probably end in `...View`, to keep consistent with our other Views.  e.g. `EditTypeItemView`.

app/src/main/java/org/wikipedia/edits/EditsContributionsFragment.kt
```diff
@@ -47,7 +47,7 @@ import java.util.concurrent.TimeUnit
 import kotlin.collections.ArrayList
 import kotlin.collections.HashMap
 
-class SuggestedEditsContributionsFragment : Fragment(), SuggestedEditsContributionsHeaderView.Callback {
+class EditsContributionsFragment : Fragment(), EditsContributionsHeaderView.Callback {
```
dbrant, line 50: Not everything needs to have the `Edits` prefix, and in fact having `Edits...` in front of all the names makes it sound weird.  This could easily be `ContributionsFragment`. You get the idea.

sharvaniharan:
changes done @dbrant

PR #1384, opened 2020-07-06
Consolidate app themes and remove redundant styles

cooltey:
- Remove redundant styles
- Combine similar styles
- Created `ViewUtil.setActionBarElevation()` for the current elevation logic to apply to the activies that have list.


cooltey:
Separate this huge PR to two PRs: #1419 #1420

PR #1389, opened 2020-07-07
Tooltip using balloon library

sharvaniharan:
Screen Shot 2020-07-07 at 12 03 19 PM
Screen Shot 2020-07-07 at 12 05 00 PM


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -98,10 +103,64 @@ class SuggestedEditsTasksFragment : Fragment() {
...
+            setText("ContributionsStatsView")
+            setTextColorResource(R.color.base100)
+            setTextIsHtml(true)
+            setIconDrawable(ContextCompat.getDrawable(requireContext(), R.drawable.ic_baseline_person_24))
+            setBackgroundColorResource(R.color.accent30)
+            setBalloonAnimation(BalloonAnimation.FADE)
+            setLifecycleOwner(lifecycleOwner)
+        }
+        editQualityStatsView.showAlignTop(balloon)
+        balloon
```
sharvaniharan, line 158: Looking for a better way to do this sequencing... just for readability...  also just a note that it is possible to define the dismiss listener.. I have just not used it because the sequencing logic overrides it..

sharvaniharan:
Finalized this library. So closing this

PR #1398, opened 2020-07-09
Tooltips - SE v5 improvements

sharvaniharan:
Note: Tooltips for user stats need to be shown sequential when we show stats the first time. Then, on long click.


sharvaniharan:
there is an explicit set-up  for RTL support, .isRtlSupport(true). Yet, it seems to be working fine without setting that. Please lmk when you both test. @cooltey @dbrant

sharvaniharan:
Robin agreed to showing tooltip on long press

sharvaniharan:
added preference @cooltey . Sorry for the delay.

app/src/main/java/org/wikipedia/util/FeedbackUtil.java
```diff
@@ -173,6 +176,19 @@ public static Toast showToastOverView(View view, CharSequence text, int duration
         return toast;
     }
 
+    @SuppressWarnings("checkstyle:magicnumber")
+    public static Balloon showTooltipBubble(@NonNull Context context, @NonNull CharSequence text) {
```
dbrant, line 180: Why call it a "bubble" when the library calls it a "balloon"? Can we just say `showTooltip()`?

app/src/main/java/org/wikipedia/views/ImageTitleDescriptionView.kt
```diff
@@ -7,20 +7,32 @@ import android.view.View
...
+    var tooltipText: String = ""
 
     init {
         View.inflate(context, R.layout.view_image_title_description, this)
         layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
     }
 
+    fun setUpTooltip(@NonNull tooltip: String) {
+        tooltipText = tooltip
+        setOnLongClickListener {
```
dbrant, line 30: Can we move this part to the constructor?

app/src/main/java/org/wikipedia/views/ImageTitleDescriptionView.kt
```diff
@@ -7,20 +7,32 @@ import android.view.View
...
 
 internal class ImageTitleDescriptionView constructor(context: Context, attrs: AttributeSet? = null) : LinearLayout(context, attrs) {
+    var tooltipText: String = ""
 
     init {
         View.inflate(context, R.layout.view_image_title_description, this)
         layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
     }
 
+    fun setUpTooltip(@NonNull tooltip: String) {
```
dbrant, line 28: Can we call this `setTooltipText`?

sharvaniharan:
Done @dbrant

app/src/main/java/org/wikipedia/views/ImageTitleDescriptionView.kt
```diff
@@ -7,18 +7,32 @@ import android.view.View
...
         layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
+        setOnLongClickListener {
+            if (!tooltipText.isEmpty()) {
+                description.showAlignBottom(FeedbackUtil.showTooltip(context, tooltipText), 0, 16)
+            }
+            true
+        }
+    }
+
+    fun setUpTooltipText(@NonNull tooltip: String) {
```
dbrant, line 34: This function is no longer necessary. We can now just use the `tooltipText =` accessor.

app/src/main/java/org/wikipedia/views/ImageTitleDescriptionView.kt
```diff
@@ -7,18 +7,32 @@ import android.view.View
...
+import org.wikipedia.util.FeedbackUtil
 
 internal class ImageTitleDescriptionView constructor(context: Context, attrs: AttributeSet? = null) : LinearLayout(context, attrs) {
+    var tooltipText: String = ""
 
     init {
         View.inflate(context, R.layout.view_image_title_description, this)
         layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT)
+        setOnLongClickListener {
+            if (!tooltipText.isEmpty()) {
```
dbrant, line 27: This can be `tooltipText.isNotEmpty()`

app/src/main/java/org/wikipedia/views/ImageTitleDescriptionView.kt
```diff
@@ -7,18 +7,32 @@ import android.view.View
...
+                description.showAlignBottom(FeedbackUtil.showTooltip(context, tooltipText), 0, 16)
```
dbrant, line 28: The alignment of the arrow might look even better if we anchor it to the view itself (`this`) instead of the `description` view.

sharvaniharan: It centers that to the whole view @dbrant so Robin wanted it to be centered to the description.

dbrant: OK, Robin wants it centered on the description _text_, not the description _View_. The description view is much wider than the text, because it is `match_parent` but should probably be `wrap_content`.
I don't think this looks right, does it?
device-2020-08-05-140134

sharvaniharan:
Correct @dbrant . Done.

dbrant:
Fab

PR #1400, opened 2020-07-09
Prepare API for editing descriptions directly in article.

dbrant:
This paves the way for submitting description edits directly into the article contents, instead of Wikidata.

- Update the Edit call to use RxJava. (This was pretty much the last call that wasn't transitioned to Rx.)
- Remove outdated `EditClient` class.
- Remove outdated unit tests for EditClient.


app/src/main/java/org/wikipedia/edit/EditSectionActivity.java
```diff
@@ -319,69 +310,76 @@ private void doSave(@NonNull String token) {
...
-                L.e(caught);
-            }
+    public void onEditFailure(@NonNull Throwable caught) {
+        showProgressBar(false);
+        if (caught instanceof MwException) {
+            handleEditingException((MwException) caught);
+            L.e(caught);
+        } else {
+            showRetryDialog(caught);
+            L.e(caught);
```
cooltey, line 382: Looks like the common `L.e(caught)` can be extracted to the bottom.

dbrant: Correct

PR #1401, opened 2020-07-09
Edit descriptions on the local wiki for English.

dbrant:


app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.java
```diff
@@ -415,4 +478,36 @@ public void onVoiceInputClick() {
             }
         }
     }
+
+    private String updateDescriptionInArticle(String articleText, String newDescription) {
```
cooltey, line 482: Would it be better if add `@NonNull` to the parameters?

app/src/main/java/org/wikipedia/descriptions/DescriptionEditUtil.java
```diff
@@ -15,6 +15,11 @@
 
     public static boolean isEditAllowed(@NonNull Page page) {
         PageProperties props = page.getPageProperties();
+        if (page.getTitle().getWikiSite().languageCode().equals("en")) {
```
cooltey, line 18: Maybe it is redundant but is it a good idea to put `&& props.canEdit()` into the `if` statement?

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.java
dbrant: yep

app/src/main/java/org/wikipedia/descriptions/DescriptionEditUtil.java
dbrant: Technically this method tells if the *description* specifically should be editable. I'm not sure we should mix up whether the entire article is editable here. (it's already handled correctly elsewhere)

cooltey (approved):
Tested and looks good to me. @sharvaniharan Please review and merge if it looks good to you too.

PR #1407, opened 2020-07-10
Fix logic of when to show Reading List sync snackbar.

dbrant:
The updates in #1397 totally changed the behavior of when the snackbar is shown -- it's now being shown for every page added to a list.  The snackbar should only be shown when the user explicitly refreshes the lists.


dbrant:
> Thanks for the finding. One thing I've noticed is that the snackbar should appear after logging in.

Do we actually want that? How many messages will the user be bombarded with when they log in for the first time?

cooltey:
> > Thanks for the finding. One thing I've noticed is that the snackbar should appear after logging in.
> 
> Do we actually want that? How many messages will the user be bombarded with when they log in for the first time?

I agreed with you. But in the ticket description, Robin mentioned that to suppress the "sync reading list" snackbar only if login through the SE login prompt and keep the snackbar. Could you please leave a comment on the ticket for the new behavior? (and then we can merge this PR)

https://phabricator.wikimedia.org/T252041

PR #1408, opened 2020-07-10
WebView right-click behavior for physical mouse devices.

dbrant:


dbrant:
@cooltey  Would you mind testing this on the Chromebook, and see how the right-click behavior works?

cooltey:
@dbrant 
The behavior of right-clicking on the non-URL text is really great (to see the default floating toolbar), but right-click on hyperlinks requires a "long" right-click to trigger the dialog.

dbrant:
> @dbrant
> The behavior of right-clicking on the non-URL text is really great (to see the default floating toolbar), but right-click on hyperlinks requires a "long" right-click to trigger the dialog.

Can this be an acceptable tradeoff, while making this functionality as simple as possible?

cooltey:
> 
> 
> > @dbrant
> > The behavior of right-clicking on the non-URL text is really great (to see the default floating toolbar), but right-click on hyperlinks requires a "long" right-click to trigger the dialog.
> 
> Can this be an acceptable tradeoff, while making this functionality as simple as possible?

Yes, of course. It is definitely a simpler solution for the Webview right-click event. Is it okay to be merged now?

dbrant:
OK to merge!

PR #1410, opened 2020-07-10
Hygiene: optimize progress bar method

cooltey:
- The `ProgressBar` is always `indeterminate=true`.
- When loading page that requires more time to load (e.g. Chinese article - [2019冠状病毒病疫情](https://zh.wikipedia.org/wiki/2019%E5%86%A0%E7%8A%B6%E7%97%85%E6%AF%92%E7%97%85%E7%96%AB%E6%83%85)), the progress bar should disappear after the metadata is loaded.


dbrant:
> * When loading page that requires more time to load (e.g. Chinese article - [2019冠状病毒病疫情](https://zh.wikipedia.org/wiki/2019%E5%86%A0%E7%8A%B6%E7%97%85%E6%AF%92%E7%97%85%E7%96%AB%E6%83%85)), the progress bar should disappear after the metadata is loaded.

I'm not sure I agree.  The progress bar should disappear when the page is fully loaded.

cooltey:
> > * When loading page that requires more time to load (e.g. Chinese article - [2019冠状病毒病疫情](https://zh.wikipedia.org/wiki/2019%E5%86%A0%E7%8A%B6%E7%97%85%E6%AF%92%E7%97%85%E7%96%AB%E6%83%85)), the progress bar should disappear after the metadata is loaded.
> 
> I'm not sure I agree. The progress bar should disappear when the page is fully loaded.

Done.

PR #1411, opened 2020-07-10
Fix: should keep selected language tab position in search screen

cooltey:
If you try to search an article in a different language, the tab position should keep on the selected language after returning from the article.

Also, if you tap a language from the "Wikipedia languages" screen, it should be automatically selected (and scroll to it) in the search screen.


app/src/main/java/org/wikipedia/views/LanguageScrollView.java
```diff
@@ -101,27 +97,34 @@ private void updateTabView(boolean selected, TabLayout.Tab tab) {
...
         for (int i = 0; i < languageCodes.size(); i++) {
             TabLayout.Tab tab = horizontalLanguageScroll.newTab();
             tab.setCustomView(createLanguageTab(languageCodes.get(i)));
             horizontalLanguageScroll.addTab(tab);
             updateTabView(false, tab);
         }
-        selectLanguageTab(position);
+
+        if (horizontalLanguageScroll != null && horizontalLanguageScroll.getTabAt(position) != null) {
+            horizontalLanguageScroll.postDelayed(() -> horizontalLanguageScroll.getTabAt(position).select(), 100);
```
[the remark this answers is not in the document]
cooltey: Done!

cooltey:
> It might be a lot less complicated, and even more useful, if we just save the last-used language position in a `Pref` (i.e. save it inside `onPause()`), and then restore it in `onStart()` where we currently set it to `0`.

Makes sense. Added!

app/src/main/java/org/wikipedia/views/LanguageScrollView.java
```diff
@@ -101,27 +97,39 @@ private void updateTabView(boolean selected, TabLayout.Tab tab) {
...
         for (int i = 0; i < languageCodes.size(); i++) {
             TabLayout.Tab tab = horizontalLanguageScroll.newTab();
             tab.setCustomView(createLanguageTab(languageCodes.get(i)));
             horizontalLanguageScroll.addTab(tab);
             updateTabView(false, tab);
         }
-        selectLanguageTab(position);
+
+        if (horizontalLanguageScroll != null && horizontalLanguageScroll.getTabAt(position) != null) {
+            horizontalLanguageScroll.postDelayed(() -> {
```
dbrant, line 121: What's the reason for `post()`ing this?  It should work fine without a post.

cooltey: It fixes the issue that the scroll view does not scroll to the selected item after tapping on the last item in the Wikipedia languages screen(the item can not be seen by default).

Add ~5 languages and make sure the last item is not able to see -> when you tapping on the "more" to go to the "Wikipedia languages" -> tap on the last item -> it should scroll to the last item.

dbrant: OK, in that case is there a reason for the 100ms delay? Why not just `post` without delay?

cooltey: Thanks, you're correct. Done!

PR #1413, opened 2020-07-13
Display only local descriptions for English Wikipedia.

dbrant:
https://phabricator.wikimedia.org/T257486


dbrant:
Closing since the API on enwiki no longer serves up descriptions that are non-local, so we don't need to worry about filtering them ourselves.

PR #1417, opened 2020-07-13
Turn ReferenceDialog into a DialogFragment.

dbrant:
The `ReferenceDialog` was actually the only remaining dialog that was inheriting from `BottomSheetDialog` instead of the more proper `BottomSheetDialogFragment`.  This was preventing the Reference dialog from properly inheriting the Activity's theme (try going into Dark mode and looking at a multiple-reference dialog -- the tab dots are messed up).

This also cleans up the logic and interconnectivity between other dialogs and PageFragment/PageActivity.


app/src/main/java/org/wikipedia/page/references/ReferenceDialog.java
```diff
@@ -26,42 +30,58 @@
...
+        return rootView;
+    }
 
-        if (adjacentReferences.size() == 1) {
+    @Override
+    public void onStart() {
+        super.onStart();
+        if (callback().getReferencesGroup().size() == 1) {
             pageIndicatorView.setVisibility(View.GONE);
             ((ViewGroup) pageIndicatorView.getParent()).removeView(pageIndicatorView);
```
cooltey, line 73: If we try to switch to the other app and back, the app will crash.

`void android.view.ViewGroup.removeView(android.view.View)' on a null object reference`

dbrant: Good catch, I don't know why we have that line anyway; it's not doing anything useful.

PR #1419, opened 2020-07-13
Consolidate app themes and remove redundant styles

cooltey:


dbrant:
A worthy consolidation!

PR #1420, opened 2020-07-13
Add elevation tool for ActionBar elevation

cooltey:
Create a method that can update the elevation manually on page with actionBar+list


cooltey:
Close since there's no immediate requirement for this change.

PR #1430, opened 2020-07-15
Add color extraction for large feed card images

cooltey:
We used to have different colors based on thumbnails' main color, and now we can still have that in Glide.


sharvaniharan:
@cooltey code lgtm but i think this was taken out on purpose because the extraction resulted in very weird unrelated colors.. @dbrant if I am mistaken please review and merge.

dbrant:
> @cooltey code lgtm but i think this was taken out on purpose because the extraction resulted in very weird unrelated colors.. @dbrant if I am mistaken please review and merge.

This is what I remember happening, too -- I'm pretty sure the color extraction was removed deliberately.

cooltey:
Got it. Thanks, @sharvaniharan  and @dbrant

PR #1445, opened 2020-07-20
Add SE feed link to "published" snackbar

cooltey:
https://phabricator.wikimedia.org/T257941


app/src/main/java/org/wikipedia/main/MainFragment.java
```diff
@@ -172,6 +172,7 @@ public void onPause() {
         requireContext().registerReceiver(downloadReceiver,
                 new IntentFilter(DownloadManager.ACTION_DOWNLOAD_COMPLETE));
         downloadReceiver.setCallback(downloadReceiverCallback);
+        Prefs.setSEFeedLinkSnackbarShownCount(0);
```
sharvaniharan, line 175: @cooltey Are we resetting to 0 every time MainFragment is resumed?

cooltey: Yes, please see the comment: https://phabricator.wikimedia.org/T257941#6326371

PR #1447, opened 2020-07-21
Use app:endIcon in TextInputLayout for the voice input icon

cooltey:
Sort of pre-processing for this ticket: https://phabricator.wikimedia.org/T256816


cooltey:
> 
> 
> The icon is not visible when the text is empty. It only appears when the first character is typed.

Fixed.

PR #1451, opened 2020-07-22
Significantly consolidate unnecessary background color setting.

dbrant:
(sorry for the size of the PR)
After some of our recent consolidation of styles, it became clear that the actual background color of all Activities is actually `paper_color`, and it should no longer need to be set explicitly as the background of all child fragments.

This reduces overdraw by at least one level in nearly all activities.


cooltey:
💯

PR #1452, opened 2020-07-22
Support adding image tags from Commons filepage

cooltey:
https://phabricator.wikimedia.org/T254142


sharvaniharan:
image
@cooltey ImageTagsProvider is not counting the tags on this. It has other claims than just P180. that is the only difference i see, although our Claim class was supposed to be able to handle that , it returns 'claims' as empty. #1479 fixes it by filtering.

cooltey:
> 
> 
> image
> @cooltey ImageTagsProvider is not counting the tags on this. It has other claims than just P180. that is the only difference i see, although our Claim class was supposed to be able to handle that , it returns 'claims' as empty. #1479 fixes it by filtering.

Thanks and it is done.

app/src/main/java/org/wikipedia/commons/FilePageFragment.kt
```diff
@@ -55,6 +63,28 @@ class FilePageFragment : Fragment() {
...
 
+    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+        super.onActivityResult(requestCode, resultCode, data)
+        if ((requestCode == ACTIVITY_REQUEST_ADD_IMAGE_CAPTION || requestCode == ACTIVITY_REQUEST_ADD_IMAGE_TAGS) && resultCode == RESULT_OK) {
+            val abTestFunnel = ABTestSuggestedEditsSnackbarFunnel()
+            val snackbar = FeedbackUtil.makeSnackbar(activity,
+                    if (requestCode == ACTIVITY_REQUEST_ADD_IMAGE_CAPTION)
+                        getString(R.string.description_edit_success_saved_image_caption_snackbar)
+                    else
+                        getString(R.string.description_edit_success_saved_image_tags_snackbar), FeedbackUtil.LENGTH_DEFAULT)
```
dbrant, line 74: @cooltey  I assume this will be updated to have the same behavior as in #1518.

PR #1460, opened 2020-07-27
Fix: reduce horizontal tap area of LOG OUT in main nav

cooltey:
https://phabricator.wikimedia.org/T258571


app/src/main/res/layout/view_main_drawer.xml
```diff
@@ -36,10 +37,9 @@
         <Button
             android:id="@+id/main_drawer_login_button"
             style="@style/TransparentButton"
-            android:layout_width="0dp"
```
sharvaniharan, line 39: If i remember correctly from the design review for this, I think Robin wanted a little more than the wrap content of the width of the button. May be an 8dp on each side?

cooltey: Not sure if we really need that since the button is quite big. Could you please share the comment from Robin? or maybe we can just merge it and see how it goes?

PR #1470, opened 2020-07-29
Using correct wikisite for network calls in Gallery

sharvaniharan:
When we click on an image in the article, we pass the page title to the gallery. GalleryActivity uses this to get the media-list. Here, the wikisite related to the page is needed. However, when we populate each image item of the list, we need to make calls using the correct wikisite, based on whether the image resides on commons or a wiki. Sending that information while starting the GalleryActivity helps us do this.


sharvaniharan:
#1476 Does it better

PR #1475, opened 2020-07-29
Add new wikis to the app

urbanecm:


dbrant:
Thanks!
(mental note: we should probably create a CI task to run these scripts periodically)

PR #1476, opened 2020-07-29
Fix: Use the correct wikisite while fetching MediaListItem details

sharvaniharan:
When we open the gallery from an article [via LeadImage, Link Preview or in-article media click], We need to send the pagetitle to gallery activity, which will be used to make api calls to fetch medis-list of that article. However, when we make further calls to fetch details of each image, we cannot use the same pagetitle to decide the wiki, because the image might reside in commons. This change introduces the detection of image location[commons v/s wikis] to assist in making that right call.


sharvaniharan:
@dbrant this good to go?

PR #1479, opened 2020-07-29
Filter claims by P180

sharvaniharan:
If image has claims other than P180, ImageTagsProvider.kt doesn't count them. Filtering them at the api call level corrects this.
Ex: https://commons.wikimedia.org/w/api.php?format=json&formatversion=2&errorformat=plaintext&action=wbgetclaims&entity=M87877968
tags count is 0.

### DIFF
--- a/app/src/main/java/org/wikipedia/commons/ImageTagsProvider.kt
+++ b/app/src/main/java/org/wikipedia/commons/ImageTagsProvider.kt
@@ -11,7 +11,7 @@ import java.util.*
 
 object ImageTagsProvider {
     fun getImageTagsObservable(pageId: Int, langCode: String): Observable<Map<String, List<String>>> {
-        return ServiceFactory.get(WikiSite(Service.COMMONS_URL)).getClaims("M$pageId")
+        return ServiceFactory.get(WikiSite(Service.COMMONS_URL)).getClaims("M$pageId", null)
                 .subscribeOn(Schedulers.io())
                 .onErrorReturnItem(Claims())
                 .flatMap { claims ->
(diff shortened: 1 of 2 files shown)


sharvaniharan:
yes... i was not able to zero-in too, because both response structures are same.

app/src/main/java/org/wikipedia/dataclient/Service.java
```diff
@@ -263,7 +263,7 @@
     @NonNull Observable<Entities> getWikidataLabels(@Query("ids") @NonNull String idList,
                                                     @Query("languages") @NonNull String langList);
 
-    @GET(MW_API_PREFIX + "action=wbgetclaims")
+    @GET(MW_API_PREFIX + "action=wbgetclaims&property=P180")
     @NonNull Observable<Claims> getClaims(@Query("entity") @NonNull String entity);
```
dbrant, line 267: Since this function is called `getClaims()`, I would expect it to return all the claims, and I would be confused if it implicitly filters out certain claims.  It would be much more intuitive if we added a `property` parameter to this function, so that the caller can specify which claim(s) they want to get.

app/src/main/java/org/wikipedia/commons/ImageTagsProvider.kt
```diff
@@ -11,7 +11,7 @@ import java.util.*
 
 object ImageTagsProvider {
     fun getImageTagsObservable(pageId: Int, langCode: String): Observable<Map<String, List<String>>> {
-        return ServiceFactory.get(WikiSite(Service.COMMONS_URL)).getClaims("M$pageId")
+        return ServiceFactory.get(WikiSite(Service.COMMONS_URL)).getClaims("M$pageId", null)
```
cooltey, line 14: Should we put `P180` instead of `null`?

sharvaniharan: whoops! sorry was testing.. changed it.

PR #1480, opened 2020-07-29
Fix: open file page if needed

cooltey:
Since we can search `File:` in the search, we should support it.


app/src/main/java/org/wikipedia/page/PageActivity.java
```diff
@@ -468,6 +471,20 @@ public void loadMainPage(TabPosition position) {
         loadPage(title, historyEntry, position);
     }
 
+    private void loadFilePageFromBackStackIfNeeded() {
```
sharvaniharan, line 474: Lgtm. Just one minor question..How do we trigger this? when will a filepage be in the backstack?

cooltey: If the tab count = 1 and tap on the tab view (the one with currently opened tabs), it will take you directly to the page.

PR #1495, opened 2020-08-03
Fix: replace getTitleOriginal() with getTitle() to prevent failing on loading media-list

cooltey:
A similar issue as #1490. Found that if we turn off the link preview, it will use the incorrect title to request endpoint when loading the Gallery sometimes. 

Steps to reproduce:
1. Turn off link preview in settings.
2. Go to https://en.wikipedia.org/wiki/List_of_countries_by_military_expenditures in Traditional Chinese
3. Click on [開發中國家] in-article link in the first paragraph. 
4. Click on the first image in article.


cooltey:
Updated the steps to reproduce in the PR description.

PR #1500, opened 2020-08-04
Use Core KTX functions.

Isira-Seneviratne:
* Switch to the Core KTX library.
* Use `ContextCompat.getSystemService()` and the `getSystemService()` extension function to retrieve system services without casting.
* Make use of other extension functions such as `String.toUri()`.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -29,6 +29,9 @@ android {
         sourceCompatibility = JAVA_VERSION
         targetCompatibility = JAVA_VERSION
     }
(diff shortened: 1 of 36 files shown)


dbrant:
Once again, this is mostly change for the sake of change, with no discernible benefit.
The only useful thing might be the use of `ContextCompat.getSystemService()`. If this PR is reduced to just that change, then it might be mergeable.

dbrant:
Please see my comments above.

PR #1478, opened 2020-07-29
Improve save to reading list UX

cooltey:
https://phabricator.wikimedia.org/T256821

TODO: update revision number once the schema is updated
https://phabricator.wikimedia.org/T256821#6352975

PR #1505, opened 2020-08-04
Replace simple bottom sheet menu with generic list view context menu in On This Day

cooltey:
This is an extra PR of #1478  that replaces the bottom sheet menu with a more generic list view context menu, which makes more sense to me.


sharvaniharan:
Lgtm @cooltey  one minor question i had..
image

Is that a thing to 'Move' from a list to another? not just for 'Saved' list, but for all lists?

cooltey:
> Lgtm @cooltey one minor question i had..
> 
> Is that a thing to 'Move' from a list to another? not just for 'Saved' list, but for all lists?
That's an expected behavior that if you save the article to another list, you still can move it to other list.

PR #1508, opened 2020-08-05
Follow up: set up UserStats tooltips with the updated user stats

cooltey:
Otherwise, the reverted edits count will always be `0`.


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -298,6 +285,23 @@ class SuggestedEditsTasksFragment : Fragment() {
         tasksContainer.visibility = VISIBLE
     }
 
+    private fun setUserStatsTooltips() {
```
dbrant, line 288: @cooltey  This function does much more than set up tooltips.  Can we give it a more meaningful name or break it up?

cooltey: Sure! Will do it.

PR #1512, opened 2020-08-07
Better, Unified search

sharvaniharan:

- Added search tab
- Removed history tab


app/src/main/java/org/wikipedia/settings/Prefs.java
```diff
@@ -957,5 +957,12 @@ public static int getSEFeedLinkSnackbarShownCount() {
         return getInt(R.string.preference_key_suggested_edits_feed_link_snackbar_shown_count, 0);
     }
 
+    public static boolean shouldShowSearchTabTooltip() {
+        return getBoolean(R.string.preference_key_pulsating_icon_event_sent, true);
```
cooltey, line 961: Can we have this also in the `developer_preferences.xml`?

app/src/main/res/layout/item_search_result.xml
```diff
@@ -61,12 +61,37 @@
...
+        android:id="@+id/page_list_button_tabs"
+        android:layout_width="16dp"
+        android:layout_height="16dp"
+        android:layout_marginStart="16dp"
+        android:layout_marginEnd="16dp"
+        android:contentDescription="@null"
+        android:foregroundGravity="center_vertical"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toStartOf="@+id/page_list_item_image"
+        app:layout_constraintTop_toTopOf="parent" />
```
cooltey, line 88: Should we make it not clickable?

app/src/main/java/org/wikipedia/main/MainFragment.java
```diff
@@ -596,4 +602,14 @@ public void aboutClick() {
         }
     }
 
+    @SuppressWarnings("checkstyle:magicnumber")
+    void maybeShowOneTimeTooltip(NavTab tab) {
```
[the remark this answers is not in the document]
sharvaniharan: Did you mean have the full code directly there?

app/src/main/java/org/wikipedia/navtab/NavTabLayout.kt
```diff
@@ -42,4 +42,15 @@ class NavTabLayout constructor(context: Context, attrs: AttributeSet) : BottomNa
...
+    fun getViewAt(position: Int): View? {
+        var view: View? = null
+        if (childCount > 0) {
+            val menuView = getChildAt(0)
+            if ((menuView as ViewGroup).childCount > position + 1) {
+                view = menuView.getChildAt(position)
+            }
+        }
+        return view
+    }
```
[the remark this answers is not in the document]
sharvaniharan: ah! got it.. i can get this by the view id... thanks.

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
```diff
@@ -578,5 +650,88 @@ private Callback callback() {
     private String getSearchLanguageCode() {
         return ((SearchFragment) getParentFragment()).getSearchLanguageCode();
     }
+
+    private void gatherSearchResultsFromAppSources() {
```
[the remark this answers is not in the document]
sharvaniharan: Whoops sorry missed this @cooltey ... I think whether we are searching cache, prefix or full text we will also need to check our app sources [tabs, lists and history]. Since we only go to full text search after the. title search returns empty, that approach wouldn't work.

cooltey:
> 
> 
> It will get a little bit laggy when there has many history items, e.g. search tab ripple animation and long-press action on history items.

Hi @sharvaniharan, I am still seeing the performance issue in this PR.

app/src/main/res/layout/view_history_list_container.xml
```diff
@@ -0,0 +1,85 @@
...
+                android:contentDescription="@string/button_clear_all_recent_searches"
+                android:focusable="true"
+                android:foreground="?attr/actionBarItemBackground"
+                android:longClickable="true"
+                android:scaleType="centerInside"
+                app:tint="@color/base30"
+                app:srcCompat="@drawable/ic_delete_white_24dp" />
+        </LinearLayout>
+
+        <androidx.recyclerview.widget.RecyclerView
```
dbrant, line 78: This `RecyclerView` is now rendering all of its contents (off screen), instead of stopping at the end of the screen.  If the user has 1000 history items, it will render all 1000 of them, grinding the app to a halt.

As we've done in the past, the proper thing to do is put the "Search" button as a different item type in the RecyclerView.

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
```diff
@@ -578,5 +650,88 @@ private Callback callback() {
...
+        if (loaderManager == null) {
+            loaderManager = LoaderManager.getInstance(requireActivity()).initLoader(SEARCH_FRAGMENT_HISTORY_LOADER_ID, null, loaderCallback);
+        } else {
+            LoaderManager.getInstance(requireActivity()).restartLoader(SEARCH_FRAGMENT_HISTORY_LOADER_ID, null, loaderCallback);
+        }
+    }
+
+    private void addSearchResultFromReadingLists() {
+        List<SearchResult> searchResultsFromAppSources = new ArrayList<>();
+        List<Object> list = ReadingListBehaviorsUtil.INSTANCE.applySearchQuery(currentSearchTerm, ReadingListDbHelper.instance().getAllLists());
```
dbrant, line 670: This is a bit wasteful because it loads the entire collection of reading lists into memory, just for the purpose of finding a single page.  You could just query the database table directly, and find a single result.  The `ReadingListDbHelper` class has a function that _almost_ does what we need (`findPageInAnyList`), so you can model it from that.

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
```diff
@@ -578,5 +650,88 @@ private Callback callback() {
...
+                SearchResult searchResult = new SearchResult(tab.getBackStackPositionTitle());
+                searchResult.setSearchResultTypeWithPriority(TAB_LIST_SEARCH_RESULT);
+                searchResultsFromAppSources.add(searchResult);
+                displayResults(searchResultsFromAppSources);
+                return;
+            }
+        }
+    }
+
+    private class LoaderCallback implements LoaderManager.LoaderCallbacks<Cursor> {
```
dbrant, line 698: Since we're only looking for a single history entry, it's not really necessary to muck around with `CursorLoader`. You can just query the database table directly to find a single result. (And `zip` it into an Rx chain with the other queries)

app/src/main/java/org/wikipedia/history/HistoryDbHelper.kt
```diff
@@ -0,0 +1,47 @@
...
+            selection = "UPPER($titleCol) LIKE UPPER(?) ESCAPE '\\'"
+            selectionArgs = arrayOf("%$searchStr%")
+        }
+        db.query(PageHistoryContract.PageWithImage.TABLES, null,
+                selection,
+                selectionArgs,
+                null, null, null).use { cursor ->
+            if (cursor.moveToFirst()) {
+                val searchResults = ArrayList<SearchResult>();
+                val indexedEntry = IndexedHistoryEntry(cursor);
```
cooltey, line 32: Please remove `;`.

app/src/main/java/org/wikipedia/history/HistoryDbHelper.kt
```diff
@@ -0,0 +1,47 @@
...
+                val pageTitle: PageTitle = indexedEntry.entry.getTitle()
```
cooltey, line 33: Please use `entry.title`

app/src/main/java/org/wikipedia/main/MainFragment.java
```diff
@@ -570,4 +574,13 @@ public void aboutClick() {
...
+       /* View tabView = tabLayout.getViewAt(tab.code());
+        if (tabView != null) {
+            FeedbackUtil.showTooltip(tabView, getString(R.string.search_tab_tooltip), true, true);
+            Prefs.setShowSearchTabTooltip(false);
+        }*/
```
cooltey, line 583: Unused code?

app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
```diff
@@ -487,6 +493,34 @@ public ReadingListPage findPageInAnyList(@NonNull PageTitle title) {
...
+            selection = "UPPER(" + titleCol + ") LIKE UPPER(?) ESCAPE '\\'";
+            selectionArgs = new String[]{"%" + normalizedQuery + "%"};
+        }
+        try (Cursor cursor = db.query(ReadingListPageContract.TABLE, null,
+                selection,
+                selectionArgs,
+                null, null, null)) {
+            if (cursor.moveToFirst()) {
+                ReadingListPage readingListPage = ReadingListPage.DATABASE_TABLE.fromCursor(cursor);
+                SearchResult searchResult = new SearchResult(new PageTitle(((ReadingListPage) readingListPage).title(), ((ReadingListPage) readingListPage).wiki(), ((ReadingListPage) readingListPage).thumbUrl()));
```
cooltey, line 515: Please remove redundant casting

app/src/main/java/org/wikipedia/search/SearchResult.java
```diff
@@ -12,9 +12,37 @@
...
+
+    public void setSearchResultTypeWithPriority(@NonNull SearchResultTypeWithPriority searchResultTypeWithPriority) {
+        this.searchResultTypeWithPriority = searchResultTypeWithPriority;
+    }
+
+    public int getPriority() {
+        return searchResultTypeWithPriority.getPriority();
+    }
+
+    public enum SearchResultTypeWithPriority {
```
cooltey, line 30: Not sure but would it be good if just name it as `SearchResultType`?

app/src/main/java/org/wikipedia/main/MainFragment.java
sharvaniharan: Whoops!

app/src/main/java/org/wikipedia/history/HistoryFragment.java
```diff
@@ -413,6 +375,78 @@ void bindItem(@NonNull String date) {
...
+        }
+    }
+
+    private class SearchCardViewHolder extends DefaultViewHolder<View> {
+        WikiCardView searchWikiCardView;
+        View searchCardView;
+        AppCompatImageView voiceSearchButton;
+        ImageView historyFilterButton;
+        ImageView clearHistoryButton;
+        SearchHeaderCallback searchHeaderCallback;
```
cooltey, line 427: Some varibles can be local varibles, and some varibles can add `private` on them.

app/src/main/java/org/wikipedia/main/MainFragment.java
```diff
@@ -137,6 +138,9 @@ public static MainFragment newInstance() {
             if (getCurrentFragment() instanceof FeedFragment && item.getOrder() == 0) {
                 ((FeedFragment) getCurrentFragment()).scrollToTop();
             }
+            if (getCurrentFragment() instanceof HistoryFragment && item.getOrder() == NavTab.SEARCH.code()) {
+                openSearchActivity(NAV_MENU, null);
```
cooltey, line 142: Can we add `return` below the `openSearchActivity()` to prevent unexpected action of scrolling to the top in the history page?

app/src/main/res/layout/item_search_result.xml
```diff
@@ -61,12 +61,38 @@
...
+        android:layout_marginEnd="16dp"
+        android:contentDescription="@null"
+        android:foregroundGravity="center_vertical"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toStartOf="@+id/page_list_item_image"
+        app:layout_constraintTop_toTopOf="parent"
+        app:tint="?attr/material_theme_de_emphasised_color"
+        app:srcCompat="@drawable/ic_bookmark_border_white_24dp" />
+
+    <org.wikipedia.views.TabCountsView
```
cooltey, line 78: Should we remove the ripple animation from it since the item is not clickable?

app/src/main/res/layout/item_search_result.xml
```diff
@@ -61,12 +61,38 @@
         app:layout_constraintTop_toBottomOf="@+id/page_list_item_redirect"
         tools:text="Lorem ipsum" />
 
+    <ImageView
```
cooltey, line 64: Looks like the view and the `TabCountsView` are not aligning to the center correctly. You can see the issue when other results have thumbnails and these views are slightly aligning to the left.

app/src/main/java/org/wikipedia/history/HistoryFragment.java
```diff
@@ -578,31 +642,9 @@ protected Context getParentContext() {
...
-            actionMode = null;
-            super.onDestroyActionMode(mode);
-        }
-    }
-
     @Nullable private Callback callback() {
         return FragmentUtil.getCallback(this, Callback.class);
     }
+
+    private static class Search{ }
```
cooltey, line 649: Would it be better if name it as `SearchBar`? I was confused by the naming. Also, it would be good if adds a space after the name like: `SearchBar { }`

app/src/main/java/org/wikipedia/main/MainFragment.java
```diff
@@ -137,6 +138,10 @@ public static MainFragment newInstance() {
...
+                return true;
```
cooltey, line 143: Maybe I was wrong because I am still seeing the content scroll to the top after back to the History page.

cooltey (changes requested):

app/src/main/java/org/wikipedia/main/MainFragment.java
sharvaniharan: @cooltey do you mean when you return back from search? during tab change it is not scrolling up.

app/src/main/res/layout/item_search_result.xml
sharvaniharan: n
@cooltey Do you mean in cases like. this. history icon and TabCountsView arent aligned?

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
```diff
@@ -347,25 +389,56 @@ private void displayResults(List<SearchResult> results) {
...
-            searchEmptyView.setVisibility(View.GONE);
-            searchResultsContainer.setVisibility(View.VISIBLE);
+            searchEmptyView.setVisibility(GONE);
+            searchResultsContainer.setVisibility(VISIBLE);
         }
 
         getAdapter().notifyDataSetChanged();
     }
 
+    private void insertSearchResultInCorrectPosition(@NonNull SearchResult newResult) {
```
dbrant, line 421: Not sure about the purpose of reshuffling the search results... If we make sure to put them in the correct order in the initial `prefixSearch` response, do we need to worry about reordering them again?

app/src/main/res/drawable/ic_baseline_history_24.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:autoMirrored="true" android:height="24dp"
```
dbrant, line 1: Any reason not to use the existing icon (`ic_restore_black_24dp`) that we were using for the History tab?

dbrant (changes requested):

app/src/main/res/layout/item_search_result.xml
cooltey: > @cooltey Do you mean in cases like this history icon and TabCountsView arent aligned?


You'll see it when there are no thumbnails on the histroy/tab items.
Screenshot (Sep 3, 2020 08 32 47)

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
sharvaniharan: @dbrant this only looks at the first 3 spots, and only for non search items. Inserting the tab, history and list item at their correct place will not work always because of cases like having to delete the less priority duplicate, a new item, say history,  getting added after a visit to the article page and back to search, or having only one of those 3 types of items, etc.

dbrant:
Merging to keep things moving, but see follow-up.

dbrant (commented):
@sharvaniharan  A couple of nitpicks I noticed post-merge.

dbrant:
@sharvaniharan  ^ just making sure you've seen my comments above. (even though this PR is already merged)

PR #1514, opened 2020-08-07
Change TabLayout background to prevent incorrect color in certain devices

cooltey:
https://phabricator.wikimedia.org/T259892

This also applies the correct ripple color to the `TabLayout`. If you re-tap on the selected indicator in `ReferenceDialog`, you'll see the purple ripple color, which is incorrect.

### DIFF
--- a/app/src/main/res/layout/fragment_onboarding_pager.xml
+++ b/app/src/main/res/layout/fragment_onboarding_pager.xml
@@ -42,6 +42,7 @@
                 android:layout_gravity="center"
                 android:background="?attr/paper_color"
                 app:tabIndicator="@null"
+                app:tabRippleColor="?attr/color_group_59"
                 app:tabPaddingStart="8dp"
                 app:tabPaddingEnd="8dp"
                 app:tabBackground="@drawable/shape_tab_dot"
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/fragment_onboarding_pager.xml
```diff
@@ -40,8 +40,9 @@
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 android:layout_gravity="center"
-                android:background="@android:color/transparent"
+                android:background="?attr/paper_color"
```
dbrant, line 43: @cooltey  The issue was already fixed, and applying `paper_color` here is unnecessary.

app/src/main/res/layout/fragment_references_pager.xml
```diff
@@ -54,8 +54,9 @@
             android:layout_width="wrap_content"
             android:layout_height="48dp"
             android:layout_gravity="center"
-            android:background="@android:color/transparent"
+            android:background="?attr/paper_color"
```
dbrant, line 57: The issue was already fixed, and applying `paper_color` here is unnecessary.

PR #1516, opened 2020-08-07
Making arrow orientation in tooltip configurable

sharvaniharan:
Need this to be configurable for cases like https://phabricator.wikimedia.org/T259717 where the arrow needs to be show in a different orientation based on the balloon position

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
@@ -304,12 +304,11 @@ class SuggestedEditsTasksFragment : Fragment() {
     }
 
     private fun showOneTimeSequentialUserStatsTooltips() {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -303,11 +304,12 @@ class SuggestedEditsTasksFragment : Fragment() {
     }
 
     private fun showOneTimeSequentialUserStatsTooltips() {
-        val balloon = FeedbackUtil.showTooltip(requireContext(), contributionsStatsView.tooltipText)
+        val arrowOrientation = ArrowOrientation.BOTTOM
```
cooltey, line 307: If we have already imported it, is it necessary to create an extra `val` for it instead of using it directly?

sharvaniharan:
done @cooltey

dbrant:
??? This totally breaks the current tooltips shown for the user stats. Please revert.

image

PR #1540, opened 2020-08-13
Prototype of Slices

cooltey:
https://developer.android.com/jetpack/androidx/releases/slice
https://developer.android.com/guide/slices/getting-started


cooltey:
Close.

PR #1541, opened 2020-08-16
Rename HistoryFragment since it is hosted in Search tab?

sharvaniharan:
Note: base is not master.


sharvaniharan:
@cooltey Agree. I felt the same. The only place it is a bit iffy is in Navtab, where for SEARCH enum, we instantiate HistoryFragment... not very readable...  but if you feel that is not a problem I will squash this PR.  Cc @dbrant

dbrant:
Agreed; this fragment is still the History fragment, it just has a Search button in it now.

PR #1547, opened 2020-08-18
Set up A/B test for showing Home vs. Explore menu item.

dbrant:
https://phabricator.wikimedia.org/T260332

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/ABTestExploreVsHomeFunnel.java
+++ b/app/src/main/java/org/wikipedia/analytics/ABTestExploreVsHomeFunnel.java
@@ -1,0 +1,20 @@
+package org.wikipedia.analytics;
+
+public final class ABTestExploreVsHomeFunnel extends ABTestFunnel {
+    public ABTestExploreVsHomeFunnel() {
+        super("exploreVsHome", ABTestFunnel.GROUP_SIZE_2);
+    }
+
+    public boolean shouldSeeHome() {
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/analytics/ABTestExploreVsHomeFunnel.java
```diff
@@ -0,0 +1,20 @@
+package org.wikipedia.analytics;
+
+public final class ABTestExploreVsHomeFunnel extends ABTestFunnel {
+    public ABTestExploreVsHomeFunnel() {
+        super("exploreVsHome", ABTestFunnel.GROUP_SIZE_2);
+    }
+
+    public boolean shouldSeeHome() {
+        boolean enrolled = isEnrolled();
+        boolean showHome = getABTestGroup() == ABTestFunnel.GROUP_1;
```
cooltey, line 10: Should we have `ReleaseUtil.isPreBetaRelease()` here for testing purposes?

dbrant: We have developer preferences for testing purposes now. (set the developer preference to an even number for the A group, or an odd number for the B group.)

cooltey: I see. I set it as `123` and that's why it does not show for me 😅

PR #1554, opened 2020-08-20
Follow-up to Tooltip on suggested edits.

dbrant:
This does all of the following:
* Enables the tooltip to be non-modal, meaning that the user can still interact with items _outside_ of the tooltip, e.g. tapping the "Edits" tab while the tooltip is still visible.
* The non-modal tooltip will be dismissed upon any touch event outside of it.
* Rearranged the tooltip code a bit, so that all of the `skydoves`-specific logic is inside the `FeedbackUtil` class. In case we ever need to switch to another tooltip library, it will be more convenient to have all the dependent logic in one place.


app/src/main/java/org/wikipedia/main/MainActivity.java
```diff
@@ -207,4 +219,16 @@ protected void setToolbarElevationDefault() {
...
     }
+
+    private void dismissCurrentTooltip() {
+        if (currentTooltip != null) {
+            currentTooltip.dismiss();
+        }
+        currentTooltip = null;
+    }
+
+    public void setCurrentTooltip(Balloon tooltip) {
```
cooltey, line 230: Can we have a `@NonNull` for the parameter?

app/src/main/res/layout/activity_main.xml
```diff
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<org.wikipedia.views.LinearLayoutTouchIntercept xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
     android:orientation="vertical"
+    android:id="@+id/main_container"
```
cooltey, line 8: Would it be better if move `android:id` to the top of `android:layout_width`?

app/src/main/java/org/wikipedia/views/LinearLayoutTouchIntercept.java
```diff
@@ -0,0 +1,34 @@
+package org.wikipedia.views;
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.view.MotionEvent;
+import android.widget.LinearLayout;
+
+public class LinearLayoutTouchIntercept extends LinearLayout {
```
cooltey, line 8: How about converting it to Kotlin

app/src/main/res/layout/activity_main.xml
dbrant: Sure!

app/src/main/java/org/wikipedia/views/LinearLayoutTouchIntercept.java
dbrant: Yep!

app/src/main/java/org/wikipedia/main/MainActivity.java
dbrant: ok!

PR #1566, opened 2020-08-25
Design tweak: Tooltip design

sharvaniharan:
- Shows the suggested edits tooltip only when edits is not the current tab
- Uses correct font for tooltip
- Scrolls to top before showing sequential tooltips
- Fixes the bug with first time user onboarding image shown on top of the tasks screen.


app/src/main/java/org/wikipedia/main/MainFragment.java
```diff
@@ -141,7 +141,9 @@ public static MainFragment newInstance() {
             return true;
         });
 
-        maybeShowEditsTooltip();
+        if (!(getCurrentFragment() instanceof SuggestedEditsTasksFragment)) {
+            maybeShowEditsTooltip();
+        }
```
cooltey, line 146: Would it be better if move this logic into the `mabyShowEditsTooltip()` method?

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -308,6 +299,7 @@ class SuggestedEditsTasksFragment : Fragment() {
     }
 
     private fun showOneTimeSequentialUserStatsTooltips() {
+        suggestedEditsScrollView.fullScroll(View.FOCUS_UP);
```
cooltey, line 302: Please remove the `;`.

Minor question: would it be possible to show the tooltips after the view is scrolled to the top?

sharvaniharan: This line scrolls it to the top @cooltey ..

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -308,6 +299,7 @@ class SuggestedEditsTasksFragment : Fragment() {
     }
 
     private fun showOneTimeSequentialUserStatsTooltips() {
+        suggestedEditsScrollView.post(Runnable { suggestedEditsScrollView.fullScroll(View.FOCUS_UP) })
```
cooltey, line 302: Still seeing the tooltips jumping from outside of the card view to the position it should be in, but I guess it should be fine. @dbrant Any thoughts?

sharvaniharan: I added a half second delay before starting the tooltip relay. Looks much better to me. Lmk your thoughts @cooltey and @dbrant

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -308,12 +299,15 @@ class SuggestedEditsTasksFragment : Fragment() {
...
 
     private fun showOneTimeSequentialUserStatsTooltips() {
-        val balloon = FeedbackUtil.getTooltip(requireContext(), contributionsStatsView.tooltipText, false, true)
-        balloon.showAlignBottom(contributionsStatsView.description)
-        balloon.relayShowAlignBottom(FeedbackUtil.getTooltip(requireContext(), editStreakStatsView.tooltipText, false, true), editStreakStatsView.description)
-                .relayShowAlignBottom(FeedbackUtil.getTooltip(requireContext(), pageViewStatsView.tooltipText, false, true), pageViewStatsView.description)
-                .relayShowAlignBottom(FeedbackUtil.getTooltip(requireContext(), editQualityStatsView.tooltipText, false, true), editQualityStatsView.description)
-        Prefs.shouldShowOneTimeSequentialUserStatsTooltip(false)
+        suggestedEditsScrollView.fullScroll(View.FOCUS_UP)
+        this.view?.postDelayed({
```
cooltey, line 303: Just curious that why don't use `suggestedEditsScrollView.postDelayed`?

cooltey (approved):
The `postDelayed` looks good to me.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
dbrant, line 303: 500ms is plenty of time for the user to close the activity just before this `post` gets executed. What will happen in that case? And how do we usually prevent it?

dbrant (changes requested):

sharvaniharan:
done @cooltey @dbrant

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -60,6 +60,17 @@ class SuggestedEditsTasksFragment : Fragment() {
     private var latestEditStreak = 0
     private var revertSeverity = 0
 
+    private val sequentialTooltipRunnable = Runnable {
+        if (isAdded) {
```
cooltey, line 64: Should we have style as the following?
```
if (!isAdded) {
    return
}
```

cooltey (approved):

PR #1572, opened 2020-08-26
Indicate search results from other languages

cooltey:
https://phabricator.wikimedia.org/T260433


app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
```diff
@@ -322,13 +335,32 @@ private void doFullTextSearch(final String searchTerm,
...
                     // full text special:
                     SearchResultsFragment.this.lastFullTextResults = results;
-                    displayResults(resultList);
+                    if (!resultList.isEmpty()) {
+                        displayResults(resultList);
+                    }
+                    return resultList.isEmpty() ? doFullTextSearchResultsCountObservable(searchTerm) : Observable.empty();
+                })
+                .toList()
+                .doAfterTerminate(() -> updateProgressBar(false))
```
dbrant, line 344: When you return an `empty()` observable, it looks like this part is never called, and the progress bar keeps spinning forever.

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
```diff
@@ -438,21 +471,77 @@ public void onAddToList(@NonNull PageTitle title, @NonNull InvokeSource source)
...
+            }
+        }
+    }
+
+    private class NoSearchResultItemViewHolder extends DefaultViewHolder<View> {
+        NoSearchResultItemViewHolder(@NonNull View itemView) {
+            super(itemView);
+        }
+
+        private ColorStateList accentColorStateList = new ColorStateList(
```
dbrant, line 515: There's a much simpler convenience method: `ColorStateList.valueOf(...)`

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
```diff
@@ -66,6 +70,7 @@
     }
 
     private static final int BATCH_SIZE = 20;
+    private static final int LARGE_BATCH_SIZE = 100;
```
dbrant, line 73: Why do we need such a large batch size here? This is pretty resource-intensive for the server.

dbrant (changes requested):

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
cooltey: I put `L.d()` in `doAfterTerminate()` and `subscribe()`, both of the logs will show on the Logcat. I believe the `toList()` will create an empty list for the `empty()` and it will not block the flow. 

If I remove the `toList()`, the flow will not go to `subscribe()` but will actually jump to the `doAfterTerminate()`.

BTW, there's a `Observable.never()` will block the flow that includes the `doAfterTerminate()`.

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
cooltey: Thanks! This is much better.

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
cooltey: Make sense. Removed.

dbrant (changes requested):
The progress bar still remains visible even when the search is complete.
- Open up search
- Search for "foo"

cooltey:
> 
> 
> The progress bar still remains visible even when the search is complete.
> 
>     * Open up search
> 
>     * Search for "foo"

Done!

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
```diff
@@ -218,7 +232,7 @@ private void doTitlePrefixSearch(final String searchTerm, boolean force) {
                     searchErrorView.setVisibility(GONE);
                     handleResults(results, searchTerm, startTime);
                 }, caught -> {
-                    searchEmptyView.setVisibility(GONE);
+                    searchErrorView.setVisibility(View.VISIBLE);
```
dbrant, line 235: This line is now identical to the next line. What was the rationale of changing it?

dbrant (changes requested):

app/src/main/java/org/wikipedia/search/SearchResultsFragment.java
cooltey: Done. I guess something went wrong when merging the `master` branch.

PR #1577, opened 2020-08-31
Wire up seen/unseen logic for Talk topics.

dbrant:


app/src/main/java/org/wikipedia/talk/TalkPageSeenDatabaseTable.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.wikipedia.database.DbUtil
+import org.wikipedia.database.column.Column
+import org.wikipedia.database.column.LongColumn
+import org.wikipedia.database.column.StrColumn
+import org.wikipedia.database.contract.AppContentProviderContract
+import org.wikipedia.dataclient.page.TalkPage.Topic
+
+private const val TABLE = "talkpageseen"
+private const val PATH = "talkpage/seen"
+private val URI = Uri.withAppendedPath(AppContentProviderContract.AUTHORITY_BASE, PATH)
```
cooltey, line 18: Having the variables above the `object` looks slightly weird to me, not sure if it is possible to make this file as a `class` and use `companion object` to reach the same approach?

dbrant: I think that would make it more complicated. Keeping this class as an `object` gives us an automatic singleton `INSTANCE` of this class that we need to use in other places.  And top-level value declarations seem to be perfectly valid in Kotlin, and useful for these kinds of cases.

app/src/main/res/layout/item_talk_topic.xml
```diff
@@ -13,9 +13,11 @@
         android:id="@+id/topicReadDot"
         android:layout_width="8dp"
         android:layout_height="8dp"
+        android:layout_marginEnd="12dp"
         app:layout_constraintTop_toTopOf="parent"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintEnd_toStartOf="@id/topic_title_text"
```
cooltey, line 20: Please update it to `@id/topicTitleText`

cooltey (changes requested):

app/src/main/res/layout/item_talk_topic.xml
dbrant: whoops

PR #1580, opened 2020-08-31
Add extra CSS to the header in mobile-web for Main page

cooltey:
> 02) Make sure titles on main pages (new tab) are not cut off by the app bar (e.g. by applying more margin)

https://phabricator.wikimedia.org/T260871

### DIFF
--- a/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
+++ b/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
@@ -138,10 +138,14 @@ object JavaScriptActionHandler {
     }
 
     @JvmStatic
-    fun mobileWebChromeShim(): String {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -138,10 +138,14 @@ object JavaScriptActionHandler {
     }
 
     @JvmStatic
-    fun mobileWebChromeShim(): String {
+    fun mobileWebChromeShim(isMainPage: Boolean): String {
+        var extraHeaderCSS = ""
+        if (isMainPage) {
```
dbrant, line 143: Any reason this is only for the main page?

cooltey: No. Removed it, thanks!

PR #1581, opened 2020-08-31
Add a null check for the revision in PageFragment

cooltey:


dbrant:
Just so this is clear, when will this happen? and what issue would it cause?

cooltey:
> Just so this is clear, when will this happen? 

Not sure I have specific reproducible steps, but I guess it happens when tapping on in-article links (with link preview off) on the main page and keep tapping in-article links and it will happen.

> and what issue would it cause?
Just an error message in the log.

PR #1584, opened 2020-09-01
Implement replies and new topics in Talk pages.

dbrant:


app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
```diff
@@ -162,10 +223,119 @@ class TalkTopicActivity : BaseActivity() {
...
+        }
+
+        // if the message is not signed, then sign it explicitly
+        if (!body.endsWith("~~~~")) {
+            body += " ~~~~"
+        }
+        if (!isNewTopic()) {
+            // add two explicit newlines at the beginning, to delineate this message as a new paragraph.
+            body = "\n\n" + body
+        }
```
cooltey, line 260: If there's no replies in a new topic, the first reply will stick with the topic.

app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
```diff
@@ -162,10 +223,119 @@ class TalkTopicActivity : BaseActivity() {
         }
     }
 
+    inner internal class ReplyTextWatcher: TextWatcher {
```
cooltey, line 226: `inner` can be removed.

cooltey (changes requested):

app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
dbrant: it cannot.

app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
dbrant: should be fixed.

app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
cooltey: You're correct. Would it be possible to add `@Suppress("RedundantInnerClassModifier")` for it?
```
    @Suppress("RedundantInnerClassModifier")
    internal inner class ReplyTextWatcher: TextWatcher {
```

dbrant: yep

cooltey (approved):
LGTM. @sharvaniharan Please review and merge if it looks good to you too.

cooltey:
@sharvaniharan I just merged it. Please create a follow-up if you see issues, thanks!

PR #1587, opened 2020-09-01
Feed Header and Footer Re-design changes

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T261667

Note: The ActionFooterView usage is not consistent. Some cards have used their own footer. So, I have just created a brand-new footer view, to be used by all cards, and subsequently the other footerviews should be deleted.


app/src/main/java/org/wikipedia/feed/view/CardFooterView.kt
```diff
@@ -0,0 +1,29 @@
...
+internal class CardFooterView constructor(context: Context, attrs: AttributeSet? = null) : ConstraintLayout(context, attrs) {
+    interface Callback {
+        fun onFooterClicked()
+    }
+
+    var callback: Callback? = null
+
+    init {
+        View.inflate(context, R.layout.view_card_footer, this)
+        ButterKnife.bind(this)
```
cooltey, line 20: Unused?

cooltey (approved):
Looks good to me. Just have one miner comment.

sharvaniharan:
done @cooltey

PR #1590, opened 2020-09-02
Revert "Launch PageActivity in default mode to maintain back-stack as expected"

dbrant:
Reverts wikimedia/apps-android-wikipedia#1586

There are very specific and extremely important reasons why `PageActivity` is set to be `singleTask`.
This would introduce much more confusion than the issue that it attempts to fix.


cooltey:
Merged. @sharvaniharan Please consider making another approach for the ticket.

PR #1591, opened 2020-09-02
Redesign featured article card & update footer layout

cooltey:
https://phabricator.wikimedia.org/T261668


cooltey:
Hi @sharvaniharan, please consider reviewing the PR as a higher priority because I updated the footer style to match the design spec.

sharvaniharan (commented):
@cooltey  Lgtm, Just one minor change needed for readability:
While openeing the mainpage from the foorter of Featuredarticle card, could we make this a method call like "openMainPage()" and anywhere else when we try to open the MainPage, can we just calculate the. pageTitle right there? Accessing the static method of Featured article to get main page is very confusing when we see it out of context in MainFragment and PageFragment.

cooltey:
> 
> 
> @cooltey Lgtm, Just one minor change needed for readability:
> While openeing the mainpage from the foorter of Featuredarticle card, could we make this a method call like "openMainPage()" and anywhere else when we try to open the MainPage, can we just calculate the. pageTitle right there? Accessing the static method of Featured article to get main page is very confusing when we see it out of context in MainFragment and PageFragment.


Make sense to me. I committed a change to make the main page `PageTitle` separately.

PR #1592, opened 2020-09-02
Load page in current stack when searched from PageActivity

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T261322


app/src/main/java/org/wikipedia/search/SearchFragment.java
```diff
@@ -301,6 +302,7 @@ public void navigateToTitle(@NonNull PageTitle title, boolean inNewTab, int posi
         funnel.searchClick(position, searchLanguageCode);
         HistoryEntry historyEntry = new HistoryEntry(title, HistoryEntry.SOURCE_SEARCH);
         startActivity(inNewTab ? PageActivity.newIntentForNewTab(requireContext(), historyEntry, historyEntry.getTitle())
+                : invokeSource == PAGE_ACTION_TAB ? PageActivity.newIntentForCurrentTab(requireContext(), historyEntry, historyEntry.getTitle(), false)
```
dbrant, line 305: You're correctly fixing the issue by passing `false` into the `squashBackstack` parameter.  But why do we want to do it only for PAGE_ACTION_TAB, and not for all other cases?

sharvaniharan:
done @dbrant

PR #1600, opened 2020-09-08
Follow-up to Better Unified Search.

dbrant:
Notice the number of lines removed.
@sharvaniharan  I have removed all the logic that re-shuffles the search results by "priority". Could you let me know why having that logic was important, and how removing it makes the search results behave differently?


sharvaniharan:
Problem with not shuffling:
1. Wrong priority in both cases:
priority
wrong pr

Also, When the history and tabs result is the same, tabs take priority, so in the shuffling logic, if a same pagetitle, but with higher priority comes in, it needs to be swapped not ignored.

prio

dbrant:
@sharvaniharan  I'm afraid I'm still not seeing the issue.  Looking in Zeplin, the priority seems to be Reading Lists -> History -> Tabs.
The screenshots you posted reflect those priorities correctly.  Are the priorities specified differently somewhere else?

sharvaniharan:
I remember the priority to be Tabs -> Reading lists-> History from a call. It was decided after the zeplin. I do not see it recorded, but I will verify it with Robin. However, The second issue will still remain. The one where you will need to display the higher priority item when we have two of the same pageTitle and different priorities

dbrant:
> However, The second issue will still remain. The one where you will need to display the higher priority item when we have two of the same pageTitle and different priorities

If I got the priorities right in this PR, then this issue disappears.

PR #1602, opened 2020-09-09
Redesign on this day card

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T261673
pulse : #1647


app/src/main/res/layout/view_on_this_day_footer.xml
```diff
@@ -6,60 +6,34 @@
...
-            android:textSize="14sp" />
-
-    </LinearLayout>
-
+        android:clickable="false"
+        android:text="@string/end_of_scroll_message"
+        android:textAllCaps="false"
+        android:textColor="?attr/primary_text_color"
+        android:textSize="20sp"
+        app:icon="@drawable/ic_translate_white_24dp"
```
cooltey, line 18: Added by accident?

app/src/main/res/layout/view_on_this_day_footer.xml
```diff
@@ -6,60 +6,34 @@
...
-            android:textColor="?attr/colorAccent"
-            android:textSize="14sp" />
-
-    </LinearLayout>
-
+        android:clickable="false"
+        android:text="@string/end_of_scroll_message"
+        android:textAllCaps="false"
+        android:textColor="?attr/primary_text_color"
+        android:textSize="20sp"
```
cooltey, line 17: The style does not align with the card footer text style.

app/src/main/res/layout/view_on_this_day_footer.xml
```diff
@@ -6,60 +6,34 @@
...
+
+    <com.google.android.material.button.MaterialButton
+        android:id="@+id/back_to_future_text_view"
+        style="@style/FlatButton.Small.Transparent"
+        android:layout_marginStart="16dp"
+        android:text="@string/back_to_future"
+        android:textAllCaps="false"
+        android:textColor="?attr/colorAccent"
+        android:textSize="20sp"
+        app:icon="@drawable/ic_baseline_arrow_up_alt_24"
```
cooltey, line 33: The icon does not align with the text

cooltey (changes requested):

sharvaniharan:
@cooltey Lmk if the footer styles look fine now.

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayActivity.java
```diff
@@ -13,21 +13,17 @@
 
 public class OnThisDayActivity extends SingleFragmentActivity<OnThisDayFragment> {
     public static final String AGE = "age";
+    public static final String YEAR = "year";
     public static final String WIKISITE = "wikisite";
 
-    public static Intent newIntent(@NonNull Context context, int age, WikiSite wikiSite, InvokeSource invokeSource) {
+    public static Intent newIntent(@NonNull Context context, int age, @NonNull WikiSite wikiSite, @NonNull InvokeSource invokeSource, int year) {
```
cooltey, line 19: Can we move the `int year` to the left of `InvokeSource invokeSource`? I'd like to see the `invokeSource` stay on the last position.

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayCardView.java
```diff
@@ -2,122 +2,92 @@
...
+    @BindView(R.id.card_footer_view) CardFooterView cardFooterView;
+    @BindView(R.id.page_list_item_image) FaceAndColorDetectImageView otdEventImage;
+    @BindView(R.id.page_list_item_title) TextView otdEventTitle;
+    @BindView(R.id.page_list_item_description) TextView otdEventDescription;
+    @BindView(R.id.otd_event_page) View otdEventView;
     private FeedFunnel funnel = new FeedFunnel(WikipediaApp.getInstance());
+    private ExclusiveBottomSheetPresenter bottomSheetPresenter = new ExclusiveBottomSheetPresenter();
 
     private int age;
+    private Context context;
```
cooltey, line 68: This is unnecessary. We can use `getContext()`.

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayCardView.java
```diff
@@ -133,42 +103,95 @@ private void header(@NonNull OnThisDayCard card) {
...
+                });
+                otdEventView.setOnLongClickListener(view -> {
+                    PageTitle pageTitle = finalChosenPage.getPageTitle(card.wikiSite());
+                    HistoryEntry entry = new HistoryEntry(pageTitle, HistoryEntry.SOURCE_ON_THIS_DAY_CARD);
+
+                    new ReadingListBookmarkMenu(view, true, new ReadingListBookmarkMenu.Callback() {
+                        @Override
+                        public void onAddRequest(boolean addToDefault) {
+                            if (addToDefault) {
+                                ReadingListBehaviorsUtil.INSTANCE.addToDefaultList((AppCompatActivity) getContext(), entry.getTitle(), NEWS_ACTIVITY,
```
cooltey, line 161: The `InvokeSource` is incorrect.

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayCardView.java
```diff
@@ -133,42 +103,95 @@ private void header(@NonNull OnThisDayCard card) {
...
+                                        readingListId ->
+                                                bottomSheetPresenter.show(((AppCompatActivity) getContext()).getSupportFragmentManager(),
+                                                        MoveToReadingListDialog.newInstance(readingListId, entry.getTitle(), NEWS_ACTIVITY)));
```
cooltey, line 164: Same

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayCardView.java
```diff
@@ -133,42 +103,95 @@ private void header(@NonNull OnThisDayCard card) {
...
+                            } else {
+                                bottomSheetPresenter.show(((AppCompatActivity) getContext()).getSupportFragmentManager(),
+                                        AddToReadingListDialog.newInstance(entry.getTitle(), NEWS_ACTIVITY));
```
cooltey, line 167: Same

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayCardView.java
```diff
@@ -133,42 +103,95 @@ private void header(@NonNull OnThisDayCard card) {
...
+                            }
+                        }
+
+                        @Override
+                        public void onMoveRequest(@Nullable ReadingListPage page) {
+                            bottomSheetPresenter.show(((AppCompatActivity) getContext()).getSupportFragmentManager(),
+                                    MoveToReadingListDialog.newInstance(page.listId(), entry.getTitle(), NEWS_ACTIVITY));
```
cooltey, line 174: Same

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
```diff
@@ -67,14 +75,15 @@
...
+    @BindView(R.id.on_this_day_title_view) TextView onThisDayTitleView;
 
     @Nullable private OnThisDay onThisDay;
     private Calendar date;
     private Unbinder unbinder;
     @Nullable private OnThisDayFunnel funnel;
-    public static final int PADDING1 = 21, PADDING2 = 38, PADDING3 = 21;
-    public static final float HALF_ALPHA = 0.5f;
     private WikiSite wiki;
+    private int yearClickedOnCardView;
```
cooltey, line 85: The naming is confusing me. Can we have a better name for it?

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
```diff
@@ -88,83 +97,81 @@ public static OnThisDayFragment newInstance(int age, WikiSite wikiSite) {
...
-                colorAnimation.start();
-            }, animDelay);
-        } else {
-            dayText.setTextColor(ResourceUtil.getThemedColor(requireContext(), R.attr.material_theme_primary_color));
+        final int animDelay = (requireActivity().getWindow().getSharedElementEnterTransition() != null
+                && savedInstanceState == null) ? 500 : 0;
+        onThisDayTitleView.postDelayed(() -> {
+            if (!isAdded() || onThisDayTitleView == null) {
+                return;
+            }
```
cooltey, line 122: A little bit messy to me if we put the code here. Would it be better if create a method of it? (and we can move the suppress `checkstyle` to the method.

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
```diff
@@ -88,83 +97,81 @@ public static OnThisDayFragment newInstance(int age, WikiSite wikiSite) {
...
+        eventsRecycler.setVisibility(GONE);
+        errorView.setVisibility(GONE);
 
         disposables.add(ServiceFactory.getRest(wiki).getOnThisDay(month + 1, date)
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
-                .doAfterTerminate(() -> progressBar.setVisibility(View.GONE))
+                .doAfterTerminate(() -> {
+                    progressBar.setVisibility(GONE);
+                    eventsRecycler.postDelayed(() -> {
```
cooltey, line 150: Adding a check to prevent a crash?

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
```diff
@@ -310,63 +315,78 @@ public int getItemViewType(int position) {
...
-                pagesRecycler.setVisibility(View.GONE);
+                pagesViewPager.setVisibility(GONE);
+                pagesIndicator.setVisibility(GONE);
             }
         }
+
+        public void animateRadioButton() {
+            Animation pulse = AnimationUtils.loadAnimation(getContext(), R.anim.pulse);
+            radioButtonImageView.startAnimation(pulse);
+        }
```
cooltey, line 349: Not sure but would it be possible to not show the animation for 2+ dots at a time?

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
```diff
@@ -310,63 +315,78 @@ public int getItemViewType(int position) {
...
-            View backToTopView = v.findViewById(R.id.back_to_top_view);
-            backToTopView.setOnClickListener(v1 -> {
+            View backToFutureView = v.findViewById(R.id.back_to_future_text_view);
+            backToFutureView.setOnClickListener(v1 -> {
                 appBarLayout.setExpanded(true);
                 eventsRecycler.scrollToPosition(0);
             });
         }
     }
+     class ViewPagerAdapter extends RecyclerView.Adapter<OnThisDayPagesViewHolder> {
```
cooltey, line 362: Add `static`?

cooltey (changes requested):

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
sharvaniharan: We tried a few different ways and Robin okayed this animation @cooltey .. when i tried to show only one by using 'firstVisibleItem' if they scrolled even a little fast they would miss the animation.

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
sharvaniharan: So ..the message that Android studio is annoying us with is because non-static inner classes hold a direct reference to outer class... and since some inner classes can live longer than the container, if the activity or fragment holding it gets destroyed, then it will not be garbage collected until this class is done with its stuff. Hence the memory leak. Here it is not an issue as we are sure it is tied to the Fragment's lifecycle. However, I think we need to take a call on whether or not we should start making them all static to get rid or warnings, or play it by the ear.. because there are other places where we have just left this... 😅

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
sharvaniharan: @cooltey yes, the 01 checks are to prevent crash.. since we dont have values in all cases for the year

cooltey: I was thinking about adding something like `isAdded` or `isDestoryed`

dbrant (changes requested):
See inline note, and a few design inconsistencies:

* The font of the date selection looks very different from the font of the other two lines. It's almost like it's "extra-bold", with super wide letter spacing.  Maybe using `MaterialButton` here is causing more problems than it solves?
image

* The year label is slightly misaligned horizontally from the rest of the text:
image

* At the bottom, the "back to the future" message is misaligned, and uses very wide letter spacing.
image

sharvaniharan:
Done @cooltey @dbrant .

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayCardView.java
```diff
@@ -133,42 +99,99 @@ private void header(@NonNull OnThisDayCard card) {
...
+                    otdEventImage.setVisibility(VISIBLE);
+                    otdEventImage.loadImage(Uri.parse(chosenPage.getThumbnailUrl()));
+                }
+                otdEventDescription.setText(chosenPage.getDescription());
+                otdEventDescription.setVisibility(TextUtils.isEmpty(chosenPage.getDescription()) ? View.GONE : View.VISIBLE);
+                otdEventTitle.setMaxLines(TextUtils.isEmpty(chosenPage.getDescription()) ? 2 : 1);
+                otdEventTitle.setText(StringUtil.fromHtml(chosenPage.getDisplayTitle()));
+                otdEventView.setOnClickListener(view -> {
+                    PageTitle pageTitle = finalChosenPage.getPageTitle(card.wikiSite());
+                    HistoryEntry entry = new HistoryEntry(pageTitle, HistoryEntry.SOURCE_ON_THIS_DAY_CARD);
```
cooltey, line 149: Since we will use the same entry here and below, would it be better if we create a method of it? (and maybe reduce the `final` variable?

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayFragment.java
```diff
@@ -95,45 +104,33 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
...
+        progressBar.setVisibility(GONE);
+        eventsRecycler.setVisibility(GONE);
+        errorView.setVisibility(GONE);
         return view;
     }
 
-    public void onBackPressed() {
-        dayText.setTextColor(ResourceUtil.getThemedColor(requireContext(), R.attr.primary_text_color));
+    @SuppressWarnings("checkstyle:magicnumber")
+    private void setUpTransitionAnimation(Bundle savedInstanceState, int age) {
```
cooltey, line 125: Add annotation?

cooltey (changes requested):

cooltey (approved):
Looks good to me. @dbrant I'll merge it soon unless you need to review the PR as well.

PR #1614, opened 2020-09-10
Redesign because you read card

cooltey:
https://phabricator.wikimedia.org/T261671


dbrant (changes requested):
A few design inconsistencies:

- The margins on the left and right of the list are too small, and the shadows around the card are not visible.
- The margins around the image inside the gradient box are inconsistent. (should be the same at the top/bottom/right)
- Too much vertical space between the title and description of the item in the gradient box.

device-2020-09-15-145008

cooltey:
The margins on the left and right of the list are too small, and the shadows around the card are not visible.

Partially done. The gradient box has a really light shadow on it, let's have a discussion tomorrow.

    The margins around the image inside the gradient box are inconsistent. (should be the same at the top/bottom/right)

Done.

    Too much vertical space between the title and description of the item in the gradient box.

Done.

PR #1616, opened 2020-09-11
Make the Search tooltip not auto-dismissable.

dbrant:
This allows the user to interact with the rest of the UI while the tooltip is shown, instead of having to dismiss the tooltip first.


sharvaniharan:
Looks likt the parameter is having no effect
https://youtu.be/RilBQZmONr0

dbrant:
@sharvaniharan   The effect is that the Search tab (to which the tooltip is pointing) becomes _clickable_ even when the tooltip is showing.  Currently the user has to dismiss the tooltip, and then click the Search tab again.

PR #1617, opened 2020-09-11
Deeply integrate Talk and User Talk namespaces.

dbrant:
This brings in static definitions for the `User:`, `User talk:`, and `Talk:` namespaces, which we unfortunately need to detect the namespace of an article by brute force.

This proceeds to deeply integrate these namespaces into PageActivity, so that the user goes directly to the native Talk interface when tapping on the footer Talk link on a User page.


app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -1097,8 +1098,13 @@ private void setupMessageHandlers() {
         bridge.addListener("footer_item", (String messageType, JsonObject messagePayload) -> {
             String itemType = messagePayload.get("itemType").getAsString();
             if ("talkPage".equals(itemType) && model.getTitle() != null) {
-                PageTitle talkPageTitle = new PageTitle("Talk", model.getTitle().getPrefixedText(), model.getTitle().getWikiSite());
-                visitInExternalBrowser(requireContext(), Uri.parse(talkPageTitle.getUri()));
+                // If we're currently looking at a User page, then go directly to the corresponding User Talk page.
+                if (model.getTitle().namespace() == Namespace.USER) {
+                    startActivity(TalkTopicsActivity.newIntent(requireActivity(), model.getTitle().getWikiSite().languageCode(), model.getTitle().getText()));
```
cooltey, line 1103: Not really related to this but I just noticed that the "back button" behavior is not expected when we heading to the user talk page.

1. Article page -> 2. User talk page -> 3. click on other user's talk page -> 4. check other user's page.

If we click on the device back button, we will go back to the `1. Article page` instead of the `3. click on other user's talk page`, which might cause confusion to the users.

cooltey (approved):
LGTM

PR #1630, opened 2020-09-16
Follow-up to Because You Read card.

dbrant:
- Consolidate a few static items.
- Perform a bit more magic with color gradient extraction to make the colors more interesting.

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/view/CardLargeHeaderView.java
+++ b/app/src/main/java/org/wikipedia/feed/view/CardLargeHeaderView.java
@@ -1,6 +1,7 @@
 package org.wikipedia.feed.view;
 
 import android.content.Context;
+import android.graphics.Color;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/feed/view/CardLargeHeaderView.java
```diff
@@ -90,10 +100,30 @@ public void onImageFailed() {
...
+
+    private static int darkenColor(@ColorInt int color) {
+        int r = (color & 0xff);
+        int g = (color & 0xff00) >> 8;
+        int b = (color & 0xff0000) >> 16;
+        r -= (r / 2);
+        g -= (g / 2);
+        b -= (b / 2);
+        return Color.rgb(r, g, b);
+    }
```
cooltey, line 121: Great!

PR #1631, opened 2020-09-16
Redesign trending card

cooltey:
https://phabricator.wikimedia.org/T261669

This PR also updates the `header()` and `footer()` methods that use the `getCard()` instead of passing a specific card object.


app/src/main/java/org/wikipedia/feed/view/GraphView.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    private val pathPaint = Paint().apply {
+        style = Paint.Style.STROKE
+        strokeWidth = GRAPH_STROKE_WIDTH
+        strokeCap = Paint.Cap.ROUND
+        pathEffect = CornerPathEffect(GRAPH_STROKE_WIDTH)
+        isAntiAlias = true
+    }
+
+    fun setData(viewHistoryList: List<MostReadArticles.ViewHistory>) {
```
dbrant, line 27: It's nice that this class has the generic name `GraphView`, which suggests we'll be able to reuse it for graphing other things.  But then it's using this extremely specific `MostReadArticles.ViewHistory` object for providing the data points. Is there a way for this View class to take a completely generic list of `floats` for graphing, regardless of where the numbers come from?

dbrant (changes requested):
See quick note inline. Also, I'm seeing the circle gradient slightly cut off along the edges:

image

cooltey:
> 
> 
> See quick note inline. Also, I'm seeing the circle gradient slightly cut off along the edges:
> 

Thanks, I guess my previous setting depends on devices. Done.

app/src/main/java/org/wikipedia/feed/view/GraphView.kt
cooltey: Make sense. Done!

PR #1645, opened 2020-09-23
Smooth transition from images in Article to Gallery

dbrant:


app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -145,4 +145,22 @@ object JavaScriptActionHandler {
                 "document.head.appendChild(style);" +
                 "})();"
     }
+
+    @JvmStatic
+    fun getElementAtPosition(x: Int, y: Int): String {
```
sharvaniharan, line 150: Very impressed! :)

app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -145,4 +145,22 @@ object JavaScriptActionHandler {
...
+                "  result.width = element.clientWidth;" +
+                "  result.height = element.clientHeight;" +
+                "  result.src = element.src;" +
+                "  return result;" +
+                "})();"
+    }
+
+    data class ImageHitInfo(val left: Float = 0f, val top: Float = 0f, val width: Float = 0f, val height: Float = 0f,
+                            val src: String = "", val centerCrop: Boolean = false) {
+    }
```
cooltey, line 165: Would it be possible to remove the `{ }` ?

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
```diff
@@ -243,16 +245,26 @@ public void onCallToActionClicked() {
...
 
-    public void openImageInGallery(@Nullable  String language) {
+    public void openImageInGallery(@Nullable String language) {
         if (getPage() != null && isLeadImageEnabled()) {
             String imageName = getPage().getPageProperties().getLeadImageName();
             if (imageName != null) {
                 String filename = "File:" + imageName;
                 WikiSite wiki = language == null ? getTitle().getWikiSite() : WikiSite.forLanguageCode(language);
+
+                JavaScriptActionHandler.ImageHitInfo hitInfo = new JavaScriptActionHandler.ImageHitInfo(pageHeaderView.image.getLeft(),
```
cooltey, line 255: If the image was face detected (e.g. Barack Obama), the image will be cropped and it looks slightly weird during the transition.

app/src/main/java/org/wikipedia/views/ViewUtil.java
```diff
@@ -55,14 +56,23 @@ public static void loadImage(@NonNull ImageView view, @Nullable String url, bool
     }
 
     public static void loadImageWithWhiteBackground(@NonNull ImageView view, @Nullable String url) {
+        loadImageWithWhiteBackground(view, url, null);
+    }
+
+    public static void loadImageWithWhiteBackground(@NonNull ImageView view, @Nullable String url, @Nullable RequestListener<Drawable> listener) {
```
cooltey, line 62: I see a "blink" when the image being loaded, is it possible to reduce the issue?

cooltey (changes requested):
The animation is pretty smooth 👍

app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
dbrant: yep

app/src/main/java/org/wikipedia/views/ViewUtil.java
dbrant: should be improved now.

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.java
dbrant: Yes, that's a bit of an issue, but I'm not sure I have a solution yet. It could be some technical debt for the future.

cooltey (changes requested):
Found one issue can be optimized:
Click on an image in landscape (or a smaller image) -> while it is loading, swipe to the left/right images quickly -> you will see the transition image overlaps other images sometimes.

dbrant:
> Found one issue can be optimized:
> Click on an image in landscape (or a smaller image) -> while it is loading, swipe to the left/right images quickly -> you will see the transition image overlaps other images sometimes.

Do you have ideas for optimizing this?

cooltey:
> 
> 
> > Found one issue can be optimized:
> > Click on an image in landscape (or a smaller image) -> while it is loading, swipe to the left/right images quickly -> you will see the transition image overlaps other images sometimes.
> 
> Do you have ideas for optimizing this?

Maybe disable the swipe before the image gets loaded?

dbrant:
OK, here's the best I've got.

PR #1660, opened 2020-09-25
Implement transition animation between feed card and page activity

cooltey:
https://phabricator.wikimedia.org/T263259

Tried to apply transition animation only on images but it does not smooth as the current approach. Please review this PR and suggestions/comments are welcome.


sharvaniharan:
Transitions look good @cooltey. Couple things I am seeing: 1. A tiny flicker when page loads.. which is inevitable i guess, but if you have not already tried, could you play around with making the page visible only after it is fully loaded, and keeping the transition screenshot alive till that point? 
2. When you open an article from feed, and change to landscape on PageActivity, and then try to go back, the transition looks weird since there will be no lead image on the landscape mode. Is it possible to have just a title transition in this case?

app/src/main/res/values/dimens.xml
```diff
@@ -95,7 +95,7 @@
     <dimen name="randomizer_text_view_height">164dp</dimen>
 
     <!-- Divider width -->
-    <dimen name="divider_width_for_article">60dp</dimen>
+    <dimen name="divider_width_for_article">55dp</dimen>
```
sharvaniharan, line 98: This still seems to be 60 dp on design

sharvaniharan (changes requested):

app/src/main/res/layout/activity_page.xml
```diff
@@ -13,6 +13,14 @@
         android:layout_width="match_parent"
         android:layout_height="match_parent">
 
+        <ImageView
```
dbrant, line 16: Couldn't the image be loaded directly into the `PageHeaderView`? (and no longer require a cached bitmap to be kept in memory?)

dbrant (changes requested):

app/src/main/res/layout/activity_page.xml
cooltey: Do you mean only the image or the image with the article summary? I have tried to implement the animation for only the image without any cached bitmap, just like what we did for the floating queue thing, I found that the animation is not as much smooth as the way we did before.

I guess that's because of the changes after we implemented `mobile-html`.

Also, looks like Robin wants the image+text animation: https://phabricator.wikimedia.org/T263259#6503265

cooltey:
> 
> 
> Transitions look good @cooltey. Couple things I am seeing: 1. A tiny flicker when page loads.. which is inevitable i guess, but if you have not already tried, could you play around with making the page visible only after it is fully loaded, and keeping the transition screenshot alive till that point?

Good point. But the flickr you've seen came from the image loading for the first time. In fact, I am setting the page visible when the page is fully loaded. 

> 2. When you open an article from feed, and change to landscape on PageActivity, and then try to go back, the transition looks weird since there will be no lead image on the landscape mode. Is it possible to have just a title transition in this case?

Thanks! Fixed.

cooltey:
@sharvaniharan and @dbrant 
I have a comparison PR of this PR, #1693, which reduces the usage of memory. Please let me know which one you preferred.

cooltey:
the v2 version is better.

PR #1663, opened 2020-09-28
Design tweaks:  Search changes

sharvaniharan:

- Placed the search bar higher to match the height of explore tab bar
- Used the 'Saved' bookmark icon for reading list search results
- aligned the delete button to the top action icon.

### DIFF
--- a/app/src/main/res/drawable/ic_tab_single_24px.xml
+++ b/app/src/main/res/drawable/ic_tab_single_24px.xml
@@ -4,6 +4,6 @@
     android:viewportWidth="24"
     android:viewportHeight="24">
   <path
-      android:fillColor="?attr/material_theme_de_emphasised_color"
+      android:fillColor="#FF000000"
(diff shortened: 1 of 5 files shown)


app/src/main/res/drawable/ic_tab_single_24px.xml
```diff
@@ -4,6 +4,6 @@
     android:viewportWidth="24"
     android:viewportHeight="24">
   <path
-      android:fillColor="#FF000000"
+      android:fillColor="?attr/material_theme_de_emphasised_color"
```
dbrant, line 7: What's the reason for this color change? It makes the tab icon look way too light. In the ticket, Robin talks about using a different icon, not a different color.

dbrant (changes requested):

app/src/main/res/drawable/ic_tab_single_24px.xml
sharvaniharan: @dbrant  was just trying something and forgot to remove. I do not see a new icon in Zeplin and have asked Robin for one. I tried looks_one.svg here https://material.io/resources/icons/?icon=looks_one&style=outline. It is not very different from what we currently have. Will update the icon on a separate PR once he provides it.

PR #1666, opened 2020-09-29
A few tweaks for User Talk pages.

dbrant:
- Adjust the Talk icon to be slightly smaller, according to the OOUI shape.
- Update link preview button text for talk pages.
- Explicitly open the User Talk activity from internal links.

### DIFF
--- a/app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.java
+++ b/app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.java
@@ -294,7 +294,9 @@ public class LinkPreviewDialog extends ExtendedBottomSheetDialogFragment
         }
         if (overlayView != null) {
             overlayView.setPrimaryButtonText(getStringForArticleLanguage(pageTitle,
-                    contents.isDisambiguation() ? R.string.button_continue_to_disambiguation : R.string.button_continue_to_article));
+                    contents.isDisambiguation() ? R.string.button_continue_to_disambiguation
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.java
```diff
@@ -153,7 +154,8 @@ public void goToLinkedPage(boolean inNewTab) {
         if (overlayView == null && containerView != null) {
             overlayView = new LinkPreviewOverlayView(getContext());
             overlayView.setCallback(new OverlayViewCallback());
-            overlayView.setPrimaryButtonText(getStringForArticleLanguage(pageTitle, R.string.button_continue_to_article));
+            overlayView.setPrimaryButtonText(getStringForArticleLanguage(pageTitle,
+                    pageTitle.namespace() == Namespace.TALK || pageTitle.namespace() == Namespace.USER_TALK ? R.string.button_continue_to_talk_page : R.string.button_continue_to_article));
```
cooltey, line 158: Screenshot_20200929-123747_Wikipedia Dev

Not sure but should this also work on the Gallery?

[Barack Obama] -> Gallery -> the credit of the 6th image

cooltey (changes requested):

app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.java
dbrant: done! good catch.

PR #1667, opened 2020-09-29
Design Tweak: Tab icon updated

sharvaniharan:


app/src/main/res/drawable/ic_tab_one_24px.xml
```diff
@@ -0,0 +1,10 @@
+<vector android:autoMirrored="true" android:height="20dp"
```
dbrant, line 1: This probably should not be `autoMirrored`, because in an RTL layout it will look like a backwards "1".

dbrant (changes requested):

PR #1668, opened 2020-09-29
Improve Gallery zoom+pager interaction.

dbrant:
There's an annoying thing when flipping through images in the Gallery:  if you start zooming in on an image, and then pan around the zoomed-in image, it sometimes causes the ViewPager to flip to another image instead.
This will make the ViewPager disable its scrolling while an image is zoomed in, and only reenable the ViewPager when the image is fully zoomed out.

https://phabricator.wikimedia.org/T260625

### DIFF
--- a/app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.java
+++ b/app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.java
@@ -18,6 +18,8 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
 
+import com.github.chrisbanes.photoview.PhotoView;
+
 import org.wikipedia.Constants;
 import org.wikipedia.R;
 import org.wikipedia.WikipediaApp;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.java
```diff
@@ -122,6 +124,9 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
                 ((GalleryActivity) requireActivity()).toggleControls();
             }
         });
+
+        imageView.setOnMatrixChangeListener(rect -> ((GalleryActivity) requireActivity()).setViewPagerEnabled(imageView.getScale() <= 1f));
```
cooltey, line 128: Adding a check for this listener?

```
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'float com.github.chrisbanes.photoview.PhotoView.getScale()' on a null object reference
        at org.wikipedia.gallery.GalleryItemFragment.lambda$onCreateView$1$GalleryItemFragment(GalleryItemFragment.java:128)
        at org.wikipedia.gallery.-$$Lambda$GalleryItemFragment$IT7mfODMYoIdODz3RDiJxJ5SW2c.onMatrixChanged(Unknown Source:2)
```

cooltey (changes requested):

app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.java
dbrant: alright!

PR #1677, opened 2020-10-02
Focus mode: hide bottom bar on scroll.

dbrant:
https://phabricator.wikimedia.org/T254771


cooltey (commented):
Pretty smooth and work as expected 👍

dbrant:
archiving this until we make a definitive decision about the behavior of the bottom bar.

PR #1680, opened 2020-10-02
Use ContextCompat.getSystemService().

Isira-Seneviratne:


app/build.gradle
```diff
@@ -29,6 +29,9 @@ android {
         sourceCompatibility = JAVA_VERSION
         targetCompatibility = JAVA_VERSION
     }
+    kotlinOptions {
+        jvmTarget = JAVA_VERSION
```
dbrant, line 33: I don't believe we need any changes to `build.gradle` to use `ContextCompat.getSystemService()`, do we?

dbrant (changes requested):

app/build.gradle
Isira-Seneviratne: That was for the `getSystemService` extension function, which inlines to `ContextCompat.getSystemService()`.

dbrant: Can we remove it?

dbrant (changes requested):

app/src/main/java/org/wikipedia/ktx/Context.kt
```diff
@@ -0,0 +1,12 @@
+@file:JvmName("ContextUtil")
+
+package org.wikipedia.ktx
+
+import android.content.ClipData
+import android.content.ClipboardManager
+import android.content.Context
+import androidx.core.content.getSystemService
+
+fun Context.setPlainText(label: CharSequence?, text: CharSequence?) {
```
dbrant, line 10: `setPlainText()` does not make sense as an extension function of `Context`.  Can we keep the file structure the way it was?

dbrant (changes requested):

app/src/main/java/org/wikipedia/ktx/Context.kt
Isira-Seneviratne: Sure.

PR #93, opened 2019-03-15
Dialog/notification concept

cooltey:
Create notification and dialog for the **Suggested edits**
https://phabricator.wikimedia.org/T217369

PR #1685, opened 2020-10-05
Redesign Explore feed announcements card

cooltey:
https://phabricator.wikimedia.org/T263482


app/src/main/java/org/wikipedia/feed/announcement/AnnouncementCardView.java
```diff
@@ -83,6 +84,7 @@ public AnnouncementCardView(@NonNull Context context) {
 
         if (card.hasFooterCaption()) {
             footerTextView.setText(StringUtil.fromHtml(card.footerCaption()));
+            RichTextUtil.removeUnderlinesFromLinks(footerTextView);
```
sharvaniharan, line 87: @cooltey Line #93 the border condition needs to be taken care of. We had thick red borders previously. We need some solution now.

sharvaniharan: Also, the space between title and text seems to be a little bit more, since we are doing this in the same string, may be use one less break <br>?

sharvaniharan (changes requested):

app/src/main/java/org/wikipedia/feed/announcement/AnnouncementCardView.java
sharvaniharan: Also, the text size is different.20sp.

cooltey: > Also, the space between title and text seems to be a little bit more, since we are doing this in the same string, may be use one less break <br>?
> Also, the text size is different.20sp.

Thanks, all the questions are waiting for Robin's feedback: https://phabricator.wikimedia.org/T263482#6519398

cooltey: > 
> 
> @cooltey Line #93 the border condition needs to be taken care of. We had thick red borders previously. We need some solution now.

Done.

PR #1686, opened 2020-10-05
Fix: crash when showing SE snackbars on the Explore feed

cooltey:
https://appcenter.ms/orgs/ci-apps-hockeyapp-f5mf/apps/Wikipedia/crashes/errors/44565870u/overview

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/FeedFragment.java
+++ b/app/src/main/java/org/wikipedia/feed/FeedFragment.java
@@ -243,8 +243,8 @@ public class FeedFragment extends Fragment implements BackPressedHandler {
             if (resultCode == RESULT_OK) {
                 if (suggestedEditsCardView != null && suggestedEditsCardView.getCard() != null) {
                     suggestedEditsCardView.refreshCardContent();
-                    SuggestedEditsSnackbars.show(requireActivity(), suggestedEditsCardView.getCard().getAction(), true,
-                            app.language().getAppLanguageCodes().get(1), true, () -> {
+                    String editLanguage = app.language().getAppLanguageCodes().size() > 1 ? app.language().getAppLanguageCodes().get(1) : app.language().getAppLanguageCode();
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/feed/FeedFragment.java
```diff
@@ -243,8 +243,8 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
             if (resultCode == RESULT_OK) {
                 if (suggestedEditsCardView != null && suggestedEditsCardView.getCard() != null) {
                     suggestedEditsCardView.refreshCardContent();
-                    SuggestedEditsSnackbars.show(requireActivity(), suggestedEditsCardView.getCard().getAction(), true,
-                            app.language().getAppLanguageCodes().get(1), true, () -> {
+                    String editLanguage = app.language().getAppLanguageCodes().size() > 1 ? app.language().getAppLanguageCodes().get(1) : app.language().getAppLanguageCode();
+                    SuggestedEditsSnackbars.show(requireActivity(), suggestedEditsCardView.getCard().getAction(), true, editLanguage, true, () -> {
                         PageTitle pageTitle = suggestedEditsCardView.getCard().getSourceSummaryForEdit().getPageTitle();
```
sharvaniharan, line 248: Wondering if the acctual cause for this crash is this lie, where we are trying to get a pageTtile from getSourceSummaryForEdit(), which in case of Image tags might be null.

cooltey: The AppCenter says it is a `java.lang.IndexOutOfBoundsException: Index: 1, Size: 1` error.

sharvaniharan: Yes .. saw that and that definitely needs fixing... but just that i found this bug during se cards re-design and fixed it there... So I guess will merge this for now, and the other thing will get fixed with re-design.

PR #1688, opened 2020-10-06
Warning message for discrepancies in edit language 

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T264000
Simple way of using the InputMethodManager does not bring up the language.

```
InputMethodManager imm = (InputMethodManager)getSystemService(Context.INPUT_METHOD_SERVICE);

InputMethodSubtype ims = imm.getCurrentInputMethodSubtype();

String locale = ims.getLocale();
```

I was able to use Firebase's language detection to do the same. Have marked this experimental, for further discussion.


sharvaniharan:
I think the build will be automatically fixed when we merge the push notifications PR

app/src/main/java/org/wikipedia/descriptions/DescriptionEditView.java
```diff
@@ -371,6 +373,18 @@ void validateText() {
         updateSaveButtonEnabled();
     }
 
+    private void checkKeyboardLanguage(String text) {
```
dbrant, line 376: This no longer has to do with the "keyboard" language, so this can be named more intuitively, e.g. `detectLanguageFromText()`

app/src/main/java/org/wikipedia/descriptions/DescriptionEditView.java
```diff
@@ -371,6 +373,18 @@ void validateText() {
...
+        FirebaseLanguageIdentification languageIdentifier =
```
dbrant, line 377: Ideally anything that references Firebase or GMS should be in a separate file, so that it can be stubbed out for builds that won't use Google Play Services.

app/src/main/res/values/strings.xml
```diff
@@ -683,6 +683,7 @@
     <string name="description_ends_with_punctuation">The text must not end with punctuation.</string>
     <string name="description_starts_with_article">Avoid starting with articles like “a” or “the”.</string>
     <string name="description_starts_with_uppercase">Start with a lowercase letter unless the first word is a proper noun.</string>
+    <string name="description_is_in_different_language">Your keyboard language and editing language are different. Are you sure you are editing in the correct language?</string>
```
dbrant, line 686: How about: `The text appears to be in the wrong language. The language should be %s.`

app/src/main/java/org/wikipedia/descriptions/DescriptionEditView.java
```diff
@@ -371,6 +373,18 @@ void validateText() {
...
+                FirebaseNaturalLanguage.getInstance().getLanguageIdentification();
+        languageIdentifier.identifyLanguage(text)
+                .addOnSuccessListener(
+                        languageCode -> {
```
dbrant, line 381: Is it possible for this `successListener` to be delayed significantly? (hint: until after this view is disposed?)

app/build.gradle
```diff
@@ -200,6 +201,9 @@ dependencies {
     implementation 'org.jsoup:jsoup:1.12.1'
     implementation 'com.github.chrisbanes:PhotoView:2.3.0'
     implementation "com.github.skydoves:balloon:1.2.1"
+    implementation 'com.google.firebase:firebase-ml-natural-language:22.0.1'
```
dbrant, line 204: It looks like language detection is no longer done through Firebase (it's deprecated), but now done directly with ML Kit. https://developers.google.com/ml-kit/language/identification/android

dbrant (changes requested):

app/src/main/res/values/strings.xml
sharvaniharan: "wrong" or "different" @dbrant

app/src/main/java/org/wikipedia/descriptions/DescriptionEditView.java
sharvaniharan: Now that the view implements a callback this won't be a problem @dbrant

app/src/main/res/values/strings.xml
dbrant: How about "unexpected"?

app/src/main/java/org/wikipedia/language/FirebaseLanguageDetector.kt
```diff
@@ -0,0 +1,20 @@
+package org.wikipedia.language
+
+import com.google.mlkit.nl.languageid.LanguageIdentification
+
+class FirebaseLanguageDetector {
```
dbrant, line 5: Just noting down what we discussed earlier today:
After the `push` PR is merged, you would move this class to the `extra` source tree, and then create an identical class in the `fdroid` source tree, except with all the functions being blank.

dbrant (changes requested):

app/src/main/res/values/strings.xml
sharvaniharan: Perfect.. `The text appears to be in an unexpected language. The language should be %s.`

app/src/main/java/org/wikipedia/descriptions/DescriptionEditView.java
```diff
@@ -438,4 +444,12 @@ public void setAction(Action action) {
         this.action = action;
         isTranslationEdit = (action == TRANSLATE_CAPTION || action == TRANSLATE_DESCRIPTION);
     }
+
+    @Override
+    public void onLanguageDetectionSuccess(@NotNull String languageCode) {
+        if (!languageCode.equals("und") && !languageCode.equals(pageSummaryForEdit.getLang())) {
```
dbrant, line 450: The condition of the language being `"und"` can probably be moved into the `FirebaseLanguageDetector`, and it doesn't have to call the `success` callback in that case.

app/src/main/java/org/wikipedia/language/FirebaseLanguageDetector.kt
```diff
@@ -0,0 +1,20 @@
...
+    interface Callback {
+        fun onLanguageDetectionSuccess(languageCode: String)
+    }
+
+    var callback: Callback? = null
+    fun detectLanguageFromText(text: String) {
+        val languageIdentifier = LanguageIdentification.getClient()
+        languageIdentifier.identifyLanguage(text)
+                .addOnSuccessListener { languageCode: String ->
+                    if (callback != null) {
```
dbrant, line 15: A null check is not necessary. You can just use `callback?.onLanguageDetectionSuccess()`

dbrant (changes requested):

app/src/main/java/org/wikipedia/descriptions/DescriptionEditView.java
sharvaniharan: @dbrant the service returns "und" [undetermined] when it is not able to determine with confidence. So we will need it here..

app/src/main/java/org/wikipedia/language/FirebaseLanguageDetector.kt
sharvaniharan: this is now done @dbrant . I have added the necessary implementations in gradle, and created the class in `extra` and a stubbed it on fdroid source tree.

sharvaniharan:
This is now ready to merge.

cooltey (approved):
It works! Excellent 👍

app/src/main/java/org/wikipedia/descriptions/DescriptionEditView.java
```diff
@@ -438,4 +444,12 @@ public void setAction(Action action) {
         this.action = action;
         isTranslationEdit = (action == TRANSLATE_CAPTION || action == TRANSLATE_DESCRIPTION);
     }
+
+    @Override
+    public void onLanguageDetectionSuccess(@NotNull String languageCode) {
```
cooltey, line 449: Just noticed that the annotation should be `@NonNull`.

PR #1693, opened 2020-10-07
Transition animation between feed card and page activity

cooltey:
https://phabricator.wikimedia.org/T263259

~~This is a comparison PR of #1660, which lets the views transit between activities.~~


sharvaniharan:
Can we do something about this landscape/device size : https://youtu.be/LQCZhGzMzuk It is a bit jarring when there is no image.. both to and from

sharvaniharan:
This is definitely smoother though :-)

cooltey:
> 
> 
> Can we do something about this landscape/device size : https://youtu.be/LQCZhGzMzuk It is a bit jarring when there is no image.. both to and from

Done!

PR #1695, opened 2020-10-07
Update the list of languages and variants

cooltey:
https://phabricator.wikimedia.org/T263471

Deeplink test: https://etherpad.wikimedia.org/p/DeepLinkForTests


dbrant (changes requested):
This is excellent -- we will now have a truly complete list of languages, and full language parity with Desktop.

However, we need to think very carefully about users who are **upgrading** from the previous version.  For example, before trying this PR, I added a few languages to my app, including Serbian, Uzbek, and Belarusian (be-tarask).  And after upgrading, this is now my list of languages:

device-2020-10-14-143809

We should probably add logic to select a _default_ variant when upgrading to this PR, when the user has a language that contains variants.

cooltey:
> 
> We should probably add logic to select a _default_ variant when upgrading to this PR, when the user has a language that contains variants.

You are correct. Done.

cooltey:
Hi @dbrant, since there's only one more language that supports the same format as the `zh` wiki, I have added the `sr` rule to the `AndroidManifest.xml` path pattern. 95bf5a2bc15b74171cfb9eeb64fe44d1744c988d

PR #1697, opened 2020-10-09
Smooth transition between Link Preview images and Gallery.

dbrant:


cooltey (changes requested):
Animation works pretty great, but I found that if the device is in landscape mode, the image will become too large during the transition to the gallery.

dbrant:
> Animation works pretty great, but I found that if the device is in landscape mode, the image will become too large during the transition to the gallery.

Done!

PR #1701, opened 2020-10-13
Finish "On This Day" activity on back pressed

zhlcbri:
**Background**
https://phabricator.wikimedia.org/T265226

Using a shared element transition between the feed to "On this day" page causes date text to overlay when user goes back to the feed.

**Change**
- [x] Call `finish()` when user goes back to the feed from "On this day" page, so all pending transition animations are cleared.

**Testing**
Before:
before

After:
after


dbrant (changes requested):
[see comments on the ticket]

PR #1702, opened 2020-10-13
Fix: correct the regular expression of getting page title from a link

cooltey:
When testing the PR #1695, I have noticed that some in-article links will not get the page title correctly.

Please test this PR with this `testwiki` article: https://test.wikipedia.org/wiki/Language_Variants

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/WikiSite.java
+++ b/app/src/main/java/org/wikipedia/dataclient/WikiSite.java
@@ -96,6 +96,9 @@ public class WikiSite implements Parcelable {
         if (languageCode.equals(AppLanguageLookUpTable.CHINESE_LANGUAGE_CODE)) {
             languageCode = LanguageUtil.getFirstSelectedChineseVariant();
         }
+
+        // TODO: should use the default language code on the authority to prevent incorrect url such as zh-tw.wikipedia.org.
+
         this.uri = new Uri.Builder()
                 .scheme(tempUri.getScheme())
                 .encodedAuthority(authority)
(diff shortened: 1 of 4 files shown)


dbrant (changes requested):
Looks like a good opportunity to add a couple of unit tests 😉

app/src/main/java/org/wikipedia/util/UriUtil.java
```diff
@@ -139,13 +139,13 @@ public static String getLanguageVariantFromUri(@NonNull Uri uri) {
...
     public static String removeInternalLinkPrefix(@NonNull String link) {
-        return link.replaceFirst("/wiki/|/zh.*/", "");
+        return link.replaceFirst("/(wiki|zh.*)/", "");
     }
 
     /** For links that could be internal or external links */
     @NonNull
     private static String removeLinkPrefix(@NonNull String link) {
-        return link.replaceFirst("^.*?/wiki/", "");
+        return link.replaceFirst("^.*?/(wiki|zh.*)/", "");
```
sharvaniharan, line 148: is it ok that it now became zh specific?

app/src/main/java/org/wikipedia/dataclient/WikiSite.java
```diff
@@ -96,6 +96,9 @@ public WikiSite(@NonNull Uri uri) {
         if (languageCode.equals(AppLanguageLookUpTable.CHINESE_LANGUAGE_CODE)) {
             languageCode = LanguageUtil.getFirstSelectedChineseVariant();
         }
+
+        // TODO: should use the default language code on the authority to prevent incorrect url such as zh-tw.wikipedia.org.
```
cooltey, line 100: This can be fixed in #1695, or maybe we can ignore this because it is an edge case.

cooltey:
> 
> 
> Looks like a good opportunity to add a couple of unit tests 😉

Done.

PR #1704, opened 2020-10-14
Use collection extensions.

Isira-Seneviratne:
Simplify some operations using the Kotlin collection extension functions.


app/src/main/java/org/wikipedia/userprofile/UserContributionsStats.kt
```diff
@@ -53,73 +56,52 @@ object UserContributionsStats {
...
-                        descLang = descArr[1]
-                        break
-                    }
-                }
-            }
-            if (descLang.isEmpty()) {
+            val descLang = userContribution.comment.split(" ")
+                    .filter { "wbsetdescription" in it }
+                    .flatMap { it.split("|") }
+                    .firstOrNull()
```
cooltey, line 62: We should get the second item from the splitted array. Maybe use `getOrNull(1)`?

app/src/main/java/org/wikipedia/userprofile/UserContributionsStats.kt
```diff
@@ -53,73 +56,52 @@ object UserContributionsStats {
...
-                qLangMap[userContribution.title] = HashSet()
-            }
-            qLangMap[userContribution.title]!!.add(descLang)
+            qLangMap.getOrPut(userContribution.title, { HashSet() }).add(descLang)
         }
 
         return ServiceFactory.get(WikiSite(Service.WIKIDATA_URL)).getWikidataLabelsAndDescriptions(qLangMap.keys.joinToString("|"))
                 .subscribeOn(Schedulers.io())
-                .flatMap {
+                .flatMap { it ->
```
cooltey, line 72: Can it be removed?

cooltey (changes requested):
Thanks for the PR, I have learned a lot. Just have a few comments inline.

app/src/main/java/org/wikipedia/suggestededits/provider/EditingSuggestionsProvider.kt
```diff
@@ -94,8 +92,9 @@ object EditingSuggestionsProvider {
                                         || !entity.sitelinks().containsKey(targetWiki.dbName())) {
                                     continue
                                 }
-                                articlesWithTranslatableDescriptionCache.push(Pair(PageTitle(entity.sitelinks()[targetWiki.dbName()]!!.title, targetWiki),
-                                        PageTitle(entity.sitelinks()[sourceWiki.dbName()]!!.title, sourceWiki)))
+                                val pair = PageTitle(entity.sitelinks()[targetWiki.dbName()]!!.title, targetWiki) to
```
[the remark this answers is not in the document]
Isira-Seneviratne: Yeah, of course.

app/src/main/java/org/wikipedia/userprofile/UserContributionsStats.kt
Isira-Seneviratne: I didn't know that, thanks!

PR #1706, opened 2020-10-16
Use the static equals() and hashCode() methods.

Isira-Seneviratne:


app/src/main/java/org/wikipedia/page/PageProperties.java
```diff
@@ -266,22 +266,24 @@ public boolean equals(Object o) {
                 && TextUtils.equals(wikiBaseItem, that.wikiBaseItem);
     }
 
+    @SuppressWarnings("NewApi")
```
cooltey, line 269: I think it would be better to wait until if we raise the minimum API level to API `24`, then every device can run this method properly.

dbrant:
No longer relevant, since we won't be using `hashCode` methods anyway soon.

PR #1707, opened 2020-10-16
Fix: show image from transition animation if the Gallery cannot draw it

cooltey:
Steps to reproduce:
1. Search [Oaxaca] in `enwiki`
2. Click on the lead image to enter the Gallery.
3. We will see the "Could not draw the image" message after the transition animation is completed.

Looks like the API cannot find the correct URL from the file name.
https://en.wikipedia.org/w/api.php?format=json&formatversion=2&errorformat=plaintext&action=query&prop=imageinfo&iiprop=timestamp|user|url|mime|extmetadata&iiurlwidth=320&titles=File:2029px-Oaxaca_in_Mexico.svg.png&iiextmetadatalanguage=en

https://commons.wikimedia.org/w/api.php?format=json&formatversion=2&errorformat=plaintext&action=query&prop=imageinfo&iiprop=timestamp|user|url|mime|extmetadata&iiurlwidth=320&titles=File:2029px-Oaxaca_in_Mexico.svg.png&iiextmetadatalanguage=en

### DIFF
--- a/app/src/main/java/org/wikipedia/util/UriUtil.java
+++ b/app/src/main/java/org/wikipedia/util/UriUtil.java
@@ -120,6 +120,19 @@ public final class UriUtil {
         return title.getUri() + "?wprov=" + context.getString(provId);
     }
 
+    @NonNull
+    public static String getFilenameFromUploadUrl(@NonNull String url) {
+        String[] splitArray = url.split("/");
+        String thumbnailName = splitArray[splitArray.length - 1];
+        if (splitArray.length > 2) {
+            String originalFilename = splitArray[splitArray.length - 2];
+            if (originalFilename.matches("^[\\w,\\s-]+\\.[A-Za-z]{3}$")) {
+                return originalFilename;
+            }
+        }
+        return thumbnailName;
+    }
+
     /**
      * Note that while this method also replaces '_' with spaces it doesn't fully decode the string.
      */
--- a/app/src/main/java/org/wikipedia/util/UriUtil.java
+++ b/app/src/main/java/org/wikipedia/util/UriUtil.java
@@ -124,11 +124,8 @@ public final class UriUtil {
     public static String getFilenameFromUploadUrl(@NonNull String url) {
         String[] splitArray = url.split("/");
         String thumbnailName = splitArray[splitArray.length - 1];
-        if (splitArray.length > 2) {
-            String originalFilename = splitArray[splitArray.length - 2];
-            if (originalFilename.matches("^[\\w,\\s-]+\\.[A-Za-z]{3}$")) {
-                return originalFilename;
-            }
+        if (url.contains("/thumb/") && splitArray.length > 2) {
+            return splitArray[splitArray.length - 2];
         }
         return thumbnailName;
     }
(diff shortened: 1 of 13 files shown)


dbrant (changes requested):
Wouldn't it be more interesting to understand why we're requesting an invalid image name, and fix it properly? And perhaps the fix might even be just one or two lines of code.

cooltey:
> 
> 
> Wouldn't it be more interesting to understand why we're requesting an invalid image name, and fix it properly? And perhaps the fix might even be just one or two lines of code.

Sure! Done.

app/src/test/java/org/wikipedia/util/UriUtilTest.java
```diff
@@ -75,4 +75,12 @@ public void testRemoveLinkPrefix() {
         assertThat(UriUtil.removeLinkPrefix("https://sr.wikipedia.org/sr-el/Барак_Обама"), is("Барак_Обама"));
         assertThat(UriUtil.removeLinkPrefix("https://sr.wikipedia.org/sr-ec/Барак_Обама"), is("Барак_Обама"));
     }
+
+    @Test
```
dbrant, line 79: excellent

app/src/test/java/org/wikipedia/util/UriUtilTest.java
```diff
@@ -75,4 +75,12 @@ public void testRemoveLinkPrefix() {
...
+    public void testFilenameFromUploadUrl() {
+        assertThat(UriUtil.getFilenameFromUploadUrl("https://upload.wikimedia.org/wikipedia/commons/thumb/8/83/Oaxaca_in_Mexico.svg/320px-Oaxaca_in_Mexico.svg.png"), is("Oaxaca_in_Mexico.svg"));
+        assertThat(UriUtil.getFilenameFromUploadUrl("https://upload.wikimedia.org/wikipedia/commons/thumb/8/8d/President_Barack_Obama.jpg/256px-President_Barack_Obama.jpg"), is("President_Barack_Obama.jpg"));
+        assertThat(UriUtil.getFilenameFromUploadUrl("https://upload.wikimedia.org/wikipedia/commons/8/8d/President_Barack_Obama.jpg"), is("President_Barack_Obama.jpg"));
+        assertThat(UriUtil.getFilenameFromUploadUrl("https://upload.wikimedia.org/wikipedia/en/thumb/0/0d/Avengers_Endgame_poster.jpg/216px-Avengers_Endgame_poster.jpg"), is("Avengers_Endgame_poster.jpg"));
+    }
```
dbrant, line 85: Try adding a test for `getFilenameFromUploadUrl("")` and see what happens.

dbrant (changes requested):

app/src/test/java/org/wikipedia/util/UriUtilTest.java
cooltey: Sure!

app/src/main/java/org/wikipedia/util/UriUtil.java
```diff
@@ -120,6 +120,19 @@ public static String getUrlWithProvenance(Context context, PageTitle title,
         return title.getUri() + "?wprov=" + context.getString(provId);
     }
 
+    @NonNull
+    public static String getFilenameFromUploadUrl(@NonNull String url) {
+        String[] splitArray = url.split("/");
+        String thumbnailName = splitArray[splitArray.length - 1];
+        if (splitArray.length > 2) {
+            String originalFilename = splitArray[splitArray.length - 2];
+            if (originalFilename.matches("^[\\w,\\s-]+\\.[A-Za-z]{3}$")) {
```
dbrant, line 129: It's not at all guaranteed that the filename extension will always be three (3) characters. Simple counterexample:
https://en.wikipedia.org/api/rest_v1/page/summary/NGC_4565

I'm pretty sure the path structure of the thumbnail URL is stable, so we can probably assume that the second-to-last path segment will be the filename. (and not require any regex)

dbrant (changes requested):

app/src/main/java/org/wikipedia/util/UriUtil.java
cooltey: You are right. 
Since I found that if I remove the regex, we will not get the correct filename from the second-to-last segment from an original file URL.
I have added a simpler check to check whether the URL contains `/thumb/` and extract the filename from it or not.

PR #1708, opened 2020-10-16
Replace the theme color button with plain circular buttons

cooltey:
We have to resize (re-style) the theme color buttons in order to add font-family buttons to the theme dialog.
https://phabricator.wikimedia.org/T264445

### DIFF
--- a/app/src/main/java/org/wikipedia/theme/ThemeChooserDialog.java
+++ b/app/src/main/java/org/wikipedia/theme/ThemeChooserDialog.java
@@ -17,6 +17,7 @@ import androidx.appcompat.widget.SwitchCompat;
 import androidx.core.content.ContextCompat;
 
 import com.google.android.material.bottomsheet.BottomSheetBehavior;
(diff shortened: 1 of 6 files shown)


sharvaniharan:
The circle looks very good. Couple things:
- The dark and Black circle colors might need to be more different. dark mode color needs to be slightly lighter i guess
- May be a single capital letter inside the circle? 'L' 'S' 'D' and 'B'?

cooltey:
> 
> 
> The circle looks very good. Couple things:
> 
>     * The dark and Black circle colors might need to be more different. dark mode color needs to be slightly lighter i guess
> 
>     * May be a single capital letter inside the circle? 'L' 'S' 'D' and 'B'?

Thanks! I'll share the comment with Robin, it's a great suggestion.

cooltey:
@dbrant @sharvaniharan I removed the Experimental tag since Robin has confirmed the design and it is ready for review now.

dbrant:
Yes! And a bunch of code removed, too.

PR #1709, opened 2020-10-17
API underpinnings for Watchlists

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T265379

**Notes:**
- We need to use token of type "watch" itself

- Watch and Unwatch use the same call, where 'unwatch' is set to null for watching.
```
 List<Integer> pages = new ArrayList<>();
 pages.add(38714);
 pages.add(21304840);
 disposables.add(ServiceFactory.get(WikiSite.forLanguageCode("en")).unWatch(null, TextUtils.join("|",pages),token)
```


app/src/main/java/org/wikipedia/watchlist/WatchListItem.kt
```diff
@@ -0,0 +1,18 @@
+package org.wikipedia.watchlist
+
+import com.google.gson.annotations.SerializedName
+
+class WatchListItem {
```
dbrant, line 5: Some of the fields in this class are awfully similar to `MwQueryPage`.  Could this be a subclass?

app/src/main/java/org/wikipedia/dataclient/Service.java
```diff
@@ -339,5 +339,21 @@
...
                                                      @NonNull @Field("batch") String batchLabels);
+    // ------- Watchlist -------
 
+    @GET(MW_API_PREFIX + "action=query&format=json&list=watchlist&continue=")
+    @NonNull Observable<MwQueryResponse> getWatchlist(@Nullable @Query("continue") String cont);
+
+    @Headers("Cache-Control: no-cache")
+    @POST(MW_API_PREFIX + "action=watch&format=json")
+    @FormUrlEncoded
+    Observable<EntityPostResponse> unWatch(@Field("unwatch") int unwatch,
```
dbrant, line 350: In this case you can just put `&unwatch=1` in the annotation, instead of as a parameter.

dbrant (changes requested):

app/src/main/java/org/wikipedia/watchlist/WatchListItem.kt
sharvaniharan: yes, it is actually a list of MwQueryPage(s)... so will use that itself.

app/src/main/java/org/wikipedia/dataclient/Service.java
```diff
@@ -339,5 +339,19 @@
...
+    // ------- Watchlist -------
 
+    @GET(MW_API_PREFIX + "action=query&list=watchlist")
+    @NonNull Observable<MwQueryResponse> getWatchlist(@Nullable @Query("continue") String cont);
+
+    @Headers("Cache-Control: no-cache")
+    @POST(MW_API_PREFIX + "action=watch")
+    @FormUrlEncoded
+    Observable<EntityPostResponse> unWatch(@Nullable @Field("unwatch") Integer unwatch,
+                                            @Field("pageids") String pageIds,
```
cooltey, line 351: Is it possible to align with the item above?

cooltey (commented):
Just have one minor comment.

app/src/main/java/org/wikipedia/dataclient/Service.java
sharvaniharan: @cooltey done. What do you think about the name? since we use this both for watching and unwatching, should we call it watchUnwatch(...)

cooltey: How about calling it `postWatch` since the `unwatch` is an `Integer` type of parameter and it is also a `post` type of action?

sharvaniharan: since we do not use 'post' method type anywhere else, i will just change it to watch(...) ... does that sound good? and when unwatch is available it is obvious what it does

cooltey: I've noticed that you used `@POST` and also `@FormUrlEncoded`, which is considered as a `post` method type. If that's not your intention, I would suggest you update it to `@GET` and remove the `@FormUrlEncoded`
```
    @POST(MW_API_PREFIX + "action=watch")
    @FormUrlEncoded
```

cooltey (changes requested):
Please see my comment in the thread: 
https://github.com/wikimedia/apps-android-wikipedia/pull/1709#discussion_r510456253

app/src/main/java/org/wikipedia/dataclient/Service.java
sharvaniharan: it needs to be POST @cooltey . This is for adding or removing a page from a user's watchlist.

sharvaniharan: > since we do not use 'post' method type anywhere else, i will just change it to watch(...) ... does that sound good? and when unwatch is available it is obvious what it does

Here I meant for naming, we not use post...(...) for any post methods, so i just called it watch(...) instead of postWatch(...)

sharvaniharan: Renamed for consistency.

cooltey (approved):

PR #1713, opened 2020-10-20
Introduce a serif font theme option

cooltey:
https://phabricator.wikimedia.org/T264445


app/src/main/java/org/wikipedia/WikipediaApp.java
```diff
@@ -301,6 +301,13 @@ public boolean setFontSizeMultiplier(int multiplier) {
         return false;
     }
 
+    public void setFontFamily(@NonNull String fontFamily) {
+        if (!fontFamily.equals(Prefs.getFontFamily())) {
+            Prefs.setFontFamily(fontFamily);
+            bus.post(new ThemeChangeEvent());
```
sharvaniharan, line 307: May be we should rename it to something more generic like ThemeFontChangeEvent() or AppearanceChangeEvent()?

sharvaniharan (changes requested):
Looks good.. I just wonder if we should show this only when we access the theme sheet from PageActivity..and not show it from settings screen, since this doesn't change overall app font, just article. <-question for @dbrant maybe

app/src/main/java/org/wikipedia/WikipediaApp.java
cooltey: Sounds good. I've updated to `ThemeFontChangeEvent()`.

app/src/main/java/org/wikipedia/analytics/AppearanceChangeFunnel.java
```diff
@@ -3,16 +3,19 @@
...
 
 import org.json.JSONObject;
+import org.wikipedia.Constants.InvokeSource;
 import org.wikipedia.WikipediaApp;
 import org.wikipedia.dataclient.WikiSite;
 import org.wikipedia.theme.Theme;
 
 public class AppearanceChangeFunnel extends Funnel {
     private static final String SCHEMA_NAME = "MobileWikiAppAppearanceSettings";
     private static final int REV_ID = 18113727;
```
dbrant, line 13: This needs to be a new revision ID, when Shay makes the schema changes.

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/AppearanceChangeFunnel.java
cooltey: Done.

cooltey:
Updated to the latest revision number
https://meta.wikimedia.org/wiki/Schema:MobileWikiAppAppearanceSettings

sharvaniharan:
Lgtm... still feel. the 'Dark' theme could be lighter, but if Robin has okayed, that is good enough. @dbrant will let you merge. Also, any thoughts on my previous comment? 

>  I just wonder if we should show this only when we access the theme sheet from PageActivity..and not show it from settings screen, since this doesn't change overall app font, just article"

PR #1720, opened 2020-10-22
Add Event Platform Client

mdholloway:
Adds the client code for interacting with the Wikimedia Event Platform (a.k.a. "Modern Event Platform").

The bulk of the new implementation is added in the new package `org.wikipedia.analytics.eventplatform`. Conceptually, the code is structured as a plain Java client "library" (represented by the `EventPlatformClient` class) and an `EventPlatformClientIntegration` class through which the `EventPlatformClient` "library" class interacts with various APIs provided by the Android platform and the app itself.

The remaining classes in the `org.wikipedia.analytics.eventplatform` package are mostly model classes, for events (`Event`), event platform intake service responses (`EventServiceResponse`), stream configurations (`StreamConfig`), and the sampling configuration components of stream configurations (`SamplingConfig`). `EventService` is a Retrofit service interface for the Event Platform intake APIs. The `DestinationEventService` enum contains the configurations for possible destination APIs designated in event stream configurations. (N.B. To match the iOS client behavior, the request for stream configurations filters for those with the destination event service `eventgate-analytics-external`, and accordingly only that event intake service will be used. The app can be updated later to adjust the filtering on stream config requests in the event that it is desired to submit app client error events to the Event Platform.)

Instruments added to the application code should interact with the `EventPlatformClient` class, specifically through the static method `EventPlatformClient.submit()`.

In `WikipediaApp.onCreate()`, the Event Platform Client is initialized. The client requests stream configurations from SharedPreferences, if available, and makes a request to the MediaWiki API for updated configurations. When an instrument submits an event to the client via `EventPlatformClient.submit()`, the client searches for an applicable stream configuration. If one is found, the client applies any sampling logic directed by the stream configuration. If the event is in sample, the client supplements the event with a timestamp, if needed, and enqueues the event for batched submission to the appropriate event platform intake API. Events are enqueued and submitted periodically in batches to limit the frequency with which the app wakes the device radio for event logging. If the queue reaches a configured maximum size, it is flushed regardless of the length of
time that has passed since the previous event submission.

A developer preference has been added to send a test event to an Event Platform intake API. The next steps are to identify a schema used by the Android app to migrate to the Event Platform, do the migration, and update the app to submit events to the migrated schema.

Since this commit is already large, session timeouts are left for a follow-up PR.

Bug: https://phabricator.wikimedia.org/T228179


app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
```diff
@@ -0,0 +1,459 @@
...
+            return PAGEVIEW_ID;
+        }
+
+        /**
+         * Generate a session identifier.
+         *
+         * @return session ID
+         *
+         * The identifier is a string of 20 zero-padded hexadecimal digits
+         * representing a uniformly random 80-bit integer.
```
mdholloway, line 303: Comment for discussion: why don't why use a (v4) UUID for session IDs (across all clients)?  It seems to be the convention for all other randomly generated IDs, and would probably also be easier and better ensure consistency across platforms.

mdholloway: I stumbled upon a partial answer to my own question. This token format is following that produced by [mw.user.generateRandomSessionId](https://github.com/wikimedia/mediawiki/blob/master/resources/src/mediawiki.user.js#L25-L75). There was some discussion around how this ID is generated in [T240452](https://phabricator.wikimedia.org/T240452).  I still think we'd be better off joining the rest of the world and using a [UUID](https://www.npmjs.com/package/uuid) for this rather than a custom ID format. (Would changing it break anything? If I understand correctly, this is only used for event logging.)

dbrant: +1 for using something that is used in virtually all other contexts where a "unique identifier" is needed.

bearloga: +1 to this. Just want to note for later:

We may want to create a `mw.user.generateUUIDv4()` (since [`generateRandomSessionId` is in use by a bunch of instruments](https://github.com/search?q=org%3Awikimedia+mw.user.generateRandomSessionId&type=code) that then use it to determine sampling) and figure out a migration plan for instruments that use `mw.user.sessionId()` (which relies on `mw.user.generateRandomSessionId()`)

mdholloway: Excellent.  Seems like we're all on the same page on this.

I imagine it would create little to no migration work for existing consumers if we just have `mw.user.generateRandomSessionId` itself call out to `mw.user.generateUUIDv4`.  If we do that, then I think the only thing we'd need to do is announce the change a little while beforehand in case anyone is doing anything weird like relying on the token format.

mdholloway: Phabbed: https://phabricator.wikimedia.org/T266813

sharvaniharan:
Thank you so much for the PR @mdholloway . This is a solid foundation for any work we will need to do on this! 
As discussed I will pull this into a feature branch.

PR #1724, opened 2020-10-23
Design tweak: News

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T261672#6567557
When we bold parts of rich text, it should be bold, and not faux bolded by converting to sans-serif-medium


cooltey:
Although it is merged, I am curious what's the different between `setTypeface(Typeface.DEFAULT_BOLD);` and `setTypeface(Typeface.create(Typeface.DEFAULT, Typeface.BOLD))`

PR #1730, opened 2020-10-26
Tweak: add more transition animation supports on Explore feed cards

cooltey:
https://phabricator.wikimedia.org/T263259#6579105

This PR also updates the "On this day" card design regarding Robin's comment: https://phabricator.wikimedia.org/T261673#6582483


app/src/main/java/org/wikipedia/feed/view/ListCardItemView.java
```diff
@@ -93,7 +94,7 @@ public ListCardItemView(Context context) {
 
     @OnClick void onClick(View view) {
         if (callback != null && entry != null && card != null) {
-            callback.onSelectPage(card, entry);
+            callback.onSelectPage(card, entry, TransitionUtil.getSharedElements(getContext(), titleView, subtitleView, imageView));
```
dbrant, line 97: Are we sure we want the `title` and `subtitle` as shared elements here? They are a different font and different size, and don't really look smooth when transitioning.

app/src/main/java/org/wikipedia/feed/view/CardLargeHeaderView.java
```diff
@@ -124,4 +126,8 @@ private void setGradientDrawableBackground(@ColorInt int leftColor, @ColorInt in
         gradientDrawable.setCornerRadius(getResources().getDimension(R.dimen.wiki_card_radius));
         borderContainer.setBackground(gradientDrawable);
     }
+
+    public Pair<View, String>[] getSharedElements() {
+        return TransitionUtil.getSharedElements(getContext(), titleView, subtitleView, imageView);
```
dbrant, line 131: The transition for this looks pretty good, except for RTL articles. Is there any way to fix it?

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/view/ListCardItemView.java
cooltey: That's a good question and I have removed them. Also I'll leave a comment on the ticket.

app/src/main/java/org/wikipedia/feed/view/CardLargeHeaderView.java
cooltey: Good catch! Done.

PR #1733, opened 2020-10-27
Only refresh feed if languages are changed in search

zhlcbri:
**Background**
https://phabricator.wikimedia.org/T198385

The feed automatically refreshes every time user enters language selection from search even if nothing is changed.

**Change**
- [x] Save and compare the initial and final language lists before and after search. Only refresh feed if they are different.

**Testing**
Add | Remove | Rearrange | No change |
--- | --- | --- | --- | 
add |  remove | rearrange | no_change |


app/src/main/java/org/wikipedia/search/SearchFragment.java
```diff
@@ -201,7 +208,12 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
         super.onActivityResult(requestCode, resultCode, data);
         if (requestCode == ACTIVITY_REQUEST_ADD_A_LANGUAGE_FROM_SEARCH) {
             int position = 0;
-            requireActivity().setResult(RESULT_LANG_CHANGED);
+
+            finalLanguageList = StringUtil.listToJsonArrayString(app.language().getAppLanguageCodes());
```
cooltey, line 212: Looks like the `finalLanguageList` can be a local variable.

cooltey (changes requested):

PR #1736, opened 2020-10-27
Bounce "thankyou." domain to external browser too

ejegg:
We've recently moved the donation Thank You pages to
thankyou.wikiPedia.org so we can set cookies which are readable
across the encyclopedias.

Bug: T259312


dbrant:
very nice; thanks!

PR #1738, opened 2020-10-27
Se card switch to recyclerview

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T261676


sharvaniharan:
No longer needs to be switched to ViewPager

PR #1740, opened 2020-10-29
Add more info for tracking unknown crashes

cooltey:
https://phabricator.wikimedia.org/T266034

### DIFF
--- a/app/src/main/java/org/wikipedia/WikipediaApp.java
+++ b/app/src/main/java/org/wikipedia/WikipediaApp.java
@@ -401,6 +401,7 @@ public class WikipediaApp extends Application {
             if (appLanguageState != null) {
                 putCrashReportProperty("app_primary_language", appLanguageState.getAppLanguageCode());
                 putCrashReportProperty("app_languages", appLanguageState.getAppLanguageCodes().toString());
+                putCrashReportProperty("app_install_id", getAppInstallID());
             }
         }
     }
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/wikipedia/activity/BaseActivity.java
```diff
@@ -110,6 +110,8 @@
         setNavigationBarColor(ResourceUtil.getThemedColor(this, R.attr.paper_color));
 
         maybeShowLoggedOutInBackgroundDialog();
+
+        WikipediaApp.getInstance().putCrashReportProperty("app_local_class_name", getLocalClassName());
```
dbrant, line 114: I don't think this will actually work:  The crash report properties are stored in memory. When a crash happens, the properties are wiped.  That's why the only properties that get sent are the ones that are initialized when `WikipediaApp` is first created.

app/src/main/java/org/wikipedia/WikipediaApp.java
```diff
@@ -401,6 +401,7 @@ private void updateCrashReportProps() {
             if (appLanguageState != null) {
                 putCrashReportProperty("app_primary_language", appLanguageState.getAppLanguageCode());
                 putCrashReportProperty("app_languages", appLanguageState.getAppLanguageCodes().toString());
+                putCrashReportProperty("app_install_id", getAppInstallID());
```
dbrant, line 404: This is a pretty good idea; it would let us look through eventlogging and see in which tables this ID pops up.

app/src/main/java/org/wikipedia/activity/BaseActivity.java
```diff
@@ -110,6 +110,8 @@
         setNavigationBarColor(ResourceUtil.getThemedColor(this, R.attr.paper_color));
 
         maybeShowLoggedOutInBackgroundDialog();
+
+        Prefs.setLocalClassName(getLocalClassName());
```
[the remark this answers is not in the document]
cooltey: Make sense! Done.

PR #1742, opened 2020-10-29
Minor SE feed card update: Adding footer

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T261676


app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsCardView.kt
```diff
@@ -10,11 +10,17 @@ import kotlinx.android.synthetic.main.view_suggested_edits_card.view.*
...
 import org.wikipedia.descriptions.DescriptionEditActivity
 import org.wikipedia.descriptions.DescriptionEditActivity.Action.*
+import org.wikipedia.feed.view.CardFooterView
 import org.wikipedia.feed.view.DefaultFeedCardView
 import org.wikipedia.feed.view.FeedAdapter
 import org.wikipedia.views.PositionAwareFragmentStateAdapter
 
-class SuggestedEditsCardView(context: Context) : DefaultFeedCardView<SuggestedEditsCard>(context), SuggestedEditsFeedClient.Callback {
+class SuggestedEditsCardView(context: Context) : DefaultFeedCardView<SuggestedEditsCard>(context)
+        , SuggestedEditsFeedClient.Callback, CardFooterView.Callback {
```
cooltey, line 19: Very minor but can we have the line break after the comma instead of before the comma?

cooltey: @sharvaniharan I'll self merge the follow-up PR soon.

app/src/main/res/layout/fragment_suggested_edits_card_item.xml
```diff
@@ -42,9 +42,12 @@
             android:id="@+id/cardItemContainer"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
+            android:addStatesFromChildren="true"
+            android:background="?attr/selectableItemBackground"
             android:clickable="true"
             android:focusable="true"
-            android:orientation="vertical">
+            android:orientation="vertical"
+            android:theme="?attr/appTouchRippleTheme">
```
cooltey, line 50: Would it be possible to just add `android:foreground="?attr/selectableItemBackground"` instead of creating a new style for only one specific ripple animation?

cooltey (changes requested):

app/src/main/res/layout/fragment_suggested_edits_card_item.xml
sharvaniharan: Dark theme doesn't work @cooltey .

sharvaniharan: Also I feel having a theme setting would be better than the api23 level foreground, may be for other places too? What do you feel?

cooltey: I tested on other views like list view or featured article view, the ripple animation does not work on black theme either. Would it be possible to have a separate PR of this? (and maybe a ticket for it as well?)

PR #1747, opened 2020-10-30
Fix: properly retry the request when getting empty source description

cooltey:
This is a proper fix of #1719, since it also fixes the broken views in the SE feed of translate description in `enwiki`.

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsItemFragment.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsItemFragment.kt
@@ -8,6 +8,7 @@ import android.view.View.GONE
 import android.view.View.VISIBLE
 import android.view.ViewGroup
 import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers
+import io.reactivex.rxjava3.core.Observable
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsItemFragment.kt
```diff
@@ -71,6 +72,13 @@ class SuggestedEditsCardsItemFragment : SuggestedEditsItemFragment() {
         when (parent().action) {
             TRANSLATE_DESCRIPTION -> {
                 disposables.add(EditingSuggestionsProvider.getNextArticleWithMissingDescription(WikiSite.forLanguageCode(parent().langFromCode), parent().langToCode, true)
+                        .map {
+                            if (it.second.description.isNullOrEmpty()) {
```
sharvaniharan, line 76: @cooltey During translation. suggestion, wont this always be empty?

cooltey: You can notice from the code below:
```
                            val source = pair.second
                            val target = pair.first
```

PR #1748, opened 2020-11-01
Only refresh feed if languages or feed configurations are changed in settings

zhlcbri:
**Background**
https://phabricator.wikimedia.org/T198385

The feed automatically refreshes every time user enters language selection or feed configuration from settings even if nothing is changed.

**Change**
- [x] Save the initial and final language lists and feed card lists when entering and leaving `SettingsActivity`. Compare these lists and only set activity results if they are different.

**Testing**
- [x] Go to Settings --> add/remove/rearrange items in languages --> go back to feed and ensure it refreshes
- [x] Go to Settings --> enable/disable/rearrange items in feed configurations --> go back to feed and ensure it refreshes
- [x] Go to Settings --> click on languages/feed configurations but do not make changes --> go back to feed and ensure it **does not refresh**

\ | Before | After |
--- | --- | --- |
Languages | languages_before | languages_after |
Feed configuration | feed_before | feed_after |

### DIFF
--- a/app/src/main/java/org/wikipedia/settings/SettingsActivity.java
+++ b/app/src/main/java/org/wikipedia/settings/SettingsActivity.java
@@ -2,10 +2,13 @@ package org.wikipedia.settings;
 
 import android.content.Context;
 import android.content.Intent;
+import android.os.Bundle;
 
 import androidx.annotation.NonNull;
 
+import org.wikipedia.WikipediaApp;
 import org.wikipedia.activity.SingleFragmentActivity;
+import org.wikipedia.util.StringUtil;
 
 import static org.wikipedia.Constants.ACTIVITY_REQUEST_ADD_A_LANGUAGE;
 import static org.wikipedia.Constants.ACTIVITY_REQUEST_FEED_CONFIGURE;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/settings/SettingsActivity.java
```diff
@@ -23,12 +34,29 @@ public SettingsFragment createFragment() {
...
+    }
+
     @Override
     public void onActivityResult(int requestCode, int resultCode, Intent data) {
         super.onActivityResult(requestCode, resultCode, data);
-        if (requestCode == ACTIVITY_REQUEST_ADD_A_LANGUAGE) {
+
+        String finalLanguageList = StringUtil.listToJsonArrayString(app.language().getAppLanguageCodes());
+        List<Boolean> finalFeedCardsEnabled = Prefs.getFeedCardsEnabled();
+        List<Integer> finalFeedCardsOrder = Prefs.getFeedCardsOrder();
```
cooltey, line 52: Would it be possible to directly use these the `Prefs.getFeedCardsEnabled()` and `Prefs.getFeedCardsOrder()` in the `if` statements instead of creating new variables?

zhlcbri: Yes we could do that, but the code would be less readable. I think keeping these variables makes the code more consistent as we have `finalLanguageList` too. Are you concerned about the performance impact?

cooltey: Although the performance impact might be really small, the `Prefs.getFeedCardsOrder()` and `Prefs.getFeedCardsEnabled()` have enough information from the names that indicate the values are the "final" or "current" data, which can compare with the "initial" ones.

PR #1758, opened 2020-11-04
Converted ClipboardUtil java class to kotlin

gouri-panda:
Converted ClipboardUtil java class to kotlin and rename getManager method to getClipboardManager.


dbrant:
Many thanks for your contributions, but just a note about "Convert X to Kotlin" pull requests:
We are gradually converting our code base to Kotlin at our own pace, and are aware that Android Studio has a convenient "convert to Kotlin" function, which we can use anytime we like.  The reason we haven't automatically converted all of our code yet is because there is virtually no benefit in many cases (like this PR) in terms of code clarity.  But there _is_ a cognitive overhead in converting to Kotlin, since each conversion will require thorough review and testing.

gouri-panda:
@dbrant Thank you for commenting in a short time. I just need to start familiar with this project but I can't find any suitable bug for this now? Can you give me any good first issue that will be great.
P.S  Should I close these two PR?

PR #1760, opened 2020-11-04
Smoothly scroll page up a bit when changing themes.

dbrant:
https://phabricator.wikimedia.org/T267243


sharvaniharan:
Scroll animation is very smooth. Just a minor concern.. the scroll almost looks like a ui element being pushed up in a buggy fashion. So wondering if a toast like "scrolling to show theme change", or scrolling back after the theme change makes sense?

dbrant:
> Scroll animation is very smooth. Just a minor concern.. the scroll almost looks like a ui element being pushed up in a buggy fashion. So wondering if a toast like "scrolling to show theme change", or scrolling back after the theme change makes sense?

Definitely no Toast, but scrolling back up is... a pretty good idea! Done.

PR #1763, opened 2020-11-05
Tweak: optimize ripple effect in black theme

cooltey:
https://phabricator.wikimedia.org/T267012
This PR also optimizes the clickable area in `FeaturedArticleCardView`, which makes the ripple effect more senses.


sharvaniharan:
Like the update to the syles! merging now..

PR #1769, opened 2020-11-06
Analytics for notifications.

dbrant:


app/src/main/java/org/wikipedia/notifications/NotificationPresenter.java
```diff
@@ -85,13 +83,19 @@ public static void showNotification(@NonNull Context context, @NonNull Notificat
     }
 
     public static void showMultipleUnread(@NonNull Context context, int unreadCount) {
-        NotificationCompat.Builder builder = getDefaultBuilder(context);
+        NotificationCompat.Builder builder = getDefaultBuilder(context, unreadCount, NotificationPollBroadcastReceiver.TYPE_MULTIPLE);
```
cooltey, line 86: Slightly confused about sending `unreadCount` as the `notification_id` here. Maybe rename the parameter or add a comment in the method is helpful?

dbrant: Comment added.

PR #1770, opened 2020-11-06
Minor adjustment to news margin

sharvaniharan:
image

### DIFF
--- a/app/src/main/res/layout/view_horizontal_scroll_list_item_card.xml
+++ b/app/src/main/res/layout/view_horizontal_scroll_list_item_card.xml
@@ -16,6 +16,8 @@
         android:layout_height="0dp"
         android:layout_marginStart="16dp"
         android:layout_marginEnd="16dp"
+        android:layout_marginTop="12dp"
+        android:layout_marginBottom="16dp"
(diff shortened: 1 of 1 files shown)


cooltey (changes requested):
The `layout_marginTop` and `layout_marginBottom` are not working properly because of the fixed height of their parent. ( `android:layout_height="288dp"` )

sharvaniharan:
it is not just tha @cooltey ... the text varies in length too, so the top and bottom spaces will vary a lot dynamically... but i think Robin's point is that when there is enough text to fill all lines, about 4 lines of text, he wants to see 12dp on top and 16dp on bottom... i would say let us merge this for now, and then you can have a discussion with Robin on Monday and tweak it?

cooltey:
> 
> 
> it is not just that @cooltey ... the text varies in length too, so the top and bottom spaces will vary a lot dynamically... but i think Robin's point is that when there is enough text to fill all lines, about 4 lines of text, he wants to see 12dp on top and 16dp on bottom... i would say let us merge this for now, and then you can have a discussion with Robin on Monday and tweak it?

Make sense to me. Will do.

PR #1775, opened 2020-11-07
Always apply white background when loading images

cooltey:
In the "In the news" item, the "Kabul University" list item has a transparent image, and which is not able to see in black mode. It would be good if we always have the white background when loading images.


cooltey:
Too many conflicts and it's easier to open a new PR #1781

PR #1781, opened 2020-11-09
Always apply white background when loading images

cooltey:
In the "In the news" item, the "Kabul University" list item has a transparent image, and which is not able to see in black mode. It would be good if we always have a white background when loading images.


sharvaniharan:
Lgtm... Just keeping it unmerged until we test more images, just in case we find an edge case where having the white bg will not work..

dbrant (changes requested):
This breaks rounded corners in thumbnail images:

image

cooltey:
> 
> 
> This breaks rounded corners in thumbnail images:
> 
> image

Done.

PR #1783, opened 2020-11-09
Fix: reset margins to zero in the Announcement dialog on the article screen

cooltey:
The recently change in #1771 added a bottom margin to the view, and we should reset it to zero in the dialog on the article screen.

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/announcement/AnnouncementCardView.java
+++ b/app/src/main/java/org/wikipedia/feed/announcement/AnnouncementCardView.java
@@ -26,6 +26,8 @@ import butterknife.BindView;
 import butterknife.ButterKnife;
 import butterknife.OnClick;
 
+import static android.view.ViewGroup.LayoutParams.MATCH_PARENT;
+
 public class AnnouncementCardView extends DefaultFeedCardView<AnnouncementCard> {
     public interface Callback {
         void onAnnouncementPositiveAction(@NonNull Card card, @NonNull Uri uri);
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/feed/announcement/AnnouncementCardView.java
```diff
@@ -104,6 +104,7 @@ public AnnouncementCardView(@NonNull Context context) {
             LinearLayout.LayoutParams layoutParams = (LinearLayout.LayoutParams) container.getLayoutParams();
             layoutParams.setMarginStart(0);
             layoutParams.setMarginEnd(0);
```
sharvaniharan, line 106: Not needed anymore

cooltey: Remove the lines will not work, because `setMargins()` only applies for `left`, `top`, `bottom`, `right`.

dbrant: Couldn't we avoid all of this by just saying  `container.setLayoutParams(new  LinearLayout.LayoutParams(MATCH_PARENT, MATCH_PARENT))`

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/announcement/AnnouncementCardView.java
cooltey: You are correct. Done!

PR #1789, opened 2020-11-10
Make Randomizer design consistent

cooltey:
This PR has the following changes:

1. Make the card view inside the Randomizer has the same design as the card in the Explore feed, and it supports the transition animation. 
2. Make the Randomizer supports specific language code from the Explore feed.
3. Optimize the transition animation smoother in `FeaturedArticleCardView`.


dbrant (changes requested):
This breaks Landscape mode:

image

cooltey:
> 
> 
> This breaks Landscape mode:
> 

I added a `ScrollView` to fix it, not sure if that is a workable solution for you.

PR #1792, opened 2020-11-11
 Transition MobileWikiAppUserContribution to the new MEP 

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T228179


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -316,7 +317,7 @@ class SuggestedEditsTasksFragment : Fragment() {
         clearContents()
         disabledStatesView.setIPBlocked()
         disabledStatesView.visibility = VISIBLE
-        UserContributionFunnel.get().logIpBlock()
```
dbrant, line 319: Don't we still want to make this old funnel call, in addition to the new one?

app/src/main/java/org/wikipedia/analytics/eventplatform/UserContributionEvent.kt
```diff
@@ -0,0 +1,108 @@
...
+        action = "ip_block"
+        submitEvent()
+    }
+
+    private fun submitEvent() {
+        WikipediaApp.getInstance().eventPlatformClient.submit(this)
+    }
+
+    companion object {
+        private var INSTANCE: UserContributionEvent? = null
```
dbrant, line 94: I understand you were modeling this class on the old `UserContributionFunnel`, but this class must _absolutely not_ be a singleton.

app/src/main/java/org/wikipedia/analytics/eventplatform/UserContributionEvent.kt
```diff
@@ -0,0 +1,108 @@
...
+        submitEvent()
+    }
+
+    fun logIpBlock() {
+        action = "ip_block"
+        submitEvent()
+    }
+
+    private fun submitEvent() {
+        WikipediaApp.getInstance().eventPlatformClient.submit(this)
```
dbrant, line 90: Think about what's happening here.
You're submitting `this`, which is a singleton. If you call additional `log...()` functions, this will get submitted again, but it's the same singleton being submitted.  It will just create multiple references to the same object in the `QUEUE`.  The queue is not emptied right away, so multiple references will get queued up.

Each of the `log()` functions modifies the `action` of this singleton. So... it will cause all of the items in the `QUEUE` to have the same action when they get sent.

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
```diff
@@ -39,7 +39,7 @@
      * Inputs: network connection state on/off, connection state bad y/n?
      * Taken out of iOS client, but flag can be set on the request object to wait until connected to send
      */
-    private static boolean ENABLED = isOnline() && isEventLoggingEnabled();
```
cooltey, line 42: Not sure but what's the reason for removing this?

app/src/main/java/org/wikipedia/analytics/eventplatform/UserContributionEvent.kt
```diff
@@ -0,0 +1,96 @@
+package org.wikipedia.analytics.eventplatform
+
+import com.google.gson.annotations.SerializedName
+
+class UserContributionEvent : Event(SCHEMA_NAME, STREAM_NAME) {
+    @SerializedName("action") private var action: String = ""
```
cooltey, line 6: Is it really necessary to add `@SerializedName("action")` if the name is already `action`?

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -320,6 +321,7 @@ class SuggestedEditsTasksFragment : Fragment() {
         disabledStatesView.setIPBlocked()
         disabledStatesView.visibility = VISIBLE
         UserContributionFunnel.get().logIpBlock()
+        UserContributionEvent().logIpBlock()
```
cooltey, line 324: If we use `UserContributionEvent()` directly, doesn't that mean we have to create an instance for every action? Does that take extra memory usages?

cooltey (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
sharvaniharan: This is per our discussion to make this check only just before we send the events @cooltey

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
sharvaniharan: That was the problem with the previous way of implementation @cooltey . When only one object is used [as in singleton], the same object was queued multiple times. So this is the workaround..

cooltey: I see. Make sense.

cooltey:
Hi @sharvaniharan, I cannot find a way to pass the test cases either. Maybe it is fine to remove the test cases.

PR #1797, opened 2020-11-17
Implement swipe-to-dismiss in Gallery

cooltey:
This is probably the easiest solution for swipe-to-dismiss gestures. 
https://phabricator.wikimedia.org/T263048

### DIFF
--- a/app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.java
+++ b/app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.java
@@ -1,5 +1,7 @@
 package org.wikipedia.gallery;
 
+import android.animation.Animator;
+import android.annotation.SuppressLint;
 import android.graphics.Bitmap;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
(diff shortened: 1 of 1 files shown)


sharvaniharan:
Only 2 minor issues:
- I can see a black flicker when - click on the lead image, go to 2nd or 3rd image, and swipe - the action creates a flicker, since we are not transitioning back to the same image..[minor]
- when you click on the lead image, open gallery, turn to landscape, and swipe... which looks like a bug although it is the correct behavior... a transition without an image to transition to.

Since both are very minor, @dbrant please merge if it is ok.

dbrant:
I suppose this "works", but only in the most technical sense.  It doesn't feel very natural, and it doesn't feel effortless. I think, if we can't make the swiping motion track with the user's finger in real time, then it's not worth having.

cooltey:
> I suppose this "works", but only in the most technical sense. It doesn't feel very natural, and it doesn't feel effortless. I think, if we can't make the swiping motion track with the user's finger in real time, then it's not worth having.

I agreed with you. Let's keep this PR open and try to find a better implementation for the swiping motion in the future.

PR #1801, opened 2020-11-18
Main Watchlist UI.

dbrant:
This implements all of the following:

* The main Watchlist screen.
* Filtering of different types of pages (pages/talk/other).
* Seamlessly combine results from different language wikis.
* Select language(s) to show.
* Launches talk activity for user names.
* Stub hooks to launch diff activity from individual items.

https://phabricator.wikimedia.org/T269445


app/src/main/res/layout/item_watchlist.xml
```diff
@@ -0,0 +1,130 @@
...
+            android:paddingStart="12dp"
+            android:paddingEnd="12dp"
+            app:layout_constraintTop_toBottomOf="@id/summaryText"
+            app:layout_constraintStart_toStartOf="parent"
+            android:fontFamily="sans-serif-medium"
+            android:textAllCaps="false"
+            android:textColor="@color/green50"
+            tools:text="+123"/>
+
+        <com.google.android.material.button.MaterialButton
```
cooltey, line 107: The IPv6 address does not fit the screen properly.
Screenshot_20201208-134621_Wikipedia Dev

app/src/main/res/layout/item_watchlist_language.xml
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
cooltey, line 2: Should we use a fixed size for language code view, just like what we do in the search screen?
Screenshot_20201208-133237_Wikipedia Dev

app/src/main/res/layout/item_watchlist.xml
```diff
@@ -0,0 +1,130 @@
...
+    android:layout_marginTop="8dp"
+    android:layout_marginBottom="8dp">
+
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:id="@+id/containerView"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:background="?attr/selectableItemBackgroundBorderless"
+        android:clickable="true"
+        android:focusable="true"
```
cooltey, line 18: Would it be possible to set the `@+id/containerView` in the parent `WikiCardView` instead of here? If so, we will no longer need to set the background explicitly since the `WikiCardView` has a proper ripple color.

cooltey (changes requested):

app/src/main/res/layout/item_watchlist.xml
dbrant: Yep, I just gave it a maxWidth, which should take care of these cases.

app/src/main/res/layout/item_watchlist_language.xml
dbrant: I fixed it so that it properly expands horizontally.

app/src/main/res/layout/item_watchlist.xml
dbrant: Whoops, forgot to do it that way.

app/src/main/java/org/wikipedia/dataclient/Service.java
```diff
@@ -126,6 +126,9 @@
     @GET(MW_API_PREFIX + "action=query&generator=unreviewedimagelabels&guillimit=10&prop=imagelabels|imageinfo&iiprop=timestamp|user|url|mime|extmetadata&iiurlwidth=" + PREFERRED_THUMB_SIZE)
     @NonNull Observable<MwQueryResponse> getImagesWithUnreviewedLabels(@NonNull @Query("uselang") String lang);
 
+    @GET(MW_API_PREFIX + "action=query&meta=siteinfo&list=watchlist&wllimit=500&wlallrev=1&wlprop=ids|title|flags|comment|parsedcomment|timestamp|sizes|user")
```
cooltey, line 129: Noticed there has two `getWatchlist()` to get the `watchlist`, should we update the existing one instead of creating a new one?

dbrant: Whoops, correct!

app/src/main/java/org/wikipedia/watchlist/WatchlistFragment.kt
```diff
@@ -0,0 +1,309 @@
...
+
+    internal inner class RecyclerAdapter() : RecyclerView.Adapter<RecyclerView.ViewHolder>() {
+        constructor(items: List<Any>) : this() {
+            this.items = items
+        }
+
+        private var items: List<Any> = ArrayList()
+        private val VIEW_TYPE_HEADER = 0
+        private val VIEW_TYPE_DATE = 1
+        private val VIEW_TYPE_ITEM = 2
```
cooltey, line 217: Should we put these `const` variables in the `companion object`?

app/src/main/java/org/wikipedia/watchlist/WatchlistFragment.kt
```diff
@@ -0,0 +1,309 @@
...
+        (requireActivity() as AppCompatActivity).supportActionBar!!.setDisplayHomeAsUpEnabled(true)
+        (requireActivity() as AppCompatActivity).supportActionBar!!.title = getString(R.string.watchlist_title)
+
+        return view
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        watchlistRefreshView.setColorSchemeResources(ResourceUtil.getThemedAttributeId(requireContext(), R.attr.colorAccent))
```
cooltey, line 57: The screen has two progress bars on it when swiping to refresh, would it be possible to avoid that?

app/src/main/java/org/wikipedia/watchlist/WatchlistFragment.kt
dbrant: yep!

app/src/main/java/org/wikipedia/watchlist/WatchlistFragment.kt
dbrant: yep!

PR #1802, opened 2020-11-18
Fix: use PX on TabCountsView text size

cooltey:
https://phabricator.wikimedia.org/T268172


dbrant:
I'm not sure I understand why this would fix the issue.
I can't reproduce it on any of my devices to check. What device or emulator are you using?

cooltey:
> I'm not sure I understand why this would fix the issue.
> I can't reproduce it on any of my devices to check. What device or emulator are you using?

I am using the Huawei tablet and set the "system text size" to the largest one.

PR #1805, opened 2020-11-19
Use buffering only whilst Offline

sharvaniharan:
Buffering while online creates unnecessary flurry of 128 network calls. We can limit this batching to 10 when online and 128 while offline

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
@@ -259,7 +259,9 @@ public final class EventPlatformClient {
                     }
                 }
                 events.removeAll(eventsList);
-                postEvent(getStreamConfig(stream), eventsList);
+                if (eventsList != null && stream != null) {
+                    postEvent(getStreamConfig(stream), eventsList);
+                }
             }
         }
     }
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
@@ -243,9 +243,6 @@ public final class EventPlatformClient {
          * @param events list of events
          */
         static void send(List<Event> events) {
-            if (!ENABLED) {
-                return;
-            }
             Map<String, ArrayList<Event>> eventsByStream = new HashMap<>();
             for (Event event : events) {
                 String stream = event.getStream();
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
@@ -19,7 +19,7 @@ import static org.wikipedia.analytics.eventplatform.EventPlatformClientIntegrati
 import static org.wikipedia.analytics.eventplatform.EventPlatformClientIntegration.getStoredSessionId;
 import static org.wikipedia.analytics.eventplatform.EventPlatformClientIntegration.getStoredStreamConfigs;
 import static org.wikipedia.analytics.eventplatform.EventPlatformClientIntegration.isOnline;
-import static org.wikipedia.analytics.eventplatform.EventPlatformClientIntegration.postEvent;
+import static org.wikipedia.analytics.eventplatform.EventPlatformClientIntegration.postEvents;
 import static org.wikipedia.analytics.eventplatform.EventPlatformClientIntegration.setStoredSessionId;
 import static org.wikipedia.analytics.eventplatform.EventPlatformClientIntegration.setStoredStreamConfigs;
 import static org.wikipedia.analytics.eventplatform.SamplingConfig.Identifier.DEVICE;
@@ -252,7 +252,7 @@ public final class EventPlatformClient {
                 eventsByStream.get(stream).add(event);
             }
             for (String stream : eventsByStream.keySet()) {
-                postEvent(getStreamConfig(stream), eventsByStream.get(stream));
+                postEvents(getStreamConfig(stream), eventsByStream.get(stream));
             }
         }
     }
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
@@ -171,9 +171,14 @@ public final class EventPlatformClient {
         private static final int WAIT_MS = 30000;
 
         /*
-         * When QUEUE.size() exceeds this value TIMER becomes non-interruptable.
+         * While coming out of offline, When QUEUE.size() exceeds this value TIMER becomes non-interruptable.
          */
-        private static final int MAX_QUEUE_SIZE = WikipediaApp.getInstance().isOnline() ? 10 : 128;
+        private static final int MAX_OFFLINE_QUEUE_SIZE = 128;
+
+        /*
+         * While online, When QUEUE.size() exceeds this value TIMER becomes non-interruptable.
+         */
+        private static final int MAX_ONLINE_QUEUE_SIZE = 10;
 
         /*
          * IMPLEMENTATION NOTE: Java Timer will provide the desired asynchronous
@@ -222,7 +227,7 @@ public final class EventPlatformClient {
             QUEUE.add(event);
 
             if (ENABLED) {
-                if (QUEUE.size() >= MAX_QUEUE_SIZE) {
+                if (QUEUE.size() >= (WikipediaApp.getInstance().isOnline() ? MAX_ONLINE_QUEUE_SIZE : MAX_OFFLINE_QUEUE_SIZE)) {
                     /*
                      * >= because while sending is disabled, any number of items
                      * could be added to QUEUE without it emptying.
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
```diff
@@ -171,7 +173,7 @@ static void addEventMetadata(Event event) {
         /*
          * When QUEUE.size() exceeds this value TIMER becomes non-interruptable.
          */
-        private static final int MAX_QUEUE_SIZE = 128;
+        private static final int MAX_QUEUE_SIZE = WikipediaApp.getInstance().isOnline() ? 10 : 128;
```
dbrant, line 176: Since this is `static` and `final`, won't it be initialized once and stay that way forever?

sharvaniharan: whoops!

mdholloway:
I have no particular objection to this, but from your questions in our meeting and in Slack the other day, I thought you'd planned to update to send batched rather than individual requests.  What are you thinking about that now?

mdholloway:
(FWIW, I sort of regret taking batching out of my original PR, not that it would be difficult to put back in.)

sharvaniharan:
Correct @mdholloway I will be batching them... Is there an upper limit to how many can be batched together?

sharvaniharan:
Done @mdholloway @dbrant @cooltey

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
```diff
@@ -239,14 +244,24 @@ static void schedule(Event event) {
...
+            while (!events.isEmpty()) {
+                List<Event> eventsList = new ArrayList<>();
+                String stream = events.get(0).getStream();
+                for (Event event : events) {
+                    if (event.getStream().equals(stream)) {
+                        eventsList.add(event);
+                    }
+                }
+                events.removeAll(eventsList);
+                postEvent(getStreamConfig(stream), eventsList);
```
cooltey, line 262: Can we have a null check at this line?

cooltey (changes requested):
Very minor comment inline.

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
```diff
@@ -169,9 +171,14 @@ static void addEventMetadata(Event event) {
...
-         * When QUEUE.size() exceeds this value TIMER becomes non-interruptable.
+         * While coming out of offline, When QUEUE.size() exceeds this value TIMER becomes non-interruptable.
+         */
+        private static final int MAX_OFFLINE_QUEUE_SIZE = 128;
+
+        /*
+         * While online, When QUEUE.size() exceeds this value TIMER becomes non-interruptable.
          */
-        private static final int MAX_QUEUE_SIZE = 128;
+        private static final int MAX_ONLINE_QUEUE_SIZE = 10;
```
mdholloway, line 181: One of the design goals of the Event Platform Client libraries is to make network requests for event logging less frequent.  Limiting the queue size to 10 items when online seems at cross-purposes with that goal.  Does batched event submission resolve the concern behind this?

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
```diff
@@ -239,14 +244,26 @@ static void schedule(Event event) {
...
+        static void send(List<Event> events) {
+            while (!events.isEmpty()) {
+                List<Event> eventsList = new ArrayList<>();
+                String stream = events.get(0).getStream();
+                for (Event event : events) {
+                    if (event.getStream().equals(stream)) {
+                        eventsList.add(event);
+                    }
+                }
+                events.removeAll(eventsList);
```
mdholloway, line 261: Is it safe to remove elements from `events` while iterating over it with `for-each`?  I would recommend avoiding mutating `events` and instead do something like this (from my [earlier implementation](https://github.com/wikimedia/apps-android-wikipedia/commit/1b9d25f4c1ce3c9a5c8f7da8f5c52ec11b2949cb#diff-806f6b0dfd5d73547d9d1bb0fd0ee4f07cd0e1a5c0f70130e0f1dca5c334df24R238-R261)):

            Map<String, ArrayList<Event>> eventsByStream = new HashMap<>();
            for (Event event : events) {
                String stream = event.getStream();
                if (!eventsByStream.containsKey(stream) || eventsByStream.get(stream) == null) {
                    eventsByStream.put(stream, new ArrayList<>());
                }
                eventsByStream.get(stream).add(event);
            }
            for (String stream : eventsByStream.keySet()) {
                postEvents(getStreamConfig(stream), eventsByStream.get(stream));
            }

Alternatively, if we're not overly concerned with future-proofing at this point, since all events will initially be going to the same event intake service (`eventgate-analytics-external`), we could simplify things by not having `postEvents` check the stream config for the destination URL and instead simply posting the entire queue contents unconditionally to its endpoint URL without any of this preliminary filtering and grouping.

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClientIntegration.java
```diff
@@ -75,27 +76,27 @@ static EventService getEventService(@NonNull StreamConfig streamConfig) {
...
      * Failures are logged remotely (as well as in Logcat). The only exception is for unexpected
      * responses, for which we crash on pre-production builds.
      *
      * @param streamConfig stream config
-     * @param event Event to be posted. Gson will take care of serializing to JSON.
+     * @param events Events to be posted. Gson will take care of serializing to JSON.
      */
-    static void postEvent(@NonNull StreamConfig streamConfig, @NonNull Event event) {
-        DISPOSABLES.add(getEventService(streamConfig).postEvent(event)
+    static void postEvent(@NonNull StreamConfig streamConfig, @NonNull List<Event> events) {
```
mdholloway, line 98: Nit: update to `postEvents`

mdholloway (changes requested):
Sorry, I must have missed your ping earlier.  I just saw it earlier today when looking this over with @linehan.

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
sharvaniharan: You are right @mdholloway , With batching, this can just go back to 128. I was curious... why did we choose 128 as the limit. The upper limit to the POST body itself is 10MB, so in theory this could be much higher?

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
sharvaniharan: I was trying to remove the elements after the for-each was exited @mdholloway . However, segregating them in a map is a better idea, so have changed them. I felt we would miss it in the future if we completely remove this filtering, so implemented per your earlier implementation instead.

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
sharvaniharan: Reverted to 128 for now

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
```diff
@@ -239,14 +237,27 @@ static void schedule(Event event) {
...
-         *
-         * @param event event
+         * If sending is enabled, attempt to send the provided events.
+         * Also batch the events ordered by their streams, as the QUEUE
+         * can contain events of different streams
+         * @param events list of events
          */
-        static void send(Event event) {
-            postEvent(getStreamConfig(event.getStream()), event);
+        static void send(List<Event> events) {
```
dbrant, line 245: A function called `send()` should do one thing: send.  If there are conditions that might cause it not to send, it should be called `maybeSend()`, or `sendIfEnabled()`.

For maximum clarity, the check for `ENABLED` should probably be the responsibility of the caller of `send()`, instead of within it.

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
```diff
@@ -171,7 +171,7 @@ static void addEventMetadata(Event event) {
         /*
          * When QUEUE.size() exceeds this value TIMER becomes non-interruptable.
          */
-        private static final int MAX_QUEUE_SIZE = 128;
+        private static final int MAX_OFFLINE_QUEUE_SIZE = 128;
```
mdholloway, line 174: This no longer has anything to do with online/offline state, does it?  Or am I missing something?

mdholloway (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
sharvaniharan: Correct! name reverted

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
sharvaniharan: Correct. The enable check might be redundant here, since the caller is checking. So removing the check and keeping the name.

sharvaniharan:
@dbrant @mdholloway i will merge this so we can have all discussions in one place on the MEP pr.

PR #1818, opened 2020-11-21
Create analytics service through ServiceFactory.

sharvaniharan:
To keep things consistent with current practices

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/ServiceFactory.java
+++ b/app/src/main/java/org/wikipedia/dataclient/ServiceFactory.java
@@ -7,6 +7,9 @@ import androidx.annotation.Nullable;
 import androidx.collection.LruCache;
 
 import org.wikipedia.WikipediaApp;
+import org.wikipedia.analytics.eventplatform.DestinationEventService;
+import org.wikipedia.analytics.eventplatform.EventService;
+import org.wikipedia.analytics.eventplatform.StreamConfig;
 import org.wikipedia.dataclient.okhttp.OkHttpConnectionFactory;
 import org.wikipedia.json.GsonUtil;
 import org.wikipedia.settings.Prefs;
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/dataclient/ServiceFactory.java
```diff
@@ -59,6 +65,29 @@ public static CoreRestService getCoreRest(@NonNull WikiSite wiki) {
...
+        if (ANALYTICS_REST_SERVICE_CACHE.get(destinationEventService.getId()) != null) {
+            s = ANALYTICS_REST_SERVICE_CACHE.get(destinationEventService.getId());
+            if (s != null) {
+                return s;
+            }
+        }
+
+        String intakeBaseUriOverride = getEventPlatformIntakeUriOverride();
+
+        Retrofit r = createRetrofit(WikiSite.forLanguageCode(WikipediaApp.getInstance().getAppOrSystemLanguageCode()),
```
cooltey, line 84: Can we have a `WikiSite wiki` parameter and use that `wiki` to create a Retrofit instance?

e.g. `getAnalyticsRest(@NonNull WikiSite wiki, @NonNull StreamConfig streamConfig)`

cooltey, line 84: Can we have a `WikiSite wiki` parameter and use that wiki to create a Retrofit instance?

e.g. `getAnalyticsRest(@NonNull WikiSite wiki, @NonNull StreamConfig streamConfig)`

sharvaniharan: @cooltey I removed it altogether since it was only needed to add the Accept-Language header which is not needed here.

app/src/main/java/org/wikipedia/dataclient/ServiceFactory.java
```diff
@@ -59,6 +66,29 @@ public static CoreRestService getCoreRest(@NonNull WikiSite wiki) {
...
+        if (ANALYTICS_REST_SERVICE_CACHE.get(destinationEventService.getId()) != null) {
+            s = ANALYTICS_REST_SERVICE_CACHE.get(destinationEventService.getId());
+            if (s != null) {
+                return s;
+            }
+        }
+
+        String intakeBaseUriOverride = getEventPlatformIntakeUriOverride();
+
+        Retrofit r = createRetrofit(null, intakeBaseUriOverride != null ? intakeBaseUriOverride
```
cooltey, line 85: Maybe it is cleaner to use `StringUtils.defaultString()` for the URL?

cooltey (approved):

sharvaniharan:
Merging this so everything is in one place..

PR #1830, opened 2020-11-30
Remove maxLines setting from the description view in POTD

cooltey:
https://phabricator.wikimedia.org/T268956


sharvaniharan:
@cooltey This only solves one of the two issues. We will now be able to see the image desc fully, but we still have to resolve the issue of the descriptions not matching on feed and on Image activvity?

cooltey:
> 
> 
> @cooltey This only solves one of the two issues. We will now be able to see the image desc fully, but we still have to resolve the issue of the descriptions not matching on feed and on Image activvity?

Please see my comment here: https://phabricator.wikimedia.org/T268956#6657399

PR #1836, opened 2020-12-02
Reinstate Main Page feed card.

dbrant:
https://phabricator.wikimedia.org/T269244


sharvaniharan:
Lgtm...checked with 3 langs.. @cooltey please feel free to test in other langs too and merge

app/src/main/java/org/wikipedia/feed/mainpage/MainPageCardView.java
```diff
@@ -0,0 +1,58 @@
...
+        super.setCard(card);
+        setLayoutDirectionByWikiSite(card.wikiSite(), containerView);
+
+        headerView.setTitle(L10nUtil.getStringForArticleLanguage(getCard().wikiSite().languageCode(), R.string.view_main_page_card_title))
+                .setLangCode(getCard().wikiSite().languageCode())
+                .setCard(getCard())
+                .setCallback(getCallback());
+
+        footerView.setCallback(this::goToMainPage);
+        footerView.setFooterActionText(L10nUtil.getStringForArticleLanguage(getCard().wikiSite().languageCode(), R.string.view_main_page_card_action));
```
cooltey, line 43: In other cards, we don't use `L10nUtil.getStringForArticleLanguage()` on the footer action text. Not sure if we are going to do this here.

dbrant: Agreed, let's keep it consistent with other cards.

PR #1839, opened 2020-12-02
Make sure using the correct Main page name for certain language variants

cooltey:
**Steps to reproduce:**

1. Add a new `zh-tw` to Wikipedia languages
2. Go to Explore feed, and scroll to the `zh-tw` "Featured article" card
3. Click on the `Today on Wikipedia`

It should open the main page of `zh-tw` instead of the definition page of the main page.


app/src/main/java/org/wikipedia/staticdata/MainPageNameData.java
```diff
@@ -5,6 +5,8 @@
```
dbrant, line 5: This is an auto-generated file, and will be overwritten the next time we update our languages.  You probably intended to edit the `scripts/templates/basichash.java.jinja` file.

dbrant (changes requested):

app/src/main/java/org/wikipedia/staticdata/MainPageNameData.java
cooltey: Thanks! You are correct. Done!

PR #1774, opened 2020-11-07
MEP

sharvaniharan:
**Phab:**  https://phabricator.wikimedia.org/T228179

PR #1844, opened 2020-12-04
EventPlatformClient class structure changes

sharvaniharan:
per CR on https://github.com/wikimedia/apps-android-wikipedia/pull/1774#discussion_r536122104

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
@@ -226,6 +226,7 @@ public final class EventPlatformClient {
                      * the countdown.
                      */
                     TIMER.cancel();
+                    TIMER.purge();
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
```diff
@@ -438,9 +433,11 @@ private static void refreshStreamConfigs() {
     /*
      * The constructor is private, so instantiation from other classes is impossible.
      */
-    private EventPlatformClient() {
+    public static void setUpStreamConfigs() {
```
cooltey, line 436: Would it be possible to just name it as `setup()` since it is the way to setup the MEP client?

sharvaniharan: @cooltey stream configs is just a part of the client so I feel this is more readable.

PR #1846, opened 2020-12-05
Build Watchlist add/remove workflow from article

cooltey:
https://phabricator.wikimedia.org/T269446


app/src/main/java/org/wikipedia/dataclient/watch/Watch.kt
```diff
@@ -0,0 +1,19 @@
+package org.wikipedia.dataclient.watch
+
+import org.wikipedia.dataclient.mwapi.MwResponse
+
+class Watch : MwResponse() {
+    private val title: String? = null
+    private val ns = 0
+    private val pageid = 0
+    private val expiry: String? = null
+    private val watched: String? = null
```
dbrant, line 10: These parameters should be `boolean`, not string.  When using the Sandbox to test API calls, don't forget to go to `format=json` and set `formatversion` to 2.  You will then see the proper API responses with booleans.

dbrant (changes requested):

app/src/main/java/org/wikipedia/dataclient/watch/Watch.kt
cooltey: You are correct. Done.

PR #1848, opened 2020-12-07
Follow-up to Talk pages.

dbrant:
This does all of the following:

- Make the namespace of talk pages no longer all-caps.
- No longer show the "lead" section of talk pages, which is often irrelevant or extraneous information.
- Add an option to view the talk page in an external browser.


app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
```diff
@@ -128,7 +131,12 @@ class TalkTopicsActivity : BaseActivity() {
                 }
                 .subscribe({ response ->
                     topics.clear()
-                    topics.addAll(response.topics!!)
+                    for (topic in response.topics!!) {
+                        if (topic.id == 0 && topic.html!!.trim().isEmpty()) {
```
cooltey, line 135: Since we are going to hide the first section of the talk page, should we show the `talkEmptyContainer` view if there are no other topics?

dbrant: Good call. Done

PR #1850, opened 2020-12-08
Watchlist Diff View

sharvaniharan:


dbrant:
A quick note on naming:
What you're building here is a generic diff-viewing activity. It doesn't actually have anything to do with watchlists. It will take the title of an article, and a revision ID. This should be all that's necessary for viewing a diff.
When complete, the diff viewing activity can be used from many other places besides watchlists.

dbrant (changes requested):
Seems to work pretty well so far; A few initial comments inline.

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,415 @@
...
+            diffCharacterCountView.setTextColor(ContextCompat.getColor(requireContext(), R.color.red50))
+            diffCharacterCountView.text = String.format("%+d", diffSize)
+        }
+    }
+
+    private fun fetchEditDetails() {
+        disposables.add(Observable.zip(ServiceFactory.get(WikiSite.forLanguageCode(languageCode)).getRevisionDetails(articleTitle, revisionId),
+                ServiceFactory.get(WikiSite.forLanguageCode(languageCode)).getWatchedInfo(articleTitle),
+                { r, w ->
+                    isWatched = w.query()!!.firstPage()!!.isWatched
```
cooltey, line 143: Would it be possible to add checks to check the `query()` and `firstPage()` to prevent untraceable errors instead of using many `!!`?

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,415 @@
...
+                    }
+                    showWatchlistSnackbar(expiry, firstWatch)
+                }) { t: Throwable? -> L.d(t) })
+    }
+
+    private fun updateWatchlistButtonUI() {
+        setButtonTextAndIconColor(watchButton, ResourceUtil.getThemedColor(requireContext(),
+                if (isWatched) R.attr.color_group_62 else R.attr.colorAccent))
+        watchButton.text = getString(if (isWatched) R.string.watchlist_details_watching_label
+        else R.string.watchlist_details_watch_label)
```
cooltey, line 215: Could we have these code reformat?

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,415 @@
...
+    }
+
+    private fun showWatchlistSnackbar(@Nullable expiry: WatchlistExpiry?, watch: Watch) {
+        isWatched = watch.watched
+        if (watch.unwatched) {
+            FeedbackUtil.showMessage(this, getString(R.string.watchlist_page_removed_from_watchlist_snackbar, articleTitle))
+            watchlistExpirySession = null
+
```
cooltey, line 223: Empty line.

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,415 @@
...
+                            getString(expiry.stringId)),
+                    FeedbackUtil.LENGTH_DEFAULT)
+            if (!watchlistExpiryChanged) {
+                snackbar.setAction(R.string.watchlist_page_add_to_watchlist_snackbar_action) {
+                    watchlistExpiryChanged = true
+                    bottomSheetPresenter.show(childFragmentManager, WatchlistExpiryDialog.newInstance(expiry))
+                }
+            }
+            watchlistExpirySession = null
+            snackbar.show()
```
cooltey, line 237: Looks like the snackbar does not show properly. Also, the "watching" button should be disabled when posting the request and enable until the process and the status is received.

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,415 @@
...
+                    .setTextColor(ResourceUtil.getThemedColor(requireContext(), R.attr.secondary_text_color))
+        }
+        dialog.show()
+    }
+
+    private fun sendThanks() {
+        ServiceFactory.get(WikiSite.forLanguageCode(languageCode)).csrfToken
+                .subscribeOn(Schedulers.io())
+                .flatMap { response: MwQueryResponse ->
+                    val csrfToken = response.query()!!.csrfToken()
```
cooltey, line 262: Is it necessary to have a `val` for the token?

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,415 @@
...
+                })
+                { Consumer { t: Throwable? -> L.e(t) } }
+    }
+
+    private fun fetchDiffText() {
+        disposables.add(ServiceFactory.getCoreRest(WikiSite.forLanguageCode(languageCode)).getDiff(olderRevisionId, revisionId)
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe({ response ->
+                    displayDiff(response.diffs)
```
cooltey, line 280: Can be simplified to `displayDiff(it.diffs)`

app/src/main/java/org/wikipedia/util/UriUtil.java
```diff
@@ -120,6 +120,16 @@ public static String getUrlWithProvenance(Context context, PageTitle title,
         return title.getUri() + "?wprov=" + context.getString(provId);
     }
 
+    public static String getUrlForDiff(PageTitle title, long newId, long oldId) {
```
cooltey, line 123: Can we add annotations to the parameters?

app/src/main/java/org/wikipedia/util/ShareUtil.java
```diff
@@ -59,6 +59,10 @@ public static void shareText(final Context context, final PageTitle title) {
         shareText(context, StringUtil.fromHtml(title.getDisplayText()).toString(), UriUtil.getUrlWithProvenance(context, title, R.string.prov_share_link));
     }
 
+    public static void shareText(final Context context, final PageTitle title, long newId, long oldId) {
```
cooltey, line 62: Can we add annotations to the parameters?

app/src/main/java/org/wikipedia/util/FeedbackUtil.java
```diff
@@ -107,6 +107,14 @@ public static void showAndroidAppEditingFAQ(Context context) {
         showAndroidAppEditingFAQ(context, R.string.android_app_edit_help_url);
     }
 
+    public static void showUserContributionsPage(Context context, String username) {
```
cooltey, line 110: Can we add annotations to the parameters?

app/src/main/java/org/wikipedia/util/FeedbackUtil.java
```diff
@@ -107,6 +107,14 @@ public static void showAndroidAppEditingFAQ(Context context) {
...
+        visitInExternalBrowser(context, Uri.parse(context.getString(R.string.user_contributions_url, username)));
+    }
+
+    public static void showUserProfilePage(Context context, String username) {
```
cooltey, line 114: Can we add annotations to the parameters?

cooltey (changes requested):

sharvaniharan:
Done @dbrant @cooltey

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,419 @@
...
+
+            val spannableString = SpannableStringBuilder(diffText.text).append(if (diff.text.isNotEmpty()) diff.text else "\n")
+            when (diff.type) {
+                DIFF_TYPE_LINE_WITH_SAME_CONTENT -> {
+                }
+                DIFF_TYPE_LINE_ADDED -> {
+                    spannableString.setSpan(BackgroundColorSpan(ResourceUtil.getThemedColor(requireContext(),
+                            R.attr.edit_green_highlight)), prefixLength, prefixLength + diff.text.length, 0)
+                    spannableString.setSpan(boldStyle, prefixLength, prefixLength + diff.text.length, 0)
+                    spannableString.setSpan(foregroundAddedColor, prefixLength, prefixLength + diff.text.length, 0)
```
cooltey, line 296: Just curious that is it possible to make these actions into a method, which may be cleaner and easier to understand?

app/src/main/res/layout/fragment_article_edit_details.xml
```diff
@@ -0,0 +1,179 @@
+<?xml version="1.0" encoding="utf-8"?>
+<ScrollView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:layout_marginStart="16dp"
+    android:layout_marginEnd="16dp"
```
cooltey, line 8: These should be removed and add `android:layout_marginHorizontal="16dp"` to the child `LinearLayout` at line `11`. The current layout has extra horizontal paddings and the scrollbar is not sticking to the right edge.

cooltey (changes requested):

sharvaniharan:
Done @cooltey

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,401 @@
...
+            } else {
+                firstPage.revisions()[1].revId
+            }
+            olderRevisionId = currentRevision!!.parentRevId
+        })
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe({
+                    updateUI()
+                }) { t: Throwable? -> L.e(t) })
```
cooltey, line 153: Would it be possible to show the `WikiErrorView` in the layout in case something goes wrong?

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,401 @@
...
+        editComment.text = currentRevision!!.comment
+
+        newerIdButton.isClickable = newerRevisionId != -1L
+        olderIdButton.isClickable = olderRevisionId != 0L
+        ImageViewCompat.setImageTintList(newerIdButton, ColorStateList.valueOf(ContextCompat.getColor(requireContext(),
+                ResourceUtil.getThemedAttributeId(requireContext(), if (newerRevisionId == -1L)
+                    R.attr.material_theme_de_emphasised_color else R.attr.primary_text_color))))
+        ImageViewCompat.setImageTintList(olderIdButton, ColorStateList.valueOf(ContextCompat.getColor(requireContext(),
+                ResourceUtil.getThemedAttributeId(requireContext(), if (olderRevisionId == 0L)
+                    R.attr.material_theme_de_emphasised_color else R.attr.primary_text_color))))
```
cooltey, line 170: Maybe this can create a method to apply the color for buttons?

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,401 @@
...
+                DIFF_TYPE_LINE_WITH_SAME_CONTENT -> {
+                }
+                DIFF_TYPE_LINE_ADDED -> {
+                    updateDiffTextDecor(spannableString, true, prefixLength, prefixLength + diff.text.length)
+                }
+                DIFF_TYPE_LINE_REMOVED -> {
+                    updateDiffTextDecor(spannableString, false, prefixLength, prefixLength + diff.text.length)
+                }
+                DIFF_TYPE_LINE_WITH_DIFF -> {
+                    for (hightlightRange in diff.highlightRanges) {
```
cooltey, line 295: typo

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,401 @@
...
+                }
+                DIFF_TYPE_PARAGRAPH_MOVED_TO -> {
+                    updateDiffTextDecor(spannableString, true, prefixLength, prefixLength + diff.text.length)
+                }
+            }
+            diffText.text = spannableString.append("\n")
+        }
+    }
+
+    private fun updateDiffTextDecor(spannableText: SpannableStringBuilder, isAddition: Boolean, start: Int, end: Int) {
```
cooltey, line 319: This is much better now 👍 

I noticed that the texts are not aligning to the center vertically (in the diff background color), can we optimize it?

cooltey (changes requested):

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
sharvaniharan: OK i finally realized this is because of the lineSpacingExtra attribute... I will try to see if there is a work around.

cooltey: Hi @sharvaniharan, please have a follow-up PR if you can resolve the issue. I'll merge this PR to make things move.

app/src/main/res/values/strings.xml
```diff
@@ -1021,7 +1021,18 @@
     <string name="talk_message_hint">Compose message</string>
     <string name="talk_message_empty">The message cannot be empty.</string>
     <string name="talk_subject_empty">The subject cannot be empty.</string>
+    <string name="menu_option_share_edit">Share edit</string>
+    <string name="menu_option_user_profile">%s\'s profile</string>
+    <string name="menu_option_user_contributions">%s\'s contributions</string>
+    <string name="watchlist_details_watching_label">Watching</string>
+    <string name="watchlist_details_watch_label">Watch</string>
+    <string name="user_contributions_url">https://en.wikipedia.org/wiki/Special:Contributions/%s</string>
```
dbrant, line 1029: This should not be a translatable string. Also, it explicitly goes to English wikipedia (?) instead of the user's language wiki.

app/src/main/res/values/strings.xml
```diff
@@ -1021,7 +1021,18 @@
...
+    <string name="user_profile_url">https://en.wikipedia.org/wiki/User:%s</string>
```
dbrant, line 1030: This should not be a translatable string. Also, it explicitly goes to English wikipedia (?) instead of the user's language wiki.

app/src/main/res/values/strings.xml
```diff
@@ -1021,7 +1021,18 @@
...
+    <string name="thank_label">Thank</string>
     <string name="go_to_user_page">Go to user page</string>
+    <string name="thank_dialog_positive_button_text">SEND THANKS</string>
```
dbrant, line 1033: Strings should never be all-caps. You can apply all-caps styling to buttons using XML.

app/src/main/res/values/styles_light.xml
```diff
@@ -78,7 +78,12 @@
         <item name="color_group_59">@color/base30</item>
         <item name="color_group_60">@color/base80</item>
         <item name="color_group_61">@color/black26</item>
+        <item name="color_group_62">@color/osage</item>
         <item name="color_group_63">@color/black54</item>
+        <item name="color_group_64">@color/base10</item>
+        <item name="color_group_65">@color/base10</item>
+        <item name="edit_green_highlight">@color/green90 </item>
```
dbrant, line 85: extra space inside tag.

app/src/main/res/layout/view_thank_dialog.xml
```diff
@@ -0,0 +1,36 @@
...
+        android:layout_height="100dp"
+        app:srcCompat="@drawable/ic_thank_illustration" />
+
+    <TextView
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="16dp"
+        android:letterSpacing="0.01"
+        android:lineSpacingExtra="8sp"
+        android:text="Publicly send \'Thanks\'"
```
dbrant, line 21: Un-resourced string.

app/src/main/res/layout/view_thank_dialog.xml
```diff
@@ -0,0 +1,36 @@
...
+        android:textSize="20sp"
+        android:textStyle="bold" />
+
+    <TextView
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="8dp"
+        android:letterSpacing="0.01"
+        android:lineSpacingExtra="8sp"
+        android:text="\'Thanks\' are an easy way to show appreciation for an editor’s work on Wikipedia. &apos;Thanks&apos; cannot be undone and are publicly viewable"
```
dbrant, line 32: Un-resourced string.

app/src/main/java/org/wikipedia/dataclient/mwapi/MwQueryPage.java
```diff
@@ -177,11 +177,25 @@ public long getLastRevId() {
...
         }
 
-        @NonNull public String content() {
+        public long getParentRevId() {
+            return parentid;
+        }
+
+        @Nullable
+        public String getUser() {
+            return user;
```
dbrant, line 194: Can we follow the pattern of returning a `NonNull` string here? (with `StringUtils.defaultString()`)

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,413 @@
...
+                }
+                DIFF_TYPE_LINE_ADDED -> {
+                    updateDiffTextDecor(spannableString, true, prefixLength, prefixLength + diff.text.length)
+                }
+                DIFF_TYPE_LINE_REMOVED -> {
+                    updateDiffTextDecor(spannableString, false, prefixLength, prefixLength + diff.text.length)
+                }
+                DIFF_TYPE_LINE_WITH_DIFF -> {
+                    for (highlightRange in diff.highlightRanges) {
+                        val highlightRangeStart = if (languageCode == "en") highlightRange.start
```
dbrant, line 308: Why are there special cases for English here?

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,413 @@
...
+        val boldStyle = StyleSpan(Typeface.BOLD)
+        val foregroundAddedColor = ForegroundColorSpan(ResourceUtil.getThemedColor(requireContext(), R.attr.color_group_64))
+        val foregroundRemovedColor = ForegroundColorSpan(ResourceUtil.getThemedColor(requireContext(), R.attr.color_group_65))
+        spannableText.setSpan(BackgroundColorSpan(ResourceUtil.getThemedColor(requireContext(),
+                if (isAddition) R.attr.edit_green_highlight else R.attr.edit_red_highlight)), start, end, 0)
+        spannableText.setSpan(boldStyle, start, end, 0)
+        spannableText.setSpan(if (isAddition) foregroundAddedColor else foregroundRemovedColor, start, end, 0)
+    }
+
+    private fun getByteInCharacters(diffText: String?, byteLength: Int, start: Int): Int {
```
dbrant, line 341: Does `diffText` need to be nullable?

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -0,0 +1,413 @@
...
+        var charCount = 0
+        var bytes = byteLength
+
+        for (pos in start until diffText!!.length - 1) {
+            val idBytes = diffText[pos].toString().toByteArray(CHARSET).size
```
dbrant, line 346: This is using something called `CHARSET`, which comes from `com.bumptech.glide.load.Key`, which seems like a weird dependency.  Can we just use `StandardCharsets.UTF_8`?

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
sharvaniharan: English characters are 1 byte each.. So can we skip the processing for it?

dbrant: So... English articles can't contain Unicode characters?
https://en.wikipedia.org/wiki/Hindi

PR #1856, opened 2020-12-11
Add autofill hints to form fields.

dbrant:
https://phabricator.wikimedia.org/T269189


cooltey:
Autofill from 1Password app works 💯

PR #1862, opened 2020-12-14
Use getWatchedInfo() to get the proper watched state of an article

cooltey:
This PR also fixes an issue that `watchlistExpirySession` should be cleared when entering another article on the page.

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
+++ b/app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
@@ -4,6 +4,7 @@ import android.widget.Toast;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.core.util.Pair;
 
 import org.wikipedia.R;
 import org.wikipedia.WikipediaApp;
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
```diff
@@ -181,27 +184,38 @@ private void pageLoadFromNetwork(final ErrorCallback errorCallback) {
 
         app.getSessionFunnel().leadSectionFetchStart();
 
-        disposables.add(ServiceFactory.getRest(model.getTitle().getWikiSite())
+        disposables.add(Observable.zip(ServiceFactory.getRest(model.getTitle().getWikiSite())
                 .getSummaryResponse(model.getTitle().getPrefixedText(), null, model.getCacheControl().toString(),
                         model.shouldSaveOffline() ? OfflineCacheInterceptor.SAVE_HEADER_SAVE : null,
-                        model.getTitle().getWikiSite().languageCode(), UriUtil.encodeURL(model.getTitle().getPrefixedText()))
+                        model.getTitle().getWikiSite().languageCode(), UriUtil.encodeURL(model.getTitle().getPrefixedText())), ServiceFactory.get(model.getTitle().getWikiSite())
```
dbrant, line 190: Can this somehow be done only if the user is logged in?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
cooltey: Done!

PR #1866, opened 2020-12-15
End A/B test of explore vs home.

dbrant:
https://phabricator.wikimedia.org/T260332


app/src/main/res/xml/developer_preferences.xml
```diff
@@ -385,8 +385,8 @@
 
         <org.wikipedia.settings.IntPreference
             style="@style/IntPreference"
-            android:key="ab_test_exploreVsHome"
-            android:title="Whether the user sees a Home icon/label, or Explore icon/label." />
+            android:key="ab_test_suggestedEditsSnackbar"
+            android:title="Whether the user sees a Snackbar action to 'Get Started' with suggested edits, after an unrelated edit." />
```
cooltey, line 389: Any reason why we keep this? It was removed in #1865

cooltey (changes requested):

app/src/main/res/xml/developer_preferences.xml
dbrant: whoops, not sure why this made its way back.

PR #1869, opened 2020-12-15
Fix the race condition on isLangaugeWrong variable in DescriptionEditView

cooltey:
Steps to reproduce:

1. Go to [日本電視台 ](https://zh.wikipedia.org/wiki/日本電視台)in `zh` wiki (you can find it at the Featured article in `zh` wiki).
2. Edit article description
3. Remove all description and enter "test".

Expected result:
Show submission warning

Actual result:
Nothing happened


dbrant (changes requested):
I'm not sure I see any race condition here. When I type "test", the language gets detected as "und" (undefined). This is why we don't see a warning for this case.

cooltey:
> 
> 
> I'm not sure I see any race condition here. When I type "test", the language gets detected as "und" (undefined). This is why we don't see a warning for this case.

Right, I didn't check the original response from `MlKitLanguageDetector` and I think your point is correct.

PR #1870, opened 2020-12-16
Make sure the Explore feed cards' header and footer have the same layout direction settings

cooltey:
https://phabricator.wikimedia.org/T269953

If we have two separate views (`TextView` + `ImageView`), it would still have the same issue when the `TextView` has multiple lines. 

I believe this PR is the proper solution, which uses an `ImageSpan` and `SpannableStringBuilder`.

-- Update --
I have udpated the PR title since it does more than before. This PR basically updates the footer and header to have the same RTL settings.


dbrant (changes requested):
You've thought about if the text has RTL characters, but not if the system UI itself is RTL:

image

cooltey:
> 
> 
> You've thought about if the text has RTL characters, but not if the system UI itself is RTL:
> 

Nice catch. I have a solution for that: detect the first character of the string for the icon's direction. I also change the layout width to match_parent for UI consistency.

dbrant (changes requested):
When the system UI is RTL, the arrow should point to the left, not right.
Wouldn't it be simpler to just do it the way we discussed, i.e. leave the arrow as an ImageView that sits next to the text?

image

cooltey:
> 
> 
> When the system UI is RTL, the arrow should point to the left, not right.
> Wouldn't it be simpler to just do it the way we discussed, i.e. leave the arrow as an ImageView that sits next to the text?
> 

I tried to use the `TextView` + `ImageView`,  maybe I was wrong but turned out it cannot fix the issue. 

The `TextView` will become a full-width layout and pushed the `ImageView` to the next line, which is not expected; or, if I keep the imageView to the right of the view, it will have the same result as the current production.

Screenshot_20201215-154003_Wikipedia Dev


I'll try to add a parent `ViewGroup` to make the layout handle the RTL properly.

cooltey:
@dbrant Done.

cooltey:
@dbrant 
Just noticed the `MainPageCardView` applies layout direction to its parent container, which does not consistent with other cards. 

I've removed the setting and here are the screenshots, and the idea is that the "header" will show in card language but the "footer" will show in device language (and it should align to the right when it's RTL).

Screenshot_1608330205
Screenshot_1608330176
Screenshot_1608330100

cooltey:
Things are getting confused, I have left a comment for Robin to finalize the layout direction decision:
https://phabricator.wikimedia.org/T269953#6702758

cooltey:
I believe all the layout direction in footer and header and also the texts have been handled properly. Please leave comment if see any issues. @dbrant @sharvaniharan

PR #1873, opened 2020-12-17
Fix: optimize loading behavior of user contributions screen

cooltey:
https://phabricator.wikimedia.org/T269863


sharvaniharan:
Nice!

PR #1876, opened 2020-12-17
Hide toolbar elevation when webview reaching to top in PageActivity

cooltey:
https://phabricator.wikimedia.org/T270168


app/src/main/java/org/wikipedia/page/ViewHideHandler.java
```diff
@@ -81,6 +83,7 @@ public void onScrollChanged(int oldScrollY, int scrollY, boolean isHumanScroll)
         if (anchoredView != null) {
             anchoredView.setTranslationY(animMargin);
         }
+        hideableView.setElevation(scrollY > 0 ? DimenUtil.dpToPx(DimenUtil.getDimension(R.dimen.toolbar_default_elevation)) : 0);
```
dbrant, line 86: This is a lot of operations to perform upon every scroll. Can we set elevation only when scrollY _reaches_ 0, and when it _leaves_ 0?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/ViewHideHandler.java
cooltey: Done.

PR #1881, opened 2020-12-26
Use ConcurrentHashMap.

Isira-Seneviratne:
Enabling API desugaring includes an implementation of `ConcurrentHashMap` that isn't affected by the bug on Android Lollipop in the app.


dbrant:
Sorry, the single usage of `ConcurrentHashMap` does not justify including the new desugaring dependency.

PR #1888, opened 2021-01-05
Wire up analytics for Watchlists.

cooltey:
https://phabricator.wikimedia.org/T267679


app/src/main/java/org/wikipedia/analytics/WatchlistFunnel.java
```diff
@@ -0,0 +1,64 @@
+package org.wikipedia.analytics;
+
+import androidx.annotation.NonNull;
+
+import org.wikipedia.WikipediaApp;
+import org.wikipedia.util.StringUtil;
+
+import java.util.List;
+
+public class WatchlistFunnel extends TimedFunnel {
```
dbrant, line 10: This is declared as a `TimedFunnel`, which means it will automatically send a `time_spent` field in addition to the other fields.  But the schema doesn't have a `time_spent` field.  If we want to log time spent, then it should be added to the schema. And if we don't need it, then this should not be a TimedFunnel.

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/WatchlistFunnel.java
cooltey: Thanks, didn't notice that. Fixed.

dbrant:
@cooltey  this looks ready, if you resolve the conflict(s)

cooltey:
> 
> 
> @cooltey this looks ready, if you resolve the conflict(s)

Done!

PR #1895, opened 2021-01-07
Stop sending language variant header for the page/related endpoint

cooltey:
The response time in the `page/related/` endpoint will become really slow in `zhwiki` when having the `Accept-Language` header. To improve that temprorarily, we can skip adding the language variant header and loading page summary endpoint to get correct page title.


dbrant (changes requested):
Actually... I have an idea to do this much more efficiently:
Just modify the `LanguageVariantHeaderInterceptor`, and only add the `Accept-Language` header if the request URL contains `/page/related`.
Would that work?

cooltey:
> 
> 
> Actually... I have an idea to do this much more efficiently:
> Just modify the `LanguageVariantHeaderInterceptor`, and only add the `Accept-Language` header if the request URL contains `/page/related`.
> Would that work?

Yes, it works! But we still need to make separate `/page/summary` calls to correct the display title and API title for the `zhwiki`.

PR #1907, opened 2021-01-13
Show animation on TabCountsView when opening a tab in backgrond on Explore feed

cooltey:
In the PageActivity, the `TabCountsView` shows an animation when opening a new tab in the background, and maybe we need this in the MainActivity, too.


sharvaniharan:
Looks fine but I am not able to trigger the animation... Looks like from code that everytime the menu item is updated we should see it... but I do not see it with any update

sharvaniharan:
@cooltey I was able to test it with long press from explore feed as we discussed offline.. Lgtm... merging now

PR #1908, opened 2021-01-13
Hygiene: remove redundant checks in BecauseYouReadClient

cooltey:


app/src/main/java/org/wikipedia/feed/becauseyouread/BecauseYouReadClient.java
```diff
@@ -61,7 +61,7 @@ private void getCardForHistoryEntry(@NonNull final HistoryEntry entry,
                 .toList()
                 .subscribe(list -> {
                     PageSummary headerPage = list.remove(0);
-                    FeedCoordinator.postCardsToCallback(cb, (list == null || list.size() == 0)
+                    FeedCoordinator.postCardsToCallback(cb, list.isEmpty()
```
sharvaniharan, line 64: we should check for null too.. so isNullOrEmpty()

dbrant: @sharvaniharan  The `list` will never be null, because this is RxJava.  It is not allowed to pass null values through the Rx pipeline.

PR #1915, opened 2021-01-14
Tweak: enlarge clickable area for overflow menu and dismiss watchlist tooltip when clicking on it

cooltey:
https://phabricator.wikimedia.org/T269448#6747169


app/src/main/java/org/wikipedia/page/PageActivity.java
```diff
@@ -238,6 +241,9 @@ public void onShowTabsButtonClicked() {
 
     @OnClick(R.id.page_toolbar_button_show_overflow_menu)
     public void onShowOverflowMenuButtonClicked() {
+        if (watchlistTooltip != null) {
+            watchlistTooltip.dismiss();
```
dbrant, line 245: should we set it to `null` here also?

app/src/main/java/org/wikipedia/page/PageActivity.java
```diff
@@ -238,6 +241,9 @@ public void onShowTabsButtonClicked() {
 
     @OnClick(R.id.page_toolbar_button_show_overflow_menu)
     public void onShowOverflowMenuButtonClicked() {
+        if (watchlistTooltip != null) {
```
dbrant, line 244: do we also want to check if it's `isShowing()`?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageActivity.java
cooltey: Yes!

app/src/main/java/org/wikipedia/page/PageActivity.java
cooltey: Yes and done.

PR #1916, opened 2021-01-14
Lowercase app_session_id expected

sharvaniharan:


app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
```diff
@@ -340,7 +340,7 @@ static void beginNewSession() {
         @SuppressWarnings("checkstyle:magicnumber")
         public static String generateRandomId() {
             Random random = new Random();
-            return String.format("%08X", random.nextInt()) + String.format("%08X", random.nextInt()) + String.format("%04X", random.nextInt() & 0xFFFF);
+            return String.format("%08X", random.nextInt()) + String.format("%08X", random.nextInt()) + String.format("%04X", random.nextInt() & 0xFFFF).toLowerCase();
```
dbrant, line 343: fyi, you can just use a lowercase `x` in those format specifiers.

dbrant (changes requested):

PR #1920, opened 2021-01-15
Fix: Cannot restart a cancelled timer

sharvaniharan:


dbrant:
Aha... looking at the source code of the `Timer` class, I think I misunderstood its intended usage, and I no longer think that it's appropriate for our purposes here. (i.e. a `Timer` is not intended to be "reused").
This is fine for now, but we should probably use some other mechanism, such as a simple delayed Runnable.

PR #1925, opened 2021-01-15
Begin migration to View Bindings.

dbrant:
It looks like the new View Binding functionality can coexist with our other usages of Synthetics, so we can convert all of our activities as we go along.
This converts just one fragment to start with, and shows the pattern that should be used for converting, namely:

### Activities

* Each Activity shall have a `lateinit var binding` field that will be assigned in `onCreate()` when the layout is inflated.
* All of the previous synthetic references can now be changed to reference the field from the binding class. For example, if the synthetic name was `myTextView`, it will now be `binding.myTextView`.

### Fragments

* Each Fragment shall have a `_binding` field that is nullable, and will be assigned when the View is inflated.
* The `_binding` field shall be set to `null` in `onDestroyView()`.
* There should also be another field called `binding` (no underscore) that is non-null, with a getter method that returns the other field (with underscore).
* All of the previous synthetic references can now be changed to reference the field from the binding class. For example, if the synthetic name was `myTextView`, it will now be `binding.myTextView`.

For more info, see the [documentation](https://developer.android.com/topic/libraries/view-binding#kotlin).


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
```diff
@@ -172,36 +176,36 @@ class SuggestedEditsCardsFragment : Fragment(), SuggestedEditsImageTagsFragment.
...
     }
 
     override fun onDestroyView() {
         funnel?.stop()
         disposables.clear()
-        cardsViewPager.unregisterOnPageChangeCallback(viewPagerListener)
-        cardsViewPager.adapter = null
+        binding.cardsViewPager.unregisterOnPageChangeCallback(viewPagerListener)
+        binding.cardsViewPager.adapter = null
         super.onDestroyView()
```
cooltey, line 209: Should we add `_binding = null` here?

dbrant: correct!

PR #1930, opened 2021-01-19
Set correct language code for watchlist items for language variants

cooltey:
If we have added language variants in the app, for example, `zh-tw` and `zh-hans`, the Watchlist will show the list items only in the default language, which is `zh`, and that's not correct.

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/Service.java
+++ b/app/src/main/java/org/wikipedia/dataclient/Service.java
@@ -336,8 +336,8 @@ public interface Service {
     @GET(MW_API_PREFIX + "action=query&prop=info&converttitles=&redirects=&inprop=watched")
     @NonNull Observable<MwQueryResponse> getWatchedInfo(@NonNull @Query("titles") String titles);
 
-    @GET(MW_API_PREFIX + "action=query&meta=allmessages&list=watchlist&wllimit=500&wlallrev=1&wlprop=ids|title|flags|comment|parsedcomment|timestamp|sizes|user")
-    @NonNull Observable<MwQueryResponse> getWatchlist(@NonNull @Query("ammessages") String languageCode);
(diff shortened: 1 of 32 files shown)


app/src/main/java/org/wikipedia/dataclient/Service.java
```diff
@@ -336,8 +336,8 @@
     @GET(MW_API_PREFIX + "action=query&prop=info&converttitles=&redirects=&inprop=watched")
     @NonNull Observable<MwQueryResponse> getWatchedInfo(@NonNull @Query("titles") String titles);
 
-    @GET(MW_API_PREFIX + "action=query&meta=siteinfo&list=watchlist&wllimit=500&wlallrev=1&wlprop=ids|title|flags|comment|parsedcomment|timestamp|sizes|user")
-    @NonNull Observable<MwQueryResponse> getWatchlist();
+    @GET(MW_API_PREFIX + "action=query&meta=allmessages&list=watchlist&wllimit=500&wlallrev=1&wlprop=ids|title|flags|comment|parsedcomment|timestamp|sizes|user")
+    @NonNull Observable<MwQueryResponse> getWatchlist(@NonNull @Query("ammessages") String languageCode);
```
dbrant, line 340: I'm not quite seeing what this is trying to do. If you pass just a language code to the `allmessages` API, it won't return anything meaningful, and will just cause load on the server.  Could you clarify what the goal is here?

dbrant (changes requested):

app/src/main/java/org/wikipedia/dataclient/Service.java
cooltey: Make senses. I change the logic that creates a temporary list to store enabled language codes and it works as expected now.

PR #1935, opened 2021-01-20
Avoid using GsonMarshaller on PageTitle objects

cooltey:
It could help us to follow the rule that converts to one Kotlin file per PR.


dbrant:
alright!

PR #1939, opened 2021-01-21
Adjust text highlighting with linespacing extra

sharvaniharan:


app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -331,12 +330,14 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback {
...
-        val foregroundAddedColor = ForegroundColorSpan(ResourceUtil.getThemedColor(requireContext(), R.attr.color_group_64))
-        val foregroundRemovedColor = ForegroundColorSpan(ResourceUtil.getThemedColor(requireContext(), R.attr.color_group_65))
-        spannableText.setSpan(BackgroundColorSpan(ResourceUtil.getThemedColor(requireContext(),
-                if (isAddition) R.attr.edit_green_highlight else R.attr.edit_red_highlight)), start, end, 0)
+        val foregroundAddedColor = ResourceUtil.getThemedColor(requireContext(), R.attr.color_group_64)
+        val foregroundRemovedColor = ResourceUtil.getThemedColor(requireContext(), R.attr.color_group_65)
+        val backgroundSpan = BackgroundColorSpanWithLineSpacing(ResourceUtil.getThemedColor(requireContext(),
+                if (isAddition) R.attr.edit_green_highlight else R.attr.edit_red_highlight),
+                if (isAddition) foregroundAddedColor else foregroundRemovedColor,
+                diffText.layout.spacingAdd)
```
cooltey, line 338: Not sure why but I am seeing the uneven top and bottom padding of highlights in different sentences. Is that expected?

dbrant (changes requested):
This does not work when diffs are more than one line:

image

PR #1942, opened 2021-01-21
Convert BaseActivity to Kotlin.

dbrant:


app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -0,0 +1,292 @@
...
+    private fun removeSplashBackground() {
+        window.setBackgroundDrawable(null)
+    }
+
+    private fun unregisterExclusiveBusMethods() {
+        if (EXCLUSIVE_BUS_METHODS != null && EXCLUSIVE_DISPOSABLE != null) {
+            EXCLUSIVE_DISPOSABLE!!.dispose()
+            EXCLUSIVE_DISPOSABLE = null
+            EXCLUSIVE_BUS_METHODS = null
+        }
```
cooltey, line 234: Is it possible to use `EXCLUSIVE_DISPOSABLE?.dispose()` here and remove the `null` checks?

dbrant: correct!

PR #1950, opened 2021-01-21
Kotlin conversion: AnimationUtil, ActiveTimer and ArrayUtils converted

sharvaniharan:


app/src/main/java/org/wikipedia/util/ArrayUtils.kt
```diff
@@ -0,0 +1,42 @@
...
+        if (index == ArrayUtils.INDEX_NOT_FOUND) {
+            return ArrayUtils.clone(array)
+        }
+        val indices = IntArray(array.size - index)
+        indices[0] = index
+        var count = 1
+        while (ArrayUtils.indexOf(array, element, indices[count - 1] + 1).also { index = it } != ArrayUtils.INDEX_NOT_FOUND) {
+            indices[count++] = index
+        }
+        return ArrayUtils.removeAll(array, *Arrays.copyOf(indices, count))
```
cooltey, line 40: The `copyOf` can be simplified.

app/src/main/java/org/wikipedia/util/ArrayUtils.kt
```diff
@@ -0,0 +1,42 @@
+package org.wikipedia.util
+
+import org.apache.commons.lang3.ArrayUtils
+import java.util.*
+
+// Some OEMs appear to have added a pre-v3.5, possibly re-v3.4, version of ArrayUtils to the system
+// path which does not contain removeAllOccurrences(). This class is a partial copy of Apache
+// Commons' ArrayUtils
+// https://rink.hockeyapp.net/manage/apps/226650/app_versions/79/crash_reasons/156658637
```
cooltey, line 9: The URL is invalid. Can we update or remove the comment?

app/src/main/java/org/wikipedia/util/ActiveTimer.kt
```diff
@@ -0,0 +1,27 @@
...
+    fun resume() {
+        startMillis -= System.currentTimeMillis() - pauseMillis
+    }
+
+    val elapsedSec: Int
+        get() = TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis() - startMillis).toInt()
+
+    init {
+        reset()
+    }
```
cooltey, line 26: Would it be possible to move `init{}` to the top of these methods?

cooltey (changes requested):

sharvaniharan:
I have removed all unnecessary comments, since our code is self-commenting. Please lmk if we can follow it going forward, or save any comments as is. @dbrant @cooltey

PR #1952, opened 2021-01-22
Fix: get fragment value properly to prevent issue when changing device orientation

cooltey:
Steps to reproduce:

1. Go to Explore Feed.
2. Change device orientation to landscape.
3. Scroll down and you will see the view overlapped.


cooltey:
#1951

PR #1958, opened 2021-01-22
Kotlin conversion: BatchUtil,ClipboardUtil, ContentProviderClientCompat

sharvaniharan:


app/src/main/java/org/wikipedia/util/BatchUtil.kt
```diff
@@ -0,0 +1,48 @@
+package org.wikipedia.util
+
+import org.wikipedia.Constants
+import org.wikipedia.page.PageTitle
+import kotlin.math.min
+
+object BatchUtil {
```
cooltey, line 7: Looks like we can remove this file since this `BatchUtil` is only used on the test case itself.

app/src/test/java/org/wikipedia/util/BatchUtilTest.java
```diff
@@ -34,7 +34,7 @@
             private List<Integer> sizes = new ArrayList<>();
             private int count;
 
-            @Override public void handleBatch(@NonNull List<PageTitle> batchTitles, int total,
+            @Override public void handleBatch(@NonNull List<? extends PageTitle> batchTitles, int total,
```
cooltey, line 37: See comment above.

app/src/main/java/org/wikipedia/util/ContentProviderClientCompat.kt
```diff
@@ -0,0 +1,16 @@
+package org.wikipedia.util
+
+import android.content.ContentProviderClient
+import android.os.Build
+
+object ContentProviderClientCompat {
+    @Suppress("DEPRECATION")
```
cooltey, line 7: It should be fine if we don't add this.

cooltey (changes requested):

app/src/main/java/org/wikipedia/util/BatchUtil.kt
sharvaniharan: Yes @cooltey I saw that, but I thought may be we could use it as a central place for batch processing... but I see your point... let us introduce it if we decide that and also would need to be able to support beyond just PageTitles.

app/src/test/java/org/wikipedia/util/BatchUtilTest.java
sharvaniharan: Done.

PR #1960, opened 2021-01-22
Fix updating of reading list metadata for articles in multiple lists.

dbrant:
A rather obscure issue, but worth fixing:
* Add an article to more than one reading list, and make sure the lists are synced to your account.
* Uninstall the app, then reinstall it and login to get your lists back, but cancel downloading of the articles as soon as it starts.
* Go to your reading lists and navigate to the article that you added (pick from one of the lists).
* Wait for the article to load, so that its description and lead image are loaded.
* Go back to your lists, and observe that the description and thumbnail for that page are not updated, or updated only for one instance, instead of all of them.

### DIFF
--- a/app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
+++ b/app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
@@ -391,8 +391,8 @@ public class ReadingListDbHelper {
         }
     }
 
-    public void updateMetadataByTitle(@NonNull String title, @NonNull String lang,
-                                      @Nullable String description, @Nullable String thumbUrl) {
+    public void updateMetadataByTitle(@NonNull ReadingListPage pageProto, @Nullable String description,
+                                      @Nullable String thumbUrl) {
         SQLiteDatabase db = getWritableDatabase();
         db.beginTransaction();
         try {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/readinglist/database/ReadingListDbHelper.java
```diff
@@ -391,6 +391,27 @@ public void updatePage(@NonNull ReadingListPage page) {
...
+    public void updateMetadataByTitle(@NonNull String title, @NonNull String lang,
+                                      @Nullable String description, @Nullable String thumbUrl) {
+        SQLiteDatabase db = getWritableDatabase();
+        db.beginTransaction();
+        try {
+            ContentValues contentValues = new ContentValues();
+            contentValues.put(ReadingListPageContract.Col.THUMBNAIL_URL.getName(), thumbUrl);
+            contentValues.put(ReadingListPageContract.Col.DESCRIPTION.getName(), description);
+            int result = db.update(ReadingListPageContract.TABLE, contentValues,
+                    ReadingListPageContract.Col.API_TITLE.getName() + " = ? AND "
```
cooltey, line 403: Would it be possible to also check `ReadingListPageContract.Col.DISPLAY_TITLE.getName()` to see if the title matches? Please see `findPageInAnyList()` and `getPageByTitle()`.

dbrant: Why is that necessary?  (When would there be multiple pages with the same API_TITLE and language, but different DISPLAY_TITLE?)

cooltey: In `ReadingListPageContract`, the `DISPLAY_TITLE` was used to be the "default" title of the data. And the reason why the checks exist is to resolve the language variant issues.

dbrant: alright!

PR #1962, opened 2021-01-22
Convert Auth package to Kotlin.

dbrant:
Note that this conversion has side-effects for other classes where it's referenced.


app/src/main/java/org/wikipedia/auth/AccountUtil.kt
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @JvmStatic
+    fun accountType(): String {
+        return WikipediaApp.getInstance().getString(R.string.account_type)
+    }
+
+    private fun createAccount(userName: String, password: String): Boolean {
+        var account = account()
+        if (account == null || TextUtils.isEmpty(account.name) || account.name != userName) {
```
cooltey, line 129: The `TextUtils.isEmpty()` can be replaced by `account.name..isNullOrEmpty()`

cooltey (changes requested):

app/src/main/java/org/wikipedia/auth/AccountUtil.kt
dbrant: correct!

PR #1967, opened 2021-01-23
Kotlin conversion: DateUtil

sharvaniharan:


app/src/main/java/org/wikipedia/util/DateUtil.kt
```diff
@@ -0,0 +1,177 @@
...
+    // TODO: Switch to DateTimeFormatter when minSdk = 26.
+    @JvmStatic
+    @Synchronized
+    fun iso8601DateFormat(date: Date?): String {
+        return getCachedDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.ROOT, true)!!.format(date!!)
+    }
+
+    @JvmStatic
+    @Synchronized
+    @Throws(ParseException::class)
```
dbrant, line 25: I don't think we need to explicitly specify `@Throws` annotations anymore.

app/src/main/java/org/wikipedia/util/DateUtil.kt
```diff
@@ -0,0 +1,177 @@
...
+import android.text.format.DateFormat
+import org.wikipedia.R
+import org.wikipedia.WikipediaApp
+import org.wikipedia.feed.model.UtcDate
+import java.text.ParseException
+import java.text.SimpleDateFormat
+import java.util.*
+
+object DateUtil {
+    private val DATE_FORMATS: MutableMap<String, SimpleDateFormat> = HashMap()
```
dbrant, line 14: Can be simplified to:
`private val DATE_FORMATS = HashMap<String, SimpleDateFormat>()`

dbrant (changes requested):

app/src/main/java/org/wikipedia/util/DateUtil.kt
```diff
@@ -0,0 +1,174 @@
...
+    @JvmStatic
+    @Synchronized
+    fun iso8601DateFormat(date: Date): String {
+        return getCachedDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.ROOT, true)!!.format(date)
+    }
+
+    @JvmStatic
+    @Synchronized
+    fun iso8601DateParse(date: String): Date {
+        return getCachedDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.ROOT, true)!!.parse(date!!)!!
```
cooltey, line 26: Please remove `!!` on `date`.

cooltey (commented):
One minor comment

PR #1968, opened 2021-01-25
Convert CommunicationBridge to Kotlin.

dbrant:


app/src/main/java/org/wikipedia/bridge/CommunicationBridge.kt
```diff
@@ -0,0 +1,190 @@
...
+    })
+
+    private class CommunicatingChrome : WebChromeClient() {
+        override fun onConsoleMessage(consoleMessage: ConsoleMessage): Boolean {
+            L.d(consoleMessage.sourceId() + ":" + consoleMessage.lineNumber() + " - " + consoleMessage.message())
+            return true
+        }
+    }
+
+    private inner class PcsClientJavascriptInterface {
```
cooltey, line 148: Would it be good if we add `@Suppress("unused")` to prevent this class from being removed accidentally?

dbrant: But it is used. I'm not seeing any warnings related to "unused" here...

cooltey: I should probably specified that the methods `onReceiveMessage()` and `setupSettings` are tagged unused. But after I tapped some `JavascriptInterface` relate action on IDE and the unused warnings are gone.

PR #1972, opened 2021-01-25
Design tweak: Diff view

sharvaniharan:
1. Introduced the bottomsheet for link preview
2. Changed submenu style to adapt to different styles
3. Ui update to avoid jump in ui
4. Hide thank button when the edit is made by the account holder


app/src/main/res/values/styles.xml
```diff
@@ -37,12 +37,19 @@
 
         <!-- TODO: remove when the components can get correct custom ?attr/colorPrimary -->
         <item name="colorPrimary">?attr/colorPrimary</item>
+        <item name="android:itemBackground">@drawable/menu_pop_up</item>
```
dbrant, line 40: Is this necessary? It seems to work correctly without it.

app/src/main/res/values/styles.xml
```diff
@@ -37,12 +37,19 @@
...
+        <item name="actionBarPopupTheme">@style/SubMenuStyle</item>
```
dbrant, line 41: This looks like the right attribute to set, and we already have a style called `ToolbarPopupTheme`, so creating a new one is not necessary.

dbrant (changes requested):

PR #1973, opened 2021-01-26
Fix: make balloon tooltip content scrollable

cooltey:
https://phabricator.wikimedia.org/T269448#6774626


dbrant:
I suppose this is fine... But this is clearly an issue in the Balloon library because it doesn't adapt to the height of a custom View.  Are the Balloon library maintainers aware of this issue?

cooltey:
> 
> 
> I suppose this is fine... But this is clearly an issue in the Balloon library because it doesn't adapt to the height of a custom View. Are the Balloon library maintainers aware of this issue?

I searched the issue on the repo and saw a similar issue (https://github.com/skydoves/Balloon/issues/112) and looks like this is a known issue, but I'll create an issue to bring this up again.

cooltey:
@dbrant I think a scrollable view would be a fine solution since I have tried the other solution but didn't get expected results.
Here is the issue of the libarary: https://github.com/skydoves/Balloon/issues/148

PR #1976, opened 2021-01-26
Remove unnecessary intercepter in OnThisDayFragment

cooltey:
When trying to convert the `DontInterceptTouchListener` to Kotlin, I noticed this file is only used in `OnThisDayFragment`, and looks like there's no reason to continue using it since the UI has been changed a lot. @sharvaniharan Do you recall the reason why you added this to the `RecyclerView`?


sharvaniharan:
If I remember it right, it was something to do with scrolling. Will test this and merge/comment. @cooltey

PR #1977, opened 2021-01-26
Kotlin Conversion:  util 4 classes

sharvaniharan:


app/src/main/java/org/wikipedia/util/MathUtil.kt
```diff
@@ -0,0 +1,32 @@
+package org.wikipedia.util
+
+object MathUtil {
+    private const val PERCENTAGE_BASE = 100
+
+    @JvmStatic
+    fun constrain(f: Float, min: Float, max: Float): Float {
```
dbrant, line 7: Kotlin already has a function called `coerceIn()`, making this unnecessary.

app/src/main/java/org/wikipedia/util/PermissionUtil.kt
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @JvmStatic
+    fun requestWriteStorageRuntimePermissions(fragment: Fragment, requestCode: Int) {
+        fragment.requestPermissions(arrayOf(Manifest.permission.WRITE_EXTERNAL_STORAGE), requestCode)
+        // once permission is granted/denied it will continue with onRequestPermissionsResult
+    }
+
+    @JvmStatic
+    fun requestWriteStorageRuntimePermissions(activity: AppCompatActivity?, requestCode: Int) {
```
dbrant, line 44: Can `activity` be non-null?

dbrant (changes requested):

PR #1978, opened 2021-01-26
T272618 Random

SIKV:
**random** package

https://phabricator.wikimedia.org/T272618


app/src/main/java/org/wikipedia/random/RandomFragment.kt
```diff
@@ -0,0 +1,301 @@
...
+        }
+    }
+
+    private var _binding: FragmentRandomBinding? = null
+    private val binding get() = _binding!!
+
+    private val disposables = CompositeDisposable()
+
+    private var wikiSite: WikiSite? = null
+    private val requireWikiSite get() = wikiSite ?: throw IllegalStateException("wikiSite is null.")
```
dbrant, line 62: I'm not sure we need this kind of construct. Let's just make `wikiSite` itself into a `lateinit var`, and let it be non-null.

app/src/main/java/org/wikipedia/random/RandomFragment.kt
```diff
@@ -0,0 +1,301 @@
...
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+        super.onCreateView(inflater, container, savedInstanceState)
+
+        _binding = FragmentRandomBinding.inflate(inflater, container, false)
+        val view = binding.root
+
+        FeedbackUtil.setButtonLongPressToast(binding.randomNextButton, binding.randomSaveButton)
+
+        wikiSite = arguments?.getParcelable(RandomActivity.INTENT_EXTRA_WIKISITE)
```
dbrant, line 79: I think the new convention is to use `requireArguments()`.

app/src/main/java/org/wikipedia/random/RandomItemFragment.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    private var _binding: FragmentRandomItemBinding? = null
+    private val binding get() = _binding!!
+
+    private val disposables = CompositeDisposable()
+
+    private var summary: PageSummary? = null
+
+    private var wikiSite: WikiSite? = null
+    private val requireWikiSite get() = wikiSite ?: throw IllegalStateException("wikiSite is null.")
```
dbrant, line 41: Same comment about making `wikiSite` itself be `lateinit`.

app/src/main/java/org/wikipedia/random/RandomItemFragment.kt
```diff
@@ -0,0 +1,144 @@
...
+
+    val isLoadComplete: Boolean get() = summary != null
+    val title: PageTitle? get() = summary?.getPageTitle(requireWikiSite)
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        wikiSite = arguments?.getParcelable(RandomActivity.INTENT_EXTRA_WIKISITE)
```
dbrant, line 49: `requireArguments()`

app/src/main/java/org/wikipedia/views/WikiArticleCardView.kt
```diff
@@ -21,7 +21,7 @@ class WikiArticleCardView constructor(context: Context, attrs: AttributeSet? = n
         View.inflate(context, R.layout.view_wiki_article_card, this)
     }
 
-    fun setTitle(title: String) {
+    fun setTitle(title: String?) {
```
dbrant, line 24: Can we keep this as a non-null parameter?

dbrant (changes requested):
Thanks! Please see some minor comments inline.

app/src/main/java/org/wikipedia/random/RandomFragment.kt
SIKV: There are two options:

1. `wikiSite = requireArguments().getParcelable(RandomActivity.INTENT_EXTRA_WIKISITE)!!`

2.
```
wikiSite = requireArguments().getParcelable(RandomActivity.INTENT_EXTRA_WIKISITE) 
                ?: throw IllegalStateException("wikiSite is null.")
```

I personally like the second more because it gives additional information.

app/src/main/java/org/wikipedia/views/WikiArticleCardView.kt
SIKV: This should be a non-null. I agree.

app/src/main/java/org/wikipedia/random/RandomFragment.kt
dbrant: I personally prefer the first one, since it will give us nearly the same information.

PR #1981, opened 2021-01-26
Convert face detect related views to Kotlin

cooltey:
The `CenterCropWithFace()` was an inner class inside `FaceAndColorDetectImageView.java`, and I think it would be better if we separate that out as an individual class.


app/src/main/java/org/wikipedia/views/FaceAndColorDetectImageView.kt
```diff
@@ -0,0 +1,84 @@
...
+class FaceAndColorDetectImageView constructor(context: Context, attrs: AttributeSet? = null) : AppCompatImageView(context, attrs) {
+
+    interface OnImageLoadListener {
+        fun onImageLoaded(palette: Palette)
+        fun onImageFailed()
+    }
+
+    private fun shouldDetectFace(uri: Uri): Boolean {
+        // TODO: not perfect; should ideally detect based on MIME type.
+        val path = StringUtils.defaultString(uri.path).toLowerCase(Locale.ROOT)
```
dbrant, line 34: Usages of `StringUtils.defaultString()` can be replaced with `.orEmpty()`

dbrant (changes requested):

app/src/main/java/org/wikipedia/views/FaceAndColorDetectImageView.kt
cooltey: Correct! Thanks.

PR #1992, opened 2021-01-28
Make the watchlist tooltip auto-cancelable.

dbrant:


cooltey:
This will let the behavior of clicking on the overflow icon that dismisses the tooltip and showing the menu does not work properly.

dbrant:
This is an improvement over the current behavior of making the tooltip sticky and only dismissable by clicking the overflow menu. The proper solution will be to make it the way it works in `MainActivity`, or move that logic up to `BaseActivity` in the future.

PR #1999, opened 2021-01-28
Fix for different log types

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T269447#6783313

### DIFF
--- a/app/src/main/java/org/wikipedia/watchlist/WatchlistItemView.kt
+++ b/app/src/main/java/org/wikipedia/watchlist/WatchlistItemView.kt
@@ -1,6 +1,8 @@
 package org.wikipedia.watchlist
 
 import android.content.Context
+import android.content.res.ColorStateList
+import android.graphics.drawable.Drawable
 import android.util.AttributeSet
 import android.view.View
 import android.view.View.OnClickListener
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/watchlist/WatchlistItemView.kt
```diff
@@ -52,17 +54,50 @@ class WatchlistItemView constructor(context: Context, attrs: AttributeSet? = nul
...
-        if (diffByteCount >= 0) {
-            diffText.setTextColor(if (diffByteCount > 0) ContextCompat.getColor(context, R.color.green50) else ResourceUtil.getThemedColor(context, R.attr.material_theme_secondary_color))
-            diffText.text = String.format("%+d", diffByteCount)
+        if (item.logType.isNotEmpty()) {
+            when (item.logType) {
+                context.getString(R.string.page_moved) -> {
+                    setButtonTextAndIconColor(context.getString(R.string.watchlist_page_moved),
+                            ResourceUtil.getThemedColor(context, R.attr.color_group_61),
+                            ContextCompat.getDrawable(context, R.drawable.ic_info_outline_black_24dp),
+                            R.attr.suggestions_background_color)
```
cooltey, line 63: Would it be possible to have the same way of sending attributes and let the method does the job?

```
R.attr.color_group_61,
R.drawable.ic_info_outline_black_24dp,
R.attr.suggestions_background_color)
```

sharvaniharan: I did it this way because we will have a chance to send null when we dont want an icon.. is it ok to have it this way so we dont need to make a null check inside the function?

cooltey: Please see the follow-up PR.

PR #2009, opened 2021-01-30
Kotlin conversion: UriUtil

sharvaniharan:


app/src/main/java/org/wikipedia/util/UriUtil.kt
```diff
@@ -0,0 +1,151 @@
...
+
+    @JvmStatic
+    fun getUrlWithProvenance(context: Context, title: PageTitle,
+                             @StringRes provId: Int): String {
+        return title.uri + "?wprov=" + context.getString(provId)
+    }
+
+    @JvmStatic
+    fun getFilenameFromUploadUrl(url: String): String {
+        val splitArray = url.split("/".toRegex()).toTypedArray()
```
cooltey, line 107: Is the `toTypedArray()` necessary? Looks like we can rewrite this method.

app/src/main/java/org/wikipedia/util/UriUtil.kt
```diff
@@ -0,0 +1,151 @@
...
+    @JvmStatic
+    fun decodeURL(url: String): String {
+        return try {
+            // Force decoding of plus sign, since the built-in decode() function will replace
+            // plus sign with space.
+            URLDecoder.decode(url.replace("+", "%2B"), "UTF-8")
+        } catch (e: IllegalArgumentException) {
+            // Swallow IllegalArgumentException (can happen with malformed encoding), and just
+            // return the original string.
+            d("URL decoding failed. String was: $url")
```
cooltey, line 34: Can we have a standard `L.d()` at this line?

cooltey (changes requested):

app/src/main/java/org/wikipedia/util/UriUtil.kt
sharvaniharan: @dbrant @cooltey Wondering if we sshould filter out spaces here too?  We might need to remove the test for empty string then : testFilenameFromUploadUrl -> assertThat(UriUtil.getFilenameFromUploadUrl(""), is(""));

app/src/main/java/org/wikipedia/util/UriUtil.kt
```diff
@@ -0,0 +1,147 @@
...
+        return removeFragment(removeLinkPrefix(url)).replace("_", " ")
+    }
+
+    /** Get language variant code from a Uri, e.g. "zh.*", otherwise returns empty string.  */
+    @JvmStatic
+    fun getLanguageVariantFromUri(uri: Uri): String {
+        if (uri.path.isNullOrEmpty()) {
+            return ""
+        }
+        val parts = StringUtils.defaultString(uri.path).split('/')
```
cooltey, line 125: Since we already have the `isNullOrEmpty()`, do we still need this `StringUtils.defaultString()` instead of using `uri.path!!`?

PR #2022, opened 2021-02-02
Fix additional UI issues in Diff screens.

dbrant:
* I recall discussing the `Locale.ROOT` issue that was causing incorrect "M01" dates to be shown, but it looks like this was never fixed?
* Apply correct sizing to the diff-count button.
* For zero-sized diffs, show "0" instead of "+0".

https://phabricator.wikimedia.org/T269447


sharvaniharan:
Thank you for this @dbrant ... I did fix the date issue here : https://github.com/wikimedia/apps-android-wikipedia/pull/1996/files
Not sure what happened there... some rebasing snafu?

PR #2027, opened 2021-02-02
Kotlin conversion: FileUtil

sharvaniharan:


app/src/main/java/org/wikipedia/util/FileUtil.kt
```diff
@@ -0,0 +1,69 @@
...
+            return stringBuilder.toString()
+        }
+    }
+
+    @JvmStatic
+    fun deleteRecursively(f: File) {
+        if (f.isDirectory) {
+            for (child in f.listFiles()) {
+                deleteRecursively(child)
+            }
```
cooltey, line 45: Can this be:
```
  f.listFiles()?.forEach {
      deleteRecursively(it)
  }
```

PR #2033, opened 2021-02-03
Follow-up: remove unused NavMenuAwareRecyclerView.

dbrant:
...we don't actually need it anymore, since we no longer have the swipeable Nav menu on the main screen.


app/src/main/java/org/wikipedia/views/FrameLayoutNavMenuTriggerer.kt
```diff
@@ -9,8 +9,10 @@ import org.wikipedia.util.DimenUtil.roundedDpToPx
 import org.wikipedia.util.L10nUtil
 import kotlin.math.abs
 
-class FrameLayoutNavMenuTriggerer @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0) :
-        FrameLayout(context, attrs, defStyleAttr) {
+class FrameLayoutNavMenuTriggerer : FrameLayout {
+    constructor(context: Context) : super(context)
+    constructor(context: Context, attrs: AttributeSet? = null) : super(context, attrs)
+    constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0) : super(context, attrs, defStyleAttr)
```
dbrant, line 15: ^ A quick note about this:
I think I'm going to start recommending that we no longer use `@JvmOverloads` for custom Views.  We have already seen that it can cause problems with styling. This is because the three different constructors might not call through to each other in the way that `@JvmOverloads` assumes, so certain style information and logic might be lost.

I believe the truly proper way is to keep three separate constructors, which call the `super` constructor with the same number of parameters.

PR #2034, opened 2021-02-03
Kotlin conversion: ThrowableUtil

sharvaniharan:


app/src/main/java/org/wikipedia/util/ThrowableUtil.kt
```diff
@@ -0,0 +1,98 @@
...
+            AppError(inner.localizedMessage!!, "")
+        } else if (throwableContainsException(e, JSONException::class.java)) {
+            AppError(context.getString(R.string.error_response_malformed),
+                    inner.localizedMessage)
+        } else {
+            // everything else has fallen through, so just treat it as an "unknown" error
+            AppError(context.getString(R.string.error_unknown),
+                    inner.localizedMessage)
+        }
+        return result
```
cooltey, line 70: Would it be possible to replace this code to inline-style? e.g.:
```
return = if (isNetworkError(e)) ...
```

app/src/main/java/org/wikipedia/util/ThrowableUtil.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @JvmStatic
+    fun isOffline(caught: Throwable?): Boolean {
+        return (caught is UnknownHostException ||
```
cooltey, line 75: Looks like the `( )` is not necessary.

app/src/main/java/org/wikipedia/util/ThrowableUtil.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @JvmStatic
+    fun is404(caught: Throwable): Boolean {
+        return caught is HttpStatusException && caught.code() == 404
+    }
+
+    @JvmStatic
+    fun isNetworkError(e: Throwable): Boolean {
+        return (throwableContainsException(e, UnknownHostException::class.java) ||
```
cooltey, line 91: Looks like the `( )` is not necessary.

app/src/main/java/org/wikipedia/util/ThrowableUtil.kt
```diff
@@ -0,0 +1,98 @@
...
+                throwableContainsException(e, TimeoutException::class.java) ||
+                throwableContainsException(e, SSLException::class.java))
+    }
+
+    @Deprecated("")
+    class AppError(val error: String, val detail: String?)
```
cooltey, line 97: Considering replace/rewrite this deprecated class?

cooltey (changes requested):

app/src/main/java/org/wikipedia/util/ThrowableUtil.kt
cooltey: @dbrant @sharvaniharan Do you think it is still reasonable to keep it as `deprecated`? or it is fine to display the raw server messages in certain cases and remove this `deprecated` tag?

sharvaniharan: I agree we can just remove the deprecation..re-writing can be a fresh task and we dont need to deprecate till then.

dbrant: yep, let's remove the `deprecated` tag for now.

PR #2035, opened 2021-02-03
Kotlin conversion: ResourceUtil

sharvaniharan:


app/src/main/java/org/wikipedia/util/ResourceUtil.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    @JvmStatic
+    @ColorInt
+    fun getThemedColor(context: Context, @AttrRes id: Int): Int {
+        val typedValue = getThemedAttribute(context, id)
+                ?: throw IllegalArgumentException("Attribute not found; ID=$id")
+        return typedValue.data
+    }
+
+    @Throws(NotFoundException::class)
```
cooltey, line 58: Would it be possible to import ` kotlin.jvm.Throws` to make this correctly?

sharvaniharan: done @cooltey

PR #2036, opened 2021-02-04
Kotlin conversion: ReleaseUtil

sharvaniharan:


app/src/main/java/org/wikipedia/util/ReleaseUtil.kt
```diff
@@ -0,0 +1,66 @@
...
+            return RELEASE_BETA
+        }
+        if (BuildConfig.APPLICATION_ID.contains("alpha")) {
+            return RELEASE_ALPHA
+        }
+        return if (BuildConfig.APPLICATION_ID.contains("dev")) {
+            RELEASE_DEV
+        } else {
+            RELEASE_PROD
+        }
```
cooltey, line 53: Can we simplify like this?
```
        return when {
            BuildConfig.APPLICATION_ID.contains("beta") -> RELEASE_BETA
            BuildConfig.APPLICATION_ID.contains("alpha") -> RELEASE_ALPHA
            BuildConfig.APPLICATION_ID.contains("dev") -> RELEASE_DEV
            else -> RELEASE_PROD
        }
```

PR #2044, opened 2021-02-05
Convert MultiSelectActionModeCallback to Kotlin

cooltey:


app/src/main/java/org/wikipedia/views/MultiSelectActionModeCallback.kt
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    override fun onDestroyActionMode(mode: ActionMode) {}
+
+    protected abstract fun onDeleteSelected()
+
+    companion object {
+        private const val ACTION_MODE_TAG = "multiSelectActionMode"
+        @JvmStatic
+        fun `is`(mode: ActionMode?): Boolean {
```
dbrant, line 37: Can we rename this to something other than `is`, since that's a reserved keyword in Kotlin?

dbrant (changes requested):

app/src/main/java/org/wikipedia/views/MultiSelectActionModeCallback.kt
cooltey: Based on our discussion, it is renamed to `isTagType`.

PR #2047, opened 2021-02-06
Dismiss login dialog to prevent showing the dialog after logged in.

cooltey:
The dialog should be dismissed after clicking on the positive button, otherwise, users will still see the dialog after logging in.

### DIFF
--- a/app/src/main/java/org/wikipedia/activity/BaseActivity.kt
+++ b/app/src/main/java/org/wikipedia/activity/BaseActivity.kt
@@ -239,7 +239,10 @@ abstract class BaseActivity : AppCompatActivity() {
                     .setCancelable(false)
                     .setTitle(R.string.logged_out_in_background_title)
                     .setMessage(R.string.logged_out_in_background_dialog)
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -239,7 +239,10 @@ abstract class BaseActivity : AppCompatActivity() {
                     .setCancelable(false)
                     .setTitle(R.string.logged_out_in_background_title)
                     .setMessage(R.string.logged_out_in_background_dialog)
-                    .setPositiveButton(R.string.logged_out_in_background_login) { _: DialogInterface?, _: Int -> startActivity(LoginActivity.newIntent(this@BaseActivity, LoginFunnel.SOURCE_LOGOUT_BACKGROUND)) }
+                    .setPositiveButton(R.string.logged_out_in_background_login) { dialog: DialogInterface, _: Int ->
+                        startActivity(LoginActivity.newIntent(this@BaseActivity, LoginFunnel.SOURCE_LOGOUT_BACKGROUND))
+                        dialog.dismiss()
```
dbrant, line 244: This should not be necessary.  An AlertDialog will automatically dismiss itself when either the Positive or Negative button is clicked.

dbrant (changes requested):

cooltey:
Closed since the issue is not from the `AlertDialog`.

PR #2050, opened 2021-02-08
Fix: show Edits tab tooltip with delayed post

cooltey:
The tooltip will not show without having a `postDelayed`.


sharvaniharan:
but do we need a half second delay?

cooltey:
> 
> 
> but do we need a half second delay?

This has the same behavior as the `maybeShowWatchlistTooltip`

PR #2053, opened 2021-02-09
Re-arrange article menu/toc items

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T273260


app/src/main/res/values/strings.xml
```diff
@@ -47,6 +47,8 @@
     <string name="menu_page_other_languages">Change language</string>
     <string name="menu_page_find_in_page">Find in article</string>
     <string name="menu_page_add_to_watchlist">Add to Watchlist</string>
+    <string name="menu_page_talk">View talk page</string>
+    <string name="menu_page_edit_history">View edit history</string>
```
cooltey, line 51: Do we really need the text "View" instead of showing "Talk page" and "Edit History"?

cooltey (changes requested):

app/src/main/res/values/strings.xml
sharvaniharan: Good catch .. looks like Jazmin has also skipped view on the ticket

sharvaniharan: I have updated it according to the strings Jazmin has put it @cooltey .

app/src/main/res/values/strings.xml
```diff
@@ -47,6 +47,8 @@
     <string name="menu_page_other_languages">Change language</string>
     <string name="menu_page_find_in_page">Find in article</string>
     <string name="menu_page_add_to_watchlist">Add to Watchlist</string>
+    <string name="menu_page_talk">Talk</string>
+    <string name="menu_page_edit_history">Edit History</string>
```
cooltey, line 51: Should we have a lowercase of `history?

sharvaniharan: She has kept it in Uppercase, probably like 'Add to Watchlist'

sharvaniharan: Per our conversation, I changed this to lowercase @cooltey . So done.

PR #2057, opened 2021-02-10
Handle all non-main namespace pages in the WebView.

dbrant:


app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
```diff
@@ -186,6 +186,24 @@ private void pageLoadFromNetwork(final ErrorCallback errorCallback) {
...
+        boolean delayLoadHtml = model.getTitle().getPrefixedText().contains(":");
+
+        if (!delayLoadHtml) {
+            bridge.resetHtml(model.getTitle());
+        }
+
+        if (model.getTitle().namespace() == Namespace.SPECIAL) {
+            // Short-circuit the entire process of fetching the Summary, since Special: pages
+            // are not supported in RestBase.
+            model.setPage(null);
```
cooltey, line 199: Is it necessary since line `189` has already set it to `null`?

app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
```diff
@@ -142,7 +142,7 @@ object JavaScriptActionHandler {
     fun mobileWebChromeShim(): String {
         return "(function() {" +
                 "let style = document.createElement('style');" +
-                "style.innerHTML = '.header-chrome { visibility: hidden; margin-top: 48px; height: 0px; } #page-secondary-actions { display: none; } .mw-footer { margin-bottom: 48px; }';" +
+                "style.innerHTML = '.header-chrome { visibility: hidden; margin-top: 48px; height: 0px; } #page-secondary-actions { display: none; } .mw-footer { margin-bottom: 48px; } .page-actions-menu { display: none; } .minerva__tab-container { display: none; }';" +
```
cooltey, line 145: Would it be possible to update `padding-bottom` in `.minerva-footer` so that the bottom toolbar will not cover the footer?

cooltey (changes requested):

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.java
dbrant: nope

app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
dbrant: yep

cooltey (commented):
Looks good to me. @sharvaniharan Please review it as well.

One more question: are we going to hide/update the overflow menu behaviors? e.g. "View talk page" and "View edit history"?

dbrant:
This can be tested by loading Category pages (search for `Category:` and pick any one), or Portal pages (search for `Portal:`), etc.  Also, this should now load article edit history pages, i.e. go to the bottom of any article and tap "View edit history".

sharvaniharan:
@dbrant lgtm.. one minor clarification... The experience of the psecial pages is a tiny bit different... I am assuming this is expected... but just wanted to make sure... for eg: https://en.wikipedia.org/wiki/Category:Companies we do not suggest the sub categories..

Also, Please re-name with 'design' in the name...  Robin will need to approve I guess..

sharvaniharan (approved):

dbrant:
> @dbrant lgtm.. one minor clarification... The experience of the psecial pages is a tiny bit different... I am assuming this is expected... but just wanted to make sure... for eg: https://en.wikipedia.org/wiki/Category:Companies we do not suggest the sub categories..

@sharvaniharan  Are you referring to the notice at the top of the Category page that it should be split into subcategories?  This is just a "message box" or "mbox", and as you know we hide all message boxes in all articles by default, including things like "needs additional citations", "candidate for deletion", for example:
https://en.wikipedia.org/wiki/Minima_Moralia
https://en.wikipedia.org/wiki/Angelo_Bertolini

PR #2059, opened 2021-02-10
Update gradle wrapper

TacoTheDank:
- Added a gradle wrapper validation action (see [here](https://github.com/gradle/wrapper-validation-action))
- Updated the gradle wrapper to 6.5 using git bash


dbrant:
Could we add the validation action to an existing workflow, i.e. `android_pr`, so that we don't thrash a poor VM unnecessarily?

TacoTheDank:
Sure! I don't know why I didn't do that in the first place lol

dbrant:
excellent 👍

PR #2068, opened 2021-02-11
Updated icons for PageActivity overflow

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T273260#6822076


app/src/main/res/drawable/ic_icon_revision_history_apps.xml
```diff
@@ -0,0 +1,13 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
```
cooltey, line 5: Should we also add `android:autoMirrored="true"` for this file?

PR #2071, opened 2021-02-12
Convert TextInputDialog and ReadingListTitleDialog to Kotlin

cooltey:


app/src/main/java/org/wikipedia/readinglist/ReadingListTitleDialog.kt
```diff
@@ -0,0 +1,56 @@
...
+                               description: String?,
+                               otherTitles: List<String?>,
+                               callback: Callback?): TextInputDialog {
+        return TextInputDialog(activity).let { textInputDialog ->
+            textInputDialog.callback = object : TextInputDialog.Callback {
+                override fun onShow(dialog: TextInputDialog) {
+                    dialog.setHint(R.string.reading_list_name_hint)
+                    dialog.setSecondaryHint(R.string.reading_list_description_hint)
+                    dialog.setText(title, true)
+                    dialog.setSecondaryText(StringUtils.defaultString(description))
```
dbrant, line 25: Can be `description.orEmpty()`

dbrant (changes requested):

app/src/main/java/org/wikipedia/readinglist/ReadingListTitleDialog.kt
cooltey: Done.

PR #2078, opened 2021-02-12
Inline reply in talk page

cooltey:
https://phabricator.wikimedia.org/T273292


dbrant:
archiving this, until the proper APIs are built.

PR #2079, opened 2021-02-13
Kotlin conversion: ThemeChooserDialog

sharvaniharan:


app/src/main/java/org/wikipedia/theme/ThemeChooserDialog.kt
```diff
@@ -0,0 +1,283 @@
...
+
+    private enum class FontSizeAction {
+        INCREASE, DECREASE, RESET
+    }
+
+    private var app: WikipediaApp? = null
+    private var funnel: AppearanceChangeFunnel? = null
+    private var invokeSource: InvokeSource? = null
+    private val disposables = CompositeDisposable()
+    private var updatingFont = false
```
dbrant, line 50: Can we have a blank line between here and `onCreateView`?

dbrant (changes requested):

PR #2080, opened 2021-02-13
Fix: show proper displayTitle for different language variants

cooltey:
Steps to reproduce:

 1. Go https://en.wikipedia.org/wiki/Lisa_the_Skeptic and change the language to `zh-hant`.
 2. Go to its talk page
 3. The title is showing `Talk: 怀疑论者莉萨`, but it should be showing `Talk: 懷疑論者莉薩`.


app/src/main/java/org/wikipedia/page/PageTitle.java
```diff
@@ -232,7 +232,8 @@ public void setText(@NonNull String convertedFromText) {
     }
 
     @NonNull public String getDisplayText() {
-        return displayText == null ? StringUtil.removeUnderscores(getPrefixedText()) : displayText;
+        return displayText == null ? StringUtil.removeUnderscores(getPrefixedText())
+                : namespace == null ? displayText : StringUtil.addUnderscores(namespace) + ":" + displayText;
     }
 
     public void setDisplayText(@Nullable String displayText) {
```
dbrant, line 239: If the `getDisplayText()` function is automatically adding the namespace, then the `setDisplayText()` function should automatically _remove_ the namespace.  We should do something like:
```
this.displayText = namespace == null ? displayText : StringUtil.removeNamespace(displayText);
```

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageTitle.java
```diff
@@ -232,7 +232,8 @@ public void setText(@NonNull String convertedFromText) {
     }
 
     @NonNull public String getDisplayText() {
-        return displayText == null ? StringUtil.removeUnderscores(getPrefixedText()) : displayText;
+        return displayText == null ? StringUtil.removeUnderscores(getPrefixedText())
+                : namespace == null ? displayText : StringUtil.addUnderscores(namespace) + ":" + displayText;
```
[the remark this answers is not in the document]
cooltey: Done.

app/src/main/java/org/wikipedia/page/PageTitle.java
cooltey: Done!

app/src/main/java/org/wikipedia/page/PageTitle.java
```diff
@@ -232,11 +232,12 @@ public void setText(@NonNull String convertedFromText) {
     }
 
     @NonNull public String getDisplayText() {
-        return displayText == null ? StringUtil.removeUnderscores(getPrefixedText()) : displayText;
+        return displayText == null ? StringUtil.removeUnderscores(getPrefixedText())
+                : namespace == null ? displayText : StringUtil.removeNamespace(namespace) + ":" + displayText;
```
dbrant, line 236: ...you mean `removeUnderscores()`?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageTitle.java
cooltey: My bad. done.

app/src/main/java/org/wikipedia/page/PageTitle.java
```diff
@@ -232,11 +232,12 @@ public void setText(@NonNull String convertedFromText) {
     }
 
     @NonNull public String getDisplayText() {
-        return displayText == null ? StringUtil.removeUnderscores(getPrefixedText()) : displayText;
+        return displayText == null ? StringUtil.removeUnderscores(getPrefixedText())
```
dbrant, line 235: Sorry, this is starting to cause underlying issues for titles that have a colon in their name, for example:

image

Can we solve your issue without making any changes to the `getDisplayText` and `setDisplayText` functions?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageTitle.java
cooltey: Sure, I'll go ahead to make another solution for the issue.

cooltey: Looks like the latest version removes the "Star trek:" in the talk page. The new commit fixes it but maybe we should consider updating the `removeNamespace()` or making another one like `removeWikiNamespace()`. 
Screenshot_20210216-152616_Wikipedia

PR #2086, opened 2021-02-17
Follow-up to page titles for Talk pages.

dbrant:


app/src/main/java/org/wikipedia/page/PageTitle.java
```diff
@@ -294,11 +296,13 @@ public boolean isSpecial() {
...
+        String talkNamespace = namespace().user() || namespace().userTalk()
+                ? UserTalkAliasData.valueFor(wiki.languageCode()) : TalkAliasData.valueFor(wiki.languageCode());
+
         PageTitle pageTitle = new PageTitle(talkNamespace, (namespace().userTalk() || namespace().user())
-                ? StringUtil.removeNamespace(getPrefixedText()) : getPrefixedText(), getWikiSite());
+                ? StringUtil.removeNamespace(getPrefixedText()) : getPrefixedText(), wiki);
         if (namespace().userTalk() || namespace().user()) {
-            pageTitle.setDisplayText(StringUtil.removeUnderscores(talkNamespace) + ": " + StringUtil.removeNamespace(getDisplayText()));
+            pageTitle.setDisplayText(StringUtil.removeUnderscores(talkNamespace) + ":" + StringUtil.removeNamespace(getDisplayText()));
         }
```
cooltey, line 306: Could you please help me to add the following? My bad that the language variant issue was not resolved in PR. You can check the issue by visiting the today's featured article on `zh-hant` wiki. (東盟成員國列表)

```
 else {
    pageTitle.setDisplayText(talkNamespace + ":" + getDisplayText());
}
```

app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
```diff
@@ -99,7 +99,18 @@ class TalkTopicsActivity : BaseActivity() {
...
+                        pageTitle.namespace() == Namespace.USER -> {
+                            UserAliasData.valueFor(WikipediaApp.getInstance().language().appLanguageCodes[pos])
+                        }
+                        pageTitle.namespace() == Namespace.USER_TALK -> {
+                            UserTalkAliasData.valueFor(WikipediaApp.getInstance().language().appLanguageCodes[pos])
+                        }
+                        else -> pageTitle.namespace
+                    }
+
+                    pageTitle = PageTitle(newNamespace, StringUtil.removeNamespace(pageTitle.prefixedText),
```
cooltey, line 113: Would it be possible to set `getDisplayText` here? Although this supposed to happen on the user talk page and it would be great to keep consistency.

app/src/main/java/org/wikipedia/page/PageTitle.java
dbrant: done

app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
dbrant: No, I don't see how this can be done without overwriting the language-specific namespace name.  To do this properly we'll need to do a major refactor of how we handle the namespace+displayText internally in PageTitle.

cooltey: Makes sense to me, thanks.

PR #2087, opened 2021-02-17
Correctly remove Wikipedia from any Sharing intents.

dbrant:
https://phabricator.wikimedia.org/T120315

This finally solves the long-standing issue of removing the Wikipedia app from the list of apps when sharing content, or when navigating to external links (if the external link is also on Wiki).


cooltey:
Excellent 💯

PR #2089, opened 2021-02-17
Streamline and consolidate invokeSource.

dbrant:
We have an enum called `InvokeSource` that serves as a universal way of specifying the source from which the user came.

...except in the case of `LanguageSettingsInvokeSource`, for some reason.

The intention might have been to have more granular invoke-sources for different features, but it's probably better to have a universal enum, for the sake of consistency.  This might even eliminate some possible crashes, if we are diligent about using `InvokeSource` instead of other ad-hoc `int` or `string` based source parameters.


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -415,7 +414,8 @@ class SuggestedEditsTasksFragment : Fragment() {
     private inner class TaskViewCallback : SuggestedEditsTaskView.Callback {
         override fun onViewClick(task: SuggestedEditsTask, isTranslate: Boolean) {
             if (WikipediaApp.getInstance().language().appLanguageCodes.size < MIN_LANGUAGES_TO_UNLOCK_TRANSLATION && isTranslate) {
-                showLanguagesActivity(LanguageSettingsInvokeSource.SUGGESTED_EDITS.text())
+                val intent = WikipediaLanguagesActivity.newIntent(requireActivity(), InvokeSource.SUGGESTED_EDITS)
+                startActivityForResult(intent, ACTIVITY_REQUEST_ADD_A_LANGUAGE)
```
cooltey, line 418: Would it be possible to combine two lines of code into one line?

PR #2100, opened 2021-02-19
Kotlin conversion: ThemeFittingRoomFragment 

sharvaniharan:


app/src/main/java/org/wikipedia/theme/ThemeFittingRoomFragment.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.wikipedia.databinding.FragmentThemeFittingRoomBinding
+import org.wikipedia.events.ChangeTextSizeEvent
+import org.wikipedia.events.WebViewInvalidateEvent
+import org.wikipedia.settings.Prefs
+
+class ThemeFittingRoomFragment : Fragment() {
+    private val disposables = CompositeDisposable()
+    private var _binding: FragmentThemeFittingRoomBinding? = null
+    private val binding get() = _binding!!
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
```
cooltey, line 22: Very minor but I would recommend that put `binding` and `_binding` on the top of these variables.

and please add an empty line between line `21` and line `22`.

cooltey (changes requested):

app/src/main/java/org/wikipedia/theme/ThemeFittingRoomFragment.kt
sharvaniharan: Done @cooltey

PR #2102, opened 2021-02-19
Kotlin conversion: ThemeFittingRoomActivity 

sharvaniharan:


app/src/main/java/org/wikipedia/theme/ThemeFittingRoomActivity.kt
```diff
@@ -0,0 +1,45 @@
...
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        if (savedInstanceState == null) {
+            themeChooserDialog = ThemeChooserDialog.newInstance(Constants.InvokeSource.SETTINGS)
+            bottomSheetPresenter.show(supportFragmentManager, themeChooserDialog!!)
+        }
+
+        // Don't let changed theme affects the status bar color and navigation bar color
+        setStatusBarColor(ContextCompat.getColor(this, R.color.black))
+        setNavigationBarColor(ContextCompat.getColor(this, R.color.black))
```
cooltey, line 24: Can we use `android.R.color.black` here instead of importing `import android.R` to prevent IDE warning?

cooltey (changes requested):

app/src/main/java/org/wikipedia/theme/ThemeFittingRoomActivity.kt
```diff
@@ -0,0 +1,44 @@
...
+import android.os.Bundle
+import androidx.core.content.ContextCompat
+import org.wikipedia.Constants
+import org.wikipedia.activity.SingleFragmentActivity
+import org.wikipedia.page.ExclusiveBottomSheetPresenter
+
+class ThemeFittingRoomActivity : SingleFragmentActivity<ThemeFittingRoomFragment>(), ThemeChooserDialog.Callback {
+    private var themeChooserDialog: ThemeChooserDialog? = null
+    private val bottomSheetPresenter = ExclusiveBottomSheetPresenter()
+    override fun onCreate(savedInstanceState: Bundle?) {
```
cooltey, line 14: Sorry I didn't notice this but is it possible to add an empty line between `onCreate` and the variables?

sharvaniharan: Done @cooltey

PR #2105, opened 2021-02-19
Convert CrashReport activity and fragment to kotlin.

dbrant:


app/src/main/java/org/wikipedia/crash/CrashReportFragment.kt
```diff
@@ -0,0 +1,35 @@
...
+import androidx.fragment.app.Fragment
+import org.wikipedia.R
+
+class CrashReportFragment : Fragment() {
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+        super.onCreateView(inflater, container, savedInstanceState)
+        val view = inflater.inflate(R.layout.fragment_crash_report, container, false)
+
+        view.findViewById<View>(R.id.crash_report_start_over).setOnClickListener {
```
dbrant, line 17: Note: this is a rare case when it's actually simpler to use `findViewById` rather than ViewBindings, because the XML for the crash report fragment uses an `include` which includes a chunk of XML that uses a `merge` node. This would make it a bit more complex to use as a ViewBinding scheme.

cooltey: Makes sense to me.

PR #2108, opened 2021-02-19
Kotlin Conversion: static data

sharvaniharan:


app/src/main/java/org/wikipedia/staticdata/FileAliasData.kt
```diff
@@ -0,0 +1,333 @@
+/* THIS IS AN AUTOMATICALLY GENERATED FILE
+   IF YOU MODIFY THIS BY HAND BE PREPARED!
+   TO HAVE YOUR CHANGES OVERWRITTEN */
```
dbrant, line 3: ^^ Please take note of the above comment. It's trying to tell us something.

sharvaniharan: I did see that @dbrant but i wanted to discuss with you and @cooltey  to see if we can change the scripts/make-templates.py to put it into a .kt file instead of .java so just converted so we could discuss more on Monday.

PR #2110, opened 2021-02-19
Use view binding on talk pages

cooltey:


app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
```diff
@@ -57,55 +58,56 @@ class TalkTopicActivity : BaseActivity(), LinkPreviewDialog.Callback {
 
     public override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        setContentView(R.layout.activity_talk_topic)
-        setSupportActionBar(replyToolbar)
+        binding = ActivityTalkTopicBinding.inflate(LayoutInflater.from(this))
```
dbrant, line 61: Can we just use the Activity's `getLayoutInflater()` function?

app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
```diff
@@ -50,27 +48,28 @@ class TalkTopicsActivity : BaseActivity() {
 
     public override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        setContentView(R.layout.activity_talk_topics)
+        binding = ActivityTalkTopicsBinding.inflate(LayoutInflater.from(this))
```
dbrant, line 51: same: `getLayoutInflater`

dbrant (changes requested):

app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
cooltey: done,

app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
cooltey: Done.

PR #2111, opened 2021-02-20
Use view binding on SuggestedEditsCardsItemFragment.kt

cooltey:


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsItemFragment.kt
```diff
@@ -28,38 +27,41 @@ import org.wikipedia.util.log.L
...
     var sourceSummaryForEdit: PageSummaryForEdit? = null
     var targetSummaryForEdit: PageSummaryForEdit? = null
     var addedContribution: String = ""
         internal set
 
-    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         super.onCreateView(inflater, container, savedInstanceState)
-        return inflater.inflate(R.layout.fragment_suggested_edits_cards_item, container, false)
+        _binding = FragmentSuggestedEditsCardsItemBinding.inflate(LayoutInflater.from(context), container, false)
```
dbrant, line 39: Can we just use the `inflater` parameter that is given to us by this function?

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsItemFragment.kt
```diff
@@ -28,38 +27,41 @@ import org.wikipedia.util.log.L
 import org.wikipedia.views.ImageZoomHelper
 
 class SuggestedEditsCardsItemFragment : SuggestedEditsItemFragment() {
+    private var _binding: FragmentSuggestedEditsCardsItemBinding? = null
```
dbrant, line 30: Should we set this to `null` in `onDestroy()`?

dbrant (changes requested):

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsItemFragment.kt
cooltey: Yes and done.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsItemFragment.kt
cooltey: Yes. Done.

PR #2113, opened 2021-02-20
Kotlin Conversion: PrefixSearchResponse

sharvaniharan:


app/src/main/java/org/wikipedia/search/PrefixSearchResponse.kt
```diff
@@ -0,0 +1,29 @@
+package org.wikipedia.search
+
+import com.google.gson.annotations.SerializedName
+import org.wikipedia.dataclient.mwapi.MwQueryResponse
+
+class PrefixSearchResponse : MwQueryResponse() {
+    @SerializedName("searchinfo")
+    private val searchInfo: SearchInfo? = null
+    private val search: Search? = null
```
sharvaniharan, line 9: I feel we should leave the unused response attributes just for completion.. Lmk

app/src/main/java/org/wikipedia/search/PrefixSearchResponse.kt
```diff
@@ -0,0 +1,30 @@
...
+    fun suggestion(): String? {
+        return searchInfo?.suggestion()
+    }
+
+    internal class SearchInfo {
+        private val suggestion: String? = null
+
+        @SerializedName("suggestionsnippet")
+        private val snippet: String? = null
+        fun suggestion(): String? {
```
dbrant, line 20: This does not need a getter function. We can simply make `suggestion` public.

app/src/main/java/org/wikipedia/search/PrefixSearchResponse.kt
dbrant: agree

dbrant (changes requested):

app/src/main/java/org/wikipedia/search/PrefixSearchResponse.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    fun suggestion(): String? {
+        return searchInfo?.suggestion
+    }
+
+    internal class SearchInfo {
+        val suggestion: String? = null
+
+        @SerializedName("suggestionsnippet")
+        private val snippet: String? = null
```
cooltey, line 19: Would it be possible to shift `snippet` and to the top of `val suggestion: String? = null`?

PR #2114, opened 2021-02-20
Kotlin conversion: RecentSearch

sharvaniharan:


app/src/main/java/org/wikipedia/search/RecentSearch.kt
```diff
@@ -0,0 +1,63 @@
+package org.wikipedia.search
+
+import android.os.Parcel
+import android.os.Parcelable
+import java.util.*
+
+class RecentSearch : Parcelable {
```
dbrant, line 7: I'm actually not sure there's any reason for this class to be `Parcelable`.  I think we can just remove everything from this class related to `Parcelable`, which is pretty much everything except the constructor.

dbrant (changes requested):

app/src/main/java/org/wikipedia/search/RecentSearch.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    override fun hashCode(): Int {
+        return text.hashCode()
+    }
+
+    override fun toString(): String {
+        return ("RecentSearch{" +
+                "text=" + text +
+                ", timestamp=" + timestamp.time +
+                '}')
```
cooltey, line 22: Would it be possible to remove the `( )` from the `return`?

And also, the line `22` uses single quotes, would it be possible to change to double quotes?

PR #2122, opened 2021-02-23
Change script to generate Kotlin Static Data files

sharvaniharan:


cooltey (changes requested):
The `jinja` works for me but looks like the code format for the static files should be refined a little bit since there are some warnings and errors.

sharvaniharan:
Thanks for checking @cooltey

PR #2123, opened 2021-02-23
Convert WikipediaLanguages related classes to Kotlin

cooltey:


app/src/main/java/org/wikipedia/settings/languages/WikipediaLanguagesItemView.kt
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    private fun updateBackgroundColor() {
+        setBackgroundColor(if (binding.wikiLanguageCheckbox.isChecked) getThemedColor(context, R.attr.multi_select_background_color) else getThemedColor(context, R.attr.paper_color))
+    }
+
+    fun setContents(langCode: String, languageLocalizedName: String?, position: Int) {
+        this.position = position
+        binding.wikiLanguageOrder.text = (position + 1).toString()
+        binding.wikiLanguageTitle.text = StringUtils.capitalize(languageLocalizedName)
```
dbrant, line 62: Can be replaced with `orEmpty().capitalize(Locale.getDefault)`

dbrant (changes requested):

app/src/main/java/org/wikipedia/settings/languages/WikipediaLanguagesItemView.kt
cooltey: Done. Thanks.

PR #2125, opened 2021-02-23
Convert all remaining things to View Bindings.

dbrant:
This removes all of the remaining usages of synthetics, and switches everything to View Bindings.  After this, we will be ready to remove the `kotlin-extensions` dependency from Gradle.

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/view/CardFooterView.kt
+++ b/app/src/main/java/org/wikipedia/feed/view/CardFooterView.kt
@@ -6,7 +6,6 @@ import android.text.Spanned
 import android.text.style.ImageSpan
 import android.util.AttributeSet
 import android.view.LayoutInflater
-import android.view.View
 import androidx.constraintlayout.widget.ConstraintLayout
 import org.wikipedia.R
 import org.wikipedia.WikipediaApp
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/feed/view/CardFooterView.kt
```diff
@@ -5,23 +5,25 @@ import android.text.SpannableStringBuilder
...
 internal class CardFooterView constructor(context: Context, attrs: AttributeSet? = null) : ConstraintLayout(context, attrs) {
     interface Callback {
         fun onFooterClicked()
     }
 
+    private val binding = ViewCardFooterBinding.inflate(LayoutInflater.from(context), this)
     var callback: Callback? = null
 
     init {
         View.inflate(context, R.layout.view_card_footer, this)
```
cooltey, line 25: Should we remove this line?

cooltey (changes requested):

app/src/main/java/org/wikipedia/feed/view/CardFooterView.kt
dbrant: whoops!

PR #2127, opened 2021-02-23
Kotlin conversion: RecentSearchDatabaseTable

sharvaniharan:


app/src/main/java/org/wikipedia/search/RecentSearchDatabaseTable.kt
```diff
@@ -0,0 +1,44 @@
+package org.wikipedia.search
+
+import android.content.ContentValues
+import android.database.Cursor
+import org.wikipedia.database.DatabaseTable
+import org.wikipedia.database.column.Column
+import org.wikipedia.database.contract.SearchHistoryContract
+
+class RecentSearchDatabaseTable : DatabaseTable<RecentSearch?>(SearchHistoryContract.TABLE, SearchHistoryContract.Query.URI) {
```
dbrant, line 9: Can this be `DatabaseTable<RecentSearch>` (not nullable?)

dbrant (changes requested):

app/src/main/java/org/wikipedia/search/RecentSearchDatabaseTable.java
```diff
@@ -1,62 +0,0 @@
...
-
-import org.wikipedia.database.DatabaseTable;
-import org.wikipedia.database.column.Column;
-import org.wikipedia.database.contract.SearchHistoryContract;
-import org.wikipedia.database.contract.SearchHistoryContract.Col;
-
-import java.util.Date;
-
-public class RecentSearchDatabaseTable extends DatabaseTable<RecentSearch> {
-    private static final int DB_VER_INTRODUCED = 5;
```
dbrant, line 16: @sharvaniharan  This constant was `5`, but you changed it to `0` in the converted code. Why?

sharvaniharan: @dbrant did not change this... just did not verify after the auto convert 😱

dbrant: I went back to the code before this change and tried the auto-conversion to Kotlin.  It does not change the 5 to a 0.

PR #2131, opened 2021-02-24
Convert PreferenceLoader related classes to Kotlin

cooltey:


app/src/main/java/org/wikipedia/settings/DeveloperSettingsActivity.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.wikipedia.activity.SingleFragmentActivity
+
+class DeveloperSettingsActivity : SingleFragmentActivity<DeveloperSettingsFragment>() {
+    public override fun createFragment(): DeveloperSettingsFragment {
+        return DeveloperSettingsFragment.newInstance()
+    }
+
+    companion object {
+        @JvmStatic
+        fun newIntent(context: Context?): Intent {
```
dbrant, line 14: Any reason for `context` to be nullable?

app/src/main/java/org/wikipedia/settings/DeveloperSettingsPreferenceLoader.kt
```diff
@@ -0,0 +1,207 @@
...
+        true
+    }
+
+    override fun loadPreferences() {
+        loadPreferences(R.xml.developer_preferences)
+        setUpMediaWikiSettings()
+        findPreference(R.string.preferences_developer_crash_key).onPreferenceClickListener = Preference.OnPreferenceClickListener { throw TestException("User tested crash functionality.") }
+        findPreference(R.string.preference_key_add_articles).onPreferenceChangeListener = Preference.OnPreferenceChangeListener { _: Preference, newValue: Any ->
+            if (!isEmptyOrZero(newValue)) {
+                createTestReadingList(TEXT_OF_TEST_READING_LIST, 1, StringUtil.trim(newValue).toInt())
```
dbrant, line 44: I'm not sure I see why we need our own `trim()` function. (Why can't we use `.toString().trim()`?)

dbrant (changes requested):

app/src/main/java/org/wikipedia/settings/DeveloperSettingsActivity.kt
cooltey: No. Done.

app/src/main/java/org/wikipedia/settings/DeveloperSettingsPreferenceLoader.kt
cooltey: Make sense. Done.

PR #2134, opened 2021-02-24
Kotlin Conversion: SearchResult

sharvaniharan:


app/src/main/java/org/wikipedia/search/SearchResult.kt
```diff
@@ -0,0 +1,46 @@
+package org.wikipedia.search
+
+import org.wikipedia.dataclient.WikiSite
+import org.wikipedia.dataclient.mwapi.MwQueryPage
+import org.wikipedia.model.BaseModel
+import org.wikipedia.page.PageTitle
+
+class SearchResult : BaseModel {
```
dbrant, line 8: I see you removed the inheritance from `Parcelable`, which I suppose is good, but can we also remove the other weird bits like `hashCode()`, `equals()`, and `toString()`?

app/src/main/java/org/wikipedia/search/SearchResult.kt
```diff
@@ -0,0 +1,46 @@
...
+    val pageTitle: PageTitle
+    val redirectFrom: String?
+    val type: SearchResultType?
```
dbrant, line 11: This whole thing can be turned into a `data class`, with these three fields in the primary constructor.  Also, `type` should not be nullable.

dbrant (changes requested):

PR #2139, opened 2021-02-25
Remove BaseModel class.

dbrant:
Why did we have it? The world will never know.


app/src/main/java/org/wikipedia/feed/FeedCoordinatorBase.java
```diff
@@ -199,7 +199,7 @@ private Card getLastCard() {
 
     private void requestProgressCard() {
         if (!(getLastCard() instanceof ProgressCard)) {
-            appendCard(new ProgressCard());
+            appendCard(progressCard);
```
dbrant, line 202: note: this was actually a bug that was exposed by removing `BaseModel`.

dbrant:
@cooltey @sharvaniharan this is now ready for review.

cooltey (approved):
Looks good to me. @sharvaniharan Please review and merge if it looks good to you too.

PR #2145, opened 2021-02-26
Convert Captcha-related classes to kotlin.

dbrant:
Notice that this also removes some of the related model classes from being `Parcelable`, and no longer saves them in the activity's `savedInstanceState`. This should no longer be an issue, since this is a bad practice anyway, and will make refactoring simpler in the future.


app/src/main/java/org/wikipedia/captcha/CaptchaHandler.kt
```diff
@@ -0,0 +1,100 @@
...
+    fun captchaWord(): String {
+        return binding.captchaText.editText?.text.toString()
+    }
+
+    fun dispose() {
+        disposables.clear()
+    }
+
+    val isActive: Boolean
+        get() = captchaResult != null
```
cooltey, line 43: Would it be possible to move this variable to the top?

app/src/main/java/org/wikipedia/captcha/CaptchaHandler.kt
```diff
@@ -0,0 +1,100 @@
...
+import org.wikipedia.views.ViewUtil.loadImage
+
+class CaptchaHandler(private val activity: Activity, private val wiki: WikiSite,
+                     captchaView: View, private val primaryView: View,
+                     private val prevTitle: String, submitButtonText: String?) {
+    private val binding: GroupCaptchaBinding = GroupCaptchaBinding.bind(captchaView)
+
+    private val disposables = CompositeDisposable()
+    var token: String? = null
+    private var captchaResult: CaptchaResult? = null
```
cooltey, line 28: Would it be possible to move `var` variable to the bottom of `private var`?

app/src/main/java/org/wikipedia/captcha/CaptchaHandler.kt
```diff
@@ -0,0 +1,100 @@
...
+    fun cancelCaptcha() {
+        if (captchaResult == null) {
+            return
+        }
+        captchaResult = null
+        binding.captchaText.editText?.setText("")
+        hideCaptcha()
+    }
+
+    init {
```
cooltey, line 78: Can we have `init{}` to the top for consistency?

app/src/main/java/org/wikipedia/edit/EditAbuseFilterResult.kt
```diff
@@ -0,0 +1,31 @@
...
+    // This case is here because, unfortunately, an admin can create an abuse filter which
+    // emits an arbitrary error code over the API.
+    // TODO: More properly handle the case where the AbuseFilter throws an arbitrary error.
+    // Oh, and, you know, also fix the AbuseFilter API to not throw arbitrary error codes.
+    val type: Int
+        get() = if (code != null && code.startsWith("abusefilter-warning")) {
+            TYPE_WARNING
+        } else if (code != null && code.startsWith("abusefilter-disallowed")) {
+            TYPE_ERROR
+        } else if (info != null && info.startsWith("Hit AbuseFilter")) {
```
cooltey, line 15: Would it be good if we can use `code.orEmpty().startWith()` on these null check type of if statement?

cooltey (changes requested):

cooltey:
@dbrant Please resolve conflicts.

app/src/main/java/org/wikipedia/captcha/CaptchaHandler.kt
dbrant: yep

app/src/main/java/org/wikipedia/edit/EditAbuseFilterResult.kt
dbrant: sure

PR #2148, opened 2021-02-26
Convert Namespace to kotlin.

dbrant:
This is actually pretty important because it fixes a crash:
https://appcenter.ms/orgs/ci-apps-hockeyapp-f5mf/apps/Wikipedia/crashes/errors/2418972845u/overview


app/src/main/java/org/wikipedia/page/Namespace.kt
```diff
@@ -0,0 +1,223 @@
...
+        return this === FILE
+    }
+
+    open fun talk(): Boolean {
+        return code and TALK_MASK == TALK_MASK
+    }
+
+    companion object {
+        @JvmStatic
+        val CODE_ENUM: CodeEnum<Namespace> = object : CodeEnum<Namespace> {
```
cooltey, line 186: Very minor but we can add `fun` to the start of the `CodeEnum` interface to enable the `lambda`.

```
fun interface CodeEnum<T> {
    fun enumeration(code: Int): T
}
```

dbrant: nice!

PR #2153, opened 2021-02-27
Convert Prefs related classes to Kotlin

cooltey:


cooltey:
Close and hold until we have completed the rest of Kotlin conversion.

PR #2156, opened 2021-03-02
Show message when no languages are selected on Watchlist page

cooltey:
https://phabricator.wikimedia.org/T275493


cooltey:
@sharvaniharan The PR should be held instead of merging it 😅

PR #2159, opened 2021-03-02
Fix: Link preview navigation

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T274012
If an existing tab is found with the same title, we should only animate the taab count but not navigate to that tab.


cooltey:
Does this PR relate to any Phab ticket? Not sure if it is expected behavior or not .

sharvaniharan:
Updates it @cooltey

cooltey (approved):
Looks good to me. I'll let @dbrant review it as well.

PR #2160, opened 2021-03-03
Convert SiteInfo related classes to Kotlin

cooltey:


app/src/main/java/org/wikipedia/settings/SiteInfoClient.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.wikipedia.staticdata.MainPageNameData
+import org.wikipedia.util.log.L
+
+object SiteInfoClient {
+    private val SITE_INFO_MAP = mutableMapOf<String, SiteInfo?>()
+
+    @JvmStatic
+    fun getMainPageForLang(lang: String): String {
+        val info = getSiteInfoForLang(lang)
+        return if (info != null && !info.mainpage.isNullOrEmpty()) {
```
sharvaniharan, line 19: Can we remove the null check like and us info?.mainpage... instead?

sharvaniharan (changes requested):

app/src/main/java/org/wikipedia/settings/SiteInfoClient.kt
cooltey: The null check is actually necessary since we are going to use its `mainpage`. But I rewrite the code a bit.

PR #2161, opened 2021-03-03
Kotlin remote config

cooltey:


app/src/main/java/org/wikipedia/settings/RemoteConfig.kt
```diff
@@ -0,0 +1,31 @@
+package org.wikipedia.settings
+
+import org.json.JSONException
+import org.json.JSONObject
+
+/**
+ * Store for config values that are retrieved from a server,
+ * and refreshed periodically.
+ */
```
sharvaniharan, line 9: Let us get rid of such not-so-helpful comments since we have adapted self-commenting code.

app/src/main/java/org/wikipedia/settings/RemoteConfig.kt
```diff
@@ -0,0 +1,31 @@
...
+        Prefs.setRemoteConfigJson(newConfig.toString())
+        curConfig = newConfig
+    }
+
+    // If there's no pref set, just give back the empty JSON Object
+    val config: JSONObject
+        get() {
+            if (curConfig == null) {
+                curConfig = try {
+                    // If there's no pref set, just give back the empty JSON Object
```
sharvaniharan, line 23: Such ones which are helpful should stay..

sharvaniharan (changes requested):

app/src/main/java/org/wikipedia/settings/RemoteConfig.kt
cooltey: Done.

PR #2167, opened 2021-03-04
Fix: Oversight of db version conversion fixed

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T276484


dbrant (changes requested):
This will not fix the crash for users who already have the app installed (with the inconsistent db structure).

sharvaniharan:
Done @dbrant

PR #2168, opened 2021-03-04
Follow-up to fixing RecentSearch table.

dbrant:


app/src/main/java/org/wikipedia/search/RecentSearchDatabaseTable.kt
```diff
@@ -40,11 +40,13 @@ class RecentSearchDatabaseTable : DatabaseTable<RecentSearch>(SearchHistoryContr
     }
 
     override fun onUpgradeSchema(db: SQLiteDatabase, fromVersion: Int, toVersion: Int) {
-        db.execSQL("CREATE TABLE IF NOT EXISTS " + SearchHistoryContract.TABLE +
-                " ( " + getColumnsAdded(dBVersionIntroducedAt).joinToString(", ") + " )")
+        if (toVersion == DB_VER_FIXED_FROM_CORRUPTION) {
```
dbrant, line 43: We don't want to attempt recreating the table on *every* upgrade, just the one that rescues it from corruption.

app/src/main/java/org/wikipedia/database/DatabaseTable.kt
```diff
@@ -85,10 +85,11 @@ abstract class DatabaseTable<T>(private val tableName: String, val baseContentUR
     }
 
     protected open fun onUpgradeSchema(db: SQLiteDatabase, fromVersion: Int, toVersion: Int) {}
-    private fun createTables(db: SQLiteDatabase) {
+
+    protected fun createTables(db: SQLiteDatabase) {
         L.i("Creating table=$tableName")
         val cols = getColumnsAdded(dBVersionIntroducedAt)
-        db.execSQL("CREATE TABLE " + tableName + " ( " + cols.joinToString(", ") + " )")
+        db.execSQL("CREATE TABLE IF NOT EXISTS " + tableName + " ( " + cols.joinToString(", ") + " )")
```
dbrant, line 92: I believe it's totally safe to add `IF NOT EXISTS` here. This also makes it symmetrical with `dropTable` which uses `IF EXISTS`.

PR #2169, opened 2021-03-04
Fix dismissability issues with tooltips that have custom View.

dbrant:


app/src/main/res/layout/view_watchlist_main_tooltip.xml
```diff
@@ -5,6 +5,7 @@
     android:orientation="vertical"
     android:paddingStart="16dp"
     android:paddingTop="16dp"
+    android:paddingBottom="16dp"
```
cooltey, line 8: Adding `16dp` looks will show too much space on the bottom (on Samsung S9). Would it be possible to change it to `8dp`?

dbrant: Here's how it looks on my Pixel 2:

image

Are we sure the Balloon updates are ready for prime time?

cooltey: I see what's happening there. 
The library uses the `width` instead of `height` to adjust the tooltip's height.

My Pixel 3 XL also has a similar issue you have on Pixel 2 (with `android:paddingBottom="16dp"` changes)

It would be better to change `android:padding='16dp'` for a proper fix.

cooltey: Should we add the `ScrollView` back to the tooltip custom views? I will report this issue to the author. 🆘

dbrant:
@cooltey   I've removed the layout changes from this PR, so this can just be fixing the logic of dismissing.  After this, we can continue to tweak the layouts separately, or pull in more updates from the balloon library.

PR #2173, opened 2021-03-05
Translation: Image recommendation strings

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T275541


app/src/main/res/values-qq/strings.xml
```diff
@@ -1009,4 +1010,32 @@
   <string name="watchlist_page_moved">Label text to indicate that a page has been moved.</string>
   <string name="watchlist_page_protected">Label text to indicate that a page has been protected.</string>
   <string name="watchlist_page_deleted">Label text to indicate that a page has been deleted.</string>
+  <string name="image_recommendations_task_title">Title shown at the top of the Image Recommendations task.</string>
```
sharvaniharan, line 1013: @dbrant Did you skip the `'suggested_edits_"` prefix for the string names because these are not really editing suggestions? otherwise they should be in the `<!-- Suggested edits --> `section with the prefix

dbrant: done

PR #2174, opened 2021-03-05
Kotlin Conversion: PageImage

sharvaniharan:


app/src/main/java/org/wikipedia/pageimages/PageImage.kt
```diff
@@ -0,0 +1,42 @@
+package org.wikipedia.pageimages
+
+import org.wikipedia.dataclient.WikiSite
+import org.wikipedia.dataclient.mwapi.MwQueryPage
+import org.wikipedia.page.PageTitle
+
+class PageImage(val title: PageTitle, val imageName: String) {
```
cooltey, line 7: Not sure but would it be better if it is a `data class`?

cooltey: @sharvaniharan Any idea about the comment above?

app/src/test/java/org/wikipedia/test/ParcelableTest.java
```diff
@@ -35,12 +34,4 @@
...
         TestParcelUtil.test(historyEntry);
     }
-
-    @Test public void testPageImage() throws Throwable {
-        WikiSite wiki = WikiSite.forLanguageCode("en");
-        PageTitle title = new PageTitle("Talk", "India", wiki);
-        PageImage pageImage = new PageImage(title, "Testing image");
-
-        TestParcelUtil.test(pageImage);
-    }
```
cooltey, line 45: If the previous `PageImage` is `Parcelable`, should we add `@Parcelize` to the `PageImage`?

sharvaniharan: @cooltey I will need to add 'kotlin-android-extensions' plugin to do that... are we ready right now or should we do it when we have a real need for Parcelizing..

dbrant: (it should be the `kotlin-parcelize` plugin, since `kotlin-android-extensions` is deprecated)

sharvaniharan: @dbrant Should I add it now then?

app/src/main/java/org/wikipedia/pageimages/PageImage.kt
```diff
@@ -0,0 +1,45 @@
+package org.wikipedia.pageimages
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+import org.wikipedia.dataclient.WikiSite
+import org.wikipedia.dataclient.mwapi.MwQueryPage
+import org.wikipedia.page.PageTitle
+
+@Parcelize
+data class PageImage(val title: PageTitle, val imageName: String) : Parcelable {
```
cooltey, line 10: Looks like these two parameters can be `nullable`.

app/src/main/java/org/wikipedia/pageimages/PageImage.kt
```diff
@@ -0,0 +1,45 @@
...
+    companion object {
+        @JvmField
+        val DATABASE_TABLE = PageImageDatabaseTable()
+
+        @JvmStatic
+        fun imageMapFromPages(wiki: WikiSite, titles: MutableList<PageTitle>, pages: MutableList<MwQueryPage>): Map<PageTitle?, PageImage> {
+            val pageImagesMap = mutableMapOf<PageTitle?, PageImage>()
+            // nominal case
+            val titlesMap = mutableMapOf<String, PageTitle>()
+            titles.forEach { title ->
```
cooltey, line 21: `title` can be removed.

app/src/main/java/org/wikipedia/pageimages/PageImage.kt
```diff
@@ -0,0 +1,45 @@
...
+
+@Parcelize
+data class PageImage(val title: PageTitle, val imageName: String) : Parcelable {
+
+    companion object {
+        @JvmField
+        val DATABASE_TABLE = PageImageDatabaseTable()
+
+        @JvmStatic
+        fun imageMapFromPages(wiki: WikiSite, titles: MutableList<PageTitle>, pages: MutableList<MwQueryPage>): Map<PageTitle?, PageImage> {
```
cooltey, line 17: Looks like the `Map<PageTitle?, PageImage>` can be definied as `Map<PageTitle, PageImage>` with an non-null `PageTitle`.

app/src/main/java/org/wikipedia/pageimages/PageImage.kt
```diff
@@ -0,0 +1,45 @@
...
+                if (!it.redirectFrom().isNullOrEmpty()) {
+                    thumbnailSourcesMap[PageTitle(null, it.redirectFrom()!!, wiki).prefixedText] = it.thumbUrl()
+                }
+            }
+
+            titlesMap.keys.forEach { key ->
+                if (thumbnailSourcesMap.containsKey(key)) {
+                    pageImagesMap[titlesMap[key]] = PageImage(titlesMap[key]!!, thumbnailSourcesMap[key]!!)
+                }
+            }
```
cooltey, line 41: Would it be possible to as the following?
```
            titlesMap.forEach {
                if (thumbnailSourcesMap.containsKey(it.key)) {
                    pageImagesMap[it.value] = PageImage(it.value, thumbnailSourcesMap[it.key])
                }
            }
```

cooltey (changes requested):

PR #2181, opened 2021-03-09
Convert ReadingListFragment to Kotlin

cooltey:


app/src/main/java/org/wikipedia/readinglist/ReadingListFragment.kt
```diff
@@ -0,0 +1,826 @@
...
+        if (selectedPages.isNotEmpty()) {
+            val titles = selectedPages.map { ReadingListPage.toPageTitle(it) }
+            bottomSheetPresenter.show(childFragmentManager,
+                    AddToReadingListDialog.newInstance(titles, InvokeSource.READING_LIST_ACTIVITY))
+            update()
+        }
+    }
+
+    private fun moveSelectedPagesToList() {
+        if (selectedPages.isNotEmpty()) {
```
dbrant, line 395: @cooltey  This is another subtle bug introduced by Kotlin conversion.  Notice that the getter for `selectedPages` actually modifies the selected state of the underlying pages. Therefore, any subsequent calls to `selectedPages` will return different results!  The proper way is to store the value of `selectedPages` once, and operate only on the returned list.

cooltey: 🤦.  Thanks for checking and fixing it!

PR #2184, opened 2021-03-10
Convert CreateAccountActivity to Kotlin

cooltey:


app/src/main/res/layout/group_captcha.xml
```diff
@@ -2,7 +2,6 @@
 
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools"
-    android:id="@+id/captcha_container"
```
dbrant, line 5: @cooltey  When making a change like this, please check if there are other components that use this layout, such as Editing.

cooltey: Sure, I will. Thanks for checking this up.

PR #2186, opened 2021-03-10
Load non-main namespace pages as mobile web in our own WebView.

dbrant:


sharvaniharan:
@dbrant The info boxes are collapsed which is fine [ maybe Robin can weigh-in if it would make sense to un-collapse them on load, only for special pages, since sometimes there is nothing else in that page and can be disorienting] 

But what I was mentioning on #2057 was:
If you look at this page https://en.wikipedia.org/wiki/Category:Companies and compare how it loads on web and on phone, 2 things are different... a title would be helpful, and, in the bottom of the page, it shows the sub-categories of the category 'Companies'. So was wondering if it would be easy to include those, or at least the title, or even worth creating a ticket for these two?

dbrant:
@sharvaniharan  Still not sure what you mean:  Here is how the page looks in mobile web (notice the `m.` domain):
https://en.m.wikipedia.org/wiki/Category:Companies

And here is how the same page looks in the app with this PR:
image

The page renders exactly the same as in mobile web (since that what we're doing in our WebView), including the title, and the subcategories.  Are you seeing something different when testing this PR?

sharvaniharan:
You are correct @dbrant My bad... I might have checked out the wrong branch... Looks good! everything checks out.

dbrant:
self-merging, since it passed design review.

PR #2188, opened 2021-03-11
Convert MainActivity and MainFragment to kotlin.

dbrant:


app/src/main/java/org/wikipedia/main/MainActivity.kt
```diff
@@ -0,0 +1,189 @@
...
+        setNavigationBarColor(getThemedColor(this, R.attr.nav_tab_background_color))
+        setSupportActionBar(binding.mainToolbar)
+        supportActionBar?.title = ""
+        supportActionBar?.setDisplayHomeAsUpEnabled(false)
+        binding.mainToolbar.navigationIcon = null
+    }
+
+    override fun onResume() {
+        super.onResume()
+        supportInvalidateOptionsMenu()
```
cooltey, line 65: Since it is deprecated, can we switch it to `invalidateOptionsMenu`?

app/src/main/java/org/wikipedia/main/MainActivity.kt
```diff
@@ -0,0 +1,189 @@
...
+            binding.mainToolbarWordmark.visibility = View.GONE
+            binding.mainToolbar.setTitle(tab.text())
+            controlNavTabInFragment = true
+        }
+        fragment.requestUpdateToolbarElevation()
+    }
+
+    override fun updateTabCountsView() {
+        showTabCountsAnimation = true
+        supportInvalidateOptionsMenu()
```
cooltey, line 123: Should switch to `invalidateOptionsMenu()`?

app/src/main/java/org/wikipedia/main/MainActivity.kt
```diff
@@ -0,0 +1,189 @@
...
+import org.wikipedia.onboarding.InitialOnboardingActivity
+import org.wikipedia.page.PageActivity
+import org.wikipedia.page.tabs.TabActivity
+import org.wikipedia.settings.Prefs
+import org.wikipedia.suggestededits.SuggestedEditsTasksFragment
+import org.wikipedia.util.DimenUtil.dpToPx
+import org.wikipedia.util.DimenUtil.getDimension
+import org.wikipedia.util.FeedbackUtil.setButtonLongPressToast
+import org.wikipedia.util.FeedbackUtil.showTooltip
+import org.wikipedia.util.ResourceUtil.getThemedColor
```
cooltey, line 27: Should we either use `DimenUtil.dpToPx()` or directly use `dpToPx()` kind of methods? I usually use the first one, which makes more sense to me.

cooltey (changes requested):

app/src/main/java/org/wikipedia/main/MainActivity.kt
dbrant: correct

app/src/main/java/org/wikipedia/main/MainActivity.kt
dbrant: Yes, this is my preference too.  For some reason the automatic Kotlin converter prefers to import individual methods, which are hard to catch and clean up.  (It does this for `L.e()`, transforming it to `e()` which is super annoying)

PR #2195, opened 2021-03-12
Expand/collapse tables for find-in-page feature in PageFragment

cooltey:
https://phabricator.wikimedia.org/T130403


cooltey:
Created another patch for fully expanding/collapsing tables, now this patch is waiting for it to be on production.

cooltey:
The PR is ready for review. Please note that the `mobile-html` should be fully refreshed and you may need to clear the cache/data if you are not able to see the tables are expanded when tapping on "find-in-page" button.

app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -919,43 +919,50 @@ void showFindInPage() {
...
+                public void onDestroyActionMode(ActionMode mode) {
+                    if (webView == null || !isAdded()) {
+                        return;
+                    }
+                    funnel.setPageHeight(webView.getContentHeight());
+                    funnel.logDone();
+                    webView.clearMatches();
+                    hideSoftKeyboard();
+                    setToolbarElevationEnabled(true);
+                    bridge.execute(JavaScriptActionHandler.expandCollapsedTables(false));
```
dbrant, line 963: I'm not sure we need to re-close the tables when this is done.
Suppose I open the find-in-page box, and it finds the text I'm looking for _inside_ a table. Then I exit out of find-in-page, but the table suddenly closes.

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageFragment.java
cooltey: Makes sense to me. Removed it.

PR #2199, opened 2021-03-15
Kotlin Conversion: PageImageDatabaseTable

sharvaniharan:


app/src/main/java/org/wikipedia/database/DatabaseTable.kt
```diff
@@ -98,6 +98,6 @@ abstract class DatabaseTable<T>(private val tableName: String, val baseContentUR
     }
 
     companion object {
-        protected const val INITIAL_DB_VERSION = 1
```
sharvaniharan, line 101: Not sure why I had to take out this protection... since PageImageDatabaseTable is inheriting from DzatabaseTable..

app/src/main/java/org/wikipedia/pageimages/PageImageDatabaseTable.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    companion object {
+        private const val DB_VER_NAMESPACE_ADDED = 7
+        private const val DB_VER_LANG_ADDED = 10
+        private const val DB_VER_DISPLAY_TITLE_ADDED = 19
+    }
+
+    override val dBVersionIntroducedAt: Int
+        get() = INITIAL_DB_VERSION
+}
```
cooltey, line 63: Would it be possible to remove the explicit type specification?

app/src/main/java/org/wikipedia/pageimages/PageImageDatabaseTable.kt
```diff
@@ -0,0 +1,63 @@
...
+    override fun getUnfilteredPrimaryKeySelectionArgs(obj: PageImage): Array<String?> {
+        return arrayOf(
+                obj.title.wikiSite.authority(),
+                obj.title.wikiSite.languageCode(),
+                obj.title.namespace,
+                obj.title.text
+        )
+    }
+
+    companion object {
```
cooltey, line 55: The `companion object` should be at the bottom of the file.

app/src/main/java/org/wikipedia/pageimages/PageImageDatabaseTable.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.wikipedia.dataclient.WikiSite
+import org.wikipedia.page.PageTitle
+
+// todo: network caching preserves images. Remove this class and drop table?
+class PageImageDatabaseTable : DatabaseTable<PageImage>(PageImageHistoryContract.TABLE, PageImageHistoryContract.Image.URI) {
+    override fun fromCursor(cursor: Cursor): PageImage {
+        val wiki = WikiSite(PageImageHistoryContract.Col.SITE.`val`(cursor), PageImageHistoryContract.Col.LANG.`val`(cursor))
+        val title = PageTitle(PageImageHistoryContract.Col.NAMESPACE.`val`(cursor), PageImageHistoryContract.Col.API_TITLE.`val`(cursor), wiki)
+        val imageName = PageImageHistoryContract.Col.IMAGE_NAME.`val`(cursor)
+        title.setDisplayText(PageImageHistoryContract.Col.DISPLAY_TITLE.`val`(cursor))
```
cooltey, line 17: Looks like the `val()` is a reserved word that requires special quotes around it. Would it be a great opportunity to also convert `StrColumn.java` to Kotlin and also rename the `val()` method?

cooltey (changes requested):

app/src/main/java/org/wikipedia/pageimages/PageImageDatabaseTable.kt
sharvaniharan: I addressed this in a separate PR #2216  Just so it is easy to review since that name change turned out to be a big chunk of change.. 😅

app/src/main/java/org/wikipedia/pageimages/PageImageDatabaseTable.kt
cooltey: Is this resolved?

app/src/main/java/org/wikipedia/pageimages/PageImageDatabaseTable.kt
cooltey: Is this resolved?

app/src/main/java/org/wikipedia/pageimages/PageImageDatabaseTable.kt
sharvaniharan: I really thought i had pushed those changes 😂  sorry

app/src/main/java/org/wikipedia/pageimages/PageImageDatabaseTable.kt
```diff
@@ -0,0 +1,63 @@
...
+    override fun getPrimaryKeySelection(obj: PageImage, selectionArgs: Array<String>): String {
+        return super.getPrimaryKeySelection(obj, PageImageHistoryContract.Col.SELECTION)
+    }
+
+    override fun getUnfilteredPrimaryKeySelectionArgs(obj: PageImage): Array<String?> {
+        return arrayOf(
+                obj.title.wikiSite.authority(),
+                obj.title.wikiSite.languageCode(),
+                obj.title.namespace,
+                obj.title.text
```
cooltey, line 51: Please fix the errors.

cooltey (changes requested):
Please try to rebuild this PR and looks like there are two classes have errors.

app/src/main/java/org/wikipedia/pageimages/PageImageDatabaseTable.kt
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    override fun getUnfilteredPrimaryKeySelectionArgs(obj: PageImage): Array<String?> {
+        return arrayOf(
+                obj.title!!.wikiSite.authority(),
+                obj.title.wikiSite.languageCode(),
+                obj.title.namespace,
+                obj.title.text
+        )
+    }
```
cooltey, line 53: As I mentioned on Slack, this can be done by the follow code:
```
override fun getUnfilteredPrimaryKeySelectionArgs(obj: PageImage): Array<String?> {
    return obj.title?.run {
        arrayOf(wikiSite.authority(), wikiSite.languageCode(), namespace, text)
    } ?: emptyArray()
}
```

PR #2215, opened 2021-03-17
Fix: add to reading list dialog will not show the onboarding view properly

cooltey:
Steps to reproduce:

1. Clear app data
2. Open the app and add any article to the reading list
3. You cannot see the list view or create a reading list button; which suppose to have an onboarding view and then enter to the list.

Discussion:
@dbrant @sharvaniharan 
Since we have changed the flow of adding an article to the "saved" list by default, looks like the users will never see this onboarding view by any chance. Should we either remove it or keep it?


dbrant:
@cooltey  Remove it 🪓

cooltey:
> 
> 
> @cooltey Remove it 🪓

Done!

PR #2219, opened 2021-03-18
Converted BackPressHandler to Kotlin

gouri-panda:


app/src/main/java/org/wikipedia/BackPressedHandler.kt
```diff
@@ -0,0 +1,5 @@
+package org.wikipedia
+
+interface BackPressedHandler {
+    fun onBackPressed(): Boolean
+}
```
cooltey, line 5: Please add a new line to the end of the file.

PR #2221, opened 2021-03-18
Converted SessionData to Kotlin

gouri-panda:


app/src/main/java/org/wikipedia/analytics/SessionData.kt
```diff
@@ -0,0 +1,70 @@
...
+    fun addPageFromBack() {
+        pagesFromBack++
+    }
+
+    fun addPageWithNoDescription() {
+        pagesWithNoDescription++
+    }
+
+    val totalPages: Int
+        get() = (pagesFromSearch + pagesFromRandom + pagesFromLangLink + pagesFromInternal +
```
cooltey, line 67: We can remove parentheses here.

app/src/main/java/org/wikipedia/analytics/SessionData.kt
```diff
@@ -0,0 +1,70 @@
...
+    var pagesFromHistory = 0
+        private set
+    var pagesFromReadingList = 0
+        private set
+    var pagesFromBack = 0
+        private set
+    var pagesWithNoDescription = 0
+        private set
+    var pagesFromSuggestedEdits = 0
+        private set
```
cooltey, line 29: It should be fine if we remove the `private set` here.

app/src/main/java/org/wikipedia/analytics/SessionData.kt
```diff
@@ -0,0 +1,60 @@
+package org.wikipedia.analytics
+
+import org.wikipedia.history.HistoryEntry
+import org.wikipedia.util.MathUtil.Averaged
+
+class SessionData {
+    private val leadLatency = Averaged<Long>()
```
cooltey, line 7: Please use `MathUtil.Averaged` instead of `Averaged` for overall code style consistency.

PR #2222, opened 2021-03-18
Follow up: LanguagesList Kotlin conversion

cooltey:


sharvaniharan (approved):

app/src/main/java/org/wikipedia/language/LanguagesListActivity.kt
```diff
@@ -130,7 +130,7 @@ class LanguagesListActivity : BaseActivity() {
 
         // To remove the already selected languages and suggested languages from all languages list
         private val nonDuplicateLanguageCodesList
-            get() = originalLanguageCodes.toMutableList().apply {
+            get() = originalLanguageCodes.apply {
```
dbrant, line 133: Nope, this introduces a bug and changes behavior.  The list returned by this getter needs to be a _copy_ of the original list. That's the purpose of the toMutableList() call.  Without making a copy, you are modifying the original list, which will cause inconsistent behavior elsewhere.

dbrant (changes requested):

app/src/main/java/org/wikipedia/language/LanguagesListActivity.kt
cooltey: I see, totally didn't notice that. Done.

PR #2223, opened 2021-03-18
Converted PreferenceMultiLine to Kotlin

gouri-panda:


app/src/main/java/org/wikipedia/settings/PreferenceMultiLine.kt
```diff
@@ -0,0 +1,41 @@
...
+import android.content.Context
+import android.util.AttributeSet
+import android.widget.TextView
+import android.widget.Toast
+import androidx.preference.Preference
+import androidx.preference.Preference.OnPreferenceClickListener
+import androidx.preference.PreferenceViewHolder
+import org.wikipedia.R
+
+class PreferenceMultiLine : Preference {
```
cooltey, line 13: We can probably add `@Suppress("unused")` to reduce the confusion.

app/src/main/java/org/wikipedia/settings/PreferenceMultiLine.kt
```diff
@@ -0,0 +1,41 @@
...
+    constructor(ctx: Context?, attrs: AttributeSet?, defStyle: Int) : super(ctx, attrs, defStyle)
+    constructor(ctx: Context?, attrs: AttributeSet?) : super(ctx, attrs)
+    constructor(ctx: Context?) : super(ctx)
+
+    override fun onBindViewHolder(holder: PreferenceViewHolder) {
+        super.onBindViewHolder(holder)
+        holder.itemView.findViewById<TextView>(android.R.id.title)?.also {
```
cooltey, line 20: Since we are not going to use the return value of the view, would it be better if we use `?.let {}` here?

app/src/main/java/org/wikipedia/settings/PreferenceMultiLine.kt
```diff
@@ -0,0 +1,41 @@
...
+            it.isSingleLine = false
+        }
+
+        // Intercept the click listener for this preference, and if the preference has an intent,
+        // launch the intent ourselves, so that we can catch the exception if the intent fails.
+        // (but only do this if the preference doesn't already have a click listener)
+        if (this.onPreferenceClickListener == null) {
+            this.onPreferenceClickListener = OnPreferenceClickListener { preference ->
```
cooltey, line 28: We can remove `this.`

cooltey (changes requested):

app/src/main/java/org/wikipedia/settings/PreferenceMultiLine.kt
```diff
@@ -0,0 +1,42 @@
...
+class PreferenceMultiLine : Preference {
+    constructor(ctx: Context, attrs: AttributeSet?, defStyle: Int) : super(ctx, attrs, defStyle)
+    constructor(ctx: Context, attrs: AttributeSet?) : super(ctx, attrs)
+    constructor(ctx: Context) : super(ctx)
+
+    override fun onBindViewHolder(holder: PreferenceViewHolder) {
+        super.onBindViewHolder(holder)
+        holder.itemView.findViewById<TextView>(android.R.id.title)?.let {
+            it.isSingleLine = false
+        }
```
cooltey, line 23: We should probably do the same thing in #2224

PR #2224, opened 2021-03-18
Converted SwitchPreferenceMultiLine to Kotlin

gouri-panda:


app/src/main/java/org/wikipedia/settings/SwitchPreferenceMultiLine.kt
```diff
@@ -0,0 +1,21 @@
...
+
+open class SwitchPreferenceMultiLine : SwitchPreferenceCompat {
+    constructor(ctx: Context?, attrs: AttributeSet?, defStyle: Int) : super(ctx, attrs, defStyle)
+    constructor(ctx: Context?, attrs: AttributeSet?) : super(ctx, attrs)
+    constructor(ctx: Context?) : super(ctx)
+
+    override fun onBindViewHolder(holder: PreferenceViewHolder) {
+        super.onBindViewHolder(holder)
+        holder.itemView.findViewById<TextView>(R.id.title)?.also { textView ->
+            textView.isSingleLine = false
```
dbrant, line 18: Couldn't this just be:
`holder.itemView.findViewById<TextView>(R.id.title).isSingleLine = true`


Also, the identifier `R.id.title` should be fully qualified. (Don't import `android.R`)

dbrant (changes requested):

PR #2226, opened 2021-03-18
Follow up: HistoryFragment Kotlin conversion

cooltey:


app/src/main/java/org/wikipedia/history/HistoryFragment.kt
```diff
@@ -277,8 +276,16 @@ class HistoryFragment : Fragment(), BackPressedHandler {
             val voiceSearchButton = itemView.findViewById<View>(R.id.voice_search_button)
             historyFilterButton = itemView.findViewById(R.id.history_filter)
             clearHistoryButton = itemView.findViewById(R.id.history_delete)
-            searchCardView.setOnClickListener { view -> (requireParentFragment() as MainFragment).openSearchActivity(Constants.InvokeSource.NAV_MENU, null, view) }
-            voiceSearchButton.setOnClickListener { (requireParentFragment() as MainFragment).onFeedVoiceSearchRequested() }
+            searchCardView.setOnClickListener {
+                if (isAdded) {
```
dbrant, line 280: Are you sure we need all these new `isAdded` checks?  Is there any way for these listeners to run outside the lifecycle of the fragment?

dbrant (changes requested):

app/src/main/java/org/wikipedia/history/HistoryFragment.kt
cooltey: I can remove these but I think I saw some crashes happened in the SearchFragment and maybe that's a different issue.

PR #2228, opened 2021-03-19
Convert onboarding activity and fragment to kotlin.

dbrant:


app/src/main/java/org/wikipedia/onboarding/OnboardingFragment.kt
```diff
@@ -0,0 +1,123 @@
...
+import org.wikipedia.databinding.FragmentOnboardingPagerBinding
+
+abstract class OnboardingFragment : Fragment(), BackPressedHandler {
+    interface Callback {
+        fun onComplete()
+    }
+
+    private var _binding: FragmentOnboardingPagerBinding? = null
+    private val binding get() = _binding!!
+    private val pageChangeCallback: PageChangeCallback = PageChangeCallback()
```
cooltey, line 24: Explicit type specification can be removed.

app/src/main/java/org/wikipedia/onboarding/OnboardingFragment.kt
```diff
@@ -0,0 +1,123 @@
...
+            advancePage()
+        }
+    }
+
+    protected abstract fun getAdapter(): FragmentStateAdapter?
+
+    @get:StringRes
+    protected abstract val doneButtonText: Int
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
```
cooltey, line 39: Should we change the `View?` to `View`?

app/src/main/java/org/wikipedia/onboarding/InitialOnboardingFragment.kt
```diff
@@ -0,0 +1,136 @@
...
+        private val callback
+            get() = FragmentUtil.getCallback(this, OnboardingPageView.Callback::class.java)
+
+        companion object {
+            fun newInstance(position: Int): ItemFragment {
+                val instance = ItemFragment()
+                val args = Bundle()
+                args.putInt("position", position)
+                instance.arguments = args
+                return instance
```
cooltey, line 104: Would it be better if we use:

```
return ItemFragment().apply { ... }
```

app/src/main/java/org/wikipedia/onboarding/InitialOnboardingFragment.kt
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+    }
+
+    internal enum class OnboardingPage(@LayoutRes val layout: Int) : EnumCode {
```
cooltey, line 109: Since the items are not really unused, would it be better if we add `@Suppress("unused")` to reduce confusion?

cooltey (changes requested):

PR #2237, opened 2021-03-22
Converted MwPostResponse  to kotlin

gouri-panda:


app/src/main/java/org/wikipedia/dataclient/mwapi/MwPostResponse.kt
```diff
@@ -0,0 +1,10 @@
+package org.wikipedia.dataclient.mwapi
+
+open class MwPostResponse : MwResponse() {
+    val pageInfo: MwQueryPage? = null
+    val options: String? = null
+    val successVal = 0
```
dbrant, line 6: You must never change the names of any fields in model classes, since these are deserialized from our JSON API, and must therefore match the server's response.

app/src/main/java/org/wikipedia/dataclient/mwapi/MwPostResponse.kt
```diff
@@ -0,0 +1,10 @@
...
+    fun success(result: String?): Boolean {
```
dbrant, line 7: Newline between fields and functions.

dbrant (changes requested):

dbrant:
These don't really make sense to be converted yet anyway. There's no benefit in terms of performance or dependencies. I would rather convert these things when we start adopting `kotlinx.serialization`.

PR #2238, opened 2021-03-22
Converted Edit.java to  kotlin

gouri-panda:


app/src/main/java/org/wikipedia/edit/Edit.kt
```diff
@@ -0,0 +1,74 @@
+package org.wikipedia.edit
+
+import org.wikipedia.dataclient.mwapi.MwPostResponse
+
+class Edit : MwPostResponse() {
+    private val edit: Result? = null
+    fun edit(): Result? {
```
dbrant, line 7: Please put a newline between field declarations and functions.

app/src/main/java/org/wikipedia/edit/Edit.kt
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    class Result {
+        private val result: String? = null
+        private val newrevid: Long = 0
+        private val captcha: Captcha? = null
+        private val code: String? = null
+        private val info: String? = null
+        private val warning: String? = null
+        private val spamblacklist: String? = null
```
dbrant, line 22: Please put a newline between field declarations and functions.

app/src/main/java/org/wikipedia/edit/Edit.kt
```diff
@@ -0,0 +1,74 @@
...
+        fun status(): String? {
```
dbrant, line 23: Most of these getter functions no longer need to exist. (kotlin automatically generates getter functions for public fields)

app/src/main/java/org/wikipedia/edit/Edit.kt
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        fun hasSpamBlacklistResponse(): Boolean {
+            return spamblacklist != null
+        }
+    }
+
+    private class Captcha {
+        private val id: String? = null
+        fun id(): String? {
```
dbrant, line 70: Also an unnecessary getter function.

dbrant (changes requested):

dbrant:
These don't really make sense to be converted yet anyway. There's no benefit in terms of performance or dependencies. I would rather convert these things when we start adopting `kotlinx.serialization`.

PR #2245, opened 2021-03-24
Reading list messages updated to include the list name

sharvaniharan:
**Phab:**  https://phabricator.wikimedia.org/T275604


app/src/main/java/org/wikipedia/readinglist/LongPressMenu.kt
```diff
@@ -85,10 +85,15 @@ class LongPressMenu(private val anchorView: View, private val existsInAnyList: B
 
     private fun deleteOrShowDialog() {
         listsContainingPage?.let { list ->
-            RemoveFromReadingListsDialog(list).deleteOrShowDialog(anchorView.context) { _: List<ReadingList?>?, _: ReadingListPage? ->
-                entry?.let {
+            RemoveFromReadingListsDialog(list).deleteOrShowDialog(anchorView.context) { readingLists: List<ReadingList?>?, _: ReadingListPage? ->
```
cooltey, line 88: The parameters can actually be non-null.
`readingLists: List<ReadingList?>?, _: ReadingListPage?`
to
`readingLists: List<ReadingList>, _: ReadingListPage`

app/src/main/java/org/wikipedia/readinglist/ReadingListBehaviorsUtil.kt
```diff
@@ -162,8 +162,8 @@ object ReadingListBehaviorsUtil {
             return
         }
         FeedbackUtil
-                .makeSnackbar(activity, if (pages.size == 1) activity.getString(R.string.reading_list_item_deleted, pages[0].title())
-                                else activity.resources.getQuantityString(R.plurals.reading_list_articles_deleted, pages.size, pages.size),
+                .makeSnackbar(activity, if (pages.size == 1) activity.getString(R.string.reading_list_item_deleted_from_list, pages[0].title(), readingList.title())
+                else activity.resources.getQuantityString(R.plurals.reading_list_articles_deleted_from_list, pages.size, pages.size, readingList.title()),
```
cooltey, line 166: Looks like the first string item has two parameters instead of one?

cooltey (changes requested):

app/src/main/java/org/wikipedia/readinglist/ReadingListBehaviorsUtil.kt
sharvaniharan: It has 2 parameters @cooltey

dbrant (changes requested):
Let's please not remove the old string resources until the next TWN sync after this is merged. This is because the new strings are very similar to the old ones, and might cause TWN to treat it as a _renaming_ of these strings, which has happened in the past.  We'll delete the old strings once the new ones are safely in TWN.

app/src/main/java/org/wikipedia/readinglist/LongPressMenu.kt
```diff
@@ -85,10 +85,15 @@ class LongPressMenu(private val anchorView: View, private val existsInAnyList: B
...
+            RemoveFromReadingListsDialog(list).deleteOrShowDialog(anchorView.context) { readingLists: List<ReadingList?>, _: ReadingListPage ->
+                entry?.let { it ->
                     if (anchorView.isAttachedToWindow) {
-                        showMessage((anchorView.context as AppCompatActivity), anchorView.context.getString(R.string.reading_list_item_deleted, it.title.displayText))
+                        val readingListNames = mutableListOf<String>()
+                        readingLists.forEach { readingList ->
+                            readingListNames.add(readingList!!.title())
+                        }
+                        showMessage((anchorView.context as AppCompatActivity), anchorView.context.getString(R.string.reading_list_item_deleted_from_list,
+                                it.title.displayText, readingListNames.joinToString(separator = ", ")))
```
dbrant, line 96: Something to consider: Not all languages use the comma `,` to separate items in a list. How will this look in Arabic?

The real proper solution will be to use the ICU `ListFormatter` class, but this can be done in a future update.

sharvaniharan:
@dbrant @cooltey I have put back the translations.. Will just get the design review started... please feel free to review as well... will come back and take care of it..

dbrant (changes requested):
@sharvaniharan  I'm still seeing the `plurals` removed from the translations. Can we put those back?

sharvaniharan:
@dbrant The tests fail if the plural string have a translation but no base string [ like the one failing now]... so are you suggesting I add back the original string in strings.xml too, but just not use it?

dbrant:
> @dbrant The tests fail if the plural strings have a translation but no base string [ like the one failing now]... so are you suggesting I add back the original string in strings.xml too, but just not use it?

Yes, exactly right -- don't remove any of the old strings or plurals, and just add a couple of new ones.

sharvaniharan:
@dbrant @cooltey It was easier to open a fresh PR... have taken care of the ListFormatter issue too.. refer #2272 please

PR #2246, opened 2021-03-24
Convert ReferenceDialog related classes to Kotlin

cooltey:


app/src/main/java/org/wikipedia/page/references/PageReferences.kt
```diff
@@ -0,0 +1,13 @@
+package org.wikipedia.page.references
+
+import com.google.gson.annotations.SerializedName
+
+data class PageReferences(val selectedIndex: Int? = 0,
```
sharvaniharan, line 5: Should this be non-null?

sharvaniharan (commented):
Just one minor question lgtm otherwise...

app/src/main/java/org/wikipedia/page/references/PageReferences.kt
cooltey: You are correct. Done.

PR #2249, opened 2021-03-25
MEP: Flush events when app is suspended

sharvaniharan:
When the app is suspended send all accumulated events so we do not lose them.


app/src/main/java/org/wikipedia/ActivityLifecycleHandler.java
```diff
@@ -79,4 +81,19 @@ public void onActivityDestroyed(@NonNull Activity activity) {
...
+    @Override
+    public void onConfigurationChanged(@NonNull Configuration configuration) {
+    }
+
+    @Override
+    public void onLowMemory() {
+    }
+
+    @Override
+    public void onTrimMemory(int i) {
```
dbrant, line 94: Naming: can we name the `i` parameter something more meaningful, such as what it's called in the documentation?

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
```diff
@@ -140,6 +140,13 @@ static void addEventMetadata(Event event) {
         event.setAppInstallId(Prefs.getAppInstallId());
     }
 
+    /**
+     * Send all scheduled-and-waiting events in case of app being suspended.
```
dbrant, line 144: This comment might be too specific: this function might be called for other reasons than the app being suspended.  I'm not sure this needs a comment at all, since `flushCachedEvents()` is a perfectly self-documenting name.

dbrant (changes requested):
An interesting strategy (hooking into `onTrimMemory`)! I think this should work pretty well for our purposes. Just a couple of nits.

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
sharvaniharan: Correct... makes sense... removed the comment.

PR #2254, opened 2021-03-26
Convert GeoMarshaller related classes to Kotlin

cooltey:


app/src/main/java/org/wikipedia/page/PageProperties.java
```diff
@@ -190,7 +191,7 @@ public void writeToParcel(Parcel parcel, int flags) {
         parcel.writeLong(revisionId);
         parcel.writeLong(lastModified.getTime());
         parcel.writeString(displayTitleText);
-        parcel.writeString(GeoMarshaller.marshal(geo));
+        parcel.writeParcelable(geo, Parcelable.PARCELABLE_WRITE_RETURN_VALUE);
```
cooltey, line 194: Feels like we do not really need to `marshal(geo)` here, and can be replaced with `writeParcelable(geo, Parcelable.PARCELABLE_WRITE_RETURN_VALUE)`.

dbrant: What about when `geo == null`?  (which is why the tests fail)

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageProperties.java
cooltey: Yes, you are correct, and thanks for committing the changes.

PR #2260, opened 2021-03-30
Convert DescriptionEditTutorial to kotlin.

dbrant:


app/src/main/java/org/wikipedia/descriptions/DescriptionEditTutorialFragment.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.core.os.bundleOf
+import androidx.fragment.app.Fragment
+import androidx.viewpager2.adapter.FragmentStateAdapter
+import org.wikipedia.R
+import org.wikipedia.onboarding.OnboardingFragment
+import org.wikipedia.onboarding.OnboardingPageView
+
+class DescriptionEditTutorialFragment : OnboardingFragment() {
+    override val doneButtonText: Int
+        get() = R.string.description_edit_tutorial_button_label_start_editing
```
cooltey, line 16: Explicit type specification can be reduced.

app/src/main/java/org/wikipedia/descriptions/DescriptionEditTutorialFragment.kt
```diff
@@ -0,0 +1,49 @@
...
+
+        override fun createFragment(position: Int): Fragment {
+            return ItemFragment().apply { arguments = bundleOf("position" to position) }
+        }
+    }
+
+    class ItemFragment : Fragment() {
+        override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+            super.onCreateView(inflater, container, savedInstanceState)
+            val position = requireArguments().getInt("position", 0)
```
cooltey, line 35: Feels like the "position" can be set as a `const val`

cooltey:
I'll just merge it since the changes are pretty minor.

PR #2261, opened 2021-03-30
Convert DescriptionEditSuccess to kotlin.

dbrant:


app/src/main/java/org/wikipedia/descriptions/DescriptionEditSuccessFragment.kt
```diff
@@ -0,0 +1,45 @@
...
+    private var _binding: FragmentDescriptionEditSuccessBinding? = null
+    private val binding get() = _binding!!
+
+    interface Callback {
+        fun onDismissClick()
+    }
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
+        super.onCreateView(inflater, container, savedInstanceState)
+        _binding = FragmentDescriptionEditSuccessBinding.inflate(layoutInflater, container, false)
```
cooltey, line 21: Can replace `layoutInflater` with `inflater`.

cooltey: I'll self merge it.

PR #2267, opened 2021-04-01
Convert ThrowableUtil methods to extensions.

Isira-Seneviratne:


dbrant:
Sorry, this gets a -1 from me, because once again this does not add a whole lot of benefit, and basically amounts to style preferences.

PR #2272, opened 2021-04-02
Reading list messages updated to include the list name

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T275604


app/src/main/java/org/wikipedia/readinglist/ReadingListBehaviorsUtil.kt
```diff
@@ -149,7 +149,7 @@ object ReadingListBehaviorsUtil {
         FeedbackUtil
                 .makeSnackbar(activity,
                         activity.getString(
-                                if (lists.size == 1) R.string.reading_list_item_deleted else R.string.reading_lists_item_deleted, page.title()),
+                                if (lists.size == 1) R.string.reading_list_item_deleted_from_list else R.string.reading_lists_item_deleted, page.title()),
```
cooltey, line 152: The first one will cause a crash since it has two parameters.

cooltey (changes requested):

app/src/main/java/org/wikipedia/readinglist/ReadingListBehaviorsUtil.kt
sharvaniharan: Thank you for catching this @cooltey. Done.

app/src/main/java/org/wikipedia/readinglist/ReadingListBehaviorsUtil.kt
```diff
@@ -146,14 +148,17 @@ object ReadingListBehaviorsUtil {
...
         }
-        FeedbackUtil
-                .makeSnackbar(activity,
-                        activity.getString(
-                                if (lists.size == 1) R.string.reading_list_item_deleted else R.string.reading_lists_item_deleted, page.title()),
-                        FeedbackUtil.LENGTH_DEFAULT)
+        val readingListNames = mutableListOf<String>()
+        lists.forEach { readingList ->
+            readingListNames.add(readingList.title())
+        }
```
cooltey, line 154: Would it be possible to simplify as below and not running the logic in the snackbar?
```
        val readingListNames = lists.map { it.title() }.run {
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
                ListFormatter.getInstance().format(this)
            } else {
                joinToString(separator = ", ")
            }
        }
```

app/src/main/java/org/wikipedia/readinglist/LongPressMenu.kt
```diff
@@ -85,10 +87,16 @@ class LongPressMenu(private val anchorView: View, private val existsInAnyList: B
...
         listsContainingPage?.let { list ->
-            RemoveFromReadingListsDialog(list).deleteOrShowDialog(anchorView.context) { _: List<ReadingList?>?, _: ReadingListPage? ->
+            RemoveFromReadingListsDialog(list).deleteOrShowDialog(anchorView.context) { readingLists, _ ->
                 entry?.let {
                     if (anchorView.isAttachedToWindow) {
-                        showMessage((anchorView.context as AppCompatActivity), anchorView.context.getString(R.string.reading_list_item_deleted, it.title.displayText))
+                        val readingListNames = mutableListOf<String>()
+                        readingLists.forEach { readingList ->
+                            readingListNames.add(readingList.title())
+                        }
```
cooltey, line 96: This could be simplified (see comment below).

cooltey (changes requested):

app/src/main/java/org/wikipedia/readinglist/LongPressMenu.kt
cooltey: Is this resolved?

sharvaniharan: yes @cooltey

sharvaniharan: Sorry resolved now

PR #2274, opened 2021-04-02
Simplify and harden CsrfTokenClient once and for all.

dbrant:
I'm confident that this should finally fix the "occasional log out" errors that some users have been seeing.
https://phabricator.wikimedia.org/T205596


app/src/main/java/org/wikipedia/csrf/CsrfTokenClient.kt
```diff
@@ -0,0 +1,87 @@
...
+                var token = ""
+                try {
+                    MUTEX.acquire()
+                    if (emitter.isDisposed) {
+                        return@create
+                    }
+                    for (retry in 0 until numRetries) {
+                        if (retry > 0) {
+                            // Log in explicitly
+                            LoginClient().loginBlocking(loginWikiSite, AccountUtil.userName!!, AccountUtil.password!!, "")
```
cooltey, line 34: Since the `loginBlocking()` requires `NonNull` username and password, would it be better if we have null checks on line `32`?

app/src/test/java/org/wikipedia/csrf/CsrfTokenClientTest.java
```diff
@@ -1,87 +1,30 @@
...
     @Test public void testRequestSuccess() throws Throwable {
         String expected = "b6f7bd58c013ab30735cb19ecc0aa08258122cba+\\";
         enqueueFromFile("csrf_token.json");
 
-        Callback cb = mock(Callback.class);
-        request(cb);
-
-        server().takeRequest();
-        assertCallbackSuccess(cb, expected);
+        (new CsrfTokenClient(wikiSite(), 1, getApiService())).getToken().test().await()
```
cooltey, line 12: Parentheses can be removed.

app/src/test/java/org/wikipedia/csrf/CsrfTokenClientTest.java
```diff
@@ -1,87 +1,30 @@
...
 
     @Test public void testRequestResponseApiError() throws Throwable {
         enqueueFromFile("api_error.json");
 
-        Callback cb = mock(Callback.class);
-        request(cb);
-
-        server().takeRequest();
-        assertCallbackFailure(cb, MwException.class);
+        (new CsrfTokenClient(wikiSite(), 1, getApiService())).getToken().test().await()
```
cooltey, line 20: Parentheses can be removed.

app/src/test/java/org/wikipedia/csrf/CsrfTokenClientTest.java
```diff
@@ -1,87 +1,30 @@
...
-    private void assertCallbackFailure(@NonNull Callback cb,
-                                       @NonNull Class<? extends Throwable> throwable) {
-        //noinspection unchecked
-        verify(cb, never()).success(any(String.class));
-        verify(cb).failure(isA(throwable));
-    }
-
-    private Call<MwQueryResponse> request(@NonNull Callback cb) {
-        return subject.request(service(Service.class), cb);
+        (new CsrfTokenClient(wikiSite(), 1, getApiService())).getToken().test().await()
```
cooltey, line 27: Parentheses can be removed.

cooltey (changes requested):

app/src/main/java/org/wikipedia/csrf/CsrfTokenClient.kt
dbrant: I'm not sure how that would be helpful.  This entire function already assumes that the user is logged in, so if `userName` or `password` turns out to be null, I would want to know about it via an error passed to the emitter.

cooltey: I see, it makes sense to me.

PR #2275, opened 2021-04-02
ImageRecs: show instruction text properly in small screen

cooltey:
https://phabricator.wikimedia.org/T278493


cooltey:
Tested on Nexus S with API 26.

dbrant (changes requested):
The trouble with this approach is that on a _large_ display the background of the text fills up the entire width, and I don't think we want that.
Hint:  keep the text as `wrap_content` but introduce a `layout_constraintHorizontal_bias` of `0`.

device-2021-04-06-104424

cooltey:
> `layout_constraintHorizontal_bias`
Great! Thanks and done.

PR #2278, opened 2021-04-05
Hide voice-search icon if speech recognition is unavailable.

dbrant:
This converts the SearchCard in the feed to Kotlin, and makes a tiny fix:
- By request of Huawei: if voice recognition is not available on the device, then we should hide the Voice icon, which becomes useless.


app/src/main/java/org/wikipedia/feed/searchbar/SearchCardView.kt
```diff
@@ -0,0 +1,29 @@
...
+import org.wikipedia.util.FeedbackUtil
+import org.wikipedia.util.ResourceUtil
+
+class SearchCardView(context: Context) : DefaultFeedCardView<SearchCard>(context) {
+    interface Callback {
+        fun onSearchRequested(view: View)
+        fun onVoiceSearchRequested()
+    }
+
+    private val binding = ViewSearchBarBinding.inflate(LayoutInflater.from(context), this, true)
```
cooltey, line 19: Can this be a local variable?

app/src/main/java/org/wikipedia/feed/searchbar/SearchCardView.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    init {
+        binding.searchContainer.setCardBackgroundColor(ResourceUtil.getThemedColor(context, R.attr.color_group_22))
+        FeedbackUtil.setButtonLongPressToast(findViewById(R.id.voice_search_button))
```
cooltey, line 23: Should we use `binding.voiceSearchButton` instead of `findViewById`?

cooltey (changes requested):

app/src/main/java/org/wikipedia/feed/searchbar/SearchCardView.kt
dbrant: yep

app/src/main/java/org/wikipedia/feed/searchbar/SearchCardView.kt
dbrant: whoops

PR #2284, opened 2021-04-06
ImageRecs preliminaries: enhance FaceAndColorImageView listener.

dbrant:
This enhances the listener of our `FaceAndColorDetectImageView` to also provide the actual dimensions of the bitmap that gets loaded into the ImageView.  This will be useful for determining the positioning of child elements in ImageViews based on the bitmap that's loaded in them.

Since `CardLargeHeaderView` is the only thing that uses this listener so far, I updated it and converted it to kotlin.

(this does not impact any other existing behavior)


sharvaniharan (approved):

app/src/main/java/org/wikipedia/views/FaceAndColorDetectImageView.kt
```diff
@@ -55,18 +55,22 @@ class FaceAndColorDetectImageView constructor(context: Context, attrs: Attribute
...
                         listener.onImageFailed()
                     }
                     return false
                 }
             })
         }
-        builder = if (shouldDetectFace(uri)) {
-            builder.transform(if (roundedCorners) FACE_DETECT_TRANSFORM_AND_ROUNDED_CORNERS else FACE_DETECT_TRANSFORM)
+        if (scaleType == ScaleType.FIT_CENTER) {
+            builder = builder.transform(WhiteBackgroundTransformation())
```
cooltey, line 67: Looks like this will make some of the article leading images show as CENTER_INSIDE with white background. Is this expected? (e.g. DMX(rapper) article).

PR #2287, opened 2021-04-06
Image Recs: Success illustration

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T278549


app/src/main/res/layout/fragment_suggested_edits_image_recommendation_item.xml
```diff
@@ -319,15 +318,14 @@
         android:elevation="14dp"
         tools:visibility="gone">
 
-        <View
-            android:id="@+id/publishBackgroundBaseView"
```
sharvaniharan, line 323: @dbrant Did not see a use for this since there is already the 'publishBackgroundView' but let me know if I missed something..

cooltey: We should keep it in order to match the design needs.
#2281

cooltey (changes requested):

app/src/main/res/layout/fragment_suggested_edits_image_recommendation_item.xml
```diff
@@ -319,25 +318,36 @@
         android:elevation="14dp"
         tools:visibility="gone">
 
+        <androidx.appcompat.widget.AppCompatImageView
```
dbrant, line 321: fyi, there's no need to use the fully-qualified `AppCompatImageView`. Just the plain `ImageView` will do the same thing.

PR #2288, opened 2021-04-06
ImgRecs: change italics to bold on the "suggestion reason"

cooltey:
https://phabricator.wikimedia.org/T278530


cooltey:
> @cooltey This is fine, but remember that your changes to the non-English translations will be overwritten with the next TWN sync. (unless you manually updated all those strings on TWN yourself?)

Yes, totally forgot to do it. I have another PR of updating string in `master` branch.
https://github.com/wikimedia/apps-android-wikipedia/pull/2291

PR #2289, opened 2021-04-07
Use View extensions.

Isira-Seneviratne:
Use Core KTX extensions for `View` and `TextView`.

### DIFF
--- a/app/src/main/java/org/wikipedia/WikipediaApp.java
+++ b/app/src/main/java/org/wikipedia/WikipediaApp.java
@@ -319,6 +319,7 @@ public class WikipediaApp extends Application {
         crashListener.logCrashManually(throwable);
     }
 
+    @NonNull
     public Handler getMainThreadHandler() {
         if (mainThreadHandler == null) {
             mainThreadHandler = new Handler(getMainLooper());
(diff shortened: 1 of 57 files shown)


anticafe (commented):
I'm against the usage of `isVisible =`, `isGone = `.
While in case true, it's easy to understand it would be Visible or Gone. But in case false, it's hard for us to know does the View become Invisible or Gone (in case from Visible), or Visible or Invisible (in case Gone).

In my real project, I would create my own extension:
- `toVisible`
- `toInvisible`
- `toGone`
- `toVisibleOtherwiseInvisible`
- `toVisibleOtherwiseGone`
- `toGoneOtherwiseVisible`
- ...and so one.

How do you think @Isira-Seneviratne ?

Isira-Seneviratne:
> I'm against the usage of `isVisible =`, `isGone = `.
> While in case true, it's easy to understand it would be Visible or Gone. But in case false, it's hard for us to know does the View become Invisible or Gone (in case from Visible), or Visible or Invisible (in case Gone).
> 
> In real project, I would create my own extensions:
> 
> * `toVisible`
> * `toInvisible`
> * `toGone`
> * `toVisibleOtherwiseInvisible(visible: Boolean)`
> * `toVisibleOtherwiseGone(Boolean)`
> * `toGoneOtherwiseVisible(Boolean)`
> * ...and so one.
> 
> How do you think @Isira-Seneviratne ?

The [documentation](https://developer.android.com/reference/kotlin/androidx/core/view/package-summary) is clear about what the extensions do, but I agree that clearer extensions could be defined, e.g. `isVisibleElseGone` or `isGoneElseVisible`.

dbrant:
Once again, this is a huge amount of style over substance, and personal preferences.  In the future please focus your pull requests to do one thing at a time, instead of 50 things at once.

PR #2290, opened 2021-04-07
Image Recs: Instrumentation on MEP

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T273057

Tested this on beta and production and it worked!

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/ImageRecommendationsEvent.kt
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/ImageRecommendationsEvent.kt
@@ -1,0 +1,20 @@
+package org.wikipedia.analytics.eventplatform
+
+class ImageRecommendationsEvent(private val lang: String, private val page_title: String, private val image_title: String,
+                                private val suggestion_source: String, private val response_type: Int, private val reasons: List<Int>,
+                                private val info_clicked: Boolean, private val details_clicked: Boolean, scrolled: Boolean,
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/analytics/eventplatform/ImageRecommendationsEvent.kt
```diff
@@ -0,0 +1,20 @@
+package org.wikipedia.analytics.eventplatform
+
+class ImageRecommendationsEvent(private val lang: String, private val page_title: String, private val image_title: String,
+                                private val suggestion_source: String, private val response_type: Int, private val reasons: List<Int>,
+                                private val info_clicked: Boolean, private val details_clicked: Boolean, scrolled: Boolean,
+                                private val time_until_click_ms: Long, private val time_until_submit_ms: Long, private val user_text: String?,
+                                private val teacher_mode_enabled: Boolean) : Event(SCHEMA_NAME, STREAM_NAME) {
```
cooltey, line 7: Just curious, are these variables are common variables that will be used in multiple log events? If not, would it be possible to optimize a bit?

sharvaniharan: The problem is that the class gets automatically serialized to send to MEP... so they all need to be fields on the class and in that order..

cooltey: Got it.

cooltey (approved):

PR #2293, opened 2021-04-07
Update recyclerview version to 1.2.0

cooltey:
https://developer.android.com/jetpack/androidx/releases/recyclerview#recyclerview-1.2.0

I am not seeing the `getAdapterPosition` has the deprecated label on it and I think we can still continue using it for a while.


sharvaniharan:
Looks like nothing is breaking in language fragment and configure where we use it.. so merging it..

PR #2299, opened 2021-04-08
ImgRecs: progress bar design tweaks

cooltey:
https://phabricator.wikimedia.org/T278528#6984374


dbrant (changes requested):
The check mark is now way too large (does not look like the design):

image

cooltey:
> The check mark is now way too large (does not look like the design):
> 

Done.

PR #2305, opened 2021-04-13
Fix edit comment when submitting short descriptions.

dbrant:
The edit comment of `suggestededit-add_1.0` was being added incorrectly even when the edit came organically from the article, not just the Suggested Edits feature.

https://phabricator.wikimedia.org/T279702


sharvaniharan:
Lgtm @dbrant but I couldn't still understand the root cause... looks like we were doing the same checks before

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -233,11 +233,7 @@ class DescriptionEditFragment : Fragment() {
                         val baseRevId = mwQueryResponse.query()!!.firstPage()!!.revisions()[0].revId
                         text = updateDescriptionInArticle(text, binding.fragmentDescriptionEditView.description!!)
                         ServiceFactory.get(wikiSite).postEditSubmit(pageTitle.prefixedText, "0", null,
-                                when (action) {
```
dbrant, line 236: @sharvaniharan   If you look here, this block does not contain the check for `if (invokeSource == InvokeSource.SUGGESTED_EDITS || invokeSource == InvokeSource.FEED)`, as the others do.

sharvaniharan: ah! ok thank you!

PR #2312, opened 2021-04-14
Don't show reply button when the topicId=-1 in TalkTopicActivity

cooltey:
As we discussed on Slack, sometimes the `id` is `-1` and we should not show the "new topic" page for it and should hide the reply button.
https://en.wikipedia.org/api/rest_v1/page/talk/Talk%3ADreamsnake

### DIFF
--- a/app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
+++ b/app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
@@ -69,7 +69,7 @@ class TalkTopicsActivity : BaseActivity() {
 
         binding.talkNewTopicButton.setOnClickListener {
             funnel.logNewTopicClick()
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
```diff
@@ -247,6 +247,7 @@ class TalkTopicsActivity : BaseActivity() {
 
     companion object {
         private const val EXTRA_PAGE_TITLE = "pageTitle"
+        const val NEW_TOPIC_ID = -2
```
sharvaniharan, line 250: Lgtm .. just a minor q... is this a randomly picked value we are giving to a new topic?

cooltey: Yes. Please let me know if you prefer another number.

sharvaniharan: oh ok got it... No .. Lgtm..

PR #2317, opened 2021-04-15
Design tweak: Image recs onboarding

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T278526


cooltey (commented):
This will make the bottom text being cut off on small screen devices. How about we make the child `LinearLayout` have the `android:layout_gravity="center_vertical"`? It works for me.

PR #2318, opened 2021-04-15
ImgRecs: add "Got it" button on sequential tooltips

cooltey:
https://phabricator.wikimedia.org/T278550


sharvaniharan:
@cooltey It looks good. Just one clarification... Does it look a bit too bright in dark / black modes?

cooltey:
Robin didn't mention about it and the mock only use plain white with 15% alpha on background.

sharvaniharan:
ah! makes sense

PR #2322, opened 2021-04-19
Image recs dialog margin update

sharvaniharan:


sharvaniharan:
Self-merging this to pass it on to design sign off... @dbrant @cooltey

PR #2333, opened 2021-04-21
Converted events java files to kotlin

Arunm619:
- Converted all files under events/ folder to Kotlin


app/src/main/java/org/wikipedia/events/ArticleSavedOrDeletedEvent.kt
```diff
@@ -0,0 +1,7 @@
+package org.wikipedia.events
+
+import org.wikipedia.readinglist.database.ReadingListPage
+
+class ArticleSavedOrDeletedEvent(val isAdded: Boolean, vararg pages: ReadingListPage) {
+    val pages: Array<ReadingListPage> = arrayOf(*pages)
```
cooltey, line 6: This can be simplified as `val pages = pages.toList()`

cooltey (changes requested):

app/src/main/java/org/wikipedia/events/ArticleSavedOrDeletedEvent.kt
Arunm619: Sure thing!

PR #2342, opened 2021-04-21
ImgRecs: progressView design tweaks

cooltey:
https://phabricator.wikimedia.org/T278528#7013728

PR #2344, opened 2021-04-22
ImgRecs: reduce spaces of daily amount string resources

cooltey:
Update string resources for image recommendation PR #2342


cooltey:
Will self-merge since it is a very straightforward change.

PR #2347, opened 2021-04-23
Image recs : Small screen fix for onboarding image being cut-off

sharvaniharan:
Currently onboarding image is getting cut-off on the top on small screens. This is because of the gravity of the layout below it. However if we make a general change, the layout doesnt render correctly in other places. So put in a specific change.
cut-off

### DIFF
--- a/app/src/main/java/org/wikipedia/onboarding/OnboardingPageView.kt
+++ b/app/src/main/java/org/wikipedia/onboarding/OnboardingPageView.kt
@@ -3,9 +3,11 @@ package org.wikipedia.onboarding
 import android.content.Context
 import android.text.TextUtils
 import android.util.AttributeSet
+import android.view.Gravity
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import android.widget.FrameLayout
 import android.widget.TextView
 import androidx.appcompat.content.res.AppCompatResources
 import androidx.constraintlayout.widget.ConstraintLayout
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/onboarding/OnboardingPageView.kt
```diff
@@ -81,6 +83,15 @@ class OnboardingPageView constructor(context: Context, attrs: AttributeSet? = nu
         }
     }
 
+    fun removeMAinTextLayoutGravity() {
```
cooltey, line 86: There are two upper cases in the function name.

cooltey (changes requested):

app/src/main/java/org/wikipedia/onboarding/OnboardingPageView.kt
sharvaniharan: Done @cooltey

app/src/main/java/org/wikipedia/suggestededits/ImageRecsOnboardingFragment.kt
```diff
@@ -64,6 +65,10 @@ class ImageRecsOnboardingFragment : OnboardingFragment(false), OnboardingPageVie
             if (OnboardingPage.PAGE_CONSENT.code() == position) {
                 view.setSwitchChecked(Prefs.isImageRecsConsentEnabled())
             }
+            if (DimenUtil.displayHeightPx < 1200) {
+                // Remove layout gravity of the text below on small screens to make centered image visible
+                view.removeMainTextLayoutGravity()
+            }
```
cooltey, line 71: Looks like this logic should be inside the `OnboardingPageView`.

sharvaniharan: We can move this inside after this release @cooltey ... but if i do now, then I will haave to test all onboarding screens to see if they are still working on small screen ... so I can add a todo. Does that sound ok?

cooltey: Sounds good.

PR #2348, opened 2021-04-23
ImgRecs: throw proper exception and provide a proper button for going to next item

cooltey:
In order to use unclear `Exception` for the empty title, it would be better if we show the "Next" button for `ExceptionEmpty`.


sharvaniharan:
Awesome! 🎉

PR #2350, opened 2021-04-24
ImgRecs: follow up: detect onboarding view height for small screen device

cooltey:
A follow up of #2347


sharvaniharan:
Awesome! checks-out!

PR #2355, opened 2021-04-26
ImgRecs: check and update daily count in SE task screen

cooltey:
https://phabricator.wikimedia.org/T278528#7033856

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -895,7 +895,7 @@
     <string name="suggested_edits_image_recommendations_task_get_started">Start</string>
     <string name="suggested_edits_image_recommendations_task_goal_progress">Great, keep going! 👍</string>
     <string name="suggested_edits_image_recommendations_task_goal_complete">Daily goal achieved! 👏</string>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -895,7 +895,7 @@
     <string name="suggested_edits_image_recommendations_task_get_started">Start</string>
     <string name="suggested_edits_image_recommendations_task_goal_progress">Great, keep going! 👍</string>
     <string name="suggested_edits_image_recommendations_task_goal_complete">Daily goal achieved! 👏</string>
-    <string name="suggested_edits_image_recommendations_task_goal_surpassed">You\'ve reached the goal</string>
+    <string name="suggested_edits_image_recommendations_task_goal_surpassed">You reached the goal</string>
```
sharvaniharan, line 898: @cooltey I think the string is "You reached the daily goal"

sharvaniharan (changes requested):

app/src/main/res/values/strings.xml
cooltey: Done.

PR #2356, opened 2021-04-26
ImgRecs: remove commented out code

cooltey:
Not sure if these codes are staying for specific reasons.


sharvaniharan:
Lgtm... but will let @dbrant merge it

PR #2357, opened 2021-04-26
Hygiene: remove unused class

cooltey:
#2331 Can be closed since this utility class is unused.


sharvaniharan:
There are tests that will need to be removed @cooltey.

cooltey:
> 
> 
> There are tests that will need to be removed @cooltey.

Ah! you are correct. Done.

PR #2368, opened 2021-04-28
Fix and improve tooltip sequencing on Main screen.

dbrant:
This improves the way we show tooltips on the Main screen. It is no longer possible to show more than one tooltip at the same time, and also multiple tooltips will be automatically "prioritized" to be shown based on the order in which they are called.

This also removes the pre-beta feature flag for the Image Recommendations tooltip.


app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -486,39 +487,36 @@ class MainFragment : Fragment(), BackPressedHandler, FeedFragment.Callback, Hist
     private fun maybeShowEditsTooltip() {
         if (currentFragment !is SuggestedEditsTasksFragment && Prefs.shouldShowSuggestedEditsTooltip() &&
                 Prefs.getExploreFeedVisitCount() == SHOW_EDITS_SNACKBAR_COUNT) {
-            binding.mainNavTabLayout.postDelayed({
-                if (!isAdded) {
-                    return@postDelayed
-                }
-                Prefs.setShouldShowSuggestedEditsTooltip(false)
+            enqueueTooltip({
```
cooltey, line 490: This can be a lambda style. `enqueueTooltip { ... } `.

app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -486,39 +487,36 @@ class MainFragment : Fragment(), BackPressedHandler, FeedFragment.Callback, Hist
     private fun maybeShowEditsTooltip() {
         if (currentFragment !is SuggestedEditsTasksFragment && Prefs.shouldShowSuggestedEditsTooltip() &&
                 Prefs.getExploreFeedVisitCount() == SHOW_EDITS_SNACKBAR_COUNT) {
```
cooltey, line 489: Since the general tooltip for Edits tab will show only when the count equals to `SHOW_EDITS_SNACKBAR_COUNT`, should we add an extra logic to the `maybeShowImageRecsTooltip` with `Prefs.getExploreFeedVisitCount() > SHOW_EDITS_SNACKBAR_COUNT`?

cooltey (changes requested):

app/src/main/java/org/wikipedia/main/MainFragment.kt
dbrant: Agreed

app/src/main/java/org/wikipedia/main/MainFragment.kt
dbrant: Whoops, it looks like my Android Studio got corrupted and stopped showing me automatic code insights.  I've now reinstalled it.

PR #2369, opened 2021-04-28
MEP: Send events immediately for testing

mdholloway:
Sends events immediately upon submission to the MEP client to match the legacy eventlogging behavior. We'll compare the resulting events with those sent via the legacy system in order to determine conclusively whether events are being leaked in the queueing process.

This PR also contains an incidental change to drop support for looking up stream names by regular expression matching. This feature is only used in practice by job-queue-related stream configurations, and the app need not support it.

Phab: https://phabricator.wikimedia.org/T281001

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.java
@@ -45,54 +45,6 @@ public final class EventPlatformClient {
      */
     private static boolean ENABLED = WikipediaApp.getInstance().isOnline();
 
-    /**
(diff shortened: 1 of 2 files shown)


mdholloway:
I haven't manually tested the client with these changes yet, but will now.

mdholloway:
OK, after some post-software-update emulator [fun](https://stackoverflow.com/questions/66455173/android-emulator-30-4-5-not-working-on-macos), I finally got this tested manually and it seems to be working well.

sharvaniharan:
Tested this too @mdholloway .. Lgtm. Will let others take a look at it too and merge before I leave for the day..

sharvaniharan (approved):

cooltey (approved):
Looks good to me.

PR #2372, opened 2021-04-28
Convert LinkMovementMethodExt to Kotlin

cooltey:


app/src/main/java/org/wikipedia/page/LinkMovementMethodExt.kt
```diff
@@ -0,0 +1,63 @@
...
+import android.text.Spannable
+import android.text.method.LinkMovementMethod
+import android.text.style.URLSpan
+import android.view.MotionEvent
+import android.widget.TextView
+import org.wikipedia.util.UriUtil
+import org.wikipedia.util.log.L
+
+class LinkMovementMethodExt : LinkMovementMethod {
+    fun interface UrlHandler {
```
sharvaniharan, line 12: Empty line above this?

cooltey: Done.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageTagsFragment.kt
```diff
@@ -88,7 +88,7 @@ class SuggestedEditsImageTagsFragment : SuggestedEditsItemFragment(), CompoundBu
 
         binding.tagsLicenseText.text = StringUtil.fromHtml(getString(R.string.suggested_edits_cc0_notice,
                 getString(R.string.terms_of_use_url), getString(R.string.cc_0_url)))
-        binding.tagsLicenseText.movementMethod = LinkMovementMethodExt.getInstance()
```
dbrant, line 91: What's the reason for this change? It doesn't seem related to Kotlin conversion.

dbrant (changes requested):

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageTagsFragment.kt
cooltey: Use `LinkMovementMethodExt.getInstance()` will show a "Unresolved reference: getInstance". 

Since this looks like a general usage of  `LinkMovementMethod` and it should be fine to use `LinkMovementMethod.getInstance()`.

sharvaniharan (approved):

PR #2381, opened 2021-04-29
Convert PageContainerLongPressHandler to Kotlin

cooltey:


app/src/main/java/org/wikipedia/page/PageContainerLongPressHandler.kt
```diff
@@ -0,0 +1,30 @@
+package org.wikipedia.page
+
+import org.wikipedia.Constants
+import org.wikipedia.LongPressHandler.WebViewMenuCallback
+import org.wikipedia.history.HistoryEntry
+import org.wikipedia.readinglist.database.ReadingListPage
+
+class PageContainerLongPressHandler(private val fragment: PageFragment) : WebViewMenuCallback {
+    override fun onOpenLink(entry: HistoryEntry) {
```
sharvaniharan, line 9: Can we have an empty line here?

app/src/main/java/org/wikipedia/LongPressHandler.kt
```diff
@@ -20,7 +20,7 @@ import org.wikipedia.util.UriUtil.isValidPageLink
 
 class LongPressHandler(view: View, private val historySource: Int, private val callback: LongPressMenu.Callback) : OnCreateContextMenuListener, OnTouchListener {
     interface WebViewMenuCallback : LongPressMenu.Callback {
```
sharvaniharan, line 22: Empty line?

app/src/main/java/org/wikipedia/page/PageContainerLongPressHandler.kt
cooltey: Done!

app/src/main/java/org/wikipedia/LongPressHandler.kt
```diff
@@ -20,7 +20,7 @@ import org.wikipedia.util.UriUtil.isValidPageLink
...
-        val wikiSite: WikiSite
```
dbrant, line 23: Why can't we keep `wikiSite` nonnull?

dbrant (changes requested):

app/src/main/java/org/wikipedia/LongPressHandler.kt
cooltey: The `override val wikiSite = fragment.title?.wikiSite` in `PageContainerLongPressHandler` has a chance to be a `null` if the `fragment.title` returns `null`.

We can do either put `!!` after `wikiSite` or change the callback to become a nonnull type of variable.

sharvaniharan (approved):

PR #2387, opened 2021-04-30
Convert remaining DescriptionEdit classes to Kotlin

cooltey:


app/src/main/java/org/wikipedia/descriptions/DescriptionEditRevertHelpView.kt
```diff
@@ -0,0 +1,51 @@
...
+import android.widget.ScrollView
+import org.wikipedia.R
+import org.wikipedia.WikipediaApp
+import org.wikipedia.databinding.ViewDescriptionEditRevertHelpBinding
+import org.wikipedia.util.DimenUtil
+import org.wikipedia.util.StringUtil
+
+class DescriptionEditRevertHelpView constructor(context: Context, attrs: AttributeSet? = null) : ScrollView(context, attrs) {
+
+    private var qNumber: String = ""
```
sharvaniharan, line 21: @cooltey Code convention: We should have the order -> constructo, any variables and then init block if there is.. what do you think..

app/src/main/java/org/wikipedia/descriptions/DescriptionEditUtil.kt
```diff
@@ -0,0 +1,20 @@
+package org.wikipedia.descriptions
+
+import org.wikipedia.page.Page
+
+object DescriptionEditUtil {
+    private const val DESCRIPTION_SOURCE_LOCAL = "local"
```
sharvaniharan, line 6: A space here?

app/src/main/res/layout/view_description_edit_revert_help.xml
```diff
@@ -5,7 +5,7 @@
     android:layout_height="wrap_content">
 
     <TextView
-        android:id="@+id/view_description_edit_revert_help_contents"
```
sharvaniharan, line 8: Why is this string changed @cooltey ..

sharvaniharan (changes requested):
Just a few minor comments... looks good otherwise.

app/src/main/res/layout/view_description_edit_revert_help.xml
cooltey: Just want to trim the string length so that it won't be too long in the code. (`viewDescriptionEditRevertHelpContents` => `helpText`).

app/src/main/java/org/wikipedia/descriptions/DescriptionEditUtil.kt
cooltey: Done

app/src/main/java/org/wikipedia/descriptions/DescriptionEditRevertHelpView.kt
cooltey: The order should be fine, as we discussed in Slack.

app/src/main/res/layout/view_description_edit_revert_help.xml
sharvaniharan: oh sorry my bad... i thought this was a string resource that was changed... sounds good!

PR #2389, opened 2021-05-01
Convert PageFragmentLoadState to Kotlin

cooltey:


app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -430,7 +429,7 @@ public void onActivityCreated(Bundle savedInstanceState) {
             new LongPressHandler(webView, HistoryEntry.SOURCE_INTERNAL_LINK, new PageContainerLongPressHandler(this));
         }
 
-        pageFragmentLoadState.setUp(model, this, webView, bridge, leadImagesHandler, getCurrentTab());
+        pageFragmentLoadState = new PageFragmentLoadState(model, this, webView, bridge, leadImagesHandler, getCurrentTab());
```
sharvaniharan, line 432: Just for safety should we just do this onCreate itself?

sharvaniharan (changes requested):
Lgtm... just one minor request.

app/src/main/java/org/wikipedia/page/PageFragment.java
cooltey: If we do this in `onCreate`, the `bridge`, `leadImagesHandler` will be `null`.

PR #2392, opened 2021-05-03
Explicitly check for edit rights when launching Edit activity.

dbrant:


cooltey (changes requested):
Found several things:
1. Should we change the current logic of blocking edit for article description for non-English pages? Looks like now an anonymous user can edit a protected page article description.
2. Should we hide the wikitext keyboard view when the page is not editable?

dbrant:
@cooltey  Both excellent observations.

> 1. Should we change the current logic of blocking edit for article description for non-English pages? Looks like now an anonymous user can edit a protected page article description.

I've updated the description editing submission logic: it will now explicitly check whether the article itself is editable by the current user, and then only submit the description to Wikidata if the article is editable. (otherwise show an error).  This way it won't be possible to edit the description of a protected article, even if the Wikidata entity is unprotected.  I've also cleaned up the way it handles AbuseFilter errors, too.

> 2. Should we hide the wikitext keyboard toolbar view when the page is not editable?

Yep.

sharvaniharan:
Lgtm @cooltey Please merge when you're ready

sharvaniharan (approved):

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -226,9 +226,13 @@ class DescriptionEditFragment : Fragment() {
 
         private fun postDescriptionToArticle(editToken: String) {
             val wikiSite = WikiSite.forLanguageCode(pageTitle.wikiSite.languageCode())
-            disposables.add(ServiceFactory.get(wikiSite).getWikiTextForSection(pageTitle.prefixedText, 0)
+            disposables.add(ServiceFactory.get(wikiSite).getWikiTextForSectionWithInfo(pageTitle.prefixedText, 0)
                     .subscribeOn(Schedulers.io())
                     .flatMap { mwQueryResponse ->
+                        if (mwQueryResponse.query()!!.firstPage()!!.getErrorForAction("edit").isNotEmpty()) {
+                            val error = mwQueryResponse.query()!!.firstPage()!!.getErrorForAction("edit")[0]
```
cooltey, line 233: Slightly concern about getting the first item from the `getErrorForAction("edit")`, would it be possible to throw an `IndexOutOfBoundsException` if it returns `Collections.emptyList()` and will not reach the `MwException()` below?

dbrant: ...but we're already checking if the list is non-empty (above). There will always be at least one item here.

cooltey: Ah! missed that 😅

PR #2395, opened 2021-05-04
Follow-up to Reading List conversion.

dbrant:


app/src/main/java/org/wikipedia/readinglist/database/ReadingList.kt
```diff
@@ -19,35 +19,22 @@ class ReadingList(var dbTitle: String,
...
 
     val sizeBytesFromPages: Long
-        get() {
-            var bytes = 0L
-            pages.forEach {
-                bytes += if (it.offline) it.sizeBytes else 0
-            }
-            return bytes
-        }
+        get() = pages.sumOf { if (it.offline) it.sizeBytes else 0 }
```
dbrant, line 31: loving these convenience constructs.

app/src/main/java/org/wikipedia/readinglist/database/ReadingList.kt
```diff
@@ -19,35 +19,22 @@ class ReadingList(var dbTitle: String,
     @Transient
     private var accentAndCaseInvariantTitle: String? = null
 
-    val isDefault = dbTitle.isEmpty()
+    val isDefault = title.isEmpty()
 
-    val title = if (isDefault) WikipediaApp.getInstance().getString(R.string.default_reading_list_name) else dbTitle
+    var title = title
```
cooltey, line 24: It looks like we cannot combine the `dbTitle` and `title` as one usage, it will cause some issues when syncing articles from the server if the current the readinglist is empty.

Steps:
1. Keep the app logged in with multiple reading lists synced.
2. Clear app data.
3. Open app again and go to the reading list.
4. Go to the reading list page and will see the order is incorrected and has multiple "Saved" lists.

app/src/main/java/org/wikipedia/readinglist/database/ReadingList.kt
cooltey: You are correct. Not sure why these didn't work for me at the moment when doing the conversion. The size numbers will not show the expected values.

cooltey (changes requested):

app/src/main/java/org/wikipedia/readinglist/database/ReadingList.kt
cooltey: I'll do another follow-up for it.

PR #2397, opened 2021-05-04
Make ReadingListDbHelper into an object class

cooltey:


dbrant:
looks good; but the test needs updating.

cooltey:
> 
> 
> looks good; but the test needs updating.

Done.

PR #2400, opened 2021-05-05
Kotlin conversion: AccessibilityCardView and MainPageCardView

sharvaniharan:


app/src/main/java/org/wikipedia/feed/accessibility/AccessibilityCardView.kt
```diff
@@ -0,0 +1,45 @@
...
+import android.widget.LinearLayout
+import org.wikipedia.databinding.ViewCardAccessibilityBinding
+import org.wikipedia.feed.model.Card
+import org.wikipedia.feed.view.FeedAdapter
+import org.wikipedia.feed.view.FeedCardView
+
+class AccessibilityCardView(context: Context?) : LinearLayout(context), FeedCardView<Card?> {
+
+    private var _binding: ViewCardAccessibilityBinding? = null
+    private val binding get() = _binding!!
```
cooltey, line 15: You can inflate the layout directly here without having an extra getter.

app/src/main/java/org/wikipedia/feed/mainpage/MainPageCardView.kt
```diff
@@ -0,0 +1,58 @@
...
+            callback!!.onSelectPage(card!!,
+                    HistoryEntry(PageTitle(getMainPageForLang(card!!.wikiSite().languageCode()), card!!.wikiSite()),
+                            HistoryEntry.SOURCE_FEED_MAIN_PAGE), false)
+        }
+    }
+
+    override fun removeView(view: View?) {
+        super.removeView(view)
+        _binding = null
+    }
```
cooltey, line 57: Not sure if it is necessary. @dbrant any thought?

app/src/main/java/org/wikipedia/feed/accessibility/AccessibilityCardView.kt
sharvaniharan: @cooltey The getter is created everywhere to manage the lifecycle, just as an extra precaution. I say we keep it both for safety and consistency..

app/src/main/java/org/wikipedia/feed/mainpage/MainPageCardView.kt
sharvaniharan: Was not sure about this myself... just wanted to make sure the binding object is available for GC as we do with other places

cooltey:
The PR still needs some code clean up, but I'll do it in #2401

PR #2405, opened 2021-05-06
Fix possible crash in SuggestedEditsFunnel after kotlin conversion.

dbrant:


app/src/main/java/org/wikipedia/analytics/SuggestedEditsFunnel.kt
```diff
@@ -169,7 +169,7 @@ class SuggestedEditsFunnel private constructor(app: WikipediaApp, private val in
         }
 
         fun get(): SuggestedEditsFunnel {
-            return if (INSTANCE?.invokeSource != InvokeSource.SUGGESTED_EDITS) {
```
dbrant, line 172: There was a null check here that was removed in #2403 but it's actually necessary.

The expression `INSTANCE?.invokeSource != InvokeSource.SUGGESTED_EDITS` will evaluate to `true` if INSTANCE is null.

PR #2408, opened 2021-05-06
Correctly re-parse certain AbuseFilter responses.

dbrant:
https://phabricator.wikimedia.org/T276139

### DIFF
--- a/app/src/main/java/org/wikipedia/util/StringUtil.kt
+++ b/app/src/main/java/org/wikipedia/util/StringUtil.kt
@@ -90,7 +90,7 @@ object StringUtil {
 
     @JvmStatic
     fun sanitizeAbuseFilterCode(code: String): String {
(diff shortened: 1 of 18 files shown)


app/src/main/java/org/wikipedia/util/StringUtil.kt
```diff
@@ -88,6 +88,11 @@ object StringUtil {
         return text.replace("<cite.*?>".toRegex(), "").replace("</cite>".toRegex(), "")
     }
 
+    @JvmStatic
+    fun sanitizeAbuseFilterCode(code: String): String {
+        return code.replace("⧼", "").replace("⧽", "")
```
cooltey, line 93: Should we use `toRegex()` here? `code.replace("<|>".toRegex(), "")`.

dbrant: yep! also added a test.

PR #2409, opened 2021-05-06
Update gradle plugins versions

cooltey:
Update com.android.tools.build:gradle to 4.2.0
Update gradle to 6.7.1


dbrant:
Can we hold on merging this for a bit; Upgrading to 4.2.0 causes the build to start failing on my Windows workstation, and I have no idea why.

dbrant:
Ha, looks like the build fails in CI, too

cooltey:
> 
> 
> Ha, looks like the build fails in CI, too

Not sure if it relates to the Kotlin plugin version, since it shows a recommended version `1.5.0-release-764` that doesn't make sense to me. Will wait and see if a newer update pops up later.

sharvaniharan:
Build id failing for me to ..

dbrant:
I traced the failure down to the way it processes our resources, and it looks like it no longer likes the `be+tarask` language.  Renaming that resource collection to `be-x-old` seems to fix it.

Personally I would just delete this language altogether, since on a real Android device there's no way to actually select this language as your system UI language.

Let's wait to merge this until my patch upstream to TWN is merged: https://gerrit.wikimedia.org/r/c/translatewiki/+/686735

cooltey:
> I traced the failure down to the way it processes our resources, and it looks like it no longer likes the `be+tarask` language.  Renaming that resource collection to `be-x-old` seems to fix it.
> 
> Personally I would just delete this language altogether, since on a real Android device there's no way to actually select this language as your system UI language.
> 
> Let's wait to merge this until my patch upstream to TWN is merged: https://gerrit.wikimedia.org/r/c/translatewiki/+/686735

Sounds good! Thanks for checking and fixing.

PR #2417, opened 2021-05-09
Avoid potential NullPointerException in OnThisDayCardView.

Isira-Seneviratne:


dbrant:
Already fixed in #2401

PR #2430, opened 2021-05-12
Convert MostRead fragment and activity to Kotlin

cooltey:


app/src/main/res/layout/fragment_most_read.xml
```diff
@@ -5,7 +5,7 @@
     android:fitsSystemWindows="true">
 
     <androidx.recyclerview.widget.RecyclerView
-        android:id="@+id/view_most_read_fullscreen_link_card_list"
+        android:id="@+id/most_read_recycler_view"
```
dbrant, line 8: good rename

PR #2432, opened 2021-05-12
Bump Gradle wrapper to 7.0.1

anticafe:


dbrant:
Thanks, but we prefer to let Android Studio update the gradle wrapper version automatically to its preferred default, which is currently 6.7.1.  Are you having any issues building app with the current wrapper?

PR #2437, opened 2021-05-13
Convert FeedFragment to Kotliin

cooltey:


app/src/main/java/org/wikipedia/feed/view/FeedAdapter.java
```diff
@@ -1,139 +0,0 @@
...
-import org.wikipedia.feed.offline.OfflineCard;
-import org.wikipedia.feed.offline.OfflineCardView;
-import org.wikipedia.feed.random.RandomCardView;
-import org.wikipedia.feed.searchbar.SearchCardView;
-import org.wikipedia.feed.suggestededits.SuggestedEditsCardView;
-import org.wikipedia.util.DimenUtil;
-import org.wikipedia.views.DefaultRecyclerAdapter;
-import org.wikipedia.views.DefaultViewHolder;
-
-public class FeedAdapter<T extends View & FeedCardView<?>> extends DefaultRecyclerAdapter<Card, T> {
```
cooltey, line 29: Somehow if we keep multiple types for the generic type specification, we'll not able to create a `FeedAdapter` in `FeedFragment` with only one type. Because of it, I change it to only extends `View` here.

PR #2443, opened 2021-05-14
Convert UtcDate model to Kotlin

anticafe:


app/src/main/java/org/wikipedia/feed/model/UtcDate.kt
```diff
@@ -0,0 +1,42 @@
...
+    private val cal: Calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"))
+    private val year: String
+    private val month: String
+    private val date: String
+
+    fun baseCalendar(): Calendar {
+        return cal
+    }
+
+    fun year(): String {
```
dbrant, line 16: None of these getter methods are necessary. That's one of the benefits of converting to Kotlin.

app/src/main/java/org/wikipedia/feed/model/UtcDate.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    init {
+        cal.add(Calendar.DATE, -age)
+        year = cal[Calendar.YEAR].toString()
+        month = pad((cal[Calendar.MONTH] + 1).toString())
+        date = pad(cal[Calendar.DATE].toString())
+    }
+
+    companion object {
+        private fun pad(value: String): String {
```
dbrant, line 36: This function is no longer necessary. Kotlin has this function built in.

app/src/main/java/org/wikipedia/feed/model/UtcDate.kt
```diff
@@ -0,0 +1,42 @@
...
+        return month
+    }
+
+    fun date(): String {
+        return date
+    }
+
+    init {
+        cal.add(Calendar.DATE, -age)
+        year = cal[Calendar.YEAR].toString()
```
dbrant, line 30: These initialization bits don't need to be here. They can be made into `get()` properties inline with the actual fields.

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/model/UtcDate.kt
anticafe: Updated via 4ae3b1e

app/src/main/java/org/wikipedia/feed/model/UtcDate.kt
anticafe: Thanks for suggestion. IDK Kotlin has built-in `padStart` method. 
Updated via 4ae3b1e

app/src/main/java/org/wikipedia/feed/model/UtcDate.kt
anticafe: Updated via 4ae3b1e

PR #2445, opened 2021-05-15
Added Option To Enable And Disable Remembering Of Search And Article History

nickthijssen1994:
I like to have the option to disable the search and article history. I added the option to disable your search history and I'm working on doing the same for article history. You can enable/disable the search history under Privacy in Settings.


nickthijssen1994:
Also added the option for article history. Disabling the options now asks for confirmation and deletes existing histories when choosing "Yes".

PR #2446, opened 2021-05-16
Use List extensions.

Isira-Seneviratne:


cooltey:
Thanks for creating this PR. It is excellent.

PR #2448, opened 2021-05-22
Added Option To Enable And Disable Remembering Of Search And Article History

nickthijssen1994:
_I'm creating this pull request again, because I didn't create a separate branch for it earlier._

I added an option to disable the search and article history. You can enable/disable these separately under Privacy in Settings. When you disable one of the two options, a dialog is shown to ask for confirmation. When choosing "Yes", the existing history is deleted and new searches or viewed articles will not be remembered. When the options are enabled, the app works like before.


dbrant:
Thanks! This feature works fine, but I'm not convinced it's necessary for the general production version of the app.
We try to strike a balance between providing essential customizations in the app, but not cluttering the app with too finely-grained settings.

We have [an item](https://phabricator.wikimedia.org/T272914) in our bug tracking system that corresponds to this feature request. I encourage you to go there and make a case for why, and for whom, this feature would be useful.  This way, other members of the team, especially our Designer, will be able to comment on the best UX for it.

dbrant:
closing per above comment.

PR #2453, opened 2021-05-24
Remove unnecessary check in push logic.

dbrant:


app/src/extra/java/org/wikipedia/push/WikipediaFirebaseMessagingService.kt
```diff
@@ -23,12 +23,8 @@ class WikipediaFirebaseMessagingService : FirebaseMessagingService() {
     override fun onMessageReceived(remoteMessage: RemoteMessage) {
         L.d("Message from: ${remoteMessage.from}")
 
-        remoteMessage.data.isNotEmpty().let {
```
dbrant, line 26: This check is actually meaningless:  this puts a `let` statement onto a Boolean return value, so even if `data` is empty, the `let` block will execute with `it=false`

Even if the check was correct, it would be unnecessary. It's perfectly fine to just call `containsValue` on an empty map.

PR #2457, opened 2021-05-25
Convert AggregatedFeedContent related classes to Kotlin

cooltey:


app/src/main/java/org/wikipedia/feed/FeedCoordinatorBase.java
```diff
@@ -301,7 +300,6 @@ private boolean isDailyCardType(@NonNull Card card) {
 
     private boolean shouldShowProgressCard(@NonNull FeedClient pendingClient) {
         return pendingClient instanceof SuggestedEditsFeedClient
-                || pendingClient instanceof AggregatedFeedContentClient
```
sharvaniharan, line 304: @cooltey why was this removed?

cooltey: Because `AggregatedFeedContentClient` is not extended from `FeedClient`, the logic is unnecessary here at the beginning.

PR #2468, opened 2021-05-26
Correctly show parsed block message in Suggested Edits screen.

dbrant:
https://phabricator.wikimedia.org/T276149

This is the final step that should close out the task of showing properly parsed block messages. Here is how a block now looks in the Edits screen:

image


sharvaniharan (approved):
So much better than before!

PR #2477, opened 2021-05-27
Use Core KTX extensions.

Isira-Seneviratne:


dbrant:
Sorry, but yet again this is more style over substance.

PR #2478, opened 2021-05-27
Use Kotlin IO extensions.

Isira-Seneviratne:


dbrant:
This, on the other hand, is a meaningful simplification.

PR #2482, opened 2021-05-28
Bounce out Special pages if they come from an external intent.

dbrant:
https://phabricator.wikimedia.org/T283713

This is an adequate solution for bouncing Special pages which only come from external `VIEW` intents, meaning it came from the external browser app.  This means that Special pages will still work in the app if they are linked internally, e.g. Article History and others.


app/src/main/java/org/wikipedia/page/PageFragment.java
```diff
@@ -489,6 +489,7 @@ public void onPageFinished(WebView view, String url) {
                     // page has now loaded and trigger the remaining logic ourselves.
                     if (!"true".equals(pcsExists)) {
                         onPageSetupEvent();
+                        bridge.onMetadataReady();
```
dbrant, line 492: This call is necessary for the MobileWeb shim javascript code to be executed when the page loads. For some reason this was missing in the latest code.

PR #2483, opened 2021-05-28
Use APIs introduced in Core 1.5.0.

Isira-Seneviratne:
Make use of APIs added in Core 1.5.0, such as [`WindowInsetsControllerCompat`](https://developer.android.com/reference/androidx/core/view/WindowInsetsControllerCompat).


dbrant:
Sorry, this is not an improvement. A function called `hideSoftKeyboard()` is infinitely more readable than `windowInsetsControllerCompat?.hide(WindowInsetsCompat.Type.ime()`

PR #2490, opened 2021-06-01
Follow up: Constants Kotlin conversion

cooltey:
Make the `const val` and `enum entry` in alphabetical order.


app/src/main/java/org/wikipedia/Constants.kt
```diff
@@ -4,98 +4,98 @@ object Constants {
 
     // Keep loader IDs unique to each loader. If the loader specified by the ID already exists, the
```
dbrant, line 5: @cooltey  What does this comment refer to now?

PR #2491, opened 2021-06-01
Convert Section to Kotlin

cooltey:


app/src/main/java/org/wikipedia/page/Section.kt
```diff
@@ -0,0 +1,44 @@
+package org.wikipedia.page
+
+import com.google.gson.annotations.SerializedName
+import org.wikipedia.json.GsonUtil
+import java.util.*
+
+data class Section(var id: Int = 0,
```
sharvaniharan, line 7: Can we have the parameters in the same line and only break to newline when the char limit on the same line is reached?

sharvaniharan (changes requested):

app/src/main/java/org/wikipedia/page/Section.kt
cooltey: Not sure but I am following the way we convert in the past for the `data class`.  I am also open to update to the way you've suggested.
https://github.com/wikimedia/apps-android-wikipedia/blob/2923c466f991e96f1e0a5119c15d510d08a19e6a/app/src/main/java/org/wikipedia/page/references/PageReferences.kt
https://github.com/wikimedia/apps-android-wikipedia/blob/2923c466f991e96f1e0a5119c15d510d08a19e6a/app/src/main/java/org/wikipedia/page/references/PageReferences.kt

PR #2506, opened 2021-06-03
Blocked error messages: towards an ultimate solution.

dbrant:
This uses some exotic API manipulation to get the wiki-specific Blocked error page, and populate it with the correct parsed parameters (no longer requiring us to use a local string resource).
The only downside of this method is that we need to use three (!) API requests to get all the necessary information. Perhaps this can be streamlined in a future update.


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -204,7 +208,7 @@ class SuggestedEditsTasksFragment : Fragment() {
                 }
                 .observeOn(AndroidSchedulers.mainThread())
                 .doAfterTerminate {
-                    if (blockInfo != null) {
+                    if (blockMessage != null) {
```
cooltey, line 211: Would it be better if we use `!blockMessage.isNullOrEmpty()`?

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -213,7 +217,7 @@ class SuggestedEditsTasksFragment : Fragment() {
                         isPausedOrDisabled = true
                     }
 
-                    if (!isPausedOrDisabled && blockInfo == null) {
+                    if (!isPausedOrDisabled && blockMessage == null) {
```
cooltey, line 220: `blockMessage.isNullOrEmpty()`?

app/src/main/java/org/wikipedia/util/ThrowableUtil.kt
```diff
@@ -91,26 +94,30 @@ object ThrowableUtil {
...
-        val app = WikipediaApp.getInstance()
-        var retStr = "<b>${app.getString(R.string.error_blocked_title)}</b>"
-        val title = PageTitle(UserAliasData.valueFor(app.appOrSystemLanguageCode), info.blockedBy, app.wikiSite)
+    @WorkerThread
+    fun getBlockMessageHtml(blockInfo: MwServiceError.BlockInfo): String {
+        var html = ""
+        Observable.zip(ServiceFactory.get(WikipediaApp.getInstance().wikiSite).userInfo,
+            ServiceFactory.get(WikipediaApp.getInstance().wikiSite).parsePage("MediaWiki:Blockedtext"),
+            ServiceFactory.get(WikipediaApp.getInstance().wikiSite).parseText(blockInfo.blockReason),
+            { userInfoResponse: MwQueryResponse, blockedParseResponse: MwParseResponse, reasonParseResponse: MwParseResponse ->
```
cooltey, line 103: Explicit type specifications can be removed.

cooltey (changes requested):

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
dbrant: correct!

app/src/main/java/org/wikipedia/util/ThrowableUtil.kt
dbrant: done!

PR #2511, opened 2021-06-04
Kotlin Conversion: LanguageUtil

sharvaniharan:


app/src/main/java/org/wikipedia/language/LanguageUtil.kt
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.commons.lang3.StringUtils
+import org.wikipedia.WikipediaApp
+import org.wikipedia.util.StringUtil
+import java.util.*
+
+object LanguageUtil {
+
+    private const val HONG_KONG_COUNTRY_CODE = "HK"
+    private const val MACAU_COUNTRY_CODE = "MO"
+    private val TRADITIONAL_CHINESE_COUNTRY_CODES = mutableListOf(Locale.TAIWAN.country, HONG_KONG_COUNTRY_CODE, MACAU_COUNTRY_CODE)
```
cooltey, line 16: The list should be non-mutable and can use `listOf()`.

app/src/main/java/org/wikipedia/language/LanguageUtil.kt
```diff
@@ -0,0 +1,142 @@
...
+    @JvmStatic
+    val firstSelectedChineseVariant: String
+        get() {
+            var firstSelectedChineseLangCode: String? = null
+            for (langCode in WikipediaApp.getInstance().language().appLanguageCodes) {
+                if (langCode.startsWith(AppLanguageLookUpTable.CHINESE_LANGUAGE_CODE)) {
+                    firstSelectedChineseLangCode = langCode
+                    break
+                }
+            }
```
cooltey, line 125: Can we use `WikipediaApp.getInstance().language().appLanguageCodes.firstOrNull(...)` here?

app/src/main/java/org/wikipedia/language/LanguageUtil.kt
```diff
@@ -0,0 +1,142 @@
...
+            return StringUtils.defaultString(firstSelectedChineseLangCode,
+                AppLanguageLookUpTable.TRADITIONAL_CHINESE_LANGUAGE_CODE)
```
cooltey, line 127: Maybe use `firstSelectedChineseLangCode.orEmpty().ifEmpty {  }`?

app/src/main/java/org/wikipedia/language/LanguageUtil.kt
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+    fun startsWithArticle(text: String, language: String): Boolean {
+        val first = text.split(" ".toRegex()).toTypedArray()[0].lowercase(Locale.getDefault()).trim { it <= ' ' }
```
cooltey, line 131: We may just use `.trim()` instead of `.trim { it <= ' ' }`

cooltey (changes requested):

app/src/main/java/org/wikipedia/language/LanguageUtil.kt
sharvaniharan: isnt this cleaner?

PR #2514, opened 2021-06-05
Optimize Kotlin code style in DescriptionEdit classes

cooltey:
Clean up unnecessary `!!` tags and update regular expression usage.


app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -375,7 +373,7 @@ class DescriptionEditFragment : Fragment() {
     }
 
     private fun updateDescriptionInArticle(articleText: String, newDescription: String): String {
-        return if (templateParsePattern.matcher(articleText).find()) {
+        return if (TEMPLATE_PARSE_REGEX.toRegex().matches(articleText)) {
```
dbrant, line 376: When making any kind of change to a feature, please try _using_ the feature and make sure it still works.

dbrant (changes requested):

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
cooltey: Fixed and done.

PR #2515, opened 2021-06-05
Remove SwitchPreferenceWithLinks and minor update of SwitchPreferenceMultiLine

cooltey:


dbrant (changes requested):
It looks like `SwitchPreferenceWithLinks` is totally unused. Should we just remove it?

cooltey:
> It looks like `SwitchPreferenceWithLinks` is totally unused. Should we just remove it?

It makes sense to me. Done.

PR #2517, opened 2021-06-07
Remove unused compat class.

dbrant:
there was... no reason for it?
This also lets us use try-with-resources in the places where it was used.


dbrant:
whoops, nevermind, the SDK just didn't have a minimum API annotation.  This will be removed when we remove ContentProviders altogether.

PR #2523, opened 2021-06-07
Follow-up to kotlin regex cleanup.

dbrant:


cooltey:
Excellent! I wasn't able to find the correct suppress tag for it.

PR #2526, opened 2021-06-08
Minor clean-up

sharvaniharan:


app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -270,15 +270,15 @@ class MainFragment : Fragment(), BackPressedHandler, FeedFragment.Callback, Hist
         }
     }
 
-    override fun onFeedMovePageToList(sourceReadingListId: Long, entry: HistoryEntry) {
+    override fun onFeedMovePageToList(sourceReadingList: Long, entry: HistoryEntry) {
         bottomSheetPresenter.show(childFragmentManager,
-                MoveToReadingListDialog.newInstance(sourceReadingListId, entry.title, InvokeSource.FEED))
+                MoveToReadingListDialog.newInstance(sourceReadingList, entry.title, InvokeSource.FEED))
     }
```
cooltey, line 276: It would be better if it is renamed on the `interface` side, which is following:

```
fun onFeedMovePageToList(sourceReadingListId: Long, entry: HistoryEntry)
```

app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -253,13 +253,13 @@ class MainFragment : Fragment(), BackPressedHandler, FeedFragment.Callback, Hist
         }
     }
 
-    override fun onFeedSelectPageWithAnimation(entry: HistoryEntry, sharedElements: Array<Pair<View, String>>) {
```
cooltey, line 256: It would be better if it is renamed on the `interface` side, which is following:
```
fun onFeedSelectPageWithAnimation(entry: HistoryEntry, sharedElements: Array<Pair<View, String>>)
```

cooltey (changes requested):

PR #2528, opened 2021-06-08
Convert CardType to Kotlin and update type name

cooltey:
Change `TrendingArticles` to `MostReadArticles`


app/src/main/java/org/wikipedia/feed/FeedContentType.kt
```diff
@@ -27,9 +27,9 @@ enum class FeedContentType(private val code: Int,
             return if (isEnabled) AggregatedFeedContentClient.FeaturedArticle(aggregatedClient) else null
         }
     },
-    TRENDING_ARTICLES(3, R.string.view_top_read_card_title, R.string.feed_item_type_trending, true) {
```
sharvaniharan, line 30: If we are renaming, should we be. more accurate and rename to "TOP_READ_ARTICCLES" per what we display?

sharvaniharan (changes requested):

app/src/main/java/org/wikipedia/feed/FeedContentType.kt
cooltey: You are correct! I did further rename on related classes and package and it all makes sense now. Thanks!

PR #2533, opened 2021-06-09
Kotlinx Serialization

sharvaniharan:


Isira-Seneviratne:
Couldn't [Moshi](https://github.com/square/moshi) be used instead?

dbrant:
> Couldn't [Moshi](https://github.com/square/moshi) be used instead? It was [created](https://medium.com/square-corner-blog/moshi-another-json-processor-624f8741f703) by some developers who worked on Gson, and it also supports Kotlin.

ok, but `kotlinx.serialization` was created by the developers of kotlin.  In what ways is Moshi better?

Isira-Seneviratne:
Moshi works similarly to Gson (albeit with [some differences](https://github.com/square/moshi#borrows-from-gson)), and Moshi also provides the ability to generate JSON adapters at compile time instead of relying on reflection.

sharvaniharan:
#2677

PR #2534, opened 2021-06-09
Convert PageViewMode to Kotlin

cooltey:


app/src/main/java/org/wikipedia/page/PageViewModel.kt
```diff
@@ -0,0 +1,21 @@
...
+class PageViewModel {
+    var page: Page? = null
+    var title: PageTitle? = null
+    var curEntry: HistoryEntry? = null
+    var readingListPage: ReadingListPage? = null
+    var hasWatchlistExpiry = false
+    var isWatched = false
+    var forceNetwork = false
+    val isInReadingList get() = readingListPage != null
+    val cacheControl get() = if (isInReadingList) OkHttpConnectionFactory.CACHE_CONTROL_FORCE_NETWORK else OkHttpConnectionFactory.CACHE_CONTROL_NONE
```
dbrant, line 16: @cooltey  This changes the logic of how `cacheControl` is done. Previously it was based on `forceNetwork`, but now it's based on `isInReadingList`, which makes `forceNetwork` unused.  Was there a reason for this change?

cooltey: Good catch! This was made by accident. I have a PR of fixing it.

PR #2542, opened 2021-06-12
Kotlin Conversion: ServiceFactory

sharvaniharan:


app/src/main/java/org/wikipedia/dataclient/ServiceFactory.kt
```diff
@@ -0,0 +1,121 @@
...
+            .client(okHttpClientBuilder.build())
+            .baseUrl(baseUrl)
+            .client(client.newBuilder().addInterceptor(LanguageVariantHeaderInterceptor(wiki)).build())
+            .addCallAdapterFactory(RxJava3CallAdapterFactory.create())
+            .addConverterFactory(GsonConverterFactory.create(GsonUtil.getDefaultGson()))
+            .build()
+    }
+
+    private class LanguageVariantHeaderInterceptor(private val wiki: WikiSite?) :
+        Interceptor {
```
cooltey, line 107: Could these be in one line?

app/src/main/java/org/wikipedia/dataclient/ServiceFactory.kt
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @JvmStatic
+    fun getAnalyticsRest(streamConfig: StreamConfig): EventService {
+        val destinationEventService = streamConfig.destinationEventService
+        if (ANALYTICS_REST_SERVICE_CACHE[destinationEventService.id] != null) {
+            return ANALYTICS_REST_SERVICE_CACHE[destinationEventService.id]!!
+        }
+        val intakeBaseUriOverride = Prefs.getEventPlatformIntakeUriOverride()
+        val r = createRetrofit(null, StringUtils.defaultString(intakeBaseUriOverride, destinationEventService.baseUri))
```
cooltey, line 68: Considering using `orEmpty().ifEmpty {} `? e.g.: `Prefs.getEventPlatformIntakeUriOverride().orEmpty().ifEmpty { destinationEventService.baseUri }`

cooltey (changes requested):

app/src/main/java/org/wikipedia/dataclient/ServiceFactory.kt
```diff
@@ -0,0 +1,119 @@
...
+            path += "/"
+        }
+        return path
+    }
+
+    private fun createRetrofit(wiki: WikiSite?, baseUrl: String): Retrofit {
+        val okHttpClientBuilder = client.newBuilder()
+        if (wiki != null) {
+            okHttpClientBuilder.addInterceptor(LanguageVariantHeaderInterceptor(wiki))
+        }
```
cooltey, line 95: Very curious that why we have to add the `LanguageVariantHeaderInterceptor` here, and then add it again below? Do you have any reason why we are doing this?

app/src/main/java/org/wikipedia/dataclient/ServiceFactory.kt
```diff
@@ -0,0 +1,119 @@
...
+        return Retrofit.Builder()
+            .client(okHttpClientBuilder.build())
+            .baseUrl(baseUrl)
+            .client(client.newBuilder().addInterceptor(LanguageVariantHeaderInterceptor(wiki)).build())
+            .addCallAdapterFactory(RxJava3CallAdapterFactory.create())
+            .addConverterFactory(GsonConverterFactory.create(GsonUtil.getDefaultGson()))
+            .build()
+    }
+
+    private class LanguageVariantHeaderInterceptor(private val wiki: WikiSite?) : Interceptor {
```
cooltey, line 105: Looks like if the `wiki` is null, then the entire process is unreachable. (or maybe will lead to a crash). Should we move the null check logic into this method?

cooltey (changes requested):

app/src/main/java/org/wikipedia/dataclient/ServiceFactory.kt
sharvaniharan: Good catch @cooltey ! done!

PR #2548, opened 2021-06-15
Convert OfflineCacheInterceptor to Kotlin

cooltey:


app/src/main/java/org/wikipedia/dataclient/okhttp/OfflineCacheInterceptor.java
```diff
@@ -1,351 +0,0 @@
...
-        public BufferedSource source() {
-            try {
-                return Okio.buffer(Okio.source(file));
-            } catch (Exception e) {
-                return null;
-            }
-        }
-    }
-
-    // TODO: Remove after 2 releases
```
cooltey, line 310: This is removed when doing the conversion.

PR #2550, opened 2021-06-15
Convert WikiSiteTypeAdapter to Kotlin

cooltey:


app/src/main/java/org/wikipedia/json/WikiSiteTypeAdapter.kt
```diff
@@ -0,0 +1,47 @@
...
+        while (reader.hasNext()) {
+            val field = reader.nextName()
+            val value = reader.nextString()
+            when (field) {
+                DOMAIN -> domain = value
+                LANGUAGE_CODE -> languageCode = value
+            }
+        }
+        reader.endObject()
+        if (domain == null) {
```
sharvaniharan, line 36: Can we use .NullOrEmpty()?

sharvaniharan (changes requested):

app/src/main/java/org/wikipedia/json/WikiSiteTypeAdapter.kt
cooltey: Done!

PR #2554, opened 2021-06-18
Added space between two resource strings

NimrodRak:
Fixed issue where there was a space missing between two of the strings in an offline reading message.
Bug: T285065


app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewErrorView.kt
```diff
@@ -29,6 +29,7 @@ class LinkPreviewErrorView : LinearLayout {
 
         if (errorType === LinkPreviewErrorType.OFFLINE) {
             val message = (resources.getString(R.string.page_offline_notice_cannot_load_while_offline) +
+                    " " +
```
dbrant, line 32: Instead of a space, let's make this a newline character (`\n`)

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewErrorView.kt
NimrodRak: Surely! Will change it by the end of the day.

dbrant: No, not really, unless we want to turn it into two TextViews that appear next to each other, but that's not necessary for now.

NimrodRak: Done.

PR #2558, opened 2021-06-21
Kotlin Conversion: MwQueryResponse

sharvaniharan:


app/src/main/java/org/wikipedia/dataclient/mwapi/MwQueryResponse.kt
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    fun success(): Boolean {
+        return query != null
+    }
+
+    @VisibleForTesting
+    protected fun setQuery(query: MwQueryResult?) {
+        this.query = query
+    }
```
cooltey, line 34: Cannot find the usage here. Is it really being used?

app/src/main/java/org/wikipedia/dataclient/mwapi/MwQueryResponse.kt
```diff
@@ -0,0 +1,35 @@
...
+        return continuation ?: emptyMap()
+    }
+
+    fun query(): MwQueryResult? {
+        return query
+    }
+
+    fun success(): Boolean {
+        return query != null
+    }
```
cooltey, line 29: Looks like it can be removed since only test cases use it and can be replaced.

app/src/main/java/org/wikipedia/dataclient/mwapi/MwQueryResponse.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.google.gson.annotations.SerializedName
+
+open class MwQueryResponse : MwResponse() {
+
+    @SerializedName("batchcomplete")
+    private val batchComplete = false
+
+    @SerializedName("continue")
+    private val continuation: Map<String, String>? = null
+    private var query: MwQueryResult? = null
```
cooltey, line 13: How about change to following:
```
    @SerializedName("batchcomplete")
    val batchComplete = false

    @SerializedName("continue")
    val continuation: Map<String, String>? = null
        get() = field ?: emptyMap()
    
    var query: MwQueryResult? = null
```

and remove the following method?

cooltey (changes requested):

PR #2562, opened 2021-06-22
Use Handler.postDelayed() and View.postDelayed() extensions.

Isira-Seneviratne:


cooltey (changes requested):
I don't see a strong reason for making the change from `Handler.postDelayed()` to `View.postDelayed()` which `View.postDelayed()` is simply using `Handler.postDelayed()` on an internal handler.

dbrant:
Neither do I. This seems to be (again) arbitrary style preferences.

PR #2563, opened 2021-06-22
Follow up: ServiceFactory conversion

cooltey:


app/src/main/java/org/wikipedia/dataclient/ServiceFactory.kt
```diff
@@ -102,10 +99,10 @@ object ServiceFactory {
     private class LanguageVariantHeaderInterceptor(private val wiki: WikiSite?) : Interceptor {
         @Throws(IOException::class)
         override fun intercept(chain: Interceptor.Chain): Response {
-            var request: Request = chain.request()
+            var request = chain.request()
 
             // TODO: remove when the https://phabricator.wikimedia.org/T271145 is resolved.
-            if (wiki != null && !request.url.encodedPath.contains("/page/related")) {
```
cooltey, line 108: Looks like we don't really need a null check here because the `WikipediaApp.getInstance().getAcceptLanguage(wiki)` accepts `null`.

PR #2564, opened 2021-06-23
Show undo snackbar in talk topic page after making a response.

cooltey:
https://phabricator.wikimedia.org/T284532


app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
```diff
@@ -323,6 +324,19 @@ class TalkTopicActivity : BaseActivity(), LinkPreviewDialog.Callback {
                 }))
     }
 
+    private fun undoSave() {
+        disposables.add(CsrfTokenClient(pageTitle.wikiSite).token
+            .subscribeOn(Schedulers.io())
+            .flatMap { token -> ServiceFactory.get(pageTitle.wikiSite).postUndoEdit(pageTitle.prefixedText, currentRevision, token) }
```
dbrant, line 330: We should not be undoing the `currentRevision`, since it's not guaranteed to be the revision that you submitted. (Someone else might have made an edit immediately after you did.)

We should create a new private field such as `revisionForUndo` and update it directly from the edit response, and then use it for undoing.

app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
```diff
@@ -356,18 +370,22 @@ class TalkTopicActivity : BaseActivity(), LinkPreviewDialog.Callback {
...
         binding.replySaveButton.isEnabled = true
         editFunnel.logSaved(newRevision)
 
+        if (showUndoSnackbar) {
+            showUndoSnackbar()
+        }
+
         if (isNewTopic()) {
             setResult(RESULT_EDIT_SUCCESS)
             finish()
```
dbrant, line 391: What happens if this branch is reached? i.e. if the user creates a "new topic" instead of replying to an existing topic?

dbrant (changes requested):

app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
cooltey: You are correct! Done!

app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
cooltey: Done!

dbrant: @cooltey I'm not sure this is resolved:  to clarify, we _should_ allow the user to undo a new topic.

cooltey: Got it. Done!

PR #2568, opened 2021-06-23
Fix: use the correct anchor in edit summary in zh.wiki

cooltey:
https://phabricator.wikimedia.org/T70265

**Steps to reproduce**
1. Go to https://zh.wikipedia.org/wiki/User:Cooltey/Sandbox on `zh-hant` wiki.
2. Change the language to `zh-hans`
3. Edit the last section `正體 中文 國家` and submit.
4. Go to https://zh.wikipedia.org/w/index.php?title=User:Cooltey/Sandbox&action=history to see if the edit summary contains the correct link to https://zh.wikipedia.org/wiki/User:Cooltey/Sandbox#正體_中文_國家


app/src/main/java/org/wikipedia/edit/EditSectionActivity.kt
```diff
@@ -202,8 +202,8 @@ class EditSectionActivity : BaseActivity() {
     }
 
     private fun doSave(token: String) {
-        var summaryText = if (sectionHeading.isNullOrEmpty() ||
-            StringUtil.addUnderscores(sectionHeading) == pageTitle.prefixedText) "/* top */" else "/* $sectionHeading */ "
+        var summaryText = if (sectionAnchor.isNullOrEmpty() || sectionAnchor == pageTitle.prefixedText) "/* top */"
```
dbrant, line 205: This comparison needs `addUnderscores(sectionAnchor)`

dbrant (changes requested):

app/src/main/java/org/wikipedia/edit/EditSectionActivity.kt
cooltey: The `sectionAnchor` has already contained underscores, should I still need to apply the `addUnderscores`?

dbrant: When I try to edit an article with spaces in the title, the `sectionAnchor` still has spaces, not underscores.  And in fact when I edit an article that has a title with HTML in it, such as today's featured article (_Grevillea juniperina_), the `sectionAnchor` has HTML tags in it.

cooltey: Got it. Thanks for checking. Done.

PR #2570, opened 2021-06-23
Kotlin Conversion: CreateAccountResponse

sharvaniharan:


app/src/main/java/org/wikipedia/dataclient/mwapi/CreateAccountResponse.kt
```diff
@@ -0,0 +1,32 @@
...
+        return createaccount!!.status
+    }
+
+    fun user(): String? {
+        return createaccount!!.username
+    }
+
+    fun message(): String? {
+        return createaccount!!.message
+    }
```
cooltey, line 17: Would it be possible to change the format to `val status get() = createaccount?.status`. And some null-related checks maybe need to revisit where they are currently on.

app/src/main/java/org/wikipedia/dataclient/mwapi/CreateAccountResponse.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    fun hasResult(): Boolean {
+        return createaccount != null
+    }
```
cooltey, line 21: Looks like this can be removed.

cooltey (changes requested):

PR #2572, opened 2021-06-24
Remove use of the deprecated SOFT_INPUT_ADJUST_RESIZE flag.

Isira-Seneviratne:


Isira-Seneviratne:
Closing this, as the use of the flag is also removed in https://github.com/wikimedia/apps-android-wikipedia/pull/2577.

PR #2576, opened 2021-06-24
Kotlin Conversion: Claims

sharvaniharan:


app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.kt
```diff
@@ -0,0 +1,101 @@
...
+import org.wikipedia.dataclient.mwapi.MwResponse
+import org.wikipedia.json.GsonUtil
+
+class Claims : MwResponse() {
+
+    private val claims: Map<String, List<Claim>>? = null
+
+    fun claims(): Map<String, List<Claim>> {
+        return claims ?: emptyMap()
+    }
```
cooltey, line 15: This can be reduced by adding `get() = field ?: emptyMap()` to the `val claims`

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    class Claim {
+
+        private val type: String? = null
+        private val id: String? = null
+        private val rank: String? = null
```
cooltey, line 21: Unused variables may be safe to be removed.

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    class MainSnak {
+
+        @SerializedName("snaktype")
+        private val snakType: String? = null
+        @SerializedName("datatype")
+        private val dataType: String? = null
+        private val property: String? = null
+        private val hash: String? = null
```
cooltey, line 33: Unused variables may be safe to be removed.

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.kt
```diff
@@ -0,0 +1,101 @@
...
+        val dataValue: DataValue? = null
+    }
+
+    class DataValue {
+
+        private val value: JsonElement? = null
+        private val type: String? = null
+
+        fun value(): String {
+            if (value != null) {
```
cooltey, line 44: This may be always `null` since you are using the `value` when the `DataValue` object is created. A proper way of doing this could be to put the logic to `get() { .... }`
```
val value: JsonElement? = null
     get() { .... }

```

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.kt
```diff
@@ -0,0 +1,101 @@
...
+        private val precision = 0.0
+        val location: Location
+            get() {
+                val loc = Location("")
+                loc.latitude = latitude
+                loc.longitude = longitude
+                loc.altitude = altitude
+                return loc
+            }
+    }
```
cooltey, line 100: Some unused variables may be removed.

cooltey (changes requested):

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.kt
sharvaniharan: I feel we should leave the unused variables as they give a good idea of what is available for use, and the cost of having them is not much.

cooltey: I see. It makes sense to me.

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.kt
cooltey, line 15: This can be one line.

`private val claims: Map<String, List<Claim>> = emptyMap()`

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.kt
```diff
@@ -0,0 +1,101 @@
...
+                }
+            }
+            return ""
+        }
+    }
+
+    class EntityIdValue {
+
+        val id: String? = null
+            get() = field.orEmpty()
```
cooltey, line 66: This can be `val id: String = ""`

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    class TimeValue {
+
+        private val timezone = 0
+        private val before = 0
+        private val after = 0
+        private val precision = 0
+        val time: String? = null
+            get() = field.orEmpty()
```
cooltey, line 76: This can be `val time: String = ""`

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    class MonolingualTextValue {
+
+        private val language: String? = null
+        val text: String? = null
+            get() = field.orEmpty()
```
cooltey, line 83: This can be `val text: String = ""`

cooltey (changes requested):

app/src/main/java/org/wikipedia/dataclient/wikidata/Claims.kt
sharvaniharan: We are getting the string value from JSON @cooltey . Not sure how to simplify.

cooltey: Ah! Yes you are correct. My bad.

PR #2578, opened 2021-06-25
Kotlin Conversion: Search

sharvaniharan:


app/src/main/java/org/wikipedia/dataclient/wikidata/Search.kt
```diff
@@ -0,0 +1,19 @@
+package org.wikipedia.dataclient.wikidata
+
+import org.wikipedia.dataclient.mwapi.MwResponse
+
+class Search : MwResponse() {
+
+    private val success = 0
+    val results: List<SearchResult>? = null
```
cooltey, line 8: This didn't work since the name should be `search` which is from the API.

cooltey (changes requested):
This entire class can be simplified to:

```
class Search : MwResponse() {

    @SerializedName("search")
    val results: List<SearchResult> = emptyList()

    class SearchResult {
        val id: String = ""
        val label: String = ""
        val description: String = ""
    }
}
```

PR #2580, opened 2021-06-25
Remove AppCenter entirely.

dbrant:
For ideal compliance with F-Droid, it's probably best to remove AppCenter integration entirely, since technically a packet gets sent to the AppCenter servers when the app is first launched, before the user gets a chance to disable crash reporting.

This means that we won't get any crash reports from F-Droid users, but that should be fine since crash reports from the main Google Play install base are plenty.


cooltey (approved):
LGTM.

PR #2581, opened 2021-06-25
Kotlin Conversion: FeedAvailability

sharvaniharan:


app/src/main/java/org/wikipedia/feed/configure/FeedAvailability.kt
```diff
@@ -0,0 +1,26 @@
...
+    val featuredPicture: List<String>? = null
+        get() = field ?: emptyList()
+
+    @SerializedName("in_the_news")
+    val news: List<String>? = null
+        get() = field ?: emptyList()
+
+    @SerializedName("on_this_day")
+    val onThisDay: List<String>? = null
+        get() = field ?: emptyList()
```
cooltey, line 25: These can be simplified as:
```
val featuredArticle: List<String> = emptyList()

and so on...
```

cooltey (changes requested):

PR #2592, opened 2021-06-29
Enable pinch-to-zoom on Explore feed cards image

cooltey:
https://phabricator.wikimedia.org/T269241


app/src/main/java/org/wikipedia/feed/news/NewsItemView.kt
```diff
@@ -50,6 +51,7 @@ class NewsItemView(context: Context) : WikiCardView(context) {
         newsItem.thumb()?.let {
             binding.horizontalScrollListItemImage.visibility = VISIBLE
             binding.horizontalScrollListItemImage.loadImage(it)
+            ImageZoomHelper.setViewZoomable(binding.horizontalScrollListItemImage)
```
dbrant, line 54: It might be helpful to add the same kind of `onLongClickListener` override logic as we have in  `FeaturedArticleView` and `OnThisDayCardView`, so the the long-press is consumed if the user is in the middle of zooming the image.

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/news/NewsItemView.kt
cooltey: Makes sense. Done.

sharvaniharan (approved):
LGTM

PR #2606, opened 2021-07-01
Update news card minimum height to avoid jump when swiping

cooltey:
https://phabricator.wikimedia.org/T268334


sharvaniharan:
@cooltey can we please make this a _design branch so that Robin can review and we can merge after... otherwise lgtm!

cooltey:
> @cooltey can we please make this a _design branch so that Robin can review and we can merge after... otherwise lgtm!

Sure.

PR #2611, opened 2021-07-02
Remove duplicate Kotlin methods.

Isira-Seneviratne:


dbrant:
Using explicit constructors for custom Views was a [conscious choice](https://medium.com/@mmlodawski/https-medium-com-mmlodawski-do-not-always-trust-jvmoverloads-5251f1ad2cfe) that we made.

PR #2618, opened 2021-07-10
Add setNavigationBarColor() method to DeviceUtil.

Isira-Seneviratne:


app/src/main/java/org/wikipedia/page/ExtendedBottomSheetDialogFragment.kt
```diff
@@ -14,15 +11,6 @@ open class ExtendedBottomSheetDialogFragment : BottomSheetDialogFragment() {
...
-            val isDarkThemeOrDarkBackground = (WikipediaApp.getInstance().currentTheme.isDark ||
-                    color == ContextCompat.getColor(requireContext(), android.R.color.black))
-
-            requireDialog().window?.run {
-                navigationBarColor = color
-                decorView.systemUiVisibility = if (isDarkThemeOrDarkBackground) decorView.systemUiVisibility and View.SYSTEM_UI_FLAG_LIGHT_NAVIGATION_BAR.inv()
-                else View.SYSTEM_UI_FLAG_LIGHT_NAVIGATION_BAR or decorView.systemUiVisibility
-            }
-        }
+        requireDialog().window?.let { DeviceUtil.setNavigationBarColor(it, color) }
```
cooltey, line 14: This `protected` method will be used in `ThemeChooserDialog`, which means you don't need to call `DeviceUtil.setNavigationBarColor()` twice in this case.

cooltey (changes requested):

cooltey:
Please resolve conflicts.

PR #2631, opened 2021-07-14
Update android.material version to 1.4.0

cooltey:
https://github.com/material-components/material-components-android/releases/tag/1.4.0

### DIFF
--- a/app/src/main/java/org/wikipedia/navtab/NavTabLayout.kt
+++ b/app/src/main/java/org/wikipedia/navtab/NavTabLayout.kt
@@ -1,16 +1,9 @@
 package org.wikipedia.navtab
 
 import android.content.Context
-import android.text.TextUtils
 import android.util.AttributeSet
 import android.view.Menu
-import android.view.ViewGroup
-import android.widget.TextView
-import androidx.core.view.forEach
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/navtab/NavTabLayout.kt
```diff
@@ -19,19 +12,5 @@ class NavTabLayout constructor(context: Context, attrs: AttributeSet) : BottomNa
...
-            (getChildAt(0) as ViewGroup).forEach {
-                val largeLabel = it.findViewById<TextView>(R.id.largeLabel)
-                largeLabel.ellipsize = TextUtils.TruncateAt.END
-                largeLabel.updatePadding(left = 0, right = 0)
-                val smallLabel = it.findViewById<TextView>(R.id.smallLabel)
-                smallLabel.ellipsize = TextUtils.TruncateAt.END
-                smallLabel.updatePadding(left = 0, right = 0)
-            }
-        }
     }
```
cooltey, line 15: They renamed the `ids` and not sure why we cannot use the new `ids`. However, by checking the source code, looks like the settings are default and it should be fine to be removed.

PR #2641, opened 2021-07-15
Fix: show local description edits stats correctly in Edits

cooltey:
https://phabricator.wikimedia.org/T286097


dbrant:
I feel like this might be overcomplicating things a bit.  Why are we only checking for "local description" edits?  i.e. why don't we just count all contributions from the user's home wiki?

cooltey:
> I feel like this might be overcomplicating things a bit. Why are we only checking for "local description" edits? i.e. why don't we just count all contributions from the user's home wiki?

We have the "article description" category in the contributions screen and that should be counted separately. If the user's home wiki edit numbers can exclude general article edits (edits that do not have `suggestededits` comment), then we can use it.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -187,22 +188,28 @@ class SuggestedEditsTasksFragment : Fragment() {
 
                     totalContributions += wikidataResponse.query?.userInfo()!!.editCount
                     totalContributions += commonsResponse.query?.userInfo()!!.editCount
+                    totalContributions += localDescriptionsContributions.size
```
dbrant, line 191: But couldn't we just say:
`totalContributions += homeSiteResponse.query?.userInfo()!!.editCount`

(and change the homeSite call to use `getUserContributions` instead of `userInfo`)
And wouldn't that automatically take care of https://phabricator.wikimedia.org/T287031 ?

cooltey: It makes sense to me. But I am also thinking that we should still keep the `localDescriptionsContributions` because we will need the article titles to get pagevIews in the past 30 days.

PR #2669, opened 2021-07-29
Use Kotlin extensions.

Isira-Seneviratne:


dbrant:
Sorry, this is changing too much at once. Please make PRs that are small, useful, and focused.

PR #2677, opened 2021-07-30
Kotlin serialization

sharvaniharan:


sharvaniharan:
@dbrant @cooltey This is now ready for review.. I would suggest waiting to review this until the conversion PRs are merged. I will also go over a couple things during our Thursday sync.

dbrant:
> @dbrant @cooltey This is now ready for review.. I would suggest waiting to review this until the conversion PRs are merged. I will also go over a couple things during our Thursday sync.

@sharvaniharan  This is looking promising, but it also looks like it's roughly half done.
* There are very many places where `@SerialName` should be specified, but isn't.  For example the `Announcement` class, `Revision`, `BlockInfo`, and more.
* Many things are not working: logging in, syncing reading lists, table-of-contents sections, etc.

Will you be finishing this PR?

PR #2679, opened 2021-07-30
Save composed text when post is undone

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T287594


app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
```diff
@@ -295,6 +305,7 @@ class TalkTopicActivity : BaseActivity(), LinkPreviewDialog.Callback {
     private fun onSaveClicked() {
         val subject = binding.replySubjectText.text.toString().trim()
         var body = binding.replyEditText.text.toString().trim()
+        undoneText = binding.replyEditText.text.toString().trim()
```
cooltey, line 308: Looks like it can be `undoneText = body` instead of getting the text from `replyEditText` again.

cooltey (changes requested):
Not sure but this should be also available when undoing the "New Topic".

sharvaniharan:
@cooltey I thought for a new topic, re-openiing the activity might be unnecessary, but let me confirm. Thankyou!

sharvaniharan:
@cooltey did that anyway... I don't think Jaz will object 😅

app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
```diff
@@ -466,14 +492,18 @@ class TalkTopicActivity : BaseActivity(), LinkPreviewDialog.Callback {
...
         private const val EXTRA_PAGE_TITLE = "pageTitle"
-        private const val EXTRA_TOPIC = "topicId"
+        const val EXTRA_TOPIC = "topicId"
+        const val EXTRA_SUBJECT = "subject"
+        const val EXTRA_BODY = "body"
         const val RESULT_EDIT_SUCCESS = 1
         const val RESULT_NEW_REVISION_ID = "newRevisionId"
 
-        fun newIntent(context: Context, pageTitle: PageTitle, topicId: Int, invokeSource: Constants.InvokeSource): Intent {
+        fun newIntent(context: Context, pageTitle: PageTitle, topicId: Int, invokeSource: Constants.InvokeSource, undoneSubject: String?, undoneBody: String?): Intent {
```
cooltey, line 501: You can set default ` = null` to the last two parameters so that you don't need to set `null` values in `TalkTopicsActivity.kt` and TalkTopicActivity.kt`

cooltey (changes requested):

PR #2681, opened 2021-07-31
Kotlin Conversion: DestinationEventService, StreamConfig, MwStreamCfigsResponse, SamplingConfig

sharvaniharan:


cooltey:
Please fix checkstyle.

app/src/main/java/org/wikipedia/analytics/eventplatform/SamplingConfig.kt
```diff
@@ -0,0 +1,38 @@
...
+ */
+ class SamplingConfig {
+
+    enum class Identifier {
+        @SerializedName("pageview")
+        PAGEVIEW,
+        @SerializedName("session")
+        SESSION,
+        @SerializedName("device")
+        DEVICE
```
cooltey, line 20: Would it be possible to have these as the following style:
```
@SerializedName("pageview") PAGEVIEW,
@SerializedName("session") SESSION,
@SerializedName("device") DEVICE
```

app/src/main/java/org/wikipedia/analytics/eventplatform/SamplingConfig.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    private var identifier: Identifier? = null
+    var rate = 1.0
+
+    // This constructor is needed for correct Gson deserialization. Do not remove!
+    constructor() {}
```
cooltey, line 27: Not sure but would it be possible to have this and the `identifier` and `rate` in the constructor as following:
```
class SamplingConfig(var rate: Double = 1.0,
                      var identifier: Identifier = Identifier.SESSION) {
    ...
}
```

app/src/main/java/org/wikipedia/analytics/eventplatform/SamplingConfig.kt
```diff
@@ -0,0 +1,38 @@
...
+import androidx.annotation.VisibleForTesting
+import com.google.gson.annotations.SerializedName
+
+/**
+ * Represents the sampling config component of a stream configuration.
+ *
+ * The boxed Double type is used instead of the double primitive because its value may be null,
+ * which denotes that the stream should always be *included*.
+ */
+ class SamplingConfig {
```
cooltey, line 12: There's an extra space at the beginning of the line.

cooltey (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/SamplingConfig.kt
sharvaniharan: Have left it as it because of the comment .... not so easy to test this as it is sampling config for MEP ... so can we leave it as is?

cooltey: Makes sense to me.

cooltey (approved):
Looks good to me. But I also noticed that the beta analytics server will send `502` back, does that mean we will need to update the `dev` MEP URL?

PR #2682, opened 2021-08-01
Use TextUtils.getLayoutDirectionFromLocale().

Isira-Seneviratne:


app/src/main/java/org/wikipedia/util/L10nUtil.kt
```diff
@@ -11,22 +12,16 @@ import org.wikipedia.page.PageTitle
...
-
-    private val RTL_LANGS = arrayOf(
-            "ar", "arc", "arz", "azb", "bcc", "bqi", "ckb", "dv", "fa", "glk", "he",
-            "khw", "ks", "lrc", "mzn", "nqo", "pnb", "ps", "sd", "ug", "ur", "yi"
-    )
-
     @JvmStatic
     val isDeviceRTL: Boolean
-        get() = isCharRTL(Locale.getDefault().displayName[0])
+        get() = TextUtils.getLayoutDirectionFromLocale(Locale.getDefault()) == View.LAYOUT_DIRECTION_LTR
```
cooltey, line 17: Should be `!=` instead of `==`.

cooltey (changes requested):

cooltey (approved):

PR #2685, opened 2021-08-02
Fix even more edge cases for database migration.

dbrant:
* Make the test parameterized, and give it versions `19` and `22` to test with.
* Roll the tests into a single `@Test` function, which makes it faster and doesn't lose any coverage.


app/src/main/java/org/wikipedia/database/AppDatabase.kt
```diff
@@ -86,8 +87,16 @@ abstract class AppDatabase : RoomDatabase() {
...
+                try {
+                    database.query("SELECT * FROM sqlite_master WHERE type='table' AND name='talkpageseen'").use {
+                        if (it.count > 0) {
+                            database.execSQL("INSERT OR REPLACE INTO TalkPageSeen_temp (sha) SELECT sha FROM talkpageseen")
+                            database.execSQL("DROP TABLE talkpageseen")
+                        }
+                    }
+                } catch (e: Exception) {
+                    // ignore further errors
+                }
```
cooltey, line 99: This makes sense to me, and would it be possible (or necessary?) to apply this logic to all other tables?

dbrant: @cooltey I'm not so sure...  I put these exception handlers only for the small remainder of crashes that I can't explain.  There haven't been any crashes in any of the other conversion statements, and if there is a problem in those statements I'd really like to see it via crash report. (Otherwise we risk data loss for the user.)

cooltey: Got it, that makes sense to me. It looks good to me so far, @sharvaniharan please review and merge if it looks good to you too.

PR #2688, opened 2021-08-03
Follow-up to showing all contributions

dbrant:
This does a few things:
* Fixes the edit streak shown in the Edits screen by taking into account edits from the user's home wiki.
* In the Contributions activity, it will now show all contributions from the user's home wiki (which we weren't doing yet??)

### DIFF
--- a/app/src/main/java/org/wikipedia/userprofile/ContributionsFragment.kt
+++ b/app/src/main/java/org/wikipedia/userprofile/ContributionsFragment.kt
@@ -31,6 +31,7 @@ import org.wikipedia.dataclient.Service
 import org.wikipedia.dataclient.ServiceFactory
 import org.wikipedia.dataclient.WikiSite
 import org.wikipedia.dataclient.mwapi.MwQueryResponse
+import org.wikipedia.diff.ArticleEditDetailsActivity
 import org.wikipedia.language.AppLanguageLookUpTable
 import org.wikipedia.userprofile.Contribution.Companion.EDIT_TYPE_ARTICLE_DESCRIPTION
 import org.wikipedia.userprofile.Contribution.Companion.EDIT_TYPE_GENERIC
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/userprofile/ContributionsFragment.kt
```diff
@@ -195,22 +193,38 @@ class ContributionsFragment : Fragment(), ContributionsHeaderView.Callback {
...
+                val cont = response.continuation["uccontinue"]
+                if (cont.isNullOrEmpty()) {
+                    continuations.remove(WikipediaApp.getInstance().wikiSite)
+                } else {
+                    continuations[WikipediaApp.getInstance().wikiSite] = cont
+                }
+                response.query?.userContributions()?.forEach {
+                    contributions.add(Contribution("", it.revid, it.title, it.title, it.title, EDIT_TYPE_GENERIC, null, it.date(),
+                        WikipediaApp.getInstance().wikiSite, 0, it.sizediff, it.top, 0))
+                }
```
cooltey, line 210: Looks like the order of all the contributions is showing incorrectly (not ordered by date).

cooltey (changes requested):

app/src/main/java/org/wikipedia/userprofile/ContributionsFragment.kt
dbrant: @cooltey  I can't reproduce. We sort all the contributions by date in the `createConsolidatedList()` function when building the list to be displayed.  Are you totally sure you saw your contributions out of order? Can you post a screenshot?

cooltey: Ah! My bad. I guess somehow I was confused by Jun and Jul. The order looks good to me now.

However, I am not sure about the edits streak. In my task screen it shows 5days edit streak, but in the contribution screen it doesn't look like I have a 5 days edit streak.

dbrant: Well, it looks like the streak logic in `getEditStreak()` is just wrong and needs to be fixed. (i.e. not related to this PR)

dbrant: @cooltey I made a quick fix that should improve the edit streak logic, although it's still naive and not expected to be perfect.

PR #2693, opened 2021-08-03
Optimize personalized search results

cooltey:
https://phabricator.wikimedia.org/T269947


sharvaniharan:
@cooltey I am getting these results when I follow the steps on the ticket:
film

And Christopher Noal filmography is not in my lists, personalized result icon suggests that it is..

cooltey:
> 
> 
> @cooltey I am getting these results when I follow the steps on the ticket:
> And Christopher Nolan filmography is not in my lists, personalized result icon suggests that it is..

Fixed! Thanks!

sharvaniharan:
Icon look good now... but I might have missed some update we made, but shouldn't it be only one result from history, readinglist and tab each?
one


I am seeing 2 results from history

cooltey:
> 
> 
> Icon look good now... but I might have missed some update we made, but shouldn't it be only one result from history, readinglist and tab each?
> I am seeing 2 results from history

That's weird, I cannot reproduce it. Could you please clear the cache and data and try again?

sharvaniharan (approved):
I was not able to repro too @cooltey ... Lgtm... Will let @dbrant test on his device once and merge!

PR #2695, opened 2021-08-05
Kotlin Conversion: PostProcessingTypeAdapter

sharvaniharan:


app/src/main/java/org/wikipedia/json/PostProcessingTypeAdapter.kt
```diff
@@ -0,0 +1,36 @@
...
+            @Throws(IOException::class)
+            override fun write(out: JsonWriter, value: T) {
+                delegate.write(out, value)
+            }
+
+            @Throws(IOException::class)
+            override fun read(`in`: JsonReader): T {
+                val obj = delegate.read(`in`)
+                if (obj is PostProcessable) {
+                    (obj as PostProcessable).postProcess()
```
cooltey, line 30: Looks like it can be simplified to `obj.postProcess()`.

cooltey (changes requested):

PR #2698, opened 2021-08-05
Enable swipe to delete recent searches in Search screen

cooltey:
https://phabricator.wikimedia.org/T276770


dbrant:
The power of Room.

PR #2702, opened 2021-08-06
Convert Prefs to Kotlin.

Isira-Seneviratne:


dbrant:
Does the `Prefs` class really benefit from being in Kotlin vs Java?

PR #2705, opened 2021-08-07
Moshi serialization

Isira-Seneviratne:
Use the [Moshi](https://github.com/square/moshi) JSON serialization library.


Isira-Seneviratne:
I tested a couple of features, such as article loading and editing, and they worked fine. I haven't tested everything, though.

PR #2717, opened 2021-08-12
Use Context.getSystemService() extension.

Isira-Seneviratne:


dbrant:
In what way is this better?

Isira-Seneviratne:
It avoids the need for a separate cast to the service class.

dbrant:
> It avoids the need for a separate cast to the service class.

No, it doesn't. The cast simply happens in a different place.

Isira-Seneviratne:
Yeah, I meant the casting doesn't have to be done in the app code itself.

PR #2718, opened 2021-08-12
Use Int.toDrawable() extension.

Isira-Seneviratne:


dbrant:
No benefit, usefulness, or reason.

PR #2719, opened 2021-08-12
Kotlin Conversion: ImageInfo

sharvaniharan:


app/src/main/java/org/wikipedia/gallery/ImageInfo.kt
```diff
@@ -0,0 +1,65 @@
...
+    val timestamp = ""
+    val size = 0
+    val width = 0
+    val height = 0
+
+    // TODO: make this smarter.
+    val bestDerivative: Derivative?
+        get() = if (derivatives.isEmpty()) {
+            null
+        } else derivatives[derivatives.size - 1]
```
cooltey, line 53: Looks like it can be non-null type of variable.

cooltey (changes requested):

app/src/main/java/org/wikipedia/gallery/ImageInfo.kt
sharvaniharan: @cooltey some code in GalleryItemFragment depends on this being sent as null when list is empty

PR #2720, opened 2021-08-12
Kotlin Conversion: TopReadArticles

sharvaniharan:


app/src/main/java/org/wikipedia/feed/topread/TopReadArticles.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.wikipedia.dataclient.page.PageSummary
+import java.util.*
+
+class TopReadArticles : PageSummary() {
+
+    private val rank = 0
+    val views = 0
+
+    @SerializedName("view_history")
+    val viewHistory: List<ViewHistory>? = null
```
cooltey, line 13: These show error messages when trying to build the apk.

cooltey (changes requested):

PR #2723, opened 2021-08-12
Update labels to wiki language

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T288020


app/src/main/java/org/wikipedia/feed/image/FeaturedImageCardView.kt
```diff
@@ -26,9 +28,16 @@ class FeaturedImageCardView(context: Context) : DefaultFeedCardView<FeaturedImag
                 image(it.baseImage())
                 description(it.description())
                 header(it)
+                updateLabelsToWikiLang()
                 setClickListeners()
             }
         }
+
+    private fun updateLabelsToWikiLang() {
+        binding.viewFeaturedImageCardDownloadButton.text = L10nUtil.getStringForArticleLanguage("en", R.string.view_featured_image_card_download)
```
dbrant, line 37: Sorry, on second thought, the Picture of the Day card should probably appear in the device's language, since Commons is language-agnostic.
It definitely makes no sense to hard-code this to English.

app/src/main/java/org/wikipedia/util/DateUtil.kt
```diff
@@ -156,18 +154,10 @@ object DateUtil {
...
-                0 -> RelativeDateTimeFormatter.getInstance().format(RelativeDateTimeFormatter.Direction.THIS, RelativeDateTimeFormatter.AbsoluteUnit.YEAR)
-                1 -> RelativeDateTimeFormatter.getInstance().format(RelativeDateTimeFormatter.Direction.LAST, RelativeDateTimeFormatter.AbsoluteUnit.YEAR)
-                -1 -> RelativeDateTimeFormatter.getInstance().format(RelativeDateTimeFormatter.Direction.NEXT, RelativeDateTimeFormatter.AbsoluteUnit.YEAR)
-                else -> RelativeDateTimeFormatter.getInstance().format(diffInYears.toDouble(), RelativeDateTimeFormatter.Direction.LAST, RelativeDateTimeFormatter.RelativeUnit.YEARS)
-            }
-        } else {
-            val context = WikipediaApp.getInstance().applicationContext
-            if (diffInYears == 0) context.getString(R.string.this_year) else context.resources.getQuantityString(R.plurals.diff_years, diffInYears, diffInYears)
-        }
+        val targetResource = L10nUtil.getResourcesForWikiLang(languageCode) ?: WikipediaApp.getInstance().resources
```
dbrant, line 159: This removes the usage of the more powerful `RelativeDateTimeFormatter`, which is the preferred formatter as of Android N.  The formatter has `getInstance()` functions that take a `Locale` parameter.

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/image/FeaturedImageCardView.kt
```diff
@@ -26,9 +28,16 @@ class FeaturedImageCardView(context: Context) : DefaultFeedCardView<FeaturedImag
...
+        binding.viewFeaturedImageCardShareButton.text = L10nUtil.getStringForArticleLanguage("en", R.string.view_featured_image_card_share)
+    }
+
```
cooltey, line 40: Please do have these settings as a method in `FeaturedImageCard` in order to keep the consistency. Also, should we replace `en` to `wikiSite().languageCode`?

app/src/main/java/org/wikipedia/util/L10nUtil.kt
```diff
@@ -67,6 +68,22 @@ object L10nUtil {
         return localizedStrings
     }
 
+    // To be used only for plural strings and strings requiring arguements
```
cooltey, line 71: Typo. Arguments.

app/src/main/java/org/wikipedia/util/L10nUtil.kt
```diff
@@ -67,6 +68,22 @@ object L10nUtil {
...
+    fun getResourcesForWikiLang(languageCode: String): Resources? {
```
cooltey, line 72: Is it possible to combine this function with `getStringsForLocale` by adding an argument for the plural strings?

cooltey (changes requested):

app/src/main/java/org/wikipedia/feed/image/FeaturedImageCardView.kt
sharvaniharan: Makes sense. I was meaning to use the app lang instead of the hardcoded "en", but I feel even that will not make sense. Removed this.

app/src/main/java/org/wikipedia/util/L10nUtil.kt
sharvaniharan: The problem with this is the arguments @cooltey

sharvaniharan:
Done @cooltey @dbrant

PR #2730, opened 2021-08-13
Kotlin Conversion: RbServiceError

sharvaniharan:


app/src/main/java/org/wikipedia/dataclient/restbase/RbServiceError.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    override fun getDetails(): String {
+        return StringUtils.defaultString(detail)
+    }
+
+    companion object {
+        fun create(rspBody: String): RbServiceError {
+            return GsonUnmarshaller.unmarshal(RbServiceError::class.java, rspBody)
+        }
+    }
```
cooltey, line 27: Would it be possible to convert `ServiceError` to Kotlin as well, and the `getTitle()` can be combined with the `val title` variable.

cooltey (changes requested):

app/src/main/java/org/wikipedia/dataclient/restbase/RbServiceError.kt
sharvaniharan: Done @cooltey

app/src/main/java/org/wikipedia/dataclient/restbase/RbServiceError.kt
```diff
@@ -0,0 +1,20 @@
...
+    private val type: String? = null
+    override val title: String = ""
+    override val details: String = ""
+    private val detail: String? = null
+    private val method: String? = null
+    private val uri: String? = null
+
+    companion object {
+        fun create(rspBody: String): RbServiceError {
+            return GsonUnmarshaller.unmarshal(RbServiceError::class.java, rspBody)
```
cooltey, line 17: Not sure but do you think it is possible to use `GsonUtil` for this line?
```
GsonUtil.getDefaultGson().fromJson(rspBody, RbServiceError::class.java)
```

PR #2731, opened 2021-08-13
Follow up: EventServiceResponse Kotlin conversion

cooltey:


app/src/main/java/org/wikipedia/analytics/eventplatform/EventServiceResponse.kt
```diff
@@ -15,5 +15,5 @@ import com.google.gson.annotations.SerializedName
 class EventServiceResponse {
 
     @SerializedName("invalid") val invalidEvents: List<Any> = emptyList()
-    @SerializedName("error") val errorEvents = emptyList<Any>()
```
sharvaniharan, line 18: Oh I have started adding the type to emptyList() to avoid unnecessary type specification 😅  May I know what the difference is?

cooltey: Should be the same and I just want to keep the consistency as the above one.

PR #2738, opened 2021-08-14
Use view binding.

Isira-Seneviratne:


cooltey:
Recommend to close this PR since we are going to update the layouts out Notification screens, and I don't see too much impact on the rest of the views by changing them to use view bindings.

PR #2741, opened 2021-08-16
Make sure the bookmark status is showing correctly when adding/removing from a list.

cooltey:
Now if we add an article to the list, the status of the bookmark icon will not be updated correctly most of the time.

This is not fixing the root cause of the issue but definitely is the easiest one.


dbrant:
@cooltey  Would you mind explaining the root cause?

cooltey:
> @cooltey Would you mind explaining the root cause?

The `EventBus` in `ReadingListDaoPage.addPageToList` will send the `ArticleSavedOrDeletedEvent` before the `ReadingListBehaviorsUtil.addToDefaultList()` hitting the `callback?.onCompleted()`, which will cause the `EventBusConsumer` class in `PageActivity` receive the event before it actually being stored to the db.

dbrant:
@cooltey  Can we just modify the `addToDefaultList()` function to not use a coroutine? (i.e. just do the DB operation on the main thread.)
That function already does DB access on the main thread anyway, and we're clearly not using coroutines correctly yet.

cooltey:
> 
> 
> @cooltey Can we just modify the `addToDefaultList()` function to not use a coroutine? (i.e. just do the DB operation on the main thread.)
> That function already does DB access on the main thread anyway, and we're clearly not using coroutines correctly yet.

It makes sense to me. Done.

PR #2745, opened 2021-08-17
Convert PageProperties to Kotlin.

Isira-Seneviratne:


app/src/main/java/org/wikipedia/page/PageProperties.kt
```diff
@@ -0,0 +1,85 @@
+package org.wikipedia.page
+
+import android.location.Location
+import android.os.Parcelable
+import kotlinx.parcelize.IgnoredOnParcel
+import kotlinx.parcelize.Parcelize
+import org.wikipedia.auth.AccountUtil.isMemberOf
```
cooltey, line 7: Please only import `import org.wikipedia.auth.AccountUtil` and use `AccountUtil.isMemberOf`

app/src/main/java/org/wikipedia/page/PageProperties.kt
```diff
@@ -0,0 +1,85 @@
...
+import org.wikipedia.dataclient.page.PageSummary
+import org.wikipedia.dataclient.page.Protection
+import org.wikipedia.util.DateUtil.iso8601DateParse
+import org.wikipedia.util.DimenUtil.calculateLeadImageWidth
+import org.wikipedia.util.ImageUrlUtil.getUrlForPreferredSize
+import org.wikipedia.util.UriUtil.decodeURL
+import org.wikipedia.util.UriUtil.resolveProtocolRelativeUrl
```
cooltey, line 14: Similar as above.

cooltey (changes requested):

PR #2764, opened 2021-08-28
Kotlin Conversion: UserInfo

sharvaniharan:


app/src/main/java/org/wikipedia/dataclient/mwapi/UserInfo.kt
```diff
@@ -0,0 +1,28 @@
...
+import androidx.collection.ArraySet
+import org.wikipedia.dataclient.mwapi.MwServiceError.BlockInfo
+import org.wikipedia.util.DateUtil
+import java.util.*
+
+@SuppressWarnings("unused")
+class UserInfo : BlockInfo() {
+    val id = 0
+    private val groups: List<String>? = null
+    private val latestcontrib: String? = null
```
dbrant, line 12: Can we keep camelCase naming for these fields, and use `@SerializedName` for the actual names?

dbrant (changes requested):

PR #2743, opened 2021-08-17
Updates to notification settings

cooltey:
https://phabricator.wikimedia.org/T287477

PR #2767, opened 2021-08-28
Kotlin Conversion: Notification

sharvaniharan:


app/src/main/java/org/wikipedia/notifications/Notification.kt
```diff
@@ -0,0 +1,135 @@
...
+    companion object {
+        const val CATEGORY_SYSTEM = "system"
+        const val CATEGORY_SYSTEM_NO_EMAIL = "system-noemail" // default welcome
+        const val CATEGORY_MILESTONE_EDIT = "thank-you-edit" // milestone
+        const val CATEGORY_EDIT_USER_TALK = "edit-user-talk"
+        const val CATEGORY_EDIT_THANK = "edit-thank"
+        const val CATEGORY_REVERTED = "reverted"
+        const val CATEGORY_LOGIN_FAIL = "login-fail"
+        const val CATEGORY_MENTION =
+            "mention" // combines "mention", "mention-failure" and "mention-success"
```
cooltey, line 133: These should be removed after #2743 merged.

PR #2771, opened 2021-08-30
Kotlin Conversion: Protection

sharvaniharan:


app/src/main/java/org/wikipedia/dataclient/page/Protection.kt
```diff
@@ -0,0 +1,20 @@
...
+
+    private val edit: String? = null
+    val type: String = ""
+    val level: String = ""
+    val expiry: String = ""
+    val firstAllowedEditorRole: String
+        get() = edit.orEmpty()
+    val editRoles: Set<String>
+        get() {
+            val roleSet: MutableSet<String> = HashSet()
```
cooltey, line 16: This can be `val roleSet = mutableSetOf<String>()`

app/src/main/java/org/wikipedia/dataclient/page/Protection.kt
```diff
@@ -0,0 +1,20 @@
...
+            edit?.let { roleSet.add(edit) }
+            return Collections.unmodifiableSet(roleSet)
```
cooltey, line 18: Looks like we can use `roleSet.toSet()`.

cooltey (changes requested):

PR #2777, opened 2021-09-02
Kotlin Conversion: Event

sharvaniharan:


app/src/main/java/org/wikipedia/analytics/eventplatform/Event.kt
```diff
@@ -0,0 +1,27 @@
...
+/** Base class for an Event Platform event.  */
+open class Event(private val schema: String, stream: String) {
+
+    @SerializedName("app_session_id")
+    var sessionId: String? = null
+
+    @SerializedName("app_install_id")
+    var appInstallId: String? = null
+
+    private val meta: Meta?
```
dbrant, line 16: If these fields are initialized in `init`, can they be nonnull?

app/src/main/java/org/wikipedia/analytics/eventplatform/Event.kt
```diff
@@ -0,0 +1,27 @@
+package org.wikipedia.analytics.eventplatform
+
+import com.google.gson.annotations.SerializedName
+import org.wikipedia.util.DateUtil.iso8601DateFormat
+import java.util.*
+
+/** Base class for an Event Platform event.  */
+open class Event(private val schema: String, stream: String) {
```
dbrant, line 8: It looks like the `schema` parameter needs a `@SerializedName` that is missing here.

dbrant (changes requested):

PR #2790, opened 2021-09-06
fix build on mac m1

HuixingWong:
https://stackoverflow.com/questions/68884589/caused-by-java-lang-exception-no-native-library-is-found-for-os-name-mac-and-o


dbrant:
As a general rule we don't use `alpha` versions of any dependencies. It's unfortunate that this dependency is breaking builds on M1, but we will need to wait until they release a stable version of Room that contains this fix.

PR #2795, opened 2021-09-07
Kotlin Conversion: MwException, MwResponse

sharvaniharan:


app/src/main/java/org/wikipedia/dataclient/mwapi/MwException.java
```diff
@@ -1,23 +0,0 @@
...
-
-    @NonNull public MwServiceError getError() {
-        return error;
-    }
-
-    @NonNull public String getTitle() {
-        return error.getTitle();
-    }
-
-    @Override @NonNull public String getMessage() {
```
dbrant, line 20: What happened to this override in the converted version?

dbrant (changes requested):

app/src/main/java/org/wikipedia/dataclient/mwapi/MwException.java
sharvaniharan: Thank you had missed it after the automatics conversion ignored it..

PR #2800, opened 2021-09-08
Follow-up to MwQueryPage conversion.

dbrant:


app/src/main/java/org/wikipedia/dataclient/mwapi/MwQueryPage.kt
```diff
@@ -91,7 +91,7 @@ class MwQueryPage {
         val comment: String = ""
 
         fun getContentFromSlot(slot: String): String {
-            return if (slots?.containsKey(slot)!!) slots[slot]!!.content else ""
+            return if (slots != null && slots.containsKey(slot)) slots[slot]!!.content else ""
```
cooltey, line 94: How about this: `return slots?.get(slot)?.content.orEmpty()`

dbrant: ok!

PR #2801, opened 2021-09-08
Update notification list view

cooltey:
https://phabricator.wikimedia.org/T288064


cooltey:
@dbrant @sharvaniharan The PR is ready for review.

dbrant (changes requested):
@cooltey  I'm continuing to review this, but I noticed a crash:  Open the Notifications activity and very quickly tap the "Mentions" tab (while the content is still being loaded).

cooltey:
> 
> 
> @cooltey I'm continuing to review this, but I noticed a crash: Open the Notifications activity and very quickly tap the "Mentions" tab (while the content is still being loaded).

Yes, I can reproduce it. Fixed.

sharvaniharan:
@cooltey The description text in the designs look like Robin expects some curation of the text like "Mention: ..." or " Message on your talk page:..." etc... it is not the same in the implementation.

cooltey:
> 
> 
> @cooltey The description text in the designs look like Robin expects some curation of the text like "Mention: ..." or " Message on your talk page:..." etc... it is not the same in the implementation. Please Lmk.

Not sure about it, if we add prefix texts to the view, the message will be useless since most of the messages will be truncated. Maybe let's leave it as it is and let Robin decide that?

sharvaniharan:
Lgtm.. @dbrant please merge when you are done.

PR #2803, opened 2021-09-09
Direct replies from Notifications.

dbrant:


app/src/main/java/org/wikipedia/talk/NotificationDirectReplyHelper.kt
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    private fun cancelNotification(context: Context, notificationId: Int) {
+        if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.O) {
+            return
+        }
+        val notificationManager = context.getSystemService<NotificationManager>()!!
+        val notifications = notificationManager.activeNotifications!!
+        for (notification in notifications) {
+            if (notification.id == notificationId) {
```
cooltey, line 107: How about using: `notifications.find { it.id == notificationId }?.run { ... }`

or this for null-safe approach

`notificationManager?.activeNotifications?.find { it.id == notificationId }?.run {`

app/src/main/java/org/wikipedia/notifications/NotificationPresenter.kt
```diff
@@ -30,11 +31,13 @@ object NotificationPresenter {
         val activityIntent = addIntentExtras(NotificationActivity.newIntent(context), n.id, n.type)
         val builder = getDefaultBuilder(context, n.id, n.type, notificationCategory)
         val title: String = StringUtil.fromHtml(if (n.contents != null) n.contents.header else "").toString()
+        val id = n.key().toInt()
 
         n.contents?.links?.let {
             it.getPrimary()?.let { primary ->
                 if (NotificationCategory.EDIT_USER_TALK.id == n.category) {
-                    addActionForTalkPage(context, builder, primary, n)
+                    // addActionForTalkPage(context, builder, primary, n)
```
cooltey, line 39: Should we keep/remove this?

app/src/main/java/org/wikipedia/talk/NotificationDirectReplyHelper.kt
dbrant: correct!

app/src/main/java/org/wikipedia/notifications/NotificationPresenter.kt
dbrant: removed!

PR #2807, opened 2021-09-13
Fix logic of ABC test.

dbrant:


app/src/main/java/org/wikipedia/analytics/ABTestFunnel.kt
```diff
@@ -14,7 +14,7 @@ class ABTestFunnel internal constructor(private val abTestName: String, private
             var group = PrefsIoUtil.getInt(AB_TEST_KEY_PREFIX + abTestName, -1)
             if (group == -1) {
                 // initialize the group if it hasn't been yet.
-                group = Random(Int.MAX_VALUE).nextInt()
```
dbrant, line 17: Rookie mistake / brain fart -- the constructor parameter is the random _seed_, not the _range_ of the random number.

PR #2811, opened 2021-09-15
Convert Service classes to Kotlin.

dbrant:
This is in preparation for refactoring our UI classes to use a ViewModel / Repository architecture, where the service calls can start becoming `suspend` functions, to be used in coroutines.


app/src/main/java/org/wikipedia/dataclient/RestService.kt
```diff
@@ -0,0 +1,210 @@
...
+        @Header(OfflineCacheInterceptor.SAVE_HEADER) saveHeader: String?,
+        @Header(OfflineCacheInterceptor.LANG_HEADER) langHeader: String?,
+        @Header(OfflineCacheInterceptor.TITLE_HEADER) titleHeader: String?
+    ): Observable<Response<MediaList>>
+
+    @GET("feed/onthisday/events/{mm}/{dd}")
+    fun getOnThisDay(@Path("mm") month: Int, @Path("dd") day: Int): Observable<OnThisDay>
+
+    @get:GET("feed/announcements")
+    @get:Headers("Accept: " + ACCEPT_HEADER_PREFIX + "announcements/0.1.0\"")
```
cooltey, line 85: Not sure but do you agree that we should use either `@GET()` and `@Headers` with a `fun` style method instead of `val` style in the `Service` related classes? Somehow the `@get:Get()` looks a bit messy to me.

dbrant: The reason I left them this way is because if we change them to functions, we would also need to change all of the places they are used to function invocation`()` style, which would require modifying 100 files.

Can we leave them like this, and then update them one-by-one as we transition them to be `suspend` functions?

cooltey: I see. It sounds good to me!

cooltey (approved):
@sharvaniharan Please review and merge if it looks good to you too.

PR #2813, opened 2021-09-16
Use ViewModel architecture in LanguagesListActivity.

dbrant:
Baby steps towards instituting a ViewModel structure throughout the project.


dbrant:
@cooltey @sharvaniharan  This is now ready, too.

app/src/main/java/org/wikipedia/language/LanguagesListViewModel.kt
```diff
@@ -0,0 +1,82 @@
...
+            val canonicalName = StringUtils.stripAccents(getCanonicalName(code).orEmpty())
+            if (filter.isEmpty() || code.contains(filter) ||
+                localizedName.lowercase(Locale.getDefault()).contains(filter) ||
+                canonicalName.lowercase(Locale.getDefault()).contains(filter)) {
+                if (first) {
+                    results.add(LanguageListItem(context.getString(R.string.languages_list_suggested_text), true))
+                    first = false
+                }
+                results.add(LanguageListItem(code))
+            }
```
cooltey, line 54: Is it possible to make this logic a method? Looks like it has pretty much similar to the logic below.

app/src/main/java/org/wikipedia/language/LanguagesListViewModel.kt
```diff
@@ -0,0 +1,82 @@
...
+    private val suggestedLanguageCodes = WikipediaApp.getInstance().language().remainingAvailableLanguageCodes
+    private val nonSuggestedLanguageCodes = WikipediaApp.getInstance().language()
+        .appMruLanguageCodes.toMutableList().also { it.removeAll(suggestedLanguageCodes) }
+
+    init {
+        fetchData()
+    }
+
+    private fun fetchData() {
+        viewModelScope.launch {
```
cooltey, line 30: Can we add a `CoroutineExceptionHandler` for the coroutine process?

cooltey (changes requested):

app/src/main/java/org/wikipedia/language/LanguagesListViewModel.kt
dbrant: sure!

app/src/main/java/org/wikipedia/language/LanguagesListViewModel.kt
dbrant: correct.

app/src/main/java/org/wikipedia/language/LanguagesListViewModel.kt
```diff
@@ -0,0 +1,88 @@
...
+            }
+        }
+    }
+
+    fun getCanonicalName(code: String): String? {
+        var canonicalName = siteListData.value?.data?.find { it.code == code }?.localname
+        if (canonicalName.isNullOrEmpty()) {
+            canonicalName = WikipediaApp.getInstance().language().getAppLanguageCanonicalName(code)
+        }
+        return canonicalName
```
cooltey, line 84: This can be simplified to:
```
siteListData.value?.data?.find { it.code == code }?.localname.orEmpty().ifEmpty { WikipediaApp.getInstance().language().getAppLanguageCanonicalName(code) }
```

app/src/main/java/org/wikipedia/language/LanguagesListViewModel.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.wikipedia.WikipediaApp
+import org.wikipedia.dataclient.ServiceFactory
+import org.wikipedia.dataclient.mwapi.SiteMatrix
+import org.wikipedia.util.Resource
+import org.wikipedia.util.log.L
+import java.util.*
+
+class LanguagesListViewModel : ViewModel() {
+
+    val siteListData = MutableLiveData<Resource<List<SiteMatrix.SiteInfo>>>()
```
cooltey, line 22: Would it be possible to move public `val` after `private val`?

cooltey (changes requested):
Looks good to me. Just have two minor comments.

app/src/main/java/org/wikipedia/language/LanguagesListViewModel.kt
dbrant: yep

app/src/main/java/org/wikipedia/language/LanguagesListViewModel.kt
dbrant: sure

PR #2826, opened 2021-09-21
Serialization overhaul

dbrant:
This will remove the old dependency on Gson, and start using `kotlinx.serialization` for our serialization needs.


cooltey:
This PR works for me and I don't see any new crashes or error messages in logcat. @sharvaniharan Please review and merge if it looks good to you too.

PR #2829, opened 2021-09-22
Use WebViewClientCompat.

Isira-Seneviratne:
Use `WebViewClientCompat` instead of `WebViewClient` in `OkHttpWebViewClient` and override the non-deprecated methods (these can be used on API levels >= 21 with an up-to-date WebView).


dbrant:
Can you describe what issue(s) this solves?

Isira-Seneviratne:
It allows some of the newer WebView APIs to be used on older versions of Android, e.g. `onReceivedHttpError`.

PR #2830, opened 2021-09-22
Fix explicit bouncing of links to external browser app.

dbrant:
https://phabricator.wikimedia.org/T291199

Here's what seems to be happening:
On Android 11 (and possibly 10?) if our app is set as the default app for our intent filter, we can no longer use `queryIntentActivities()` to get a list of other external apps that can handle links.  The guidance from the documentation is to use a `<queries>` element in our manifest to state that we need to query activities that match a certain intent.  But even with a proper `<queries>` tag, the system is still not returning other apps that can handle the intent, if the intent is "tightly bound" to our app (i.e. if our app is the explicit default).

This PR makes the following workaround:
If our current way of bouncing to an external browser fails (i.e. if the intent chooser is empty), we will query apps that can handle a _generic_ link ("example.com"), and explicitly open the _first_ app in this list, which should ideally be the user's "default" browser app, feeding it the URI that needs to be opened.

This should be considered a temporary workaround, while we continue to research the most proper method of enumerating external activities.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -43,6 +43,13 @@
     -->
     <uses-feature android:name="android.hardware.touchscreen" android:required="false" />
 
+    <queries>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:scheme="https" />
+        </intent>
+    </queries>
+
     <application
         android:allowBackup="true"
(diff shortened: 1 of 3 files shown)


cooltey:
It works for me. 

Another note here is that I have two browsers installed (Chrome + Firefox), and in `main` branch, even the app is set to default, it will show the "Firefox" option without any issue in this case.

PR #2836, opened 2021-09-23
Tooltip for the new placement of notifications

cooltey:
https://phabricator.wikimedia.org/T288823


sharvaniharan:
Lgtm... Should we merge this to the feature branch instead of main? keeping in mind that it should all go out together..

cooltey:
> Lgtm... Should we merge this to the feature branch instead of main? keeping in mind that it should all go out together..

This is related to the ABC tests which are in the production now, should we still merge this to the feature branch?
cc @dbrant

PR #2845, opened 2021-09-27
Use Char extensions.

Isira-Seneviratne:


dbrant:
Why/how is this an improvement?

PR #2856, opened 2021-09-29
Open native diff page when clicking on an diff page url

cooltey:
https://phabricator.wikimedia.org/T288064#7386546


sharvaniharan:
Lgtm... Will let @dbrant take a look and merge.

PR #2860, opened 2021-09-30
Tweak: long press action on notification list view

cooltey:
https://phabricator.wikimedia.org/T288066#7390343


dbrant (changes requested):
Looks very good, but this makes the circular progress bar (while loading notifications) hidden behind the bottom system navigation buttons. Can we move the ProgressBar up a bit more, or even put it in the center of the layout?

cooltey:
> Looks very good, but this makes the circular progress bar (while loading notifications) hidden behind the bottom system navigation buttons. Can we move the ProgressBar up a bit more, or even put it in the center of the layout?

Makes sense! Done.

PR #2864, opened 2021-10-02
Update setup-java action to v2 for workflows

nuhkoca:
Updating the [setup-java](https://github.com/actions/setup-java) action to v2. 

- V2 supports custom distributions and provides support for `Zulu OpenJDK`, `Eclipse` `Temurin` and `Adopt OpenJDK`. V1 supports only `Zulu OpenJDK`.
- V1 supported legacy Java syntax such as 1.8 (same as 8) and 1.8.0.212 (same as 8.0.212). V2 dropped support for legacy syntax.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -10,9 +10,10 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Set up JDK 11
-      uses: actions/setup-java@v1
+      uses: actions/setup-java@v2
       with:
-        java-version: 11
+        distribution: 'adopt'
+        java-version: '11'
     - name: Check for missing qq strings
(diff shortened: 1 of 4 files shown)


nuhkoca:
Would be good if I could get some Hactoberfest points for this PR 😬

dbrant:
The documentation says:

> Adopt OpenJDK got moved to Eclipse Temurin and won't be updated anymore. It is highly recommended to migrate workflows from `adopt` to `temurin` to keep receiving software and security updates.

Can we use `temurin` instead?

nuhkoca:
@dbrant good point! I was just going to mention `Temurin`, as well. I will change the distributor then. Thanks!

nuhkoca:
@dbrant ready for review!

PR #2872, opened 2021-10-05
Support offline notifications

cooltey:


sharvaniharan:
Looks very good... so seamless. 

- When we click on a notification whilst offline, can we also make it unclickable along with showing the toast that the functionality is not available? No use showing an error page
- Every once in a while i encounter this while offline. Haven't figured out reproducible steps yet. will try:
offline
-

cooltey:
>     * When we click on a notification whilst offline, can we also make it unclickable along with showing the toast that the functionality is not available? No use showing an error page

Nice catch. Somehow it is a bug and I have a PR for it. https://github.com/wikimedia/apps-android-wikipedia/pull/2973

Ideally, it should not show the toast message on click actions since pages might have been visited before.

cooltey:
Looks like after updating the Kotlin version to `1.6.0`, something was wrong with coroutine and Room integration. In d90700b I have removed the `suspend` from `insert`, `delete`, and `update` types of action. The functionality has not been impacted, and hopefully, they can fix that issue soon.

PR #2878, opened 2021-10-07
Edit notices.

dbrant:
https://phabricator.wikimedia.org/T201597


dbrant:
@sharvaniharan @cooltey This is now ready for review.

app/src/main/res/values/preference_keys.xml
```diff
@@ -123,4 +123,6 @@
     <string name="preference_key_excluded_wiki_codes_notification">filterNotificationExcludedWikiCodes</string>
     <string name="preference_key_excluded_type_codes_notification">filterNotificationExcludedTypeCodes</string>
     <string name="preference_key_page_notification_tooltip_shown">pageNotificationTooltipShown</string>
+    <string name="preference_key_auto_show_edit_notices">autoShowEditNotices</string>
+    <string name="preference_key_edit_notices_tooltip_shown">editNoticesTooltipShown</string>
```
cooltey, line 127: Should we add this to `developer_preferences.xml` as well?

app/src/main/java/org/wikipedia/views/EditNoticesDialog.kt
```diff
@@ -0,0 +1,72 @@
...
+        val wikiSite: WikiSite,
+        val editNotices: List<String>,
+        context: Context
+) : AlertDialog(context) {
+
+    private val binding = DialogEditNoticesBinding.inflate(LayoutInflater.from(context))
+
+    private val movementMethod = LinkMovementMethodExt { urlStr ->
+        L.v("Link clicked was $urlStr")
+        UriUtil.visitInExternalBrowser(context, Uri.parse(UriUtil.resolveProtocolRelativeUrl(wikiSite, urlStr)))
```
cooltey, line 34: Would it be possible to implement the `LinkHandler` here so that some links can be opened in the app? e.g. talk pages or Wikipedia articles.

app/src/main/java/org/wikipedia/views/EditNoticesDialog.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.wikipedia.util.log.L
+
+@SuppressLint("ViewConstructor")
+class EditNoticesDialog constructor(
+        val wikiSite: WikiSite,
+        val editNotices: List<String>,
+        context: Context
+) : AlertDialog(context) {
+
+    private val binding = DialogEditNoticesBinding.inflate(LayoutInflater.from(context))
```
cooltey, line 30: We can use `layoutInflater` instead.

cooltey (changes requested):

app/src/main/res/values/preference_keys.xml
dbrant: Sure!

app/src/main/java/org/wikipedia/views/EditNoticesDialog.kt
dbrant: Correct

app/src/main/java/org/wikipedia/views/EditNoticesDialog.kt
dbrant: This is actually done deliberately. If a link goes to a regular article, it will open `PageActivity`, and because PageActivity is `singleTask`, it will automatically close the Editing activity, and the user will lose their place, and won't be able to go back to editing.

cooltey: Ah! Right, that makes sense.

PR #2898, opened 2021-10-14
Design tweak: Search and Filter

sharvaniharan:
**Phab:**  https://phabricator.wikimedia.org/T288068#7425074


sharvaniharan:
Np @cooltey I would have missed that!

PR #2911, opened 2021-10-19
Notifications for anonymous IP editors.

dbrant:
https://phabricator.wikimedia.org/T291481


app/src/main/java/org/wikipedia/page/PageActivity.kt
```diff
@@ -733,6 +738,17 @@ class PageActivity : BaseActivity(), PageFragment.Callback, LinkPreviewDialog.Ca
             } else {
                 binding.pageToolbarButtonNotifications.setUnreadCount(0)
             }
+        } else if (!AccountUtil.isLoggedIn && AnonymousNotificationHelper.isWithinAnonNotificationTime()) {
+            binding.pageToolbarButtonNotifications.isVisible = true
```
cooltey, line 742: Is it possible to save the language code for anonymous editors once they received notification from a specific wiki site? 

For example, if the IP editor received a notification from `en` wiki site, then only show the bell icon for `en` articles instead of all the wiki sites. It would be less confusing when reading articles in different languages since the talk page will be empty on other wiki sites.

cooltey (commented):
Works for me! Only a comment below:

sharvaniharan:
Looks good to me... @cooltey feel free to merge after the above issue is resolved.

sharvaniharan (approved):

app/src/main/java/org/wikipedia/page/PageActivity.kt
dbrant: Done -- this now saves the language of the "first" talk page message that was detected for the current IP. The notification will continue to be shown (and will launch the Talk page for the saved language) even if you browse articles in other languages.

PR #2950, opened 2021-10-29
Optimize and streamline Notifications api call sequencing.

dbrant:
There is no reason for the filtering logic to affect the kind of API call that we make. The filtering is a post-processing step that should happen after the API results are received.

We can now get all notifications from all wikis, and then filter them based on any selected filters.

This way, if the user selects a new filter, it can be applied _immediately_ to the results that are already received, instead of making a whole new API call.


cooltey (changes requested):
Found two issues:

1. The `zh` language code does not appear correctly in the list for the `zh-hant` notifications.
2. If I only select `Traditional Chinese` in the `NotificationFilterActivity`, I will not see any notification in the list, which is incorrect. 

Note: please use `getDefaultLanguageCode()` in `AppLanguageState` to get its default language code.

PR #2951, opened 2021-11-01
Update to API 31.

dbrant:
Some required updates:
* Added the `autoVerify` attribute to our intent filter, which will need to be paired with a corresponding Digital Asset link file on the server side. https://phabricator.wikimedia.org/T294776
* All activities, services, and broadcast receivers now require the `exported=` attribute in the manifest.
* All `PendingIntent`s now require mutability flags.
* This required an update of the LeakCanary dependency because they just recently updated their Activity to be `exported`.


cooltey (approved):
LGTM. I'll let @sharvaniharan review it as well and merge it.

PR #2955, opened 2021-11-02
Make Notification wiki code box expand to text length.

dbrant:
This makes the wiki-code box properly wrap around the language code text, making it much more readable, especially with three-letter (or greater) language codes:

device-2021-11-02-104903


cooltey (commented):
Not sure but should we make the height always fit `24dp` for the language code block?

dbrant:
> Not sure but should we make the height always fit `24dp` for the language code block?

I think the important thing is to have a constant amount of padding around the text. If we make the height always `24dp`, we would need to make the text always the same size, which would make `zh-yue` a bit too large.  I think this a good balance -- we'll confirm with Robin later, but surely this is an improvement over the previous style.

PR #2959, opened 2021-11-02
Additional design tweaks for read/unread notifications.

dbrant:
(not a blocker for Beta)

* Remove the blue dot for unread notifications.
* Update text typefaces and colors for read vs. unread notifications, to create a better contrast.


cooltey:
The PR looks good to me. Do we have a ticket or comment that indicated this change so that Robin will know about this (and probably will have other thoughts)?

dbrant:
> The PR looks good to me. Do we have a ticket or comment that indicated this change so that Robin will know about this (and probably will have other thoughts)?

Of course: https://phabricator.wikimedia.org/T294369

PR #2969, opened 2021-11-08
Design Tweak: Add or edit languages

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T294035

### DIFF
--- a/app/src/main/java/org/wikipedia/notifications/NotificationFilterItemView.kt
+++ b/app/src/main/java/org/wikipedia/notifications/NotificationFilterItemView.kt
@@ -2,6 +2,7 @@ package org.wikipedia.notifications
 
 import android.content.Context
 import android.content.res.ColorStateList
+import android.graphics.Typeface
 import android.os.Build
 import android.util.AttributeSet
 import android.view.LayoutInflater
@@ -70,10 +71,15 @@ class NotificationFilterItemView constructor(context: Context, attrs: AttributeS
     }
 
     fun setSingleLabel(text: String) {
-        binding.notificationFilterLanguageCode.visibility = View.INVISIBLE
-        binding.notificationFilterWikiLogo.visibility = View.GONE
+        binding.notificationFilterLanguageCode.visibility = View.GONE
+        binding.notificationFilterWikiLogo.visibility = View.VISIBLE
+        binding.notificationFilterWikiLogo.imageTintList =
+            ColorStateList.valueOf(ContextCompat.getColor(context, ResourceUtil.getThemedAttributeId(context, R.attr.colorAccent)))
+        binding.notificationFilterWikiLogo.setImageDrawable(AppCompatResources.getDrawable(context, R.drawable.ic_add_gray_themed_24dp))
         binding.notificationFilterCheck.visibility = View.GONE
-        binding.notificationFilterTitle.text = text
+        binding.notificationFilterTitle.setTextColor(ResourceUtil.getThemedColor(context, R.attr.colorAccent))
+        binding.notificationFilterTitle.setTypeface(binding.notificationFilterTitle.typeface, Typeface.BOLD)
+        binding.notificationFilterTitle.text = text.uppercase()
     }
 
     private fun getTitleCodeFor(filterCode: String): String? {
--- a/app/src/main/java/org/wikipedia/notifications/NotificationFilterItemView.kt
+++ b/app/src/main/java/org/wikipedia/notifications/NotificationFilterItemView.kt
@@ -22,7 +22,6 @@ import org.wikipedia.util.DimenUtil
 import org.wikipedia.util.ResourceUtil
 import org.wikipedia.views.ViewUtil
 
-
 class NotificationFilterItemView constructor(context: Context, attrs: AttributeSet? = null) : LinearLayout(context, attrs) {
 
     interface Callback {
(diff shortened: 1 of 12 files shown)


app/src/main/res/values-qq/strings.xml
```diff
@@ -1087,4 +1087,5 @@
   <string name="watchlist_page_moved">Label text to indicate that a page has been moved.</string>
   <string name="watchlist_page_protected">Label text to indicate that a page has been protected.</string>
   <string name="watchlist_page_deleted">Label text to indicate that a page has been deleted.</string>
+  <string name="add_app_language_text">Label text to add or edit apps langugaes list.</string>
```
cooltey, line 1090: Would it be possible to have the string name has a `notification_filters` prefix in it?

cooltey (changes requested):
Does Robin want to have a consistent style across the app? Looks like the text size is different from the "add language" button in the app languages screen.

app/src/main/res/values-qq/strings.xml
sharvaniharan: I added it to the `Multilingual` section so it could be used other places too, so no prefix would keep it generic... what do you think?

sharvaniharan:
> ooks like the text size and typeface (bold vs medium) are different from the "add language" button in the app languages screen.

Will clarify this with Robin... If we dont need to create a new viewholder it will be better..

app/src/main/res/values-qq/strings.xml
cooltey: I recall there was a ticket about *not* recommending using the same strings because of countries/geolocations.

sharvaniharan:
@cooltey Robin is fine with just removing the bold, but I thought 14sp looked better than 16sp as well.. please take a look at the changes and

cooltey (changes requested):
I am seeing this:

sharvaniharan:
@cooltey fixed it!

app/src/main/java/org/wikipedia/notifications/NotificationFilterItemView.kt
```diff
@@ -70,10 +71,16 @@ class NotificationFilterItemView constructor(context: Context, attrs: AttributeS
...
+        binding.notificationFilterLanguageCode.visibility = View.GONE
+        binding.notificationFilterWikiLogo.visibility = View.VISIBLE
+        binding.notificationFilterWikiLogo.imageTintList =
+            ColorStateList.valueOf(ContextCompat.getColor(context, ResourceUtil.getThemedAttributeId(context, R.attr.colorAccent)))
+        binding.notificationFilterWikiLogo.setImageDrawable(AppCompatResources.getDrawable(context, R.drawable.ic_add_gray_themed_24dp))
         binding.notificationFilterCheck.visibility = View.GONE
-        binding.notificationFilterTitle.text = text
+        binding.notificationFilterTitle.setTextColor(ResourceUtil.getThemedColor(context, R.attr.colorAccent))
+        binding.notificationFilterTitle.text = text.uppercase()
+        binding.notificationFilterTitle.setTextSize(TypedValue.COMPLEX_UNIT_SP, textSize)
```
cooltey, line 83: How about just setting `14f` here instead of having another `val` for it?

app/src/main/java/org/wikipedia/notifications/NotificationFilterItemView.kt
```diff
@@ -70,10 +71,16 @@ class NotificationFilterItemView constructor(context: Context, attrs: AttributeS
...
+        val textSize = 14f
+        binding.notificationFilterLanguageCode.visibility = View.GONE
+        binding.notificationFilterWikiLogo.visibility = View.VISIBLE
+        binding.notificationFilterWikiLogo.imageTintList =
+            ColorStateList.valueOf(ContextCompat.getColor(context, ResourceUtil.getThemedAttributeId(context, R.attr.colorAccent)))
+        binding.notificationFilterWikiLogo.setImageDrawable(AppCompatResources.getDrawable(context, R.drawable.ic_add_gray_themed_24dp))
         binding.notificationFilterCheck.visibility = View.GONE
-        binding.notificationFilterTitle.text = text
+        binding.notificationFilterTitle.setTextColor(ResourceUtil.getThemedColor(context, R.attr.colorAccent))
+        binding.notificationFilterTitle.text = text.uppercase()
```
cooltey, line 82: Looks like the `TypeFace` should be `Roboto-Medium`. Now it is in regular Roboto.

app/src/main/java/org/wikipedia/notifications/NotificationFilterItemView.kt
```diff
@@ -70,10 +71,16 @@ class NotificationFilterItemView constructor(context: Context, attrs: AttributeS
     }
 
     fun setSingleLabel(text: String) {
-        binding.notificationFilterLanguageCode.visibility = View.INVISIBLE
-        binding.notificationFilterWikiLogo.visibility = View.GONE
+        val textSize = 14f
+        binding.notificationFilterLanguageCode.visibility = View.GONE
+        binding.notificationFilterWikiLogo.visibility = View.VISIBLE
+        binding.notificationFilterWikiLogo.imageTintList =
+            ColorStateList.valueOf(ContextCompat.getColor(context, ResourceUtil.getThemedAttributeId(context, R.attr.colorAccent)))
```
cooltey, line 78: Can we have the `colorAccent` as a `val` so that we don't need to use the method twice?

cooltey (changes requested):

PR #2972, opened 2021-11-10
Fix: handle redirected title properly before saving into database.

cooltey:
https://phabricator.wikimedia.org/T184828

The way of preventing saving an incorrect prefix title to the database is by calling the `/page/summary` endpoint to get the correct title. 

In this case, the user tended to store "B*-tree" article, but it is actually a section inside the "B-tree" article, and the API request will be redirected to "B-tree#Variants" and that's why the offline reading list does not work properly.

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageFragment.kt
+++ b/app/src/main/java/org/wikipedia/page/PageFragment.kt
@@ -1131,14 +1131,17 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
 
     fun addToReadingList(title: PageTitle, source: InvokeSource, addToDefault: Boolean) {
         if (addToDefault) {
+            var finalPageTitle = title
             // Make sure handle redirected title before saving into database
             disposables.add(ServiceFactory.getRest(title.wikiSite).getSummary(null, title.prefixedText)
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
-                .subscribe({ summary ->
-                    val finalPageTitle = summary.getPageTitle(title.wikiSite)
+                .doAfterTerminate {
                     ReadingListBehaviorsUtil.addToDefaultList(requireActivity(), finalPageTitle, source) { readingListId ->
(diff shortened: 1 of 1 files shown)


sharvaniharan:
Works and is fixed.. just wondering if we can collect this information when we already make a call to pageSummary endpoint in PageFragmentLoadState, by storing it into PageViewModel on PageFragment?

app/src/main/java/org/wikipedia/page/PageFragment.kt
```diff
@@ -1131,8 +1131,17 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
...
+            // Make sure handle redirected title before saving into database
+            disposables.add(ServiceFactory.getRest(title.wikiSite).getSummary(null, title.prefixedText)
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe({ summary ->
+                    val finalPageTitle = summary.getPageTitle(title.wikiSite)
+                    ReadingListBehaviorsUtil.addToDefaultList(requireActivity(), finalPageTitle, source) { readingListId ->
+                        moveToReadingList(readingListId, finalPageTitle, source, false) }
+                }, {
+                    L.e(it)
```
dbrant, line 1143: If fetching the redirected title fails for any reason, can we fall back to saving the original un-redirected title?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageFragment.kt
cooltey: Yes we can! Done.

cooltey (commented):
> 
> 
> Works and is fixed.. just wondering if we can collect this information when we already make a call to pageSummary endpoint in PageFragmentLoadState, by storing it into PageViewModel on PageFragment?

The idea of calling `page/summary` is to prevent saving incorrect titles when long-pressing in-article links, so explicitly calling the API is somehow required to fix the issue.

PR #2973, opened 2021-11-10
Skip markReadItems action when clicking on offline notification items

cooltey:
Tweak PR that based on the comment in [2872](https://github.com/wikimedia/apps-android-wikipedia/pull/2872#issuecomment-963740971), which will not show the toast message and skip the "markReadItems" action since maybe some pages will be available if the user visited them before.


sharvaniharan:
This doesn't seem to resolve the issue completely @cooltey ... I was thinking more like disabling click on any notification item, because we know it is going to fail... what do you think... Also a question you could ask on the task may be?

cooltey:
> 
> 
> This doesn't seem to resolve the issue completely @cooltey ... I was thinking more like disabling click on any notification item, because we know it is going to fail... what do you think... Also a question you could ask on the task may be?

As per the description on this PR, some items may have been visited while online and we can visit the pages even offline. It does not make sense to me if we make all items unclickable when offline. 

And yes, here's the comment for folks to discuss: https://phabricator.wikimedia.org/T288358#7497511

PR #2980, opened 2021-11-16
Update error message when User page doesn't exist

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T295071


sharvaniharan:
Ready for review @cooltey @dbrant

PR #2981, opened 2021-11-16
Fix and simplify counting of unread notifications in tab headers.

dbrant:
https://phabricator.wikimedia.org/T295785

### DIFF
--- a/app/src/main/java/org/wikipedia/notifications/NotificationActivity.kt
+++ b/app/src/main/java/org/wikipedia/notifications/NotificationActivity.kt
@@ -279,7 +279,6 @@ class NotificationActivity : BaseActivity() {
         val selectedFilterTab = binding.notificationTabLayout.selectedTabPosition
         val filteredList = notificationList
             .filter { if (Prefs.hideReadNotificationsEnabled) it.isUnread else true }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/notifications/NotificationActivity.kt
```diff
@@ -288,7 +287,20 @@ class NotificationActivity : BaseActivity() {
...
+        var mentionsUnreadCount = 0
+
         for (n in filteredList) {
+            val isMention = NotificationCategory.isMentionsGroup(n.category)
+            if (n.isUnread) {
+                allUnreadCount++
+                if (isMention) {
+                    mentionsUnreadCount++
+                }
+            }
```
cooltey, line 300: They will not be counted correctly if the `excludedTypeCodes` is not empty. Should we move the logic after line `315` or some other place?

cooltey (changes requested):

app/src/main/java/org/wikipedia/notifications/NotificationActivity.kt
dbrant: Whoops, that's correct! How about now?

PR #2983, opened 2021-11-16
Update Kotlin gradle plugin version to 1.6.0

cooltey:


app/src/main/java/org/wikipedia/history/HistoryEntry.kt
```diff
@@ -20,12 +20,12 @@ import java.util.*
 @TypeParceler<Date, DateParceler>()
 @Entity
 class HistoryEntry(
```
dbrant, line 22: Would you mind adding a comment here saying something like "TODO: change these properties back to `val` when HistoryEntry is no longer serializable. (i.e. when we update Tabs to be in the database instead of Prefs)"

dbrant (changes requested):

app/src/main/java/org/wikipedia/history/HistoryEntry.kt
cooltey: Of course!

PR #3004, opened 2021-11-23
Share design tweak

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T295808#7522705


app/src/main/res/menu/menu_talk.xml
```diff
@@ -4,22 +4,39 @@
...
+        app:showAsAction="always" />
+
+
     <item
-        android:id="@+id/menu_talk_topic_share"
-        android:icon="@drawable/ic_share_material_secondary_themed"
-        android:title="@string/menu_option_share_talk_page"
-        app:showAsAction="never" />
+        android:id="@+id/menu_notifications_group"
+        android:icon="@drawable/ic_more_vert_themed_24dp"
```
cooltey, line 20: Looks like the `themed` icons are using different color attributes here. 

The `ic_translate_themed_24dp` is using `material_theme_secondary_color` but the notifications and more menu are using `toolbar_icon_color`. Could you please check which one is the correct one?

cooltey (changes requested):

PR #3005, opened 2021-11-24
Fix: show the "no results" item properly in the NotificationActivity

cooltey:
https://phabricator.wikimedia.org/T296261


cooltey:
I cannot find the solution for this fix yet, will re-release to beta without fixing this.

cooltey:
> This still doesn't seem to solve the issue of the "No results" message getting hidden behind the toolbar when the user scrolls through the notifications before typing the search term:

Done!

sharvaniharan:
Looks good now..

PR #3011, opened 2021-11-28
Use Map entry destructuring declarations.

Isira-Seneviratne:


cooltey:
PR looks good to me. Please fix conflicting files.

PR #3020, opened 2021-12-02
Enable editing Talk page as wikitext

cooltey:
https://phabricator.wikimedia.org/T295809


sharvaniharan:
Looks good @cooltey ... Just a couple of things:
1. If a user edits source and replies to a talk page message, on success of publishing the response, on the topic screen the newly added message is not shown. We have to refresh to see it. I feel it will make sense too automatically refresh so the user sees the message they added.
2. In the preview we are showing the tilde ~~~~ as part of the message. I feel this is not an accurate preview. May be we should either not show it at all, or somehow indicate that a timestamp will appear there - Maybe a question for Robin.

sharvaniharan (changes requested):
Looks good @cooltey ... Just a couple of things:

If a user edits source and replies to a talk page message, on success of publishing the response, on the topic screen the newly added message is not shown. We have to refresh to see it. I feel it will make sense too automatically refresh so the user sees the message they added.
In the preview we are showing the tilde ~~~~ as part of the message. I feel this is not an accurate preview. May be we should either not show it at all, or somehow indicate that a timestamp will appear there - Maybe a question for Robin.

cooltey:
> Looks good @cooltey ... Just a couple of things:
> 
> If a user edits source and replies to a talk page message, on success of publishing the response, on the topic screen the newly added message is not shown. We have to refresh to see it. I feel it will make sense too automatically refresh so the user sees the message they added. In the preview we are showing the tilde ~~~~ as part of the message. I feel this is not an accurate preview. May be we should either not show it at all, or somehow indicate that a timestamp will appear there - Maybe a question for Robin.

Fixed and left a comment to the ticket to address the question. Thanks!

PR #3022, opened 2021-12-03
Create Customizable Toolbar

cooltey:
https://phabricator.wikimedia.org/T296753


cooltey:
@dbrant @sharvaniharan The PR is ready for review.

sharvaniharan (changes requested):
Looks very good and is very seamless! 
One minor change.. the icon tint looks off from what i can see on the ticket..

sharvaniharan:
One other thing, please review and merge this: #3040 and wire-it up on exit of this customization screen.

cooltey:
> Looks very good and is very seamless! One minor change.. the icon tint looks off from what i can see on the ticket..

Done!

app/src/main/java/org/wikipedia/page/customize/CustomizeQuickActionsActivity.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.wikipedia.analytics.SuggestedEditsFunnel
+
+class CustomizeQuickActionsActivity : SingleFragmentActivity<CustomizeQuickActionsFragment>() {
+    public override fun createFragment(): CustomizeQuickActionsFragment {
+        return CustomizeQuickActionsFragment.newInstance()
+    }
+
+    companion object {
+        fun newIntent(context: Context): Intent {
+            SuggestedEditsFunnel.get().contributionsOpened()
```
sharvaniharan, line 15: Please remove this @cooltey

cooltey: Done

cooltey:
@sharvaniharan @dbrant This PR passed the design review and is ready for code review now.

app/src/main/java/org/wikipedia/theme/ThemeChooserDialog.kt
```diff
@@ -86,7 +86,12 @@ class ThemeChooserDialog : ExtendedBottomSheetDialogFragment() {
         updateComponents()
         disableBackgroundDim()
         requireDialog().window?.let {
-            DeviceUtil.setNavigationBarColor(it, getThemedColor(requireContext(), R.attr.paper_color))
+            DeviceUtil.setNavigationBarColor(it, ResourceUtil.getThemedColor(requireContext(), R.attr.paper_color))
+        }
+
+        // TODO: test only
```
dbrant, line 92: Is this still a TODO?

dbrant (changes requested):
Functionally this works very well! Just a few requests and observations:

* When rearranging the menu items in the list, it looks like grabbing anywhere in an item causes it to start rearranging. This makes it impossible to _scroll_ the whole list up and down. Can we make it so that the items start rearranging only when grabbing the rearrange button on the right? (similar to how the Language list works)

* Have we thought about how we should handle long captions on toolbar buttons that overflow onto multiple lines?

image

* I'm not a designer, but the visual appearance of the "Customize toolbar" button clashes with almost every other UI element in the app. The color doesn't match anything in the WMF guidelines. Are we sure about this?

image

app/src/main/java/org/wikipedia/theme/ThemeChooserDialog.kt
sharvaniharan: Removed in my PR @dbrant https://github.com/wikimedia/apps-android-wikipedia/pull/3060/files#diff-c1eae10dbe3c866e252b9d1e4a1386098e0313ee4b3e23c11f23d02360b2ab59L92

cooltey:
> Functionally this works very well! Just a few requests and observations:
> 
>     * When rearranging the menu items in the list, it looks like grabbing anywhere in an item causes it to start rearranging. This makes it impossible to _scroll_ the whole list up and down. Can we make it so that the items start rearranging only when grabbing the rearrange button on the right? (similar to how the Language list works)

Yes, I think this also makes sense to me. Updated! 

And I'll leave a comment on the ticket.

>     * Have we thought about how we should handle long captions on toolbar buttons that overflow onto multiple lines?
> 

This too, I'll let Robin and Jazmin know about it.

cooltey:
> Functionally this works very well! Just a few requests and observations:
> 
>     * When rearranging the menu items in the list, it looks like grabbing anywhere in an item causes it to start rearranging. This makes it impossible to _scroll_ the whole list up and down. Can we make it so that the items start rearranging only when grabbing the rearrange button on the right? (similar to how the Language list works)
> 

This one is resolved. Please let me know how it looks to you.

PR #3024, opened 2021-12-04
Replaced hardcoded dispatchers with dependency injection in pageavailableofflinehandler

Cerozob:
Best practice according to Android documentation


dbrant (changes requested):
This pull request is doing two different things: (1) replacing hard-coded dispatchers, which is not necessary, and (2) using a Binding class in a view holder. We may consider merging the latter, if you focus this pull request just on that.

PR #3025, opened 2021-12-04
Removed fixed cache size in ServiceFactory

Cerozob:
Not taking into account the current available memory can lead to an OOM


dbrant:
This feels like premature optimization. Have you actually profiled the memory impact of creating Retrofit objects? Have you observed OOM errors due to this?

PR #3026, opened 2021-12-05
Use LocalDate in On This Day.

Isira-Seneviratne:


dbrant:
Closing, since we are not switching to `java.time` yet.

PR #3029, opened 2021-12-07
Dynamically resolve Talk page names.

dbrant:
This will now automatically "resolve" the Talk page that corresponds to a regular page, without needing to construct the Talk title ourselves.
The Talk activity will now accept any regular page title, and will automatically query the API to get the correct Talk page.

https://phabricator.wikimedia.org/T297124

### DIFF
--- a/app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
+++ b/app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
@@ -254,6 +254,11 @@ class TalkTopicsActivity : BaseActivity() {
                     resolveTitleRequired = false
                     response.query?.namespaces?.let { namespaces ->
                         response.query?.firstPage()?.let { page ->
+                            // In MediaWiki, namespaces that are even-numbered are "regular" pages,
+                            // and namespaces that are odd-numbered are the "Talk" versions of the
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
```diff
@@ -232,9 +247,23 @@ class TalkTopicsActivity : BaseActivity() {
...
                 .subscribeOn(Schedulers.io())
                 .observeOn(AndroidSchedulers.mainThread())
+                .flatMap { response ->
+                    resolveTitleRequired = false
+                    response.query?.namespaces?.let { namespaces ->
+                        response.query?.firstPage()?.let { page ->
+                            val newNs = namespaces.values.find { it.id == page.namespace().code() + 1 }
+                            if (page.namespace().code() % 2 == 0 && newNs != null) {
+                                pageTitle.namespace = newNs.name
+                            }
```
cooltey, line 260: Would it be possible to add some comments here to explain the logic?

dbrant: Sure! I agree this is a little too much voodoo without adequate comments.

PR #3030, opened 2021-12-08
Reading focus mode.

dbrant:
https://phabricator.wikimedia.org/T254771


app/src/main/java/org/wikipedia/views/ObservableWebView.kt
```diff
@@ -135,8 +135,10 @@ class ObservableWebView : WebView {
             MotionEvent.ACTION_UP -> {
                 if (abs(event.x - touchStartX) <= touchSlop &&
                         abs(event.y - touchStartY) <= touchSlop) {
-                    if (onClickListeners.any { it.onClick(event.x, event.y) }) {
-                        return true
+                    if (hitTestResult.type == HitTestResult.UNKNOWN_TYPE) {
```
cooltey, line 138: Could you please add a note here to explain why we should check here and why `HitTestResult.UNKNOWN_TYPE`?

cooltey (commented):
Looks good to me and only one comment below:

app/src/main/java/org/wikipedia/views/ObservableWebView.kt
dbrant: done!

cooltey (approved):
Looks good to me now. Just curious that is it possible to send the reading focus mode variable to the `mobile-html` in the initial setup? So that we will not see the edit buttons pop up and disappear every time when we jump to another article.

dbrant:
> Looks good to me now. Just curious that is it possible to send the reading focus mode variable to the `mobile-html` in the initial setup? So that we will not see the edit buttons pop up and disappear every time when we jump to another article.

@cooltey Good point, but I don't think the page-library currently supports passing an "editable" variable in the initial setup. I agree this would be useful, but we'll need to add it to PCS, and then follow it up with another client-side patch.

PR #3032, opened 2021-12-08
Poll notifications in a JobService

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T296772


app/src/main/java/org/wikipedia/util/JobsUtil.kt
```diff
@@ -0,0 +1,16 @@
+package org.wikipedia.util
+
+import android.app.job.JobInfo
+import android.app.job.JobScheduler
+import android.content.ComponentName
+import android.content.Context
+import org.wikipedia.notifications.PollNotificationService
+
+object JobsUtil {
+    fun schedulePollNotificationJob(context: Context) {
```
dbrant, line 10: Can this be placed in a companion object of `PollNotificationService` itself, instead of a separate file?

app/src/main/java/org/wikipedia/notifications/PollNotificationService.kt
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    override fun onStopJob(p0: JobParameters?): Boolean {
+        return true
+    }
+
+    private fun pollNotifications(context: Context) {
+        CoroutineScope(Dispatchers.Default).launch(CoroutineExceptionHandler { _, t ->
+            if (t is MwException && t.error.title == "login-required") {
+                NotificationPollBroadcastReceiver.assertLoggedIn()
```
dbrant, line 32: The `assertLoggedIn()` function can be moved into this class, and made private.

app/src/main/java/org/wikipedia/notifications/PollNotificationService.kt
```diff
@@ -0,0 +1,75 @@
...
+            val lastNotificationTime =
+                response.query?.notifications?.list?.maxOfOrNull { it.utcIso8601 }.orEmpty()
+            if (lastNotificationTime > Prefs.remoteNotificationsSeenTime) {
+                Prefs.remoteNotificationsSeenTime = lastNotificationTime
+                retrieveNotifications(context)
+            }
+        }
+    }
+
+    suspend fun retrieveNotifications(context: Context) {
```
dbrant, line 46: Can be private.

dbrant (changes requested):
Great; This is basically what it needs to be, but there's one little bit that's missing:
As I understand it, if we're going to use `JobService`, we need to take care to call the `jobFinished()` function at the end of our job, to signal to the system that the job is complete. Otherwise the system will hold on to the wakelock longer than necessary.

(let's call `jobFinished()` at the end of the success case and the error case in the `pollNotifications` coroutine.)

Additional stuff inline.

app/src/main/java/org/wikipedia/notifications/PollNotificationService.kt
```diff
@@ -0,0 +1,106 @@
...
+
+    private var jobParameters: JobParameters? = null
+
+    override fun onStartJob(jobParameters: JobParameters?): Boolean {
+        pollNotifications(this)
+        return true
+    }
+
+    override fun onStopJob(p0: JobParameters?): Boolean {
+        jobParameters = p0
```
dbrant, line 33: Shouldn't we assign `jobParameters` when the job is started, instead of stopped?

app/src/main/java/org/wikipedia/notifications/PollNotificationService.kt
```diff
@@ -0,0 +1,106 @@
...
+    private fun assertLoggedIn() {
+        // Attempt to get a dummy CSRF token, which should automatically re-log us in explicitly,
+        // and should automatically log us out if the credentials are no longer valid.
+        CsrfTokenClient(WikipediaApp.getInstance().wikiSite).token
+            .subscribeOn(Schedulers.io())
+            .subscribe()
+    }
+
+    companion object {
+        var pollNotificationService = PollNotificationService()
```
dbrant, line 94: Instances of `Service` classes are created by the system. Keeping our own instance of a Service class is not correct.

app/src/main/java/org/wikipedia/notifications/PollNotificationService.kt
```diff
@@ -0,0 +1,106 @@
...
+
+        fun schedulePollNotificationJob(context: Context) {
+            val serviceComponent = ComponentName(context, pollNotificationService.javaClass)
+            val builder = JobInfo.Builder(0, serviceComponent)
+            val jobScheduler = context.getSystemService(Context.JOB_SCHEDULER_SERVICE) as JobScheduler
+            jobScheduler.schedule(builder.build())
+        }
+         fun pollNotificationJobFinished() {
+            pollNotificationService.jobFinished(pollNotificationService.jobParameters, false)
```
dbrant, line 103: You're finishing the job of a Service class that was never created by the OS, so this will not do anything.

app/src/main/java/org/wikipedia/notifications/PollNotificationService.kt
```diff
@@ -0,0 +1,106 @@
...
+import org.wikipedia.dataclient.mwapi.MwException
+import org.wikipedia.settings.Prefs
+import org.wikipedia.util.log.L
+
+class PollNotificationService : JobService() {
+
+    private var jobParameters: JobParameters? = null
+
+    override fun onStartJob(jobParameters: JobParameters?): Boolean {
+        pollNotifications(this)
```
dbrant, line 28: You can pass `jobParameters` on to the `pollNotifications()` function, and finish the job at the end of the coroutine (success case and error case) in that function.  (And not need to keep `jobParameters` as a class property.)

dbrant (changes requested):

Isira-Seneviratne:
Couldn't [AndroidX Work](https://developer.android.com/reference/androidx/work/package-summary) be used for this?

PR #3036, opened 2021-12-14
Use AndroidX Work.

Isira-Seneviratne:


cooltey:
Thanks for submitting the PR. Could you please explain briefly the benefit of using `CoroutineWorker` instead of `JobService` and how does it affect the project?

Isira-Seneviratne:
`WorkManager` provides several [advantages](https://developer.android.com/topic/libraries/architecture/workmanager#features) over `JobScheduler`, such as the ability to chain work units together and make use of work constraints (e.g. running only on Wi-Fi).

In addition, `JobIntentService` has been deprecated in favor of `WorkManager`.

cooltey (approved):
Worked for me. @dbrant @sharvaniharan Please review and merge if it looks good to you too.

PR #3040, opened 2021-12-16
MEP: Wiring for Customize toolbar schema

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T297818


app/src/main/java/org/wikipedia/analytics/eventplatform/CustomizeToolbarEvent.kt
```diff
@@ -0,0 +1,30 @@
...
+@Serializable
+@SerialName("/analytics/mobile_apps/android_customize_toolbar_interaction/1.0.0")
+class CustomizeToolbarEvent(private var is_anon: Boolean,
+                            private var is_rfm_enabled: Boolean,
+                            private var source: String,
+                            private var favorites_order: List<Int>,
+                            private var menu_order: List<Int>,
+                            private var time_spent_ms: Int) : TimedEvent(STREAM_NAME) {
+
+    fun logCustomization(source: String, favorites_order: List<Int>, menu_order: List<Int>, is_rfm_enabled: Boolean) {
```
cooltey, line 17: Would it be possible to make the parameters have a naming style like this: `favoritesOrder`?

cooltey (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/CustomizeToolbarEvent.kt
sharvaniharan: The parameter names have to match the MEP tables Cooltey ... and they have a snake_case convention..

app/src/main/java/org/wikipedia/analytics/eventplatform/TimedEvent.kt
```diff
@@ -0,0 +1,36 @@
...
+open class TimedEvent(@Transient private val streamName: String = "") : Event(streamName) {
+
+    @Transient
+    private var startTime = 0L
+
+    @Transient
+    private var pauseTime = 0L
+
+    @Transient
+    val duration = System.currentTimeMillis() - startTime
```
dbrant, line 16: @sharvaniharan  this field is never updated, and will therefore always be incorrect. You probably intended for this to be a `get()` property?

sharvaniharan: Thank you.. fixed here: #3237

PR #3041, opened 2021-12-19
Support for East Asian numeric prefix

akihironagai:
Support for East Asian numeric prefix in "Top read".

In East Asian language, numbers are grouped into fours.

万 (萬, 만) = 10^4

1,000 = 1 k = 1000
1,0000 = 10 k = 1 万
1,00000 = 100 k = 10 万
1,000,000 = 1 M = 100 万
1,000,0000 = 10 M = 1000 万

|Before|After|
|---|---|
|Screenshot_20211220-002322_Wikipedia Beta|Screenshot_20211220-002317_Wikipedia Alpha|


dbrant:
@akihironagai Thanks for the PR.  As it is right now, this doesn't work for us because our string translation files are auto-generated from our crowdsourced translation site (https://translatewiki.net).

Can all this logic be made even simpler by using `icu.text.CompactDecimalFormat` instead?

akihironagai:
> Can all this logic be made even simpler by using icu.text.CompactDecimalFormat instead?

I've implemented the logic using CompactDecimalFormat.
As a result, `view_top_read_card_pageviews_k_suffix` and `view_top_read_card_pageviews_m_suffix` are no longer needed from `strings.xml`.

cooltey (approved):

app/build.gradle
```diff
@@ -216,6 +216,7 @@ dependencies {
     implementation "com.github.bumptech.glide:okhttp3-integration:$glideVersion"
     kapt("com.github.bumptech.glide:compiler:$glideVersion")
 
+    implementation 'com.ibm.icu:icu4j:70.1'
```
dbrant, line 219: We would rather not import a whole new dependency just for this single feature. We can just use `android.icu.text.CompactDecimalFormat`, and make it conditional for API 24+.

dbrant (changes requested):

app/build.gradle
akihironagai: I've fixed it.

PR #3043, opened 2021-12-21
MEP: Wire-up the app to send events to 'editattemptstep'

sharvaniharan:
**Phab:**https://phabricator.wikimedia.org/T297817

LEgacy schemas on MEP do not accept app-specific info like app_session_id and app_install_id. So the app architecture has been tweaked to have a base `Event` and our regular app schemas will be sent `AppsEvent` and in the future if the legacy schemas have special reqs, we can have `LegacyEvent` class extending the base `Event`. Modeled after the mep packages under `analytics`, `legacy`,`mobile_apps`, etc

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/AppsEvent.kt
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/AppsEvent.kt
@@ -5,7 +5,7 @@ import kotlinx.serialization.Serializable
 import kotlinx.serialization.Transient
 
 @Serializable
-sealed class AnalyticsEvent(@Transient val streamName: String = "") : Event(streamName) {
+sealed class AppsEvent(@Transient val streamName: String = "") : Event(streamName) {
 
     @SerialName("app_session_id")
     var sessionId: String? = null
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/EditAttemptStepEvent.kt
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/EditAttemptStepEvent.kt
@@ -4,37 +4,37 @@ import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import org.wikipedia.R
 import org.wikipedia.WikipediaApp
-import org.wikipedia.analytics.eventplatform.ActionType.*
-import org.wikipedia.analytics.eventplatform.EditorInterfaceType.WIKITEXT
-import org.wikipedia.auth.AccountUtil
+import org.wikipedia.analytics.eventplatform.EditorInterfaceType.OTHER
 
 @Serializable
 @SerialName("/analytics/legacy/editattemptstep/1.2.0")
 class EditAttemptStepEvent(private val event: EditAttemptStepInteractionEvent) : Event(STREAM_NAME) {
+
     companion object {
+
         fun logInit() {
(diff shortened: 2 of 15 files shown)


app/src/main/java/org/wikipedia/analytics/eventplatform/EditAttemptStepEvent.kt
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        private fun submitEditAttemptEvent(action: ActionType) {
+            EventPlatformClient.submit(EditAttemptStepEvent(EditAttemptStepInteractionEvent(action.valueString, "example",
+                OTHER.valueString, INTEGRATION_ID, "example", PLATFORM, 0, 1, 1)))
+        }
+
+        private const val STREAM_NAME = "eventlogging_EditAttemptStep"
+        private const val INTEGRATION_ID = "app-android"
+        private val PLATFORM = WikipediaApp.getInstance().getString(R.string.device_type).lowercase()
```
cooltey, line 42: Would it be possible to have these variables at the top of the `companion object`.

app/src/main/java/org/wikipedia/analytics/eventplatform/AppsEvent.kt
```diff
@@ -0,0 +1,15 @@
+package org.wikipedia.analytics.eventplatform
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.Transient
+
+@Serializable
+sealed class AppsEvent(@Transient val streamName: String = "") : Event(streamName) {
```
cooltey, line 8: Would it be possible to make it as `AppEvent` instead of a plural name of it?

cooltey (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/AppsEvent.kt
sharvaniharan: So these are the events that only the `apps` send Cooltey... then there is a provision for `legacy` events... just maintaining the same naming as the MEP repository... actually they name it `mobile_apps` so you feel it is better to name `MobileAppsEvent` on our side too?

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -108,7 +108,7 @@ class DescriptionEditFragment : Fragment() {
         val type = if (pageTitle.description == null) DescriptionEditFunnel.Type.NEW else DescriptionEditFunnel.Type.EXISTING
         funnel = DescriptionEditFunnel(WikipediaApp.getInstance(), pageTitle, type, invokeSource)
         funnel.logStart()
-        EditAttemptStepEvent.logInit()
+        EditAttemptStepEvent.logInit(pageTitle.wikiSite.languageCode)
```
cooltey, line 111: Would it be possible to initialize the `EditAttemptStepEvent` with the `WikiSite` on the top so we don't have to pass `languageCode` every time when we call it.

app/src/main/java/org/wikipedia/analytics/eventplatform/AppsEvent.kt
cooltey: `MobileAppsEvent` sounds good to me.

cooltey (approved):
Looks good to me now. Please consider the suggestion in the follow-up PR: https://github.com/wikimedia/apps-android-wikipedia/pull/3043#pullrequestreview-838844504

PR #3048, opened 2021-12-25
Rename the Abkhaz language

amire80:
Phabricator bug: https://phabricator.wikimedia.org/T298309


dbrant:
@amire80 Thanks for this! Our list of languages is auto-generated by a script that we run periodically. Once that change is actually deployed to abwiki, it will be automatically picked up the next time we run our script.

PR #3059, opened 2022-01-06
Fix a few warnings and deprecations.

dbrant:


app/src/main/java/org/wikipedia/analytics/EventLoggingService.kt
```diff
@@ -43,7 +43,7 @@ class EventLoggingService private constructor() {
 
         // https://github.com/wikimedia/mediawiki-extensions-EventLogging/blob/8b3cb1b/modules/ext.eventLogging.core.js#L57
         private const val MAX_URL_LEN = 2000
-        private val EMPTY_REQ = RequestBody.create(null, ByteArray(0))
```
dbrant, line 46: This was deprecated, and moved to an extension function.

app/src/main/java/org/wikipedia/gallery/MediaListItem.kt
```diff
@@ -34,11 +34,10 @@ class MediaListItem constructor(val title: String = "",
 
     @Serializable
     @Parcelize
-    class ImageSrcSet : Parcelable {
-
-        @SerialName("scale")
-        private val _scale: String? = null
-        val src: String = ""
```
dbrant, line 41: These parameters were not being parceled correctly, and need to be in the primary constructor.

app/src/main/java/org/wikipedia/views/CustomDatePicker.kt
```diff
@@ -91,10 +91,10 @@ class CustomDatePicker : DialogFragment() {
 
             init {
                 binding.root.setOnClickListener {
-                    selectedDay[Calendar.DATE] = adapterPosition + 1
```
dbrant, line 94: The getter for `adapterPosition` is deprecated, but is equivalent to `bindingAdapterPosition`.

PR #3060, opened 2022-01-06
RFM: Updates to theme bottom sheet

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T297134


app/src/main/res/xml/developer_preferences.xml
```diff
@@ -384,6 +384,10 @@
             android:key="@string/preference_key_show_search_tab_tooltip"
             android:title="@string/preference_key_show_search_tab_tooltip" />
 
+        <SwitchPreferenceCompat
+            android:key="@string/preference_key_customize_toolbar_tooltip"
+            android:title="@string/preference_key_customize_toolbar_tooltip" />
```
cooltey, line 389: Could you please move this tooltip under "Onboarding" category? (If possible, please also move the search tab tooltip as well).

cooltey (changes requested):

app/src/main/java/org/wikipedia/page/PageFragment.kt
```diff
@@ -278,8 +277,8 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
         updateQuickActionsAndMenuOptions()
     }
 
-    fun getPageActionsTabLayoutViews(): List<TextView> {
-        return binding.pageActionsTabLayout.tabViewsList
+    fun getPageActionsTabLayoutViews(): PageActionTabLayout {
```
cooltey, line 280: Can it be renamed to `getPageActionTabLayout`?

app/src/main/java/org/wikipedia/theme/ThemeChooserDialog.kt
```diff
@@ -100,7 +99,7 @@ class ThemeChooserDialog : ExtendedBottomSheetDialogFragment() {
 
     override fun onStart() {
         super.onStart()
-        BottomSheetBehavior.from(requireView().parent as View).peekHeight = DimenUtil.roundedDpToPx(DimenUtil.getDimension(R.dimen.themeChooserSheetPeekHeight))
+        BottomSheetBehavior.from(requireView().parent as View).peekHeight = (DimenUtil.displayHeightPx * 0.75).toInt()
```
cooltey, line 102: We can have another set here to expand the bottom sheet by default.
```
BottomSheetBehavior.from(requireView().parent as View).state = BottomSheetBehavior.STATE_EXPANDED
```

cooltey (changes requested):

app/src/main/java/org/wikipedia/theme/ThemeChooserDialog.kt
sharvaniharan: but Robin wants it to be at .75 of the screen cooltey.... if i set only the state then it might expand beyond .75 on small screens

PR #3070, opened 2022-01-13
Fix: TimedFunnel to send the correct duration

sharvaniharan:


app/src/main/java/org/wikipedia/analytics/TimedFunnel.kt
```diff
@@ -12,7 +12,8 @@ abstract class TimedFunnel @JvmOverloads constructor(app: WikipediaApp, schemaNa
     private var pauseTime = 0L
     /** Override me for deviant implementations.  */
     private val durationFieldName = "time_spent"
-    private val duration = System.currentTimeMillis() - startTime
+    private val duration: Long
+        get() = System.currentTimeMillis() - startTime
     private val durationSeconds = TimeUnit.MILLISECONDS.toSeconds(duration)
```
dbrant, line 17: `durationSeconds` also probably needs to have a getter.

dbrant (changes requested):

sharvaniharan:
Done @dbrant

PR #3071, opened 2022-01-13
At-mentions!

dbrant:
https://phabricator.wikimedia.org/T297912


app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
```diff
@@ -531,6 +540,50 @@ class TalkTopicActivity : BaseActivity(), LinkPreviewDialog.Callback {
...
+                    break
+                }
+                val name = it.html!!.substring(start, end)
+                val title = PageTitle(name, pageTitle.wikiSite)
+                if (title.namespace() == Namespace.USER || title.namespace() == Namespace.USER_TALK) {
+                    userList.add(0, StringUtil.removeUnderscores(title.text))
+                }
+                start = end
+            }
+            userList.forEach { userNames.add(it) }
```
cooltey, line 582: Can we use `userNames.addAll(userList)`?

app/src/main/java/org/wikipedia/views/UserMentionEditText.kt
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Returns the text of this field, with properly expanded user mentions.
+     * Each user mention (in th form of @{username}) will be expanded into a wiki link to the
+     * user page of that user, i.e. [[User:username|@username]]
+     */
+    fun getParsedText(wikiSite: WikiSite): String {
+        if (text == null) {
+            return ""
+        }
+        var str = text!!.toString()
```
cooltey, line 165: Would it be possible to assign a value of `text` so that we can reduce the non-null symbol on it? e.g.: `val editText = text!!`

app/src/main/java/org/wikipedia/views/UserMentionInputView.kt
```diff
@@ -0,0 +1,154 @@
...
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe({ response ->
+                    userNameList.clear()
+
+                    userNameHints.forEach {
+                        if (it.startsWith(prefix, ignoreCase = true)) {
+                            userNameList.add(it)
+                        }
+                    }
```
cooltey, line 96: Would it be possible to simply it as following:
```
userNameList.addAll(userNameHints.filter { it.startsWith(prefix, ignoreCase = true) })
```

cooltey (changes requested):

app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
dbrant: yes!

app/src/main/java/org/wikipedia/views/UserMentionEditText.kt
dbrant: yep!

app/src/main/java/org/wikipedia/views/UserMentionInputView.kt
dbrant: sure!

cooltey (changes requested):
Found 1 issue and one discussion will be needed:
1. When clicking on "Reply" and changing the orientation of the device, the bottom "cc" license note will remain on the screen while the `EditText` is `GONE`.
2. When visiting a user's own talk page, should we not @ the user him/she self?

dbrant:
> 1. When clicking on "Reply" and changing the orientation of the device, the bottom "cc" license note will remain on the screen while the `EditText` is `GONE`.

Done!

> 2. When visiting a user's own talk page, should we not @ the user him/she self?

Good point! I added some logic where if the user is replying to their own thread, or on their own page, the text is not pre-populated. I think this makes sense, and probably doesn't need additional discussion.

cooltey (approved):
LGTM. @sharvaniharan Please also review and merge if it looks good to you too.

PR #3077, opened 2022-01-18
ListActivity for Edit History

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T299178


app/src/main/java/org/wikipedia/page/EditHistoryListModelView.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    init {
+        fetchData()
+    }
+
+    private fun fetchData() {
+        viewModelScope.launch(handler) {
+            withContext(Dispatchers.IO) {
+                disposables.add(ServiceFactory.get(WikiSite.forLanguageCode("en")).getEditHistoryDetails("Earth").subscribeOn(Schedulers.io()).observeOn(AndroidSchedulers.mainThread()).subscribe({
```
dbrant, line 36: Quick note: one of the benefits of using coroutines and ViewModels is no longer needing RxJava, and making our network calls into `suspend` functions.

dbrant (changes requested):

cooltey:
Hi @sharvaniharan, would it be possible to create a sub-package folder called `history` or `edit_history` in the `page`?

sharvaniharan:
> Hi @sharvaniharan, would it be possible to create a sub-package folder called `history` or `edit_history` in the `page`?

ok @cooltey will make it `edit_history`

app/src/main/java/org/wikipedia/page/edit_history/EditHistoryListActivity.kt
```diff
@@ -0,0 +1,166 @@
...
+        fun bindItem(listItem: String) {
+            itemView.findViewById<TextView>(id.section_header_text).text = listItem
+        }
+    }
+
+    private inner class EditHistoryListItemHolder constructor(itemView: View) :
+        ViewHolder(itemView) {
+        fun bindItem(oldRevision: Revision?, listItem: Revision) {
+            CoroutineScope(Dispatchers.IO).launch(CoroutineExceptionHandler { _, msg -> run { L.e(msg) } }) {
+                val editDetails: EditDetails = viewModel.fetchEditDetails(pageTitle.wikiSite.languageCode, oldRevision?.revId ?: 0, listItem.revId)
```
dbrant, line 135: I'm not seeing the necessity of making a separate call to fetch the details of each specific revision. Don't we already have all this data in the list of revisions that you fetch initially?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/edit_history/EditHistoryListActivity.kt
sharvaniharan: Correct @dbrant. I will fetch it on need-to basis, but yes talking to Robin about how best to handle the ui without having to make another call for each row, especially since the click takes them to the full diff details view. Your suggestion on the task is a good middle ground.

app/src/main/java/org/wikipedia/page/edit_history/EditHistoryListActivity.kt
```diff
@@ -0,0 +1,166 @@
...
+import androidx.core.content.ContextCompat
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView.Adapter
+import androidx.recyclerview.widget.RecyclerView.ViewHolder
+import com.google.android.material.button.MaterialButton
+import kotlinx.coroutines.CoroutineExceptionHandler
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import org.wikipedia.R.*
```
cooltey, line 21: Can we import only `org.wikipedia.R` so that we can use the resources consistently?

cooltey (changes requested):

app/src/main/java/org/wikipedia/page/edit_history/EditHistoryListActivity.kt
sharvaniharan: @dbrant I am still making the api call for the diffSize, wondering if that should be removed as well, and just show in bytes, which is part of the first call anyway?

dbrant: @sharvaniharan Absolutely, let's use the `size` field that we get for free from the initial call.

app/src/main/java/org/wikipedia/page/edit_history/EditHistoryListActivity.kt
```diff
@@ -0,0 +1,139 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        binding = ActivityEditHistoryBinding.inflate(layoutInflater)
+        setContentView(binding.root)
+        pageTitle = intent.getParcelableExtra(INTENT_EXTRA_PAGE_TITLE)!!
+        viewModel.fetchData(pageTitle)
+        binding.editHistoryLoadProgress.visibility = View.VISIBLE
+        viewModel.editHistoryListData.observe(this) {
+            if (it is Success) {
```
dbrant, line 42: What if `it is Error`?

app/src/main/java/org/wikipedia/page/edit_history/EditHistoryListActivity.kt
```diff
@@ -0,0 +1,139 @@
...
+    private lateinit var editHistoryListAdapter: EditHistoryListAdapter
+    private lateinit var pageTitle: PageTitle
+    private val viewModel: EditHistoryListViewModel by viewModels()
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        binding = ActivityEditHistoryBinding.inflate(layoutInflater)
+        setContentView(binding.root)
+        pageTitle = intent.getParcelableExtra(INTENT_EXTRA_PAGE_TITLE)!!
+        viewModel.fetchData(pageTitle)
```
dbrant, line 39: One of the benefits of using a ViewModel is that network calls don't need to be repeated unnecessarily, e.g. when the activity is recreated due to screen rotation, etc.
The way you have it here, the data will be fetched every time the activity is recreated, even if the ViewModel already contains valid data for this `pageTitle`.

dbrant (changes requested):

dbrant:
Merging to keep things moving.

PR #3078, opened 2022-01-18
Fix: Consistency while loading edit history

sharvaniharan:
When we load edit history from the footer of the article, we load the "Special: History/%title" directly into the webview, but when loading it from the menu we are opening the external browser. They should be consistent with each other.


cooltey:
Not sure if it is expected. Could you please confirm with Robin and Jazmin? Thanks.

PR #3083, opened 2022-01-20
Add drop-down menu when tapping on usernames in Talk topics.

dbrant:
This is independent of @-mentions. This should work in the general case of tapping on any username link while browsing talk topics.

https://phabricator.wikimedia.org/T299644

### DIFF
--- a/app/src/main/java/org/wikipedia/talk/UserTalkPopupHelper.kt
+++ b/app/src/main/java/org/wikipedia/talk/UserTalkPopupHelper.kt
@@ -1,0 +1,64 @@
+package org.wikipedia.talk
+
+import android.annotation.SuppressLint
+import android.view.MenuItem
+import android.view.View
+import android.view.ViewGroup
+import androidx.appcompat.app.AppCompatActivity
+import androidx.appcompat.view.menu.MenuBuilder
+import androidx.appcompat.view.menu.MenuPopupHelper
+import org.wikipedia.Constants
+import org.wikipedia.R
+import org.wikipedia.history.HistoryEntry
+import org.wikipedia.page.ExclusiveBottomSheetPresenter
+import org.wikipedia.page.Namespace
+import org.wikipedia.page.PageActivity
+import org.wikipedia.page.PageTitle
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/talk/UserTalkPopupHelper.kt
```diff
@@ -0,0 +1,64 @@
...
+            (rootView as ViewGroup).addView(anchorView)
+
+            val builder = MenuBuilder(activity)
+            activity.menuInflater.inflate(R.menu.menu_user_talk_popup, builder)
+            builder.setCallback(object : MenuBuilder.Callback {
+                override fun onMenuItemSelected(menu: MenuBuilder, item: MenuItem): Boolean {
+                    when (item.itemId) {
+                        R.id.menu_user_profile_page -> {
+                            val entry = HistoryEntry(title, HistoryEntry.SOURCE_TALK_TOPIC)
+                            activity.startActivity(PageActivity.newIntentForNewTab(activity, entry, title))
```
cooltey, line 41: Looks like this does not work properly. If you go to the user profile and press the back button on the article page, it should be back to the talk page instead of the previous page of the current article page.

cooltey (changes requested):

app/src/main/java/org/wikipedia/talk/UserTalkPopupHelper.kt
dbrant: @cooltey I'm unable to reproduce this...
That's why I'm launching the intent in a new tab, so that pressing Back would close the one and only page in the tab. Can you provide a recording?

cooltey: Not sure if it is intended. Please see here: https://www.youtube.com/watch?v=6xnEsDGa8zM

cooltey: Got it. Please make sure to have another PR for it. 😉. The PR looks good to me now.

cooltey (approved):

PR #3084, opened 2022-01-21
Solve data loss

aalmds:
Fix for data loss when an orientation change happens: the dialog that is been displayed in the screen disappears because the activity is recreated.

The following videos show the behaviour before and after the modification, respectively:

https://user-images.githubusercontent.com/61888475/150534582-b284bf94-07e5-426c-ab6c-b005338c7215.mp4

https://user-images.githubusercontent.com/61888475/150534603-9c0a51b1-5f47-4746-bd5b-0cfb466d9dc7.mp4

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -73,6 +73,7 @@
             android:name=".main.MainActivity"
             android:windowSoftInputMode="adjustResize"
             android:theme="@style/AppTheme.Splash"
+            android:configChanges="orientation|screenSize"
             android:exported="true">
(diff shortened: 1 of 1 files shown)


dbrant:
Thanks! However, this is not an acceptable solution. (There's also no "data loss" happening; it's just a dialog box that disappears.)
Merely adding the `configChanges` flag does not solve the underlying issue. In fact it would probably introduce more issues than it solves. The proper solution would be to no longer use `AlertDialog`s, and replace them with `DialogFragment`s, so that the fragment manager would restore them automatically upon screen rotation.

PR #3094, opened 2022-01-31
Don't show "external link" icon for edit history.

dbrant:
Since we're now loading Edit History inside the WebView (via #3078), we should no longer show the "external" icon next to that menu item.


dbrant:
no longer necessary, after the Customize Toolbar patch.

PR #3096, opened 2022-01-31
Follow-up to Customize Toolbar.

dbrant:
(to be reviewed by @cooltey)

* Correctly make items swipeable with long-press action, same as Feed configuration.
* De-duplicate default list of menu items.
* Don't use fully qualified component names in XML.


sharvaniharan (approved):
@cooltey will let you merge.

PR #3102, opened 2022-02-04
Refactor ArticleEditDetailsFragment to use ViewModel

dbrant:
In preparation for: https://phabricator.wikimedia.org/T299180

Using the new ViewModel pattern in this fragment means:
* No more RxJava, only coroutines + LiveData
* I introduced a new `SingleLiveData` helper class, to allow LiveData objects that should be consumed _once_ only, for example the result of a network call that should trigger a Snackbar to be shown. (we don't want the snackbar to be shown again upon screen rotation, which would happen with the standard `MutableLiveData`)

### DIFF
--- a/app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
+++ b/app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
@@ -67,11 +67,11 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback, L
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        retainInstance = true
         revisionId = requireArguments().getLong(EXTRA_EDIT_REVISION_ID, 0)
         languageCode = requireArguments().getString(EXTRA_EDIT_LANGUAGE_CODE, AppLanguageLookUpTable.FALLBACK_LANGUAGE_CODE)
         wikiSite = WikiSite.forLanguageCode(languageCode)
         articlePageTitle = PageTitle(requireArguments().getString(EXTRA_ARTICLE_TITLE, ""), wikiSite)
+        viewModel.setup(articlePageTitle, revisionId)
     }
 
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/wikipedia/diff/ArticleEditDetailsViewModel.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    fun setup(pageTitle: PageTitle, revisionId: Long) {
+        if (curTitle == null) {
+            curTitle = pageTitle
+            getWatchedStatus(pageTitle)
+            getRevisionDetails(pageTitle, revisionId)
+        }
+    }
+
+    fun getWatchedStatus(title: PageTitle) {
```
cooltey, line 46: This can be `private`

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -528,4 +462,17 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback, L
...
+    companion object {
+        const val EXTRA_ARTICLE_TITLE = "articleTitle"
+        const val EXTRA_EDIT_REVISION_ID = "revisionId"
+        const val EXTRA_EDIT_LANGUAGE_CODE = "languageCode"
+
+        fun newInstance(articleTitle: String, revisionId: Long, languageCode: String): ArticleEditDetailsFragment {
+            val articleEditDetailsFragment = ArticleEditDetailsFragment()
+            articleEditDetailsFragment.arguments = bundleOf(EXTRA_ARTICLE_TITLE to articleTitle,
+                    EXTRA_EDIT_REVISION_ID to revisionId, EXTRA_EDIT_LANGUAGE_CODE to languageCode)
+            return articleEditDetailsFragment
```
cooltey, line 475: Would it be possible to use `return ArticleEditDetailsFragment().apply { }`, just like what we have done in other places?

cooltey (changes requested):
Looks pretty good! Just have some minor code review comments inline.

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsViewModel.kt
dbrant: done

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
dbrant: yep!

PR #3112, opened 2022-02-08
Remove RichTextUtil.getSpans().

Isira-Seneviratne:


dbrant (changes requested):
This doesn't actually remove `RichTextUtils.getSpans()` as claimed. Can it?

PR #3113, opened 2022-02-08
New UX for arbitrary diffs.

dbrant:
https://phabricator.wikimedia.org/T299180


app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -463,16 +442,72 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback, L
...
+    }
+
+    companion object {
+        fun newInstance(articleTitle: String, revisionFrom: Long, revisionTo: Long,
+                        languageCode: String): ArticleEditDetailsFragment {
             return ArticleEditDetailsFragment().apply {
-                arguments = bundleOf(EXTRA_ARTICLE_TITLE to articleTitle,
-                        EXTRA_EDIT_REVISION_ID to revisionId,
-                        EXTRA_EDIT_LANGUAGE_CODE to languageCode)
+                arguments = bundleOf(ArticleEditDetailsActivity.EXTRA_ARTICLE_TITLE to articleTitle,
```
cooltey, line 507: Found a language variant issue when trying to fix this one: #3183, would it be possible to pass `PageTitle` instead of plain `String` here? And maybe we can reduce some redundant work of sending `languageCode` or creating `WikiSite` if we send `PageTitle` here.

dbrant: correct!

dbrant:
@cooltey @sharvaniharan This is now ready for review. If there are further design tweaks, they will be very minor follow-ups.

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -221,40 +251,41 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback, L
...
-        binding.olderIdButton.isClickable = olderRevisionId != 0L
-        setEnableDisableTint(binding.newerIdButton, newerRevisionId == -1L)
-        setEnableDisableTint(binding.olderIdButton, olderRevisionId == 0L)
+        binding.progressBar.isVisible = true
+        binding.revisionDetailsView.isVisible = false
+        binding.diffRecyclerView.isVisible = false
+    }
+
+    private fun updateAfterRevisionFetchSuccess() {
+        if (viewModel.revisionFrom != null) {
```
cooltey, line 260: Is it possible to use `viewModel.revisionFrom?.let` so that we don't need to add `!!` in the following places where we use it.

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -221,40 +251,41 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback, L
...
+            binding.usernameFromButton.text = viewModel.revisionFrom!!.user
+            binding.revisionFromTimestamp.text = DateUtil.getDateAndTimeWithPipe(DateUtil.iso8601DateParse(viewModel.revisionFrom!!.timeStamp))
+            binding.overlayRevisionFromTimestamp.text = binding.revisionFromTimestamp.text
+            binding.revisionFromEditComment.text = StringUtil.fromHtml(viewModel.revisionFrom!!.parsedcomment.trim())
+        }
+
+        binding.usernameToButton.text = viewModel.revisionTo!!.user
```
cooltey, line 267: Same as above?

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -322,46 +345,52 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback, L
...
-                DiffResponse.DIFF_TYPE_PARAGRAPH_MOVED_FROM -> {
-                    diffSize -= diff.text.length + 1
-                    updateDiffTextDecor(spannableString, false, prefixLength, prefixLength + diff.text.length)
-                }
-                DiffResponse.DIFF_TYPE_PARAGRAPH_MOVED_TO -> {
-                    diffSize += diff.text.length + 1
-                    updateDiffTextDecor(spannableString, true, prefixLength, prefixLength + diff.text.length)
+    private fun showUndoDialog() {
+        val dialog = UndoEditDialog(requireActivity())
+        dialog.callback = object : UndoEditDialog.Callback {
```
cooltey, line 350: We can add `fun` to the interface if it only has one method so that we can use lambda here.

```
    fun interface Callback {
        fun onSuccess(text: CharSequence)
    }
```

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsViewModel.kt
```diff
@@ -15,67 +14,128 @@ import org.wikipedia.dataclient.mwapi.MwQueryResponse
...
+                    revisionFrom = responseFrom?.query?.firstPage()!!.revisions[0]
+                    revisionTo = pageTo.revisions[0]
+                    canGoForward = revisionTo!!.revId < pageTo.lastrevid
+                } else {
+                    val response = ServiceFactory.get(pageTitle.wikiSite).getRevisionDetailsWithInfo(pageTitle.prefixedText, 2, revisionIdTo)
+                    val page = response.query?.firstPage()!!
+                    val revisions = page.revisions
+                    revisionTo = revisions[0]
+                    canGoForward = revisions[0].revId < page.lastrevid
+                    revisionFrom = if (revisions.size > 1) { revisions[1] } else null
```
cooltey, line 90: We can use `revisions.getOrNull(1)` here

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsViewModel.kt
```diff
@@ -15,67 +14,128 @@ import org.wikipedia.dataclient.mwapi.MwQueryResponse
...
+        viewModelScope.launch(CoroutineExceptionHandler { _, throwable ->
+            revisionDetails.postValue(Resource.Error(throwable))
+        }) {
+            withContext(Dispatchers.IO) {
+                val response = ServiceFactory.get(pageTitle.wikiSite).getRevisionDetailsAscending(pageTitle.prefixedText, 2, revisionIdFrom)
+                val page = response.query?.firstPage()!!
+                val revisions = page.revisions
+
+                revisionFrom = revisions[0]
+                revisionTo = if (revisions.size > 1) { revisions[1] } else revisions[0]
```
cooltey, line 118: We can do this `revisions.getOrElse(1) { revisions.first() }`

cooltey (changes requested):

sharvaniharan:
A couple minor things:
1. Would it be worth it to check if user is anon before sending a `thank` post?
anon_thank

2. The date "buttons" at the bottom, next to compare button, are not actually buttons but look very clickable. Should it just be text without border and background?
3. 
anon_btn

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
dbrant: yep!

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
dbrant: yep!

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
dbrant: yep!

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsViewModel.kt
dbrant: yep!

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsViewModel.kt
dbrant: yep!

dbrant:
@sharvaniharan 
> 1. Would it be worth it to check if user is anon before sending a `thank` post?

Done! good observation.

> 2. The date "buttons" at the bottom, next to compare button, are not actually buttons but look very clickable. Should it just be text without border and background?

I asked, and Robin insisted.

sharvaniharan (approved):
Lgtm, @cooltey please merge if you are done.

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsViewModel.kt
```diff
@@ -15,67 +14,128 @@ import org.wikipedia.dataclient.mwapi.MwQueryResponse
...
+                    var responseFrom: MwQueryResponse? = null
+                    var responseTo: MwQueryResponse? = null
+                    coroutineScope {
+                        launch {
+                            responseFrom = ServiceFactory.get(pageTitle.wikiSite).getRevisionDetailsWithInfo(pageTitle.prefixedText, 2, revisionIdFrom)
+                        }
+                        launch {
+                            responseTo = ServiceFactory.get(pageTitle.wikiSite).getRevisionDetailsWithInfo(pageTitle.prefixedText, 2, revisionIdTo)
+                        }
+                    }
```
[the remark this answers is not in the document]
dbrant: correct!

dbrant:
> Not sure if you want to handle the UI issue in this PR. The user name will be cut off sometimes, please see the following: 

@cooltey Let's tackle that in a follow-up PR... I think there's a lot of nuanced usage of `ConstraintLayout` and `Flow` that we still need to learn.

PR #3115, opened 2022-02-10
Add metrics REST API for showing graph in Edit history page

cooltey:
Based on Toni's comment, this API call can help us draw the graph on the edit history page.
https://phabricator.wikimedia.org/T299181#7698567

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/restbase/Metrics.kt
+++ b/app/src/main/java/org/wikipedia/dataclient/restbase/Metrics.kt
@@ -12,16 +12,16 @@ class Metrics {
 
     @Serializable
     class Items {
-        val project: String? = null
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/dataclient/restbase/Metrics.kt
```diff
@@ -0,0 +1,27 @@
...
+        @SerialName("editor-type") val editorType: String? = null
+        @SerialName("page-title") val pageTitle: String? = null
+        val granularity: String? = null
+        val results: List<Results> = emptyList()
+    }
+
+    @Serializable
+    class Results {
+        val timestamp: String? = null
+        val edits: Int? = null
```
dbrant, line 25: Unless there's good reason, all fields can be nonnull and initialized with `""` or `0`

PR #3117, opened 2022-02-10
Guard against unnecessary POSTs of options API.

dbrant:
https://phabricator.wikimedia.org/T301433

The problem is that we were effectively making this API call every time the app is launched, which is quite excessive. It really only needs to be done upon successful login, or when changing the list of app languages.


Krinkle:
Note to self - this was a follow-up to https://github.com/wikimedia/apps-android-wikipedia/pull/2780.

PR #3127, opened 2022-02-17
Use IO extensions.

Isira-Seneviratne:


dbrant:
Just because an extension method exists, doesn't mean it must be used all the time, in all cases, in every place.

PR #3134, opened 2022-02-18
Only for design review: Diff edit design

sharvaniharan:


app/src/main/res/values/strings.xml
```diff
@@ -51,6 +51,7 @@
     <string name="page_edit_history_activity_label">Page history</string>
     <string name="page_edit_history_minor_edit"><![CDATA[<b>m</b> %s]]></string>
     <string name="page_edit_history_comment_placeholder">There is no comment associated with this edit</string>
+    <string name="page_edit_history_item_size_text">%d bytes</string>
```
dbrant, line 54: Do we really need to say "n bytes"? Why not just the number, like before? I don't think it's very relevant whether it's bytes or characters.

PR #3139, opened 2022-02-23
Add talk page side panel in the article

cooltey:
https://phabricator.wikimedia.org/T299033


cooltey:
@dbrant @sharvaniharan This is ready for review.

app/src/main/java/org/wikipedia/page/SidePanelHandler.kt
```diff
@@ -11,41 +12,61 @@ import android.view.ViewGroup
...
-                                      private val drawerLayout: FixedDrawerLayout,
-                                      private val scrollerView: PageScrollerView,
-                                      private val bridge: CommunicationBridge) :
+class SidePanelHandler internal constructor(private val fragment: PageFragment,
+                                            private val drawerLayout: FixedDrawerLayout,
+                                            private val scrollerView: PageScrollerView,
+                                            private val bridge: CommunicationBridge) :
         ObservableWebView.OnClickListener, ObservableWebView.OnScrollChangeListener, OnContentHeightChangedListener {
 
+    private val containerView = drawerLayout.findViewById<FrameLayout>(R.id.side_panel_container)
```
dbrant, line 52: Instead of getting all these Views through `findViewById()`, can we just pass the `binding` object from PageActivity into here?

app/src/main/java/org/wikipedia/page/SidePanelHandler.kt
```diff
@@ -87,8 +108,80 @@ class ToCHandler internal constructor(private val fragment: PageFragment,
...
+
+        talkRecyclerView.layoutManager = LinearLayoutManager(fragment.requireContext())
+        talkRecyclerView.addItemDecoration(FooterMarginItemDecoration(0, 120))
+        talkRecyclerView.addItemDecoration(DrawableItemDecoration(fragment.requireContext(), R.attr.side_panel_list_separator_drawable, drawStart = true, drawEnd = false))
+        talkRecyclerView.adapter = talkTopicsAdapter
+        talkErrorView.backClickListener = View.OnClickListener {
+            hide()
+        }
+
+        TalkTopicsProvider(newPageTitle).load(object : TalkTopicsProvider.Callback {
```
dbrant, line 136: If the `TalkTopicsProvider` takes a long time to complete the network requests, and the Fragment is destroyed, these callbacks will cause crashes.
We should somehow "clean up" by calling the `TalkTopicsProvider.cancel()` function when the Fragment is destroyed.

app/src/main/java/org/wikipedia/talk/TalkTopicsProvider.kt
```diff
@@ -0,0 +1,92 @@
...
+                }
+                callback.onUpdatePageTitle(pageTitle)
+                ServiceFactory.get(pageTitle.wikiSite).getLastModified(pageTitle.prefixedText)
+            }
+            .observeOn(AndroidSchedulers.mainThread())
+            .flatMap {
+                it.query?.firstPage()?.revisions?.firstOrNull()?.let { revision ->
+                    callback.onReceivedRevision(revision)
+                }
+                ServiceFactory.getRest(pageTitle.wikiSite).getTalkPage(pageTitle.prefixedText)
```
dbrant, line 75: Is it possible to call `getLastModified()` and `getTalkPage()` concurrently, instead of sequentially?

dbrant (changes requested):
Sorry for the delay! Generally looks very good - just some observations.

app/src/main/res/layout/activity_page.xml
```diff
@@ -102,49 +102,181 @@
...
+                            android:textColor="?colorAccent"
+                            android:textSize="20sp"
+                            android:lineSpacingExtra="5sp"
+                            android:textStyle="bold"
+                            android:background="?android:selectableItemBackground"
+                            android:paddingStart="16dp"
+                            android:paddingVertical="14dp"/>
+
+                        <ImageView
+                            android:id="@+id/tall_fullscreen_button"
```
dbrant, line 195: spelling ("tall")

cooltey: Done!

app/src/main/java/org/wikipedia/talk/TalkTopicsProvider.kt
cooltey: Yes!

app/src/main/java/org/wikipedia/page/SidePanelHandler.kt
cooltey: Makes sense and done.

app/src/main/java/org/wikipedia/page/SidePanelHandler.kt
cooltey: Done!

PR #3141, opened 2022-02-24
Add password is same as username check in app (T211639)

svenvdvoort:
Hello!

This PR adds a check to the CreateAccount and ResetPassword activities that will warn the user if the new password is the same as the username ([T211639](https://phabricator.wikimedia.org/T211639)). This [constraint is enforced](https://meta.wikimedia.org/wiki/Password_policy) by the Wikipedia site and it's useful to give users direct feedback on these constraints.

Yours sincerely, Sven.

P.S. I'm part of group of 4 students who will spend 10 weeks investigating the software architecture of wikipedia-android as part of the [Software Architecture MSc course at TU Delft](https://se.ewi.tudelft.nl/delftswa/2022/). We'll be writing [several blog posts](https://desosa2022.netlify.app/projects/wikipedia-android/) about our findings and make contributions to the project. If possible, it would be great to get in contact with core developers, both at the Wikimedia Foundation and volunteers, so please reach out! :)


app/src/main/java/org/wikipedia/createaccount/CreateAccountActivity.kt
```diff
@@ -319,6 +324,8 @@ class CreateAccountActivity : BaseActivity() {
                 return ValidateResult.INVALID_USERNAME
             } else if (password.length < PASSWORD_MIN_LENGTH) {
                 return ValidateResult.PASSWORD_TOO_SHORT
+            } else if (password.toString() == username.toString()) {
```
dbrant, line 327: Let's make the comparison case-insensitive.

app/src/main/res/values/strings.xml
```diff
@@ -159,6 +159,7 @@
     <string name="dialog_create_account_checking_progress">Verifying</string>
     <string name="create_account_email_hint">Email (Optional)</string>
     <string name="create_account_password_repeat_hint">Repeat password</string>
+    <string name="create_account_password_is_username">Password can\'t be the same as username</string>
```
dbrant, line 162: When adding a string resource to `strings.xml`, you must also add the same resource to `values-qq\strings.xml`, with an explanation of the string.

This is part of our community translation architecture ([translatewiki.net](https://translatewiki.net)) that helps community members translate strings into all the languages we support.

app/src/test/java/org/wikipedia/createaccount/CreateAccountActivityTest.java
```diff
@@ -34,6 +34,11 @@
                 is(ValidateResult.PASSWORD_MISMATCH));
     }
 
+    @Test public void testValidateInputPasswordIsUsername() {
+        assertThat(validateInput("password", "password", "password", ""),
```
dbrant, line 38: Let's also add an assertion for case-insensitive comparison.

dbrant (changes requested):
Thanks! See comments inline.

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivity.kt
svenvdvoort: Yes that's a good idea!

app/src/main/res/values/strings.xml
svenvdvoort: Thanks for pointing that out, the string is added now

PR #3150, opened 2022-02-25
Design tweaks: Edit History list

sharvaniharan:
- Corrected ripple effect on the item
- Added different icons for anon and logged in users
- Updated placeholder text


app/src/main/res/layout/item_edit_history.xml
```diff
@@ -57,6 +59,7 @@
             android:paddingBottom="8dp"
             android:textAlignment="viewStart"
             android:textAllCaps="false"
+            app:rippleColor="@null"
```
cooltey, line 62: If you would like to make this "button" unclickable, does it mean it should be a `Textview`?

cooltey (changes requested):

app/src/main/res/layout/item_edit_history.xml
sharvaniharan: It is easier to style the rounded edges @cooltey but I could probably not use the style Flatbutton.Small and just have it inline. Do you feel that is better?

cooltey: Makes sense to me. Or, maybe we can use `rounded_12dp_corner_base90_fill` drawable to achieve this since making a `Button` not clickable also with some other extra setting to the view does not seem right to me. How do you think @dbrant?

sharvaniharan: Did this @cooltey

app/src/main/res/layout/item_edit_history.xml
```diff
@@ -83,9 +85,10 @@
             app:iconPadding="8dp"
             app:iconSize="16dp"
             app:iconTint="?attr/material_theme_secondary_color"
+            app:layout_constraintBaseline_toBaselineOf="@+id/diffText"
             app:layout_constraintStart_toEndOf="@id/diffText"
-            app:layout_constraintTop_toTopOf="@id/diffText"
             app:layout_constraintWidth_max="220dp"
+            app:rippleColor="@null"
```
cooltey, line 91: Looks like you missed this one. We can have `MaterialTextView` for this one.

dbrant: We'll discuss this during sync, but I kind of disagree that this should no longer be a button, because (a) it looks like a button, and (b) it would be very useful to tap a username and provide options to go to the user's Talk page or Profile page, as we do in other places.

PR #3152, opened 2022-02-25
Multi-selection of Edit History items for comparison.

dbrant:
https://phabricator.wikimedia.org/T299180


app/src/main/java/org/wikipedia/page/edithistory/EditHistoryItemView.kt
```diff
@@ -13,11 +18,30 @@ import org.wikipedia.util.ResourceUtil
...
+        }
+        binding.clickTargetView.setOnLongClickListener {
+            listener?.onLongClick()
+            true
+        }
+        binding.selectButton.setOnClickListener {
+            listener?.onToggleSelect()
+        }
+        binding.userNameText.setOnClickListener {
+            // TODO
```
cooltey, line 43: Should we show a menu here?

cooltey (changes requested):
The code looks good to me, but some places may be optimized:
1. When in "compare" mode, the back button should be able to cancel the "compare" mode.
2. Should we make the whole item togglable when in "compare" mode instead of leading the user to the detail view when clicking on the item? Feel like only making the radio button togglable is not enough.
3. According to the design, the first item should be yellow, and the second one should be blue.

app/src/main/java/org/wikipedia/page/edithistory/EditHistoryItemView.kt
dbrant: This will be handled in #3113

dbrant:
> 1. When in "compare" mode, the back button should be able to cancel the "compare" mode.

Done!

> 2. Should we make the whole item togglable when in "compare" mode instead of leading the user to the detail view when clicking on the item? Feel like only making the radio button togglable is not enough.
> 3. According to the design, the first item should be yellow, and the second one should be blue.

These two items will be discussed in Design review, since there are some inconsistencies in the designs.

PR #3154, opened 2022-02-27
Fix invalid Strings file plural entries in Slovak

Jeehut:
As per the docs, the `other` plural item is required but it was not provided a value here, which was invalid. This fixes it.
See also: https://developer.android.com/guide/topics/resources/string-resource#Plurals

### DIFF
--- a/app/src/main/res/values-sk/strings.xml
+++ b/app/src/main/res/values-sk/strings.xml
@@ -542,14 +542,12 @@
   <plurals name="reading_list_article_offline_message">
     <item quantity="one">Tento článok bude teraz dostupný bez pripojenia.</item>
     <item quantity="few">Tieto články budú teraz dostupné bez pripojenia.</item>
-    <item quantity="many">Tieto články budú teraz dostupné bez pripojenia.</item>
-    <item quantity="other"/>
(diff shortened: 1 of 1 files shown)


dbrant:
Thanks! Our string resources are automatically synced from our crowd-sourced translation site, translatewiki.net.
Here are the relevant strings for you to edit:

https://translatewiki.net/wiki/Wikimedia:Wikipedia-android-strings-reading_list_article_offline_message/sk
https://translatewiki.net/wiki/Wikimedia:Wikipedia-android-strings-reading_list_article_not_offline_message/sk

Jeehut:
I think the issue lies in that site allowing plurals without the "other" variant, which is disallowed. The "other" is always required as per the specs. Fixing these Strings doesn't prevent from the same issues repeating in the future again.

dbrant:
The "other" variant should go after the final pipe `|` symbol and before the closing `}}` symbol, which is currently empty.

Jeehut:
I do not have an account there and can't find an option to register. I just found this issue while working on a Android Strings file parser where I included several open source projects in my tests. I thought, I should report it in case you're interested in a fix. I'm not using this app myself, nor do I speak Slovak.

PR #3156, opened 2022-02-28
OTRS to VRT in README.md

jhsoby:


dbrant (changes requested):
Thanks for catching that. But actually we no longer use either OTRS or AppCenter. (and I don't know what "VRT" is.)
So, we can just shorten the whole thing to say "Optional logcat output."

jhsoby:
VRT is OTRS, it was changed for branding rights reasons I believe.

Added a second commit to remove references to OTRS and AppCenter – I don't know how to squash them, I think that's something you do when you merge, right?

dbrant:
Correct! Thanks again.

PR #3158, opened 2022-03-01
Hygiene: remove unused preference key and rename a key with a proper name

cooltey:
While doing this ticket: https://phabricator.wikimedia.org/T301426

Rename `preference_key_multilingual_search_tutorial_enabled` to `preference_key_multilingual_search_tooltip_shown` and also its related method names, since it is showing a tooltip.

### DIFF
--- a/app/src/main/res/values/preference_keys.xml
+++ b/app/src/main/res/values/preference_keys.xml
@@ -12,7 +12,7 @@
     <string name="preference_key_app_channel">channel</string>
     <string name="preference_key_language_mru">languageMru</string>
     <string name="preference_key_language_app">languageApp</string>
-    <string name="preference_key_multilingual_search_tutorial_enabled">multilingualSearchTutorialEnabled</string>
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/preference_keys.xml
```diff
@@ -12,7 +12,7 @@
     <string name="preference_key_app_channel">channel</string>
     <string name="preference_key_language_mru">languageMru</string>
     <string name="preference_key_language_app">languageApp</string>
-    <string name="preference_key_multilingual_search_tutorial_enabled">multilingualSearchTutorialEnabled</string>
+    <string name="preference_key_multilingual_search_tooltip_shown">multilingualSearchTooltipShown</string>
```
dbrant, line 15: If we change the actual value of the key, it means that the user will see the tooltip again, even if they've already seen it. Are we sure we want to do that?

dbrant (changes requested):

app/src/main/res/values/preference_keys.xml
cooltey: You are correct. Fixed.

PR #3160, opened 2022-03-03
Adjustments to EditAttempStep to collect more info

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T302943

- Wired-up the app to send page namespace and page title to EditAttemptStep schema to assist in analysis


app/src/main/java/org/wikipedia/analytics/eventplatform/EditAttemptStepEvent.kt
```diff
@@ -17,29 +18,30 @@ class EditAttemptStepEvent(private val event: EditAttemptStepInteractionEvent) :
...
+        fun logSaveFailure(pageTitle: PageTitle) {
+            submitEditAttemptEvent(ActionType.SAVE_FAILURE, pageTitle)
         }
 
-        private fun submitEditAttemptEvent(action: ActionType, wikiCode: String) {
+        private fun submitEditAttemptEvent(action: ActionType, pageTitle: PageTitle) {
             EventPlatformClient.submit(EditAttemptStepEvent(EditAttemptStepInteractionEvent(action.valueString, "", OTHER.valueString,
-                INTEGRATION_ID, "", PLATFORM, 0, if (AccountUtil.isLoggedIn) AccountUtil.getUserIdForLanguage(wikiCode) else 0, 1)))
+                INTEGRATION_ID, "", PLATFORM, 0, if (AccountUtil.isLoggedIn) AccountUtil.getUserIdForLanguage(pageTitle.wikiSite.languageCode) else 0,
+                1, pageTitle.displayText, pageTitle.namespace().code())))
```
cooltey, line 44: Not sure but should we show `pageTitle.prefixText`?

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageTagsFragment.kt
```diff
@@ -326,7 +326,7 @@ class SuggestedEditsImageTagsFragment : SuggestedEditsItemFragment(), CompoundBu
         publishSuccess = false
 
         funnel?.logSaveAttempt()
-        EditAttemptStepEvent.logSaveAttempt(callback().getLangCode())
+        EditAttemptStepEvent.logSaveAttempt(PageTitle(page!!.title, WikiSite(Service.COMMONS_URL)))
```
cooltey, line 329: Is it possible to have a `var pageTitle` for this one? so that we don't have to create `PageTitle` for the following log events.

cooltey (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/EditAttemptStepEvent.kt
dbrant: +1 to this. Using `prefixText` would be consistent with how other platforms use this schema.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageTagsFragment.kt
dbrant: +1 to this too. It can live alongside `page`, and updated whenever `page` is assigned.

dbrant (changes requested):

PR #3170, opened 2022-03-03
Add functionality to see statistics in Edit history

cooltey:
https://phabricator.wikimedia.org/T299181


app/src/main/java/org/wikipedia/page/edithistory/EditHistoryListActivity.kt
```diff
@@ -90,6 +90,11 @@ class EditHistoryListActivity : BaseActivity() {
                 loadHeader.loadState = it.refresh
                 loadFooter.loadState = it.append
             }
+            viewModel.editHistoryStatsFlow.collectLatest {
```
dbrant, line 93: As I understand it, you can't put more than one `collect...` function in a single `launchWhenCreated` call, because the `collect` function never actually returns.
Each `collect` function must be inside its own `launchWhenCreated` scope.

dbrant (changes requested):
@cooltey  Please see #3175, which is a PR to be merged into this one. It puts the EditStats view into a compound adapter into the RecyclerView, instead of a separate view inside a NestedScrollView.
Usage of NestedScrollView is causing the PagingSource to be called infinitely, and reduces performance. It's no longer recommended to be used.

app/src/main/java/org/wikipedia/page/edithistory/EditHistoryListActivity.kt
cooltey: Yes, you are correct! Thanks for the PR and tips 👍

cooltey:
Hi @dbrant @sharvaniharan, this is ready for review!

PR #3174, opened 2022-03-04
Add 'More' option and bottom sheet to article toolbar

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T302397


sharvaniharan:
Variant B was chosen

PR #3181, opened 2022-03-08
Enable filter and search function in edit history list

cooltey:
https://phabricator.wikimedia.org/T299233
https://phabricator.wikimedia.org/T299234


cooltey:
Looks like using the rest `page/history` endpoint also has some disadvantages, since it cannot filter "editors" edits, and we can't either do local filtering since the object does not contain a `isBot` value. I'll go to reuse the original API call.

app/src/main/java/org/wikipedia/page/edithistory/EditHistoryListViewModel.kt
```diff
@@ -123,8 +139,37 @@ class EditHistoryListViewModel(bundle: Bundle) : ViewModel() {
...
+
+                val revision = response.query!!.pages?.first()?.revisions!!
+
+                // The `ususers` parameter has a size limit of 50 usernames.
+                val userNamesList = revision.asSequence().filterNot { it.isAnon }.map { it.user }.distinct().chunked(50).toList()
+
+                // Combining two async API responses, finding out all bot usernames, and joining them into one list.
+                val botUserList = withContext(Dispatchers.IO) {
+                    val result = userNamesList.map {
+                        async { ServiceFactory.get(pageTitle.wikiSite).getUserInfoList(it.joinToString("|")) }
```
dbrant, line 152: This is a very costly and unnecessary API call. I believe we discussed that we will not be supporting filtering by Bot edits? This is not only because the API does not support it, but also because no one actually needs it.

cooltey:
@dbrant Please check this way of caching data so that we don't need to filter the entire list at the top of the page (after changing filter type).
https://github.com/wikimedia/apps-android-wikipedia/pull/3181/commits/bbcfbe23a71e9c57f9080b5ca2c0f556d1700c93

dbrant:
> @dbrant Please check this way of caching data so that we don't need to filter the entire list at the top of the page (after changing filter type)

Yes! That's it.

cooltey:
@dbrant @sharvaniharan This is ready for code review.

PR #3183, opened 2022-03-09
Fix: update displayText value when reset the namespace

cooltey:
When going to set the `namesapce`, we will also reset the `_displayText` to `null`, it will cause a language variant issue.

**Steps to reproduce:**

1. Go to "World War II" in the English wiki
2. Change the language to the Traditional Chinese wiki
3. Go to its talk page

The title should be "Talk: 第二次世界大戰" instead of "Talk:第二次世界大战"

This PR also fixes incorrect usage of calling `ArticleEditDetailsActivity.newIntent()`, which we should always use `pageTitle.prefixedText` for calling APIs.

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageTitle.kt
+++ b/app/src/main/java/org/wikipedia/page/PageTitle.kt
@@ -51,7 +51,7 @@ data class PageTitle(
         get() = _namespace.orEmpty()
         set(value) {
             _namespace = value
-            _displayText = if (value.isEmpty()) _displayText else StringUtil.addUnderscores(value) + ":" + _displayText
+            _displayText = if (value.isEmpty()) _displayText else StringUtil.removeUnderscores(value) + ":" + _displayText
         }
 
     val isFilePage: Boolean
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/wikipedia/page/PageTitle.kt
```diff
@@ -49,7 +49,10 @@ data class PageTitle(
 
     var namespace: String
         get() = _namespace.orEmpty()
-        set(value) { _namespace = value; _displayText = null }
+        set(value) {
+            _namespace = value
+            _displayText = if (value.isEmpty()) _displayText else StringUtil.addUnderscores(value) + ":" + _displayText
```
dbrant, line 54: The `_displayText` should not contain underscores, since it's for display purposes. Should this be `removeUnderscores()` instead?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageTitle.kt
cooltey: Correct! Done!

PR #3188, opened 2022-03-10
Follow-up to Statistics in Edit History.

dbrant:
* Consolidate a couple of API calls.
* When scrolling, increase the activity title text to 2 lines.
* Remove basically duplicate string resource.

### DIFF
--- a/app/src/main/res/layout/activity_edit_history.xml
+++ b/app/src/main/res/layout/activity_edit_history.xml
@@ -10,10 +10,11 @@
     <androidx.appcompat.widget.Toolbar
         android:id="@+id/toolbar"
         android:layout_width="match_parent"
-        android:layout_height="?attr/actionBarSize"
+        android:layout_height="wrap_content"
+        android:minHeight="?attr/actionBarSize"
         style="@style/ToolbarStyle">
 
-        <LinearLayout
+        <androidx.constraintlayout.widget.ConstraintLayout
             android:layout_width="match_parent"
             android:layout_height="match_parent"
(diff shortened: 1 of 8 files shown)


app/src/main/res/layout/activity_edit_history.xml
```diff
@@ -22,11 +23,16 @@
...
+                app:layout_constraintBottom_toBottomOf="parent"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintEnd_toStartOf="@id/compareButton"
                 android:layout_marginEnd="8dp"
+                android:layout_marginTop="8dp"
+                android:layout_marginBottom="8dp"
                 android:lineSpacingExtra="4sp"
                 android:fontFamily="sans-serif"
-                android:maxLines="1"
+                android:maxLines="2"
```
cooltey, line 35: Per Robin's comment, he preferred to set the max line to 1 and show the ellipsis if no more spaces.

cooltey (changes requested):

app/src/main/res/layout/activity_edit_history.xml
dbrant: ok, I've changed it back to 1 line for this PR, but I'd be interested to confirm with Robin if this is really desired.
With only 1 line, the title will always be taken up by the "Revision history:" portion, with maybe three characters of the actual title.
image

cooltey: Thanks! Here's what it used to be but Robin thought making it a single line is better 😅

PR #3190, opened 2022-03-10
Fix the namespace sent to TalkFunnel

sharvaniharan:
Currently we are sending the namepace of the page, and not the talkPage of the original page. This works fine when we directly open a user talk page, but when we open an article of any other namespace, and try to access its talk, then we end up sending the wrong namespace number to analytics. 
**Phab:** https://phabricator.wikimedia.org/T296310#7768256


sharvaniharan:
@dbrant @cooltey Should we do it this way, or initialize the funnel in the beginning itself, but just increment the number by 1 for its respective talk namespace?

cooltey (commented):
Not sure but I would put the `funnel` to `nullable` and add `?` null-safe to prevent possible crashes when calling methods.

sharvaniharan:
> Not sure but I would put the `funnel` to `nullable` and add `?` null-safe to prevent possible crashes when calling methods.

Done @cooltey

PR #3198, opened 2022-03-14
Convert LangLinksActivity to use ViewModel, and improve efficiency.

dbrant:
Prior to this PR, the LangLinksActivity was doing these not-so-good things:
* Making network requests every time the activity is started, including on screen rotations.
* Making a network request for each and every language variant title, which could add up to a lot. For the Barack Obama article, this was making **34** requests total.

Switching to a ViewModel (and requesting variant titles lazily) solves all of this.


app/src/main/java/org/wikipedia/language/LangLinksActivity.kt
```diff
@@ -242,23 +169,27 @@ class LangLinksActivity : BaseActivity() {
...
+                LangLinksItemViewHolder(view)
             }
         }
 
         override fun onBindViewHolder(holder: DefaultViewHolder, pos: Int) {
-            holder.bindItem(pos)
+            if (variantTitlesToUpdate.contains(languageEntries[pos])) {
+                viewModel.fetchLangVariantLink(languageEntries[pos])
+                variantTitlesToUpdate.remove(languageEntries[pos])
+            }
```
cooltey, line 191: Looks like there's a scrolling issue when language variants are added, not sure they are related.

Steps to reproduce:
1. Search "Island groups of the Philippines" 
2. Change language to "Traditional Chinese" (Please make sure there's only one `zh` app language)
3. Click on "Language" again and scroll down to the bottom of the list.
   
https://www.youtube.com/watch?v=IGJ_EzhZRVE

cooltey (changes requested):

app/src/main/java/org/wikipedia/language/LangLinksActivity.kt
dbrant: It doesn't seem like a very significant issue (and I can't reproduce it on my device). It's probably because I'm using `notifyItemRangeChanged()`, which is more efficient but produces an animation, instead of `notifyDataSetChanged()`. If it's not too distracting, and if all the text is correct, can we keep it this way?

cooltey: Got it. If it is possible, can we apply `binding.langlinksRecycler.itemAnimator = null` to the `RecyclerView` so that people will not see the variant titles pop up?

dbrant: yep!

PR #3199, opened 2022-03-14
Analytics wiring for talk bubble

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T303772

What to measure:
Measure clicks on the bubble in MobileWikiAppTalk as "action" = "bubble_click" and "source" will be "article_bubble"
Add another new source as "article_side_panel" and log it as source when Talk page is opened from bubble->side panel -> talk page and "action" as always will be "open_talk"
This will equip us to measure the number of users who clicked the bubble, and of them, how many actually opened the full view talk page.

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageFragment.kt
+++ b/app/src/main/java/org/wikipedia/page/PageFragment.kt
@@ -18,7 +18,6 @@ import androidx.core.animation.doOnEnd
 import androidx.core.app.ActivityCompat
 import androidx.core.app.ActivityOptionsCompat
 import androidx.core.graphics.Insets
-import androidx.core.view.forEach
 import androidx.fragment.app.Fragment
 import androidx.swiperefreshlayout.widget.SwipeRefreshLayout.OnRefreshListener
 import com.google.android.material.bottomsheet.BottomSheetDialogFragment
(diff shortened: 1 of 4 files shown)


sharvaniharan:
@cooltey @dbrant this can be reviewed and merged to Cooltey's PR and there is a hold on that, just to keep it moving.

dbrant (changes requested):
@sharvaniharan  I don't think I'm seeing where this hooks up the funnel to the side panel, i.e. when the user taps an item in the side list.

sharvaniharan:
Done @dbrant . Had missed the item click.

app/src/main/java/org/wikipedia/page/PageFragment.kt
```diff
@@ -807,14 +809,17 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
...
+                        sidePanelHandler.showTalkTopics()
+                    }
                 }
             }
         }
         bridge.addListener("footer_item") { _, messagePayload ->
             messagePayload?.let { payload ->
                 when (payload["itemType"]?.jsonPrimitive?.content) {
-                    "talkPage" -> model.title?.run { startTalkTopicActivity(this) }
+                    "talkPage" -> model.title?.run { startTalkTopicActivity(this, PAGE_SIDE_PANEL) }
```
cooltey, line 822: This is actually from the footer item on the article page instead of the side panel. Would you like to confirm with Robin or Jazmin about this? (e.g. does the footer "Talk page" button open the side panel or go to the talk page directly?).

cooltey:
Please fix the lint error.

PR #3200, opened 2022-03-15
Use rest.php page/history API for the EditHistory page

cooltey:
In order to support proper list filtering in #3181, it would be great if we switch to using the `rest.php/page/{title}/history`.

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/CoreRestService.kt
+++ b/app/src/main/java/org/wikipedia/dataclient/CoreRestService.kt
@@ -2,8 +2,10 @@ package org.wikipedia.dataclient
 
 import org.wikipedia.dataclient.restbase.DiffResponse
 import org.wikipedia.dataclient.restbase.EditCount
+import org.wikipedia.dataclient.restbase.PageHistory
 import retrofit2.http.GET
 import retrofit2.http.Path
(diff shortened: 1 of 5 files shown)


dbrant:
Good find! It even looks like this will let us filter by Bot edits 😲 
However, some things to keep in mind about this API:
* It returns only 20 revisions per call, while the regular API can return up to 500. This will make it much less useful for searching: typing different search terms will cause the PagingSource to call this API again and again, because not enough items are loaded at one time.
* It does not return a `parsedcomment` field (a parsed HTML version of the `comment` field), which we might need in the future if we want to show a properly parsed comment in the Edit History list.
* When filtering by Anon/Bot/etc, only one filter can be specified at a time. This is fine, but the design will need to be adjusted a bit.

@cooltey Can we merge this PR onto your Filtering PR instead, so that we can get comfortable with how this API feels, and then build the Searching PR on top of it, to confirm whether it will work for us?

cooltey:
> @cooltey Can we merge this PR onto your Filtering PR instead, so that we can get comfortable with how this API feels, and then build the Searching PR on top of it, to confirm whether it will work for us?

Thanks for the comments @dbrant and yes, we can merge this PR to the filtering PR to see how it works!

PR #3205, opened 2022-03-16
Categories.

dbrant:
...using all the latest stuff: PagingSource, ViewModel, etc!

https://phabricator.wikimedia.org/T73966


dbrant:
@sharvaniharan @cooltey This is now ready for review.

app/src/main/java/org/wikipedia/categories/CategoryActivityViewModel.kt
```diff
@@ -0,0 +1,55 @@
...
+    ) : PagingSource<String, PageTitle>() {
+        override suspend fun load(params: LoadParams<String>): LoadResult<String, PageTitle> {
+            return try {
+                val response = ServiceFactory.get(WikiSite.forLanguageCode(pageTitle.wikiSite.languageCode))
+                        .getCategoryMembers(pageTitle.prefixedText, resultType, params.loadSize, params.key)
+                val titles = response.query!!.pages!!.map { page ->
+                    PageTitle(page.title, pageTitle.wikiSite).also {
+                        it.description = page.description.orEmpty()
+                        it.thumbUrl = page.thumbUrl()
+                    }
```
cooltey, line 36: Looks like it will cause a language variant issue here if we don't set the correct display title.

Would it be possible to call `page/summary` to get the correct `displaytitle`?

app/src/main/java/org/wikipedia/categories/CategoryDialogViewModel.kt
```diff
@@ -0,0 +1,43 @@
...
+        fetchCategories()
+    }
+
+    private fun fetchCategories() {
+        viewModelScope.launch(CoroutineExceptionHandler { _, throwable ->
+            categoriesData.postValue(Resource.Error(throwable))
+        }) {
+            withContext(Dispatchers.IO) {
+                val response = ServiceFactory.get(pageTitle.wikiSite).getCategories(pageTitle.prefixedText)
+                categoriesData.postValue(Resource.Success(response.query!!.firstPage()!!.categories!!.filter { !it.hidden }))
```
cooltey, line 32: Would it be possible to also call the `page/summary` to get the correct `displaytitle`? Looks like the API response only contains the original title of the page.

cooltey (changes requested):
Found the following issues:
1. The bottom sheet will pop up (with the original article category list) after navigating to another article from the category list.
2. The bottom of the list in the category dialog has the default purple ripple color, can we get rid of it?

app/src/main/java/org/wikipedia/categories/CategoryDialog.kt
```diff
@@ -0,0 +1,135 @@
...
+            binding.categoriesNoneFound.visibility = View.VISIBLE
+            binding.categoriesRecycler.visibility = View.GONE
+        }
+        binding.categoriesRecycler.visibility = View.VISIBLE
+        binding.categoriesNoneFound.visibility = View.GONE
+        binding.categoriesError.visibility = View.GONE
+        binding.categoriesRecycler.adapter = CategoryAdapter(categoryList)
+    }
+
+    private inner class CategoryItemHolder constructor(itemView: PageItemView<PageTitle>) : RecyclerView.ViewHolder(itemView) {
```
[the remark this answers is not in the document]
dbrant: yep

app/src/main/java/org/wikipedia/categories/CategoryActivityViewModel.kt
dbrant: I have updated API call to also return variant titles, similarly to how we do it in Search.

app/src/main/java/org/wikipedia/categories/CategoryDialogViewModel.kt
dbrant: Found a better API!

dbrant:
> 1. The bottom sheet will pop up (with the original article category list) after navigating to another article from the category list.
> 2. The bottom of the list in the category dialog has the default purple ripple color, can we get rid of it?

Done!

cooltey (changes requested):
Found the following issue:
https://youtu.be/lBE1PADPYak


Steps to reproduce:
1. Go to "Computer Space" in `en` wiki
2. Change language to `zh-hant` wiki
3. Go to the first item "街機遊戲" of the categories.
4. In "街機遊戲", switch to subcategories.
5. Click on the first item "分類:電動機械", and you will see the error message.

By checking the category, the [分類:電動機械](https://zh.wikipedia.org/wiki/Category:%E9%9B%BB%E5%8B%95%E6%A9%9F%E6%A2%B0) only has one subcategory and empty article.

And, the "Back" button is not responding, should we add a listener to the `ErrorView` back button?

dbrant:
@cooltey Done! Excellent observations -- I've added correct empty-state handling.

cooltey (commented):
Looks good to me now!

If we selected the subcategories tab and change the device's orientation, the tab will be changed to the articles tab. Should we fix that?

dbrant:
> Looks good to me now!
> 
> If we selected the subcategories tab and change the device's orientation, the tab will be changed to the articles tab. Should we fix that?

Done!!

PR #3227, opened 2022-03-28
Further optimize API calls to Commons and Wikidata.

dbrant:
* Further reduce usage of the `wbgetentities` call.
* Combine unnecessary combinations of `wbgetentities` and `wbgetclaims`.


app/src/main/java/org/wikipedia/commons/FilePageFragment.kt
```diff
@@ -94,15 +93,11 @@ class FilePageFragment : Fragment() {
         binding.filePageView.visibility = View.GONE
         binding.progressBar.visibility = View.VISIBLE
 
-        disposables.add(Observable.zip(getImageCaptions(pageTitle.prefixedText),
-                ServiceFactory.get(WikiSite(Service.COMMONS_URL)).getImageInfo(pageTitle.prefixedText,
-                    pageTitle.wikiSite.languageCode), { caption, response ->
-                    // set image caption to pageTitle description
-                    pageTitle.description = caption[pageTitle.wikiSite.languageCode]
-                    response
-                })
```
cooltey, line 103: Looks like this API call will get the correct image caption from the file, which should be as the same as the one we saw in the gallery.

Steps to reproduce:
1. Go to [Ruby Lin](https://en.wikipedia.org/wiki/Ruby_Lin) in English Wikipedia
2. Change language to Traditional Chinese
3. In the first image of the Gallery, the image caption is in Traditional Chinese - "林心如 2008 年於北京出席記者招待會"
4. Go to the image page
5. Check the image description section, it should show the one in `3.` instead of the English one.

cooltey (changes requested):

app/src/main/java/org/wikipedia/commons/FilePageFragment.kt
dbrant: Correct! And... done.

PR #3229, opened 2022-03-28
Article more menu variant B

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T306709

This is the design version 'B' for user testing of more menu options on article screen.


app/src/main/java/org/wikipedia/page/PageActivity.kt
```diff
@@ -78,6 +78,7 @@ class PageActivity : BaseActivity(), PageFragment.Callback, LinkPreviewDialog.Ca
     private val listDialogDismissListener = DialogInterface.OnDismissListener { pageFragment.updateBookmarkAndMenuOptionsFromDao() }
     private val isCabOpen get() = currentActionModes.isNotEmpty()
     private var exclusiveTooltipRunnable: Runnable? = null
+    private var themeTooltipBalloon: Balloon? = null
```
cooltey, line 81: Would it be possible to change the name to `customizeToolbarTooltip`?

app/src/main/java/org/wikipedia/page/PageActivity.kt
```diff
@@ -664,23 +666,20 @@ class PageActivity : BaseActivity(), PageFragment.Callback, LinkPreviewDialog.Ca
...
+                        FeedbackUtil.Callback {
+                        override fun onDismissClicked() {
+                            Prefs.showOneTimeCustomizeToolbarTooltip = false
+                            Prefs.toolbarTooltipVisible = false
+                        }
+                    })
+                themeTooltipBalloon?.showAlignBottom(binding.pageToolbarButtonShowOverflowMenu)
+                Prefs.toolbarTooltipVisible = true
+            }
+        }, 2000)
```
cooltey, line 682: Not sure if the delay is really necessary if we are going to point the tooltip to the overflow menu.

app/src/main/java/org/wikipedia/page/PageActivity.kt
```diff
@@ -664,23 +666,20 @@ class PageActivity : BaseActivity(), PageFragment.Callback, LinkPreviewDialog.Ca
...
-        Prefs.showOneTimeCustomizeToolbarTooltip = false
+        binding.pageToolbarButtonShowOverflowMenu.postDelayed({
+            if (!isDestroyed) {
+                themeTooltipBalloon =
+                    FeedbackUtil.getTooltip(this, getString(R.string.theme_chooser_menu_item_short_tooltip), arrowAnchorPadding = -DimenUtil.roundedDpToPx(12f), topOrBottomMargin = -12, aboveOrBelow = true, autoDismiss = false, showDismissButton = true, callback = object :
+                        FeedbackUtil.Callback {
+                        override fun onDismissClicked() {
+                            Prefs.showOneTimeCustomizeToolbarTooltip = false
+                            Prefs.toolbarTooltipVisible = false
+                        }
```
cooltey, line 677: Would it be possible to just use `themeTooltipBalloon.setOnBalloonDismissListener { }`  instead of creating a new interface?

cooltey (changes requested):
Looks like the arrow is a little bit over the edge of the tooltip.

app/src/main/java/org/wikipedia/page/PageActivity.kt
```diff
@@ -666,20 +666,16 @@ class PageActivity : BaseActivity(), PageFragment.Callback, LinkPreviewDialog.Ca
...
-                    FeedbackUtil.getTooltip(this, getString(R.string.theme_chooser_menu_item_short_tooltip), arrowAnchorPadding = -DimenUtil.roundedDpToPx(12f), topOrBottomMargin = -12, aboveOrBelow = true, autoDismiss = false, showDismissButton = true, callback = object :
-                        FeedbackUtil.Callback {
-                        override fun onDismissClicked() {
-                            Prefs.showOneTimeCustomizeToolbarTooltip = false
-                            Prefs.toolbarTooltipVisible = false
-                        }
-                    })
-                themeTooltipBalloon?.showAlignBottom(binding.pageToolbarButtonShowOverflowMenu)
-                Prefs.toolbarTooltipVisible = true
+        if (!isDestroyed) {
```
cooltey, line 669: Since the logic is not inside the `postDelayed`, can we remove the logic of checking `isDestroyed`?

cooltey (changes requested):

PR #3234, opened 2022-03-29
Article clicks and Edit history analytics wiring

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T304335
**Phab:** https://phabricator.wikimedia.org/T304336


sharvaniharan:
@cooltey @dbrant ready for review!

app/src/main/java/org/wikipedia/analytics/eventplatform/ArticleInteractionEvent.kt
```diff
@@ -0,0 +1,150 @@
...
+
+    fun logCategoriesClick() {
+        action = CATEGORIES.valueString
+        submitEvent()
+    }
+
+    private fun submitEvent() {
+        timeSpentMs = duration.toInt()
+        isAnon = !AccountUtil.isLoggedIn
+        EventPlatformClient.submit(this)
```
dbrant, line 121: This doesn't seem to follow the principles of OOP.  With every call of this `submitEvent()` function, you re-submit the same object (`this`) to be sent. But all the different `log...` functions _modify_ the `this` object. This means that if multiple references to `this` get queued up to be sent, they'll all contain the same information, even if they were trying to log different things.

Every item submitted into `EventPlatformClient` must be a unique instance of an event. For an example, see `UserContributionEvent`.

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/ArticleInteractionEvent.kt
sharvaniharan: @dbrant We had the exact opposite of this conversation in a pair programming session for this PR https://github.com/wikimedia/apps-android-wikipedia/pull/3040/commits and as you can see, in the first commit i had sent unique instances of the event, but then we decided that since we will never queue them up, we will just manipulate the object and keep re-sending.. so which way should we do this? Shall we decide once for all, that some day we will queue events and then have all MEP classes send unique instances? I have even made a ticket to change them all to send `this` after that conversation we had  https://phabricator.wikimedia.org/T299240

sharvaniharan: So the series of decisions was  - we were not finding parity between legacy and MEP events - we made many other code tightening changes, but it remained the same - Then we decided to try the not queuing up solution to make it more of apples-to-apples  comparison, which in turn fixed how we were seeing the events. So the conclusion was that there must be something not fully flushed out at the MEP end about handling batched events [ speculation] as we were doing what was expected at our end. We should definitely discuss the path forward with batching. This makes sense #3240 , so will merge it and make similar changes for this schema as well.

sharvaniharan:
@dbrant ready for review.

app/src/main/java/org/wikipedia/analytics/eventplatform/ArticleInteractionEvent.kt
```diff
@@ -0,0 +1,154 @@
+package org.wikipedia.analytics.eventplatform
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import org.wikipedia.analytics.eventplatform.ArticleInteractionEvent.ActionType.*
+import org.wikipedia.auth.AccountUtil
+
+class ArticleInteractionEvent(private var wikiDb: String, private var pageId: Int) : TimedEvent() {
+
+    private lateinit var action: String
```
dbrant, line 10: We no longer need this field here. We can just make `submitEvent(...)` take an action as a parameter.

And does the action really need to be an Enum? (Can it just be literal string values? Are they used anywhere else?)

app/src/main/java/org/wikipedia/analytics/eventplatform/EditHistoryInteractionEvent.kt
```diff
@@ -0,0 +1,70 @@
+package org.wikipedia.analytics.eventplatform
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import org.wikipedia.analytics.eventplatform.EditHistoryInteractionEvent.ActionType.*
+import org.wikipedia.auth.AccountUtil
+
+class EditHistoryInteractionEvent(private var wikiDb: String, private var pageId: Int) : TimedEvent() {
+
+    private lateinit var action: String
```
dbrant, line 10: Same

app/src/main/java/org/wikipedia/page/PageFragment.kt
```diff
@@ -127,6 +128,7 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
     private lateinit var leadImagesHandler: LeadImagesHandler
     private lateinit var pageFragmentLoadState: PageFragmentLoadState
     private lateinit var bottomBarHideHandler: ViewHideHandler
+    lateinit var articleInteractionEvent: ArticleInteractionEvent
```
dbrant, line 131: Since this is a `TimedEvent`, do we need to `pause()` and `resume()` it when the PageFragment is paused and resumed?

app/src/main/java/org/wikipedia/page/PageActivity.kt
```diff
@@ -110,23 +110,27 @@ class PageActivity : BaseActivity(), PageFragment.Callback, LinkPreviewDialog.Ca
         clearActionBarTitle()
         supportActionBar?.setDisplayHomeAsUpEnabled(true)
         binding.pageToolbarButtonSearch.setOnClickListener {
+            pageFragment.articleInteractionEvent.logSearchWikipediaClick()
```
dbrant, line 113: All of these interactions will crash if they happen before the article is loaded, since `articleInteractionEvent` will not yet be initialized.  Instead of making it `lateinit`, let's make it nullable, and change all these calls to `articleInteractionEvent?.log...`

dbrant (changes requested):

PR #3235, opened 2022-03-30
Even further optimize edit history filter and search.

dbrant:
@cooltey  I think this might be an even better solution:

* Change our local cache back to a simple list, instead of a map of lists.
* The real performance issue was actually in the Adapter. Nothing in the ViewModel is very expensive. Even if we keep 10000 items in our cache, and perform the filtering on all those items, that's not expensive. The heavy part actually happens in the Adapter on the Activity side. When we call `refresh()` on the adapter, it actually keeps all the current items in it! And when it receives an updated "page" of items, it checks whether each item is "the same" as every other item currently in the adapter, which becomes a O(n²) operation.


cooltey:
It is so great! Excellent job 👍

PR #2627, opened 2021-07-13
Kotlin Conversion: PageSummary

sharvaniharan:

PR #3236, opened 2022-03-30
Fix deserialization of pageId.

dbrant:
@sharvaniharan When you converted the PageSummary class to Kotlin in #2627, the `pageId` field was not converted correctly.
It's gone undetected for so long because we haven't actually needed it until now.


cooltey:
Looks good to me. Please review and merge it @sharvaniharan

sharvaniharan:
Thanks @dbrant

PR #3240, opened 2022-03-31
Improve OOP logic for analytics events containing state information.

dbrant:
An example of how we can have container "`Event`" objects that contain state, and yet send unique immutable `Event` objects to the EventPlatformClient.


app/src/main/java/org/wikipedia/analytics/eventplatform/CustomizeToolbarEvent.kt
```diff
@@ -7,29 +7,29 @@ import org.wikipedia.auth.AccountUtil
...
-@Serializable
-@SerialName("/analytics/mobile_apps/android_customize_toolbar_interaction/1.0.0")
-class CustomizeToolbarEvent : TimedEvent(STREAM_NAME) {
-    @SerialName("is_anon") private var isAnon: Boolean? = null
-    @SerialName("is_rfm_enabled") private var isRfmEnabled: Boolean? = null
-    @SerialName("favorites_order") private var favoritesOrder: List<Int> = emptyList()
-    @SerialName("menu_order") private var menuOrder: List<Int> = emptyList()
-    @SerialName("time_spent_ms") private var timeSpentMs: Int = 0
-    private var source: String = ""
+class CustomizeToolbarEvent : TimedEvent() {
```
dbrant, line 10: Note that this class itself (as well as `TimedEvent`) no longer inherits from `Event`, but is now a container for an implementation model class (`...Impl`) that will be instantiated immutably whenever an event needs to be sent.

cooltey (approved):

PR #3241, opened 2022-04-01
Fix: update talk page pageTitle after changing the language

cooltey:
Make sure to create a new `TalkTopicsProvider` after changing the language.


app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
```diff
@@ -145,6 +145,8 @@ class TalkTopicsActivity : BaseActivity() {
 
                     pageTitle = PageTitle(newNamespace, StringUtil.removeNamespace(pageTitle.prefixedText),
                             WikiSite.forLanguageCode(WikipediaApp.getInstance().language().appLanguageCodes[pos]))
+
+                    talkTopicsProvider = TalkTopicsProvider(pageTitle)
```
dbrant, line 149: Should we `cancel()` the previous provider before creating a new one?

dbrant (changes requested):

app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
cooltey: Yes we should. Done.

PR #3246, opened 2022-04-04
Add snackbar informing user of redirect

svenvdvoort:
Hello!

This PR will add functionality to display a short-lived snackbar when the user is redirected on clicking a link in the app or when opening the app by clicking a link. An example of redirecting can be seen when navigating to [Transportation](https://en.m.wikipedia.org/wiki/Transportation) which redirects to [Transport](https://en.m.wikipedia.org/wiki/Transport).
This PR solves issue [T153583](https://phabricator.wikimedia.org/T153583).

Kind regards,
Sven.


app/src/main/java/org/wikipedia/page/PageTitle.kt
```diff
@@ -32,7 +32,8 @@ data class PageTitle(
     var description: String? = null,
     // TODO: remove after the restbase endpoint supports ZH variants.
     @SerialName("displayText") private var _displayText: String? = null,
-    var extract: String? = null
+    var extract: String? = null,
+    var redirectedFrom: String? = null,
```
dbrant, line 36: Adding a `redirectedFrom` field here is not a clean way of passing around this information. It should be easy enough to pass the redirected name directly from `PageFragmentLoadState` into `onPageMetadataLoaded()`.
(And don't worry about preserving the `redirectedFrom` name from the link preview dialog.)

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageTitle.kt
svenvdvoort: Thanks for the feedback! I have moved the `redirectedFrom` parameter to the `onPageMetadataLoaded(...)` method. Now the app only shows the redirected snackbar when clicking a Wikipedia link in an external application.

Wikigoagec: > Adding a `redirectedFrom` field here is not a clean way of passing around this information. It should be easy enough to pass the redirected name directly from `PageFragmentLoadState` into `onPageMetadataLoaded()`. (And don't worry about preserving the `redirectedFrom` name from the link preview dialog.)

PR #3258, opened 2022-04-07
Update ktlint version to 0.45.2

cooltey:
https://github.com/pinterest/ktlint/releases


cooltey:
Close due to the new Kotlin format syntax rules, which will need lots of time to fix them.

PR #3260, opened 2022-04-11
Use drawToBitmap() extension.

Isira-Seneviratne:


app/src/main/java/org/wikipedia/page/tabs/TabActivity.kt
```diff
@@ -321,40 +321,13 @@ class TabActivity : BaseActivity() {
...
-                    bmp = cacheBmp.scale(width, height)
-                }
-                if (!wasCacheEnabled) {
-                    view.isDrawingCacheEnabled = false
-                }
-            } catch (e: OutOfMemoryError) {
-                // don't worry about it
-            }
-            FIRST_TAB_BITMAP = bmp
+            FIRST_TAB_BITMAP = view.drawToBitmap()
```
cooltey, line 330: Since we'd like to make it low quality, can we set the `Bitmap.Config.RGB_565`? cc @dbrant

cooltey (changes requested):

app/src/main/java/org/wikipedia/page/tabs/TabActivity.kt
```diff
@@ -321,40 +321,13 @@ class TabActivity : BaseActivity() {
...
-                    bmp = cacheBmp.scale(width, height)
-                }
-                if (!wasCacheEnabled) {
-                    view.isDrawingCacheEnabled = false
-                }
-            } catch (e: OutOfMemoryError) {
-                // don't worry about it
-            }
-            FIRST_TAB_BITMAP = bmp
+            FIRST_TAB_BITMAP = view.drawToBitmap(Bitmap.Config.RGB_565)
```
cooltey, line 330: Can you please also add try and catch to prevent crashes from `OutOfMemoryError`?  Thanks!

cooltey (changes requested):

PR #3268, opened 2022-04-16
Use asReversed() extension.

Isira-Seneviratne:


dbrant:
No clear benefit described, therefore closing.

PR #3270, opened 2022-04-16
Use ConcurrentHashMap.

Isira-Seneviratne:
It is recommended to use `ConcurrentHashMap` over `Hashtable`, and it removes the need to manually make the methods in `DateUtil` synchronized.

Note: Core library desugaring is enabled to fix a bug in `ConcurrentHashMap` on Android 5.0 where updates can be lost on thread contention.


cooltey:
Hi @Isira-Seneviratne, thanks for the patch. Would you be able to share a document that indicates the recommendation of using `ConcurrentHashMap` over `HashMap`? And also I would like to know what would be the performance of using `ConcurrentHashMap` vs`HashMap`

Isira-Seneviratne:
> Hi @Isira-Seneviratne, thanks for the patch. Would you be able to share a document that indicates the recommendation of using `ConcurrentHashMap` over `HashMap`? And also I would like to know what would be the performance of using `ConcurrentHashMap` vs`HashMap`

Hi @cooltey, [this guide](https://www.baeldung.com/java-concurrent-map#4-performance) has a decent overview of the advantages of `ConcurrentHashMap` over `HashMap` in multi-threaded environments.

app/build.gradle
```diff
@@ -180,6 +182,8 @@ dependencies {
     String espressoVersion = '3.4.0'
     String serialization_version = '1.4.0'
 
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.2.2'
```
cooltey, line 185: I noticed a newer version of this library, would you please be able to update it?

Isira-Seneviratne: That requires AGP 7.4.0.

PR #3272, opened 2022-04-17
Use ArraySet.

Isira-Seneviratne:


dbrant:
No clear benefit described, therefore closing.

Isira-Seneviratne:
Sorry about that, I added a description.

PR #3274, opened 2022-04-18
Use ArraySet.

Isira-Seneviratne:
`ArraySet` is a more memory-efficient implementation of `Set` than `HashSet`, and it supports index-based value retrieval.


cooltey (approved):

app/build.gradle
```diff
@@ -196,6 +196,7 @@ dependencies {
 
     implementation "com.google.android.material:material:1.5.0"
     implementation 'androidx.appcompat:appcompat:1.4.1'
+    implementation "androidx.collection:collection-ktx:1.2.0"
```
dbrant, line 199: It doesn't seem like we need to declare this dependency explicitly here.

app/src/main/java/org/wikipedia/page/PageActivity.kt
```diff
@@ -391,9 +393,9 @@ class PageActivity : BaseActivity(), PageFragment.Callback, LinkPreviewDialog.Ca
     }
 
     override fun onPageCloseActionMode() {
-        val actionModesToFinish = HashSet(currentActionModes)
-        for (mode in actionModesToFinish) {
-            mode.finish()
+        val actionModesToFinish = ArraySet(currentActionModes)
+        for (i in actionModesToFinish.indices) {
```
dbrant, line 397: Why not keep the previous (cleaner) syntax of `for (mode in actionModesToFinish)`?

dbrant (changes requested):

app/build.gradle
Isira-Seneviratne: In the older version, the return type of `valueAt` was marked as nullable.

app/src/main/java/org/wikipedia/page/PageActivity.kt
Isira-Seneviratne: The documentation [recommends](https://developer.android.com/reference/androidx/collection/ArraySet#iterator()) against using the iterator.

dbrant:
Same comments/findings as #3290.

PR #3280, opened 2022-04-25
Wiring for analytics schemas for Mar 4 -20 cycle

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T306406


app/src/main/java/org/wikipedia/page/PageFragment.kt
```diff
@@ -1050,7 +1051,8 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
                     return@evaluate
                 }
                 val funnel = FindInPageFunnel(app, title.wikiSite, model.page?.run { pageProperties.pageId } ?: -1)
-                val findInPageActionProvider = FindInWebPageActionProvider(this, funnel)
+                val articleFindInPageInteractionEvent = ArticleFindInPageInteractionEvent(title.wikiSite.dbName(), model.page?.run { pageProperties.pageId } ?: -1)
```
dbrant, line 1054: This can be `model.page?.pageProperties?.pageId ?: -1`

app/src/main/java/org/wikipedia/analytics/eventplatform/AppearanceSettingInteractionEvent.kt
```diff
@@ -0,0 +1,40 @@
+package org.wikipedia.analytics.eventplatform
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import org.wikipedia.Constants.InvokeSource
+import org.wikipedia.auth.AccountUtil
+import org.wikipedia.theme.Theme
+
+class AppearanceSettingInteractionEvent(private var source: InvokeSource) {
```
dbrant, line 9: All constructor parameters in all of the Event classes and all Impl classes can be `val` instead of `var`.

PR #3286, opened 2022-04-29
Add a temporary fix for reading focus mode on article description

cooltey:
Per discussion from Robin, this should be included in the reading focus mode.

This is a temporary fix, the actual fix should be on the `mobile-html` side.


cooltey:
Close since there's a patch that fixes the issue.

PR #3289, opened 2022-05-02
Use resource methods.

Isira-Seneviratne:


app/src/main/java/org/wikipedia/watchlist/WatchlistItemView.kt
```diff
@@ -90,8 +90,7 @@ class WatchlistItemView constructor(context: Context, attrs: AttributeSet? = nul
         val themedTint = ColorStateList.valueOf(ResourceUtil.getThemedColor(context, R.attr.color_group_61))
         binding.diffText.text = text
         binding.diffText.setTextColor(themedTint)
-        binding.diffText.icon = if (iconResourceDrawable == null) null
-        else ContextCompat.getDrawable(context, iconResourceDrawable)
+        iconResourceDrawable?.let { binding.diffText.setIconResource(it) }
```
dbrant, line 93: Subtle bug:
There are cases when we **want** to explicitly set the icon resource to null. The logic here should be exactly the same as before.

dbrant (changes requested):

PR #3290, opened 2022-05-02
Use ArrayMap.

Isira-Seneviratne:


app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
```diff
@@ -165,8 +166,9 @@ object EventPlatformClient {
             if (!Prefs.isEventLoggingEnabled) {
                 return
             }
-            QUEUE.groupBy { it.stream }.forEach { (stream, events) ->
-                sendEventsForStream(STREAM_CONFIGS[stream]!!, events)
+            val eventsPerStream = QUEUE.groupByTo(arrayMapOf()) { it.stream }
+            for (i in 0 until eventsPerStream.size) {
```
dbrant, line 170: Why not keep using the `forEach` construct here? (and in other places where you've switched to iterating through `indices`?)

dbrant (changes requested):

dbrant:
I don't believe we have any particular issues with memory efficiency with our usage of regular MutableMaps.
After some minimal performance profiling, ArrayMap is about 50% **slower** than MutableMap, so I'm not sure this is necessary.

PR #3292, opened 2022-05-02
Analytics wiring for edit history

sharvaniharan:
Analytics wiring for showing edit history screen, showing revision screen and the related actions like compare and thank.

### DIFF
--- a/app/src/main/java/org/wikipedia/diff/ArticleEditDetailsViewModel.kt
+++ b/app/src/main/java/org/wikipedia/diff/ArticleEditDetailsViewModel.kt
@@ -7,7 +7,6 @@ import androidx.lifecycle.ViewModelProvider
 import androidx.lifecycle.viewModelScope
 import kotlinx.coroutines.*
 import org.wikipedia.analytics.WatchlistFunnel
-import org.wikipedia.analytics.eventplatform.EditHistoryInteractionEvent
 import org.wikipedia.dataclient.ServiceFactory
 import org.wikipedia.dataclient.WikiSite
 import org.wikipedia.dataclient.mwapi.MwQueryPage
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/diff/ArticleEditDetailsViewModel.kt
```diff
@@ -132,13 +133,17 @@ class ArticleEditDetailsViewModel(bundle: Bundle) : ViewModel() {
...
+                   revisionId: Long,
+                   editHistoryInteractionEvent: EditHistoryInteractionEvent?) {
         viewModelScope.launch(CoroutineExceptionHandler { _, throwable ->
             thankStatus.postValue(Resource.Error(throwable))
+            editHistoryInteractionEvent?.logThankFail()
         }) {
             withContext(Dispatchers.IO) {
                 val token = ServiceFactory.get(wikiSite).getCsrfToken().query?.csrfToken()
                 thankStatus.postValue(Resource.Success(ServiceFactory.get(wikiSite).postThanksToRevision(revisionId, token!!)))
+                editHistoryInteractionEvent?.logThankSuccess()
```
cooltey, line 146: This should be put in the `viewModel.thankStatus.observe` block?

app/src/main/java/org/wikipedia/page/PageFragment.kt
```diff
@@ -1429,7 +1427,7 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
 
         override fun onViewEditHistorySelected() {
             title?.run {
-                startActivity(EditHistoryListActivity.newIntent(requireContext(), this))
+                startActivity(EditHistoryListActivity.newIntent(requireContext(), this, model.page?.run { pageProperties.pageId } ?: -1))
```
cooltey, line 1430: Don't have to use `run {}` for getting variables. Just simply using the null check with express symbols is fine.

```
model.page?.pageProperties?.pageId ?: -1
```

cooltey (changes requested):

PR #3293, opened 2022-05-03
Follow-up to EditHistory analytics (properly get pageId)

dbrant:
@sharvaniharan 
* The EditHistory activity itself gets the `pageId` of the current page _for free_ from the API, so it doesn't need to be fed into the EditHistory activity by callers.
* Same for the EditDetails fragment.
* Since these activities get the `pageId` for free, why not include it in the Events that we send?


sharvaniharan:
That is great use of the api response. Thank you.

PR #3294, opened 2022-05-03
Complete the edit history analytics wiring

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T307388


app/src/main/java/org/wikipedia/analytics/eventplatform/EditHistoryInteractionEvent.kt
```diff
@@ -37,6 +41,54 @@ class EditHistoryInteractionEvent(private var wikiDb: String, private var pageId
...
+    fun logUndoSuccess() {
+        submitEvent("undo_success")
+    }
+
+    fun logUndoFail() {
+        submitEvent("undo_fail")
+    }
+
+    fun logOlderEditChevronClick() {
+        submitEvent("undo_fail")
```
dbrant, line 73: Incorrect string.

app/src/main/java/org/wikipedia/analytics/eventplatform/EditHistoryInteractionEvent.kt
```diff
@@ -37,6 +41,54 @@ class EditHistoryInteractionEvent(private var wikiDb: String, private var pageId
...
+    }
+
+    fun logNewerEditChevronClick() {
+        submitEvent("undo_fail")
```
dbrant, line 77: Incorrect string. (for all the rest).

app/src/main/java/org/wikipedia/diff/UndoEditDialog.kt
```diff
@@ -41,5 +42,6 @@ class UndoEditDialog constructor(context: Context, callback: Callback) : AlertDi
     override fun onDetachedFromWindow() {
         super.onDetachedFromWindow()
         binding.textInput.removeTextChangedListener(watcher)
+        editHistoryInteractionEvent?.logUndoCancel()
```
dbrant, line 45: This will always be called, even if they didn't actually cancel.

dbrant (changes requested):

PR #3298, opened 2022-05-05
MEP: Factor out `is_anon` parameter into a base class.

dbrant:
Using the magic of object-oriented programming™ we can create a base class that contains an `is_anon` parameter, and make child classes that inherit from it, so that we don't have to re-declare `is_anon` in every single class.

Additionally, the `sessionId`, `appInstallId`, and `dt` parameters can be initialized on instantiation, which simplifies the submission of events.


cooltey (approved):
LGTM

sharvaniharan:
Lgtm. My PR #3307 is ready. @dbrant once you merge it here, we can merge this as a whole.

PR #3302, opened 2022-05-06
Error handling in FeedCoordinatorBase.kt.

PatrickBarsch:
Use requestOfflineCard(),
instead of inserting the card unchecked with appendCard(),
to prevent duplication.

https://phabricator.wikimedia.org/T282388


dbrant:
A reasonable fix, thanks.

PR #3309, opened 2022-05-07
Use AppCompatResources.getDrawable().

Isira-Seneviratne:


dbrant:
Why is this better?

Isira-Seneviratne:
> Why is this better?

The method includes some fixes for issues with vector drawables on API level 21.

PR #3312, opened 2022-05-09
Accept Pairs in log().

Isira-Seneviratne:


dbrant:
We are in the process of deprecating and/or moving the `Funnel` classes to a different logging system, so any kind of style enhancements will not be necessary there.

PR #3327, opened 2022-05-11
Add option to show categories of categories.

dbrant:
https://phabricator.wikimedia.org/T307785


cooltey (changes requested):
Looks good to me. But I also found an issue when the bottom sheet shows the error view on the top of the categories.

Here's the screenshot when I visited "User: Cooltey" in `zhwiki` and continually tapped on the top category of the list.

dbrant:
> Looks good to me. But I also found an issue when the bottom sheet shows the error view on the top of the categories.

Great catch, and... done.

PR #3332, opened 2022-05-12
Refactor: RandomActivity to use ViewModel

Lampotrias:
Hi all!

It's my version refactoring RandomActivity with ViewModel.
https://phabricator.wikimedia.org/T303942

Thanks in advance for the code review!


app/src/main/java/org/wikipedia/random/RandomViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+class RandomViewModel(bundle: Bundle) : ViewModel() {
+
+	private val funnel: RandomizerFunnel
+	private var saveShareJob: Job? = null
+
+	private val handler = CoroutineExceptionHandler { _, throwable ->
+		L.w(throwable)
+	}
+
+	private val _savaShareState = MutableStateFlow(false)
```
dbrant, line 27: misspelling

app/src/main/java/org/wikipedia/random/RandomViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+	val savaShareState: StateFlow<Boolean> = _savaShareState
```
dbrant, line 28: misspelling

app/src/main/java/org/wikipedia/random/RandomViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+
+	private val _saveToDefaultList = MutableStateFlow<SaveSharedState<PageTitle>>(Initial())
```
dbrant, line 30: What's the reason for having underscore `_` and non-underscore versions of all these fields?

app/src/main/java/org/wikipedia/random/RandomViewModel.kt
```diff
@@ -0,0 +1,103 @@
...
+		_saveToDefaultList.value = Result(title)
+	}
+
+	fun saveToCustomList(title: PageTitle) {
+		_saveToCustomList.value = Result(title)
+	}
+
+	fun movePageToList(sourceReadingListId: Long, title: PageTitle) {
+		_movePageToList.value = Result(MoveDataSource(sourceReadingListId, title))
+	}
```
dbrant, line 56: I'm not sure I see the benefit of the above three functions. You're passing in a PageTitle, which gets passed to a Flow, which gets passed right back to the Fragment. Why not keep these interactions within the Fragment as before?

app/src/main/java/org/wikipedia/random/RandomViewModel.kt
Lampotrias: The non-underscored fields used only observe values in other entities, such fragments or activities. The underscored field guarantees emit values only from view model. Unfortunately now not other way for it.

app/src/main/java/org/wikipedia/random/RandomViewModel.kt
Lampotrias: I'll fix it. I created these methods with the expectation that the logic could be complicated in the future

Lampotrias:
Thank you for your feedback.

I fixed all defects in pull request. In RandomItemFragment was added viewmodel and removed `retainInstance = true`. I tested case with rotation of screen and saw no problem with it.

dbrant (changes requested):

app/src/main/java/org/wikipedia/random/RandomItemViewModel.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.wikipedia.dataclient.WikiSite
+import org.wikipedia.dataclient.page.PageSummary
+import org.wikipedia.util.Resource
+
+class RandomItemViewModel(bundle: Bundle) : ViewModel() {
+
+	private val wikiSite: WikiSite = bundle.getParcelable(RandomActivity.INTENT_EXTRA_WIKISITE)!!
+
+	private val _saveShareState = MutableStateFlow<SaveSharedState>(Initial)
+	val saveShareState = _saveShareState.asStateFlow()
```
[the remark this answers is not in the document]
Lampotrias: Yes, you are right.

app/src/main/java/org/wikipedia/random/RandomItemFragment.kt
```diff
@@ -33,7 +34,7 @@ class RandomItemFragment : Fragment() {
     private var _binding: FragmentRandomItemBinding? = null
     private val binding get() = _binding!!
 
-    private val disposables = CompositeDisposable()
+    private val viewModel: RandomItemViewModel by viewModels { RandomItemViewModel.Factory(requireArguments()) }
 
     private lateinit var wikiSite: WikiSite
```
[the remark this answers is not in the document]
Lampotrias: Now, I don't need `wikiSite`  in the ViewModel, I planning use `WikipediaApp.getInstance().wikiSite`. Currently, `wikiSite` use in `RandomItemFragment` for call:
```L10nUtil.setConditionalLayoutDirection(view, wikiSite.languageCode)```, as it was before the refactoring

File: app/src/main/java/org/wikipedia/random/RandomItemViewModel.kt [hunk 7]
```diff
@@ -0,0 +1,58 @@
...
+
+	private val disposables = CompositeDisposable()
+
+	init {
+		getRandomPage()
+	}
+
+	fun getRandomPage() {
+		val d = ServiceFactory.getRest(wikiSite).randomSummary
```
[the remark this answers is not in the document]
Lampotrias: I have question with select MediaWiki Action API. Do I need to make a request like `https://en.wikipedia.org/w/api.php?action=query&format=json&list=random&formatversion=2` and then get an instance of PageSummary

app/src/main/java/org/wikipedia/random/RandomItemFragment.kt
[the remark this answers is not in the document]
Lampotrias: Ok, now `wikiSite` is given from `viewModel`.

Lampotrias:
Thank you for your feedback.

I removed all code, that was created for the future, such Factory.

And sorry about the copy-paste code. This is my first PR, I'm a little worried.

PR #3333, opened 2022-05-12
Breadcrumbs for user actoins

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T305575

### DIFF
--- a/app/src/main/java/org/wikipedia/activity/BaseActivity.kt
+++ b/app/src/main/java/org/wikipedia/activity/BaseActivity.kt
@@ -11,8 +11,12 @@ import android.net.Uri
 import android.os.Build
 import android.os.Bundle
 import android.provider.Settings
+import android.util.Log
 import android.view.MenuItem
 import android.view.MotionEvent
+import android.view.View
+import android.view.View.OnTouchListener
+import android.view.ViewGroup
 import androidx.annotation.ColorInt
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatActivity
@@ -45,8 +49,10 @@ import org.wikipedia.util.PermissionUtil
 import org.wikipedia.util.ResourceUtil
 import org.wikipedia.util.log.L
 import org.wikipedia.views.ImageZoomHelper
+import kotlin.math.abs
 
-abstract class BaseActivity : AppCompatActivity() {
+
+abstract class BaseActivity : AppCompatActivity(), OnTouchListener {
     private lateinit var exclusiveBusMethods: ExclusiveBusConsumer
     private val networkStateReceiver = NetworkStateReceiver()
     private var previousNetworkState = WikipediaApp.getInstance().isOnline
@@ -91,6 +97,22 @@ abstract class BaseActivity : AppCompatActivity() {
         maybeShowLoggedOutInBackgroundDialog()
 
         Prefs.localClassName = localClassName
+
+        val decorView = window.decorView
+        decorView.viewTreeObserver.addOnGlobalLayoutListener { printClickedViews(window.decorView) }
+    }
+
+    private fun printClickedViews(currentView: View?) {
+        if (currentView == null) {
+            return
+        }
+        currentView.setOnTouchListener(this)
+        if (currentView is ViewGroup) {
+            val viewGroup = currentView
+            for (i in 0 until viewGroup.childCount) {
+                printClickedViews(viewGroup.getChildAt(i))
+            }
+        }
     }
 
     override fun onDestroy() {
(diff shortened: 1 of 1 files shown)


dbrant (changes requested):
Architecturally this looks very promising, but will need some more refinement. Some things I'm noticing right away include:
* This is sending `logClick` events when clicking _anything_, even those Views that are not clickable, e.g. when tapping on empty space, or when tapping on plain text that doesn't respond to clicks.
* This doesn't seem to differentiate between clicks and long-clicks (will send a `logClick` event when long-pressing a button)
* This won't catch clicks that don't come from touch input, i.e. clicks caused by keyboard navigation or voice navigation.
* This breaks Views that respond to touch interactions, such as rearrangeable items: Feed Configuration, and Language List, where the user can touch-and-drag the items to rearrange their order.

Let's keep brainstorming a bit further on how to generalize our system so that the above points might be addressed.

On another note, we should try to make the event string have a dot-delimited structure/hierarchy, because (**hint**) this will be essential when we upgrade to the Metrics Platform.

So for example, when the onboarding Continue button is clicked, the action string could be: `"InitialOnboardingActivity.PAGE_WELCOME.continueButton.click"`.

When the usage-data Switch component is turned ON, the event string could be: `"InitialOnboardingActivity.PAGE_USAGE_DATA.switchView.on"`. 

...and so on.

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -90,6 +97,15 @@ abstract class BaseActivity : AppCompatActivity() {
         maybeShowLoggedOutInBackgroundDialog()
 
         Prefs.localClassName = localClassName
+
+        val decorView = window.decorView
+        decorView.viewTreeObserver.addOnGlobalLayoutListener { ViewUtil.setTouchListenersToViews(window.decorView, this) }
```
cooltey, line 102: Please use `decorView` in the `addOnGlobalLayoutListener` instead of `window.decorView`.

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -298,4 +319,34 @@ abstract class BaseActivity : AppCompatActivity() {
         private var EXCLUSIVE_BUS_METHODS: ExclusiveBusConsumer? = null
         private var EXCLUSIVE_DISPOSABLE: Disposable? = null
     }
+
+    private var startX = 0f
+    private var startY = 0f
+    private var startSystemTime = 0L
+    override fun onTouch(view: View?, event: MotionEvent?): Boolean {
```
cooltey, line 326: Can we move this function above the `companion object`?

app/src/main/java/org/wikipedia/analytics/eventplatform/BreadCrumbViewUtil.kt
```diff
@@ -0,0 +1,124 @@
...
+        return ""
+    }
+
+    private fun getVisibleFragment(activity: BaseActivity): Fragment? {
+        val fragmentManager = activity.supportFragmentManager
+
+        if (activity is SingleFragmentActivity<*>) {
+            return fragmentManager.findFragmentById(R.id.fragment_container)
+        } else {
+            val fragments: List<Fragment> = fragmentManager.fragments
```
cooltey, line 117: `List<Fragment>` can be removed.

app/src/main/java/org/wikipedia/views/ViewUtil.kt
```diff
@@ -23,11 +27,14 @@ import org.wikipedia.util.DimenUtil.roundedDpToPx
...
 import org.wikipedia.util.WhiteBackgroundTransformation
 import java.util.*
+import kotlin.math.abs
 
 object ViewUtil {
     private val CENTER_CROP_ROUNDED_CORNERS = MultiTransformation(CenterCrop(),
             WhiteBackgroundTransformation(), RoundedCorners(roundedDpToPx(2f)))
     val ROUNDED_CORNERS = RoundedCorners(roundedDpToPx(15f))
+    const val CLICK_ACTION_THRESHOLD = 200
+    const val CLICK_TIME_THRESHOLD = 400
```
cooltey, line 37: Looks like these two variables can be `private`. And would it be possible to move these two variables to the top of the first `private val` variable?

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -298,4 +319,34 @@ abstract class BaseActivity : AppCompatActivity() {
         private var EXCLUSIVE_BUS_METHODS: ExclusiveBusConsumer? = null
         private var EXCLUSIVE_DISPOSABLE: Disposable? = null
     }
+
+    private var startX = 0f
+    private var startY = 0f
+    private var startSystemTime = 0L
```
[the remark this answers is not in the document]
sharvaniharan: I felt it was more readable right above the function? Lmk if that sounds good..

sharvaniharan:
> "InitialOnboardingActivity.PAGE_WELCOME.continueButton.click"

@dbrant Fixed most issues. Created a ticket to capture the rest: https://phabricator.wikimedia.org/T309802. Please let me know if it looks ready to merge given that it is also behind pre-prod flag now..

dbrant:
@sharvaniharan  Just so I understand, is detecting "swipe" actions part of the product/analytics requirements? i.e. is it necessary? It looks like this is using a rather heavy `GestureDetector` at all times, in all activities, but only cares about logging swipes in the InitialOnboardingActivity. Does this add value to the kind of data we want to get?

sharvaniharan:
@dbrant Since it is hard to make sure that each. and every action is caught on all screens, this first pass is to try to capture as many events as possible in the initial onboarding flow. To which, the. swipe on the screen is important. I have now changed it to attach the gestureDetector only to the initialOnboarding screen.

sharvaniharan:
We can also just do the treatment just to the Onboarding Page. so please merge this if you feel that way is better: #3378

PR #3346, opened 2022-05-19
Convert RemoteConfig to use kotlinx.serialization.

dbrant:
This is pretty much the last thing that was using the old `JSONObject` routines, instead of our new `kotlinx.serialization` logic.

The only other stuff that still uses `JSONObject` is all of the `Funnel` classes, but those are all considered deprecated, and will be going away eventually anyway.


app/src/main/java/org/wikipedia/analytics/Funnel.kt
```diff
@@ -16,8 +16,6 @@ abstract class Funnel @JvmOverloads internal constructor(protected val app: Wiki
                                                          private val revision: Int, private val sampleRate: Int = SAMPLE_LOG_ALL,
                                                          private val wiki: WikiSite? = null) {
 
-    private val sampleRateRemoteParamName = schemaName + "_rate"
```
dbrant, line 19: (this dynamically-programmable sample rate was never used in RemoteConfig, and never will be.)

PR #3347, opened 2022-05-19
Convert WikipediaApp to kotlin, at long last.

dbrant:
Now that we have a slight pause in between major features, let's just get this done, like ripping off a bandage.

The number of files changed is eye-popping, but aside from the actual changes to `WikipediaApp.kt`, it's all literally find/replace.


sharvaniharan:
Lgtm, @cooltey please take a look and merge.

app/src/main/java/org/wikipedia/WikipediaApp.kt
```diff
@@ -0,0 +1,318 @@
...
+        WikipediaFirebaseMessagingService.updateSubscription()
+
+        EventPlatformClient.setUpStreamConfigs()
+    }
+
+    /**
+     * @return the value that should go in the Accept-Language header.
+     */
+    fun getAcceptLanguage(wiki: WikiSite?): String {
+        val wikiLang = if (wiki == null || "meta" == wiki.languageCode) "" else StringUtils.defaultString(wiki.languageCode)
```
cooltey, line 194: Should we either show `wiki.languageCode` or `wiki.languageCode.orEmpty()` instead of using `StringUtils.defaultString()`?

dbrant: yes!

PR #3352, opened 2022-05-23
Use lazy properties in WikipediaApp.

Isira-Seneviratne:


app/src/main/java/org/wikipedia/WikipediaApp.kt
```diff
@@ -42,17 +42,22 @@ import org.wikipedia.util.log.L
...
-        private set
-    lateinit var sessionFunnel: SessionFunnel
-        private set
-    lateinit var userAgent: String
-        private set
+    val mainThreadHandler by lazy { Handler(mainLooper) }
+    val languageState by lazy { AppLanguageState(this) }
+    val funnelManager by lazy { FunnelManager(this) }
+    val sessionFunnel by lazy { SessionFunnel(this) }
+    val userAgent: String by lazy {
```
cooltey, line 49: `: String` can be removed.

PR #3357, opened 2022-05-25
NavTabs: Better naming for breadcrumbs

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T305575


app/src/main/res/values/ids.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <item name="explore" type="id" />
+    <item name="reading_lists" type="id" />
+    <item name="search" type="id" />
+    <item name="edits" type="id" />
```
cooltey, line 6: Is it possible to add prefixes like `nav_tab_` to each id?

PR #3367, opened 2022-05-31
Follow-up to Rollback API underpinnings.

dbrant:
Performing a rollback requires a special `rollback` token, which is acquired using the same call as a `csrf` token.


sharvaniharan:
Lgtm, will let @cooltey take a look and merge.

PR #3368, opened 2022-05-31
Add namespace hints when searching.

dbrant:
https://phabricator.wikimedia.org/T308750


dbrant:
@cooltey This is ready for review.

app/src/main/java/org/wikipedia/search/RecentSearchesFragment.kt
```diff
@@ -155,4 +181,34 @@ class RecentSearchesFragment : Fragment() {
...
         }
     }
+
+    private inner class NamespaceItemViewHolder constructor(itemView: View) : RecyclerView.ViewHolder(itemView), View.OnClickListener {
+        private var ns: Namespace? = null
+
+        fun bindItem(ns: Namespace?) {
+            itemView.setOnClickListener(this)
+            this.ns = ns
+            (itemView as TextView).text = if (ns == null) getString(R.string.search_namespaces) else namespaceMap[callback?.getLangCode()]?.get(ns).orEmpty() + ":"
```
cooltey, line 191: Would it be possible to set `ns == null` as a variable so that we don't have to add the logic in the following code?

app/src/main/java/org/wikipedia/search/RecentSearchesFragment.kt
```diff
@@ -155,4 +181,34 @@ class RecentSearchesFragment : Fragment() {
...
+        override fun getItemCount(): Int {
+            return namespaceHints.size + 1
+        }
+
+        override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): NamespaceItemViewHolder {
+            return NamespaceItemViewHolder(LayoutInflater.from(parent.context).inflate(R.layout.item_namespace, parent, false))
+        }
+
+        override fun onBindViewHolder(holder: NamespaceItemViewHolder, pos: Int) {
+            holder.bindItem(if (pos > 0) namespaceHints[pos - 1] else null)
```
cooltey, line 211: We can use `namespaceHints.getOrNull(pos - 1)` here.

app/src/main/java/org/wikipedia/search/RecentSearchesFragment.kt
```diff
@@ -12,29 +12,35 @@ import androidx.lifecycle.lifecycleScope
...
+        fun getLangCode(): String
     }
 
     private var _binding: FragmentSearchRecentBinding? = null
     private val binding get() = _binding!!
     var callback: Callback? = null
-    var recentSearchList = mutableListOf<RecentSearch>()
+    val recentSearchList = mutableListOf<RecentSearch>()
+    private val namespaceHints = listOf(Namespace.USER, Namespace.PORTAL, Namespace.HELP)
+    private val namespaceMap = ConcurrentHashMap<String, Map<Namespace, String>>()
```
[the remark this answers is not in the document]
dbrant: yep

app/src/main/java/org/wikipedia/search/RecentSearchesFragment.kt
dbrant: yep

app/src/main/java/org/wikipedia/search/RecentSearchesFragment.kt
dbrant: yes!

app/src/main/java/org/wikipedia/search/RecentSearchesFragment.kt
```diff
@@ -101,13 +103,37 @@ class RecentSearchesFragment : Fragment() {
         callback?.onAddLanguageClicked()
     }
 
+    fun onLangCodeChanged() {
+        lifecycleScope.launch {
```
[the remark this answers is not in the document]
dbrant: correct!

dbrant:
> Found a minor issue that when changing the device orientation, the keyboard will overlap the bottom namespace suggestion view.

Yeah, that's pretty much unavoidable, and acceptable for this iteration.

PR #3380, opened 2022-06-03
Follow-up to breadcrumbs.

dbrant:
* Remove a lot of unnecessary null checks and `let { }` blocks.
* Greatly simplify click detection, with system-provided touch slop and timeout values.
* Revert changes to `getTooltip` function, where the new `anchor` parameter did not do what it says.


sharvaniharan (changes requested):
Everything looks good. Only one issue is that the click and long click are not getting recorded...

PR #3381, opened 2022-06-03
Greatly simplify/streamline breadcrumb click detection.

dbrant:
* Use a single TouchListener in BaseActivity to listen for click events, instead of applying a separate TouchListener to every view upon every layout change.
* We no longer need to use reflection to check for touch listeners on individual views.


cooltey:
Excellent work!

PR #3395, opened 2022-06-07
Use HandlerCompat.hasCallbacks().

Isira-Seneviratne:


dbrant:
Why is this necessary? Please write a _detailed description_ of the reason for the change in all your pull requests.

PR #3397, opened 2022-06-07
Make HistoryEntryDao methods suspend.

Isira-Seneviratne:


app/src/main/java/org/wikipedia/history/HistoryFragment.kt
```diff
@@ -204,7 +216,9 @@ class HistoryFragment : Fragment(), BackPressedHandler {
         val message = if (entries.size == 1) getString(R.string.history_item_deleted, entries[0].title.displayText) else getString(R.string.history_items_deleted, entries.size)
         val snackbar = FeedbackUtil.makeSnackbar(requireActivity(), message)
         snackbar.setAction(R.string.history_item_delete_undo) {
-            AppDatabase.instance.historyEntryDao().insert(entries)
+            lifecycleScope.launch(Dispatchers.IO) {
+                AppDatabase.instance.historyEntryDao().insert(entries)
+            }
             reloadHistoryItems()
```
cooltey, line 222: If we press the `UNDO` button on the snackbar after seeing `%s removed from history`, the deleted item will not be inserted back into the list.

PR #3403, opened 2022-06-09
Fix association of Talk topic by ID.

dbrant:
When tapping on a topic in the `Topics` activity, and finding it again in the `Topic` activity, it's actually more consistent to use the `id` of the topic rather than the `name`. (The `name` is still important, and is used for subscribing.)

https://phabricator.wikimedia.org/T297617#7991442


cooltey:
I notice the latency when entering the `User:Sadads` talk topic, it takes about 1 or 2 seconds to get into the thread.

cooltey:
> > I notice the latency when entering the `User:Sadads` talk topic, it takes about 1 or 2 seconds to get into the thread, could you please check the `User:Sadads` user talk to see if you can reproduce it?
> 
> That doesn't seem to be related to this PR. It's probably because in `TalkTopicsActivity` you're using a RecyclerView inside a NestedScrollView which is causing all the items in the RecyclerView to get rendered and bound (instead of only the items that are on the screen).
> 
> Ideally in `TalkTopicsActivity` the header and footer views should be extra items in the RecyclerView (in a CompoundAdapter).

Makes sense to me. I can create a separate ticket to fix it.

PR #3374, opened 2022-06-01
Add Rollback button and Undo to Diff View

cooltey:
https://phabricator.wikimedia.org/T303547

PR #3406, opened 2022-06-09
Fix RTL consistency in Talk activities, and menu order.

dbrant:
* Apply explicit RTL to specific subviews, not the root view.
* Update order of overflow menu items, and fix tinting.

https://phabricator.wikimedia.org/T297617#7991442

### DIFF
--- a/app/src/main/res/menu/menu_talk.xml
+++ b/app/src/main/res/menu/menu_talk.xml
@@ -22,40 +22,34 @@
                 android:icon="@drawable/ic_translate_white_24dp"
                 android:title="@string/langlinks_activity_title"
                 android:visible="false"
-                app:iconTint="?attr/material_theme_secondary_color"
                 app:showAsAction="never" />
             <item
                 android:id="@+id/menu_view_user_page"
                 android:icon="@drawable/ic_user_avatar"
                 android:title="@string/menu_option_user_page"
                 android:visible="false"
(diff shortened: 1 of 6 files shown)


app/src/main/res/menu/menu_talk.xml
```diff
@@ -17,40 +17,45 @@
...
+                android:title="@string/menu_page_watch"
+                android:visible="false"
+                app:iconTint="?attr/material_theme_secondary_color"
                 app:showAsAction="never" />
             <item
                 android:id="@+id/menu_talk_topic_share"
-                android:icon="@drawable/ic_share_material_secondary_themed"
+                android:icon="@drawable/ic_share"
                 android:title="@string/menu_page_article_share"
+                app:iconTint="?attr/material_theme_secondary_color"
```
cooltey, line 58: Looks like we can apply the tint here:
https://github.com/wikimedia/apps-android-wikipedia/pull/3374/files#diff-1e051f2b0330608e92eba8edc616403ec7bdaa76e29ce0cfb794700686715925R118-R121

Do you think it would be better if you update the tint color in your PR? e.g. add `color_group_9` to `android:tint` in `popupTheme`?  

BTW, the color group should be `color_group_9`, per Robin's comment here: 
https://phabricator.wikimedia.org/T303547#7989356

dbrant: Excellent; I've added that style change to this PR.

PR #3407, opened 2022-06-09
Implement coffee-roll headers in Talk pages.

dbrant:
This hooks up to the latest updates in the API, which gives us content for improperly-signed topics, so that we can display them, even though we can't reply or subscribe to them.

The updated API also gives us the header ("coffee-roll") section, which we should display at the top of the list of topics.


dbrant:
> How should we handle the coffee roll when the `actionMode` is on? should we hide it or show it and allow searching the content of it?

Yep! The coffee-roll should be searchable like the other topics. (Done)

PR #3440, opened 2022-06-22
Fix: do not show empty lead image view in TalkTopicsActivity

cooltey:
Now it shows an empty block of the view.


cooltey:
Self-merged this PR.

PR #3443, opened 2022-06-23
Update schema revision for MobileWikiAppEdit

cooltey:


cooltey:
Self merge for re-release to beta.

PR #3451, opened 2022-06-30
Add archived talk pages access in the app

cooltey:
https://phabricator.wikimedia.org/T275917

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/Service.kt
+++ b/app/src/main/java/org/wikipedia/dataclient/Service.kt
@@ -59,6 +59,18 @@ interface Service {
         @Query("gsroffset") gsrOffset: String?
     ): Observable<MwQueryResponse>
 
+    @GET(
+        MW_API_PREFIX + "action=query&converttitles=" +
+                "&prop=description|pageprops|info&ppprop=mainpage|disambiguation" +
+                "&generator=search&gsrnamespace=0&gsrwhat=text" +
+                "&inprop=varianttitles" +
+                "&gsrinfo=&gsrprop=redirecttitle&pilicense=any"
+    )
+    suspend fun searchSubPages(@Query("gsrsearch") searchTerm: String?,
+                               @Query("gsrlimit") gsrLimit: Int,
+                               @Query("continue") cont: String?,
+                               @Query("gsroffset") gsrOffset: String?): MwQueryResponse
+
     @GET(MW_API_PREFIX + "action=query&list=allusers&auwitheditsonly=1")
     fun prefixSearchUsers(
             @Query("auprefix") prefix: String,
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/wikipedia/dataclient/Service.kt
```diff
@@ -59,6 +59,18 @@ interface Service {
...
     ): Observable<MwQueryResponse>
 
+    @GET(
+        MW_API_PREFIX + "action=query&redirects=" +
+                "&converttitles=&prop=description|info" +
+                "&generator=prefixsearch&gpsnamespace=0&list=search&srnamespace=0" +
+                "&inprop=varianttitles" +
+                "&srwhat=text&srinfo=suggestion&srprop=&sroffset=0&srlimit=1"
+    )
+    suspend fun searchSubPages(@Query("gpssearch") searchTerm: String?,
```
tacsipacsi, line 69: A trailing slash is missing. For example, https://de.wikipedia.org/w/api.php?action=query&list=prefixsearch&pssearch=Diskussion:Gl%C3%A4ubiger returns `Diskussion:Gläubigerschutz` as well, even though it’s definitely not an archive of [Diskussion:Gläubiger](https://de.wikipedia.org/wiki/Diskussion:Gl%C3%A4ubiger). https://de.wikipedia.org/w/api.php?action=query&list=prefixsearch&pssearch=Diskussion:Gl%C3%A4ubiger/ returns the expected result (nothing, as the given talk page has no archives).

app/src/main/res/values/strings.xml
```diff
@@ -90,6 +90,7 @@
     <string name="menu_page_new_tab">New tab</string>
     <string name="menu_page_reading_lists">Reading lists</string>
     <string name="menu_page_recently_viewed">Recently viewed</string>
+    <string name="menu_page_archive">Archive</string>
```
tacsipacsi, line 93: Why singular? There may be several archive pages, and for this reason it’s usually in plural (cf. [Template:Archives on enwiki](https://en.wikipedia.org/wiki/Template:Archives)).

app/src/main/res/values/strings.xml
```diff
@@ -1196,6 +1197,7 @@
     <string name="talk_footer_view_user_page">View user page</string>
     <string name="talk_footer_view_article">View article</string>
     <string name="talk_footer_last_modified"><![CDATA[Last edited %1$s by %2$s]]></string>
+    <string name="talk_archived_title">Talk archive: %s</string>
```
tacsipacsi, line 1200: This should also be plural, as it lists the several archive**s**.

app/src/main/java/org/wikipedia/dataclient/Service.kt
dbrant: Already handled in #3475.

PR #3453, opened 2022-06-30
User Contributions screen.

dbrant:
https://phabricator.wikimedia.org/T310270


app/src/main/java/org/wikipedia/usercontrib/UserContribItemView.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.wikipedia.databinding.ItemUserContribBinding
+import org.wikipedia.dataclient.mwapi.UserContribution
+import org.wikipedia.util.DateUtil
+import org.wikipedia.util.ResourceUtil
+import org.wikipedia.util.StringUtil
+
+class UserContribItemView(context: Context) : FrameLayout(context) {
+    interface Listener {
+        fun onClick()
+        fun onLongClick()
```
cooltey, line 21: Do we need to implement the `onLongClick` action for the list item?

cooltey (changes requested):
Looks good to me! Only one question about the longClick listener.

app/src/main/java/org/wikipedia/usercontrib/UserContribItemView.kt
dbrant: Nope! Removed.

PR #3472, opened 2022-07-08
Update androidx.fragment:fragment-ktx version to 1.5.0

cooltey:
https://developer.android.com/jetpack/androidx/releases/fragment#version_15_2


dbrant:
@cooltey  It looks like they've updated `fragment-ktx` to 1.5.2, which may have fixed the issues that we saw. It may be worth reviving this PR and continuing our upgrade path.

PR #3479, opened 2022-07-08
Downgrade the androidx.fragment library to 1.4.1

cooltey:
Version `1.5.0` breaks the menu items behaviors in some fragments, we will need to wait until they fix the issue.


dbrant:
👍 Probably a good idea to no longer use any libraries that are version `*.0`, and always wait for `*.1`.

PR #3502, opened 2022-07-26
Fix: temporary fix the issue of showing redundant namespace in talk page title

cooltey:
https://phabricator.wikimedia.org/T299033#8104360


dbrant:
@cooltey What do you think of this solution: #3503

PR #3508, opened 2022-07-28
Optimization idea: exclude qq strings from build.

dbrant:
Here's the idea:
We have a large number of localized `strings` resources that correspond to different languages. One of those "languages" is `qq`, but it's not really a language - it's just hints for the translators and will never be used in the app itself, and yet the `qq` strings still get packaged into the APK, because the compiler believes that `qq` is just another language.

So, is there a way for us to _exclude_ `qq` from the build? This turned out to be much more difficult than I thought. The solution that I finally arrived at works like this:

The Gradle plugin offers a parameter called `resConfigs` that lets us specify which languages to _include_ (but not exclude). So we need a way to enumerate all the languages that we have, except `qq`, and pass them into `resConfigs`. I am therefore using Groovy to literally list the directory contents of the `res` folder, and filter out the localization directories, and feed them into the `resConfigs` setting.


dbrant:
Closing some stale pull requests. (the branches can remain in the repo, just not cluttering pull requests.)

PR #3510, opened 2022-08-01
Detect whether page is protected before allowing editing description.

dbrant:
https://phabricator.wikimedia.org/T313267


cooltey (changes requested):
Found a tiny issue that the warning dialog pops up when the keyboard is on and off. Would it be possible to hide the keyboard before showing the dialog?

dbrant:
> Found a tiny issue that the warning dialog pops up when the keyboard is on and off. Would it be possible to hide the keyboard before showing the dialog?

Done!

PR #3522, opened 2022-08-12
Show warning message from AbuseFilter when editing description

cooltey:
https://phabricator.wikimedia.org/T313994


app/src/main/java/org/wikipedia/dataclient/mwapi/MwServiceError.kt
```diff
@@ -41,7 +41,11 @@ class MwServiceError(val code: String? = null,
     }
 
     @Serializable
-    class Data(val messages: List<Message>? = null, val blockinfo: BlockInfo? = null)
+    class Data(
```
dbrant, line 44: Do we actually need to include/deserialize this structure? We're not using any of these fields, and it looks like your addition of prioritizing "`abusefilter-`" codes is all that's necessary here.

If we deserialize things without using them, we risk hidden errors if the structure given by the server isn't exactly what we expect.

dbrant (changes requested):

app/src/main/java/org/wikipedia/dataclient/mwapi/MwServiceError.kt
```diff
@@ -72,4 +76,11 @@ class MwServiceError(val code: String? = null,
                 return expiry.after(now)
             }
     }
+
+    @Serializable
+    open class AbuseFilter {
```
dbrant, line 81: Do we actually need to include/deserialize this structure? We're not using any of these fields, and it looks like your addition of prioritizing "`abusefilter-`" codes is all that's necessary here.

If we deserialize things without using them, we risk hidden errors if the structure given by the server isn't exactly what we expect.

dbrant (changes requested):

app/src/main/java/org/wikipedia/dataclient/mwapi/MwServiceError.kt
cooltey: Thanks! Forgot to remove the class after testing. Done!

PR #3524, opened 2022-08-12
Correctly stop / don't seek MediaPlayer on error.

dbrant:
Certain types of audio files are not supported by the Android MediaPlayer component. In these cases the component invokes the `onError()` callback, where we explicitly call the `stop()` function. But inside the `stop()` function we actually have a call to `seek()` to the beginning, which is not correct, and is causing `onError()` to be called in an endless loop.

The call to `seek(0)` was likely to work around very old versions of MediaPlayer not having correct behavior when `stop()` is called.

https://phabricator.wikimedia.org/T315062


app/src/main/java/org/wikipedia/page/PageFragment.kt
```diff
@@ -1274,8 +1275,11 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
             updateProgressBar(false)
         }
 
-        override fun onError() {
-            avPlayer?.stop()
+        override fun onError(code: Int, extra: Int) {
+            if (avPlayer?.isPlaying == true) {
+                avPlayer?.stop()
+            }
+            FeedbackUtil.showMessage(this@PageFragment, getString(R.string.media_playback_error, code.toString()))
```
cooltey, line 1282: The error message for the article "Nikola Jokic" will show "Error playing media: 1". 

Do you think the error code "1" make sense to users or can we show the error message without the error code?

dbrant: Yeah, I suppose you're right; done!
I was going back and forth about it. And actually, the error code "1" just means "UNKNOWN", so it's not helpful anyway. If a user sees this error, they can send us a screenshot, and we'll see which article they're on, so we can reproduce the code ourselves.

PR #3527, opened 2022-08-14
Use EnumSet for enum storage.

Isira-Seneviratne:
Use [`EnumSet`](https://docs.oracle.com/javase/8/docs/api/java/util/EnumSet.html) to store enum values, as it is optimized for use with enums.


dbrant:
This makes lookups from `EnumSet`s significantly slower than before (because of the `get()` operation which works in linear time), therefore unfortunately not suitable for our purposes.

PR #3529, opened 2022-08-16
Remove "Profile" from the user page menu option

cooltey:
https://phabricator.wikimedia.org/T315242

This PR also removes one of the unused "profile" related menu option.


app/src/main/res/values/strings.xml
```diff
@@ -1215,7 +1214,7 @@
     <string name="thank_dialog_positive_button_text">Send thanks</string>
     <string name="thank_dialog_negative_button_text">Cancel</string>
     <string name="thank_success_message">Your \'Thanks\' was sent to %s 👍</string>
-    <string name="menu_option_view_user_profile">View user profile page</string>
```
dbrant, line 1218: Do we really need to change the name of the string resource itself? (what about all the other language versions of it?)

dbrant (changes requested):

PR #3533, opened 2022-08-17
Radically improve performance of wikitext editor.

dbrant:
This makes our wikitext editing experience much more performant, and more usable on a wider range of devices.
There are a few key insights that led to this improvement:

1. The `androidx.appcompat` library adds a lot of extra baggage to the `AppCompatEditText` component, which actually cripples its performance with large amounts of text. Our new editing component will _no longer inherit_ from `AppCompatEditText`, and instead inherit from the base platform `EditText`. This accounts for ~**80%** of the performance issue.
2. Syntax highlighting (as well as find-in-page) uses Spans, and the longer the text, the more Spans there will be, which also impacts performance. Instead of applying syntax highlighting to the entire text every time it's updated, we will now apply syntax highlighting _only inside the currently-visible text_. This has a few tradeoffs (e.g. if you scroll very quickly, you will see un-highlighted text for a moment before it becomes highlighted; you might also see a little jumping because certain highlights affect the line height of text), but the benefits are tremendous. This accounts for another 10% of the performance issue.
3. By default the `EditText` component enables typing suggestions, which can have an impact on performance with large amounts of text, particularly on Samsung devices. We will now disable typing suggestions by default, and will provide a preference to enable them in our next update. (Typing suggestions are not very beneficial for wikitext editing anyway.)

https://phabricator.wikimedia.org/T164936

### DIFF
--- a/app/src/main/java/org/wikipedia/edit/FindInEditorActionProvider.kt
+++ b/app/src/main/java/org/wikipedia/edit/FindInEditorActionProvider.kt
@@ -80,7 +80,7 @@ class FindInEditorActionProvider(private val scrollView: View,
 
     private fun scrollToCurrentResult() {
         setMatchesResults(currentResultIndex, resultPositions.size)
-        val textPosition = if (resultPositions.isEmpty()) 0 else resultPositions[currentResultIndex]
+        val textPosition = resultPositions.getOrElse(currentResultIndex) { 0 }
         textView.setSelection(textPosition, textPosition + searchQuery.orEmpty().length)
         val r = Rect()
         textView.getFocusedRect(r)
--- a/app/src/main/java/org/wikipedia/edit/FindInEditorActionProvider.kt
+++ b/app/src/main/java/org/wikipedia/edit/FindInEditorActionProvider.kt
@@ -4,18 +4,20 @@ import android.graphics.Rect
 import android.view.ActionMode
 import android.view.MenuItem
 import android.view.View
-import android.widget.ScrollView
 import org.wikipedia.edit.richtext.SyntaxHighlighter
 import org.wikipedia.util.DeviceUtil
 import org.wikipedia.util.DimenUtil
 import org.wikipedia.views.FindInPageActionProvider
 import org.wikipedia.views.FindInPageActionProvider.FindInPageListener
-import org.wikipedia.views.PlainPasteEditText
+import java.util.*
 
-class FindInEditorActionProvider(private val scrollView: ScrollView,
-                                 private val textView: PlainPasteEditText,
+class FindInEditorActionProvider(private val scrollView: View,
+                                 private val textView: SyntaxHighlightableEditText,
                                  private val syntaxHighlighter: SyntaxHighlighter,
                                  private val actionMode: ActionMode) : FindInPageActionProvider(textView.context), FindInPageListener {
+
+    private val resultPositions = mutableListOf<Int>()
+    private var currentResultIndex = 0
     private var searchQuery: String? = null
 
     init {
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/wikipedia/edit/SyntaxHighlightableEditText.kt
```diff
@@ -0,0 +1,208 @@
...
+            applyPaddingForLineNumbers()
+        }
+
+    constructor(context: Context) : super(context)
+    constructor(context: Context, attrs: AttributeSet?) : super(context, attrs)
+    constructor(context: Context, attrs: AttributeSet?, defStyle: Int) : super(context, attrs, defStyle)
+
+    init {
+        // The MIME type(s) need to be set for onReceiveContent() to be called.
+        ViewCompat.setOnReceiveContentListener(this, arrayOf("text/*"), null)
```
cooltey, line 61: I notice the `this` has some warning message. Should we make the class `final`?

app/src/main/java/org/wikipedia/edit/SyntaxHighlightableEditText.kt
```diff
@@ -0,0 +1,208 @@
...
+    fun enqueueNoScrollingLayoutChange() {
+        allowScrollToCursor = false
+        postDelayed({
+            if (isAttachedToWindow) {
+                allowScrollToCursor = true
+            }
+        }, 1000)
+    }
+
+    fun enableTypingSuggestions(enabled: Boolean) {
```
cooltey, line 84: Should we make this function `private`?

app/src/main/java/org/wikipedia/views/PlainPasteEditText.kt
```diff
@@ -48,7 +36,7 @@ open class PlainPasteEditText : TextInputEditText {
     }
 
     override fun onCreateInputConnection(outAttrs: EditorInfo): InputConnection? {
-        inputConnection = super.onCreateInputConnection(outAttrs)
+        val inputConnection = super.onCreateInputConnection(outAttrs)
```
cooltey, line 39: Would it be possible to move `super.onCreateInputConnection(outAttrs)` to line `48`?

cooltey (changes requested):
Looks pretty good! Just have some minor comments.

app/src/main/java/org/wikipedia/edit/FindInEditorActionProvider.kt
```diff
@@ -31,37 +33,24 @@ class FindInEditorActionProvider(private val scrollView: ScrollView,
         return view
     }
 
-    private fun findInPage(text: String) {
-        textView.findListener = object : PlainPasteEditText.FindListener {
-            override fun onFinished(activeMatchOrdinal: Int, numberOfMatches: Int, textPosition: Int, findingNext: Boolean) {
```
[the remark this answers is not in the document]
dbrant: yep!

app/src/main/java/org/wikipedia/edit/FindInEditorActionProvider.kt
```diff
@@ -70,12 +59,32 @@ class FindInEditorActionProvider(private val scrollView: ScrollView,
...
+                }
+            } while (position >= 0)
         }
-        searchQuery = text
+        scrollToCurrentResult()
+    }
+
+    private fun scrollToCurrentResult() {
+        setMatchesResults(currentResultIndex, resultPositions.size)
+        val textPosition = if (resultPositions.isEmpty()) 0 else resultPositions[currentResultIndex]
```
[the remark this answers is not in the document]
dbrant: correct

app/src/main/java/org/wikipedia/edit/SyntaxHighlightableEditText.kt
dbrant: yep

app/src/main/java/org/wikipedia/edit/SyntaxHighlightableEditText.kt
dbrant: Technically yes, but this function is intended to be used by external classes, and will be used in the next update, so not much point in making it private now.

app/src/main/java/org/wikipedia/views/PlainPasteEditText.kt
dbrant: yep!

PR #3536, opened 2022-08-19
Improve edit summaries

cooltey:
https://phabricator.wikimedia.org/T296952


app/src/main/java/org/wikipedia/edit/summaries/EditSummaryFragment.kt
```diff
@@ -1,82 +1,179 @@
...
+                requireActivity().invalidateOptionsMenu()
+            }
+        }
+
+        binding.editSummaryTextLayout.setEndIconOnClickListener {
+            launchVoiceInput()
+        }
+
+        binding.learnMoreButton.setOnClickListener {
+            CustomTabsUtil.openInCustomTab(requireContext(), getString(R.string.preview_edit_learn_more_url))
```
dbrant, line 85: It doesn't look like CustomTabs will work for these cases. Because these are Wikipedia URLs, it still brings up the default app-chooser, and lets the user select the Wikipedia app. If they select Wikipedia, it will launch PageActivity, and will erase the current edit that the user was making.

These URLs should probably be handled by calling `visitInExternalBrowser()`.

dbrant (changes requested):
Excellent; Just a minor note.

app/src/main/java/org/wikipedia/edit/summaries/EditSummaryFragment.kt
cooltey: Makes sense. I'll let Robin know about this. Thanks!

PR #3539, opened 2022-08-23
Add info links to article call to action(s)

sharvaniharan:
Added info links to editing vies to provide guidance to users while adding article descriptions or image captions from article CTA. Also available when editing from Suggested edits.
**Phab:** https://phabricator.wikimedia.org/T315702


cooltey:
@sharvaniharan Would it be possible to move this PR to a `_design` branch so that Robin can download an auto-generated APK to do the design review?

PR #3548, opened 2022-08-29
Add info links to article call to action(s) 

sharvaniharan:
Added info links to editing vies to provide guidance to users while adding article descriptions or image captions from article CTA. Also available when editing from Suggested edits.
**Phab**: https://phabricator.wikimedia.org/T315702


app/src/main/res/layout/view_description_edit.xml
```diff
@@ -180,6 +180,23 @@
                         </com.google.android.material.textfield.TextInputLayout>
                     </FrameLayout>
 
+                    <com.google.android.material.button.MaterialButton
+                        android:id="@+id/learnMoreButton"
+                        style="@style/FlatButton.Small.Transparent"
+                        android:layout_width="match_parent"
+                        android:layout_marginStart="12dp"
```
cooltey, line 187: Can we make the start and the end margins balanced? Only applying `marginStart` will make the ripple animation look aligned to the right instead of to the middle.

cooltey (changes requested):

app/src/main/res/layout/view_description_edit.xml
sharvaniharan: Done @cooltey However, I feel just matching the margin numbers might not resolve the issue on all device sizes. It definitely looks better.

PR #3552, opened 2022-08-30
Go to the existing tab when clicking on "open in new tab"

cooltey:
https://phabricator.wikimedia.org/T311325

This PR also does some minor code cleanup.


sharvaniharan:
Lgtm.. but I was just wondering may be we should apply the same logic while opening articles from search. This below scenario looks rather weird, with same article in both tabs:
https://youtube.com/shorts/ynuJ_DLcMdI
Just food for thought or to discuss with @schoenbaechler

PR #3560, opened 2022-09-02
Export reading list into a csv file

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T316839


dbrant:
A couple of (unsolicited) notes and hints:

* Requesting permissions is only necessary for API <29. For API 29+, scoped storage will work automatically.
* Opening a file in the Downloads directory automatically can be done with `getExternalStoragePublicDirectory(...)` for <29, and for 29+ using `contentResolver.insert(...)`.
* When showing a notification after the file is written, the intent of the notification can go directly to the Downloads directory: `intent = Intent(DownloadManager.ACTION_VIEW_DOWNLOADS)`

dbrant:
@sharvaniharan Here's some rough code for creating a file in the Downloads folder using scoped storage (for API 29+):

```
val fileName = "foo.csv"

val contentResolver = context.contentResolver
val contentValues = ContentValues()
contentValues.put(MediaStore.MediaColumns.DISPLAY_NAME, fileName) // The file name is just the name, not the full path!
contentValues.put(MediaStore.MediaColumns.MIME_TYPE, "text/csv")
val uri = contentResolver.insert(MediaStore.Downloads.getContentUri(MediaStore.VOLUME_EXTERNAL_PRIMARY), contentValues)

val outputStream = contentResolver.openOutputStream(uri!!)!!
```

app/src/main/java/org/wikipedia/readinglist/ReadingListItemView.kt
```diff
@@ -182,8 +196,76 @@ class ReadingListItemView : ConstraintLayout {
...
+        val sanitizedTitle: String = StringUtil.removeUnderscores(title)
+        return if (sanitizedTitle.contains(",")) {
+            context.getString(R.string.reading_list_csv_comma_title, sanitizedTitle)
+        } else {
+            sanitizedTitle
+        }
+    }
+
+    private fun getNotificationBuilder(intent: Intent, listName: String): NotificationCompat.Builder {
+        return NotificationCompat.Builder(context, NotificationCategory.SYSTEM.id)
```
dbrant, line 251: Hint: in order to make the notification "pop" on top of the screen (instead of the just the icon appearing silently), you need to use a notification category that is not `SYSTEM`. It needs to be another category that has an importance of `IMPORTANCE_HIGH`. We can co-opt an existing category like `MENTION`, or create a new category just for these notifications.

app/src/main/java/org/wikipedia/readinglist/ReadingListItemView.kt
```diff
@@ -182,8 +196,76 @@ class ReadingListItemView : ConstraintLayout {
                     list?.let { callback?.onRemoveAllOffline(it) }
                     return true
                 }
+                R.id.menu_reading_list_export -> {
+                    AlertDialog.Builder(context)
```
dbrant, line 200: I'm not sure we need to show a dialog box before exporting the list. We can just export it.

dbrant (changes requested):

app/src/main/java/org/wikipedia/util/FileUtil.kt
```diff
@@ -19,6 +27,33 @@ object FileUtil {
         return destinationFile
     }
 
+    fun createFileInDownloadsFolder(context: Context, filename: String, stringBuilder: StringBuilder) {
+        val appExportsFolderName = WikipediaApp.instance.getString(R.string.app_name)
```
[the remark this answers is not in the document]
sharvaniharan: We do that while downloading media, so can we keep it consistent here too?

app/src/main/java/org/wikipedia/readinglist/ReadingListItemView.kt
```diff
@@ -182,8 +196,76 @@ class ReadingListItemView : ConstraintLayout {
...
+            it.pages.forEach { page ->
+                val pageTitle = ReadingListPage.toPageTitle(page)
+                val uri = pageTitle.uri
+                val language = pageTitle.wikiSite.languageCode
+                stringBuilder.append(context.getString(R.string.reading_list_csv_entry, getSanitizedPageTitle(StringUtil.removeUnderscores(pageTitle.prefixedText)), language, uri)).appendLine()
+            }
+            val intent = Intent(DownloadManager.ACTION_VIEW_DOWNLOADS)
+            FileUtil.createFileInDownloadsFolder(context, "${readingList?.listTitle}.csv", stringBuilder)
+            context.getSystemService<NotificationManager>()?.notify(id, getNotificationBuilder(intent, readingList?.listTitle!!).build())
+            FeedbackUtil.showMessage(activity, R.string.reading_list_export_completed_message)
```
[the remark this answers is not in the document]
sharvaniharan: If they turn-off notifications at a system level, they will not have any feedback after clicking export.. May be leave it until Robin weighs in?

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -54,7 +63,11 @@ abstract class BaseActivity : AppCompatActivity() {
     private var startTouchY = 0f
     private var startTouchMillis = 0L
     private var touchSlopPx = 0
+    var callback: Callback? = null
 
+    val requestPermissionLauncher = registerForActivityResult(ActivityResultContracts.RequestPermission()) { isGranted: Boolean ->
```
cooltey, line 68: Do you think it would be simpler to use  `it` instead of assigning a `Boolean` variable for it?

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -54,7 +63,11 @@ abstract class BaseActivity : AppCompatActivity() {
...
+            callback?.onPermissionResult(isGranted)
+    }
```
cooltey, line 70: Please add an empty line below this line.

app/src/main/java/org/wikipedia/readinglist/ReadingListItemView.kt
```diff
@@ -182,8 +195,68 @@ class ReadingListItemView : ConstraintLayout {
...
+            }
+            FileUtil.createFileInDownloadsFolder(context, "${readingList?.listTitle}.csv", stringBuilder)
+            val intent = Intent(DownloadManager.ACTION_VIEW_DOWNLOADS)
+            context.getSystemService<NotificationManager>()?.notify(id, getNotificationBuilder(intent, readingList?.listTitle!!).build())
+            FeedbackUtil.showMessage(activity, R.string.reading_list_export_completed_message)
+        }
+    }
+
+    private fun getSanitizedPageTitle(title: String): String {
+        val sanitizedTitle: String = StringUtil.removeUnderscores(title)
```
cooltey, line 233: The `: String` can be removed.

cooltey (changes requested):

app/src/main/java/org/wikipedia/readinglist/ReadingListItemView.kt
sharvaniharan: Have repurposed `MENTION`, but wondering if we should keep it consistent with the system notification for media download, which is not a pop, but has a toast too in combination with a silent notification?

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
sharvaniharan: I feel its more readable.

dbrant:
@sharvaniharan One of the requirements in the phab task is to export multiple lists, which means implementing "multi-select" logic in the reading lists screen. Will that be part of this PR, or a follow-on PR?

sharvaniharan:
@dbrant will be doing that in a separate PR.

PR #3561, opened 2022-09-02
Properly use custom FileProvider class.

dbrant:
According to the documentation it's bad practice to declare the base `androidx.core.content.FileProvider` in our manifest. Instead it should always be a subclass, even if the subclass doesn't do anything special.


cooltey:
Great catch!

PR #3564, opened 2022-09-02
Improve/fix usage and utilization of displayTitle in reading list pages.

dbrant:
https://phabricator.wikimedia.org/T316957

It seems that the `displayTitle` that we get from the `summary` endpoint can contain extra HTML such as `<span>` tags, which we ignore completely when displaying it. But when we want to search or sort the titles using the displayTitle, the HTML tags are (incorrectly) included in the search.

This makes sure that searching is done on the invariant title of the page, and also reduces unnecessary usage of displayTitle in queries and other logic.


cooltey (changes requested):
This change will break the logic for saving articles with different language variants. I recall the reason why we filtered with `displayTitle` because of the language variant reason.

Steps to reproduce:
1. Find any Chinese Wikipedia article, for example: Taiwan in `zh-hant`
2. Save the article to reading list
3. Change the language to `zh-hans` and save the article again.
4. Go to the reading list and you will find two saved articles are having the same title and cannot go to the correct one.

dbrant:
> This change will break the logic for saving articles with different language variants. I recall the reason why we filtered with `displayTitle` because of the language variant reason.
> 
> Steps to reproduce:
> 
> 1. Find any Chinese Wikipedia article, for example: Taiwan in `zh-hant`
> 2. Save the article to reading list
> 3. Change the language to `zh-hans` and save the article again.
> 4. Go to the reading list and you will find two saved articles are having the same title and cannot go to the correct one.

Thanks, interesting catch. In that case, we should fix the underlying root cause, instead of introducing unnecessary complexity. Does this latest update fix the issue?

cooltey:
> > This change will break the logic for saving articles with different language variants. I recall the reason why we filtered with `displayTitle` because of the language variant reason.
> > Steps to reproduce:
> > 
> > 1. Find any Chinese Wikipedia article, for example: Taiwan in `zh-hant`
> > 2. Save the article to reading list
> > 3. Change the language to `zh-hans` and save the article again.
> > 4. Go to the reading list and you will find two saved articles are having the same title and cannot go to the correct one.
> 
> Thanks, interesting catch. In that case, we should fix the underlying root cause, instead of introducing unnecessary complexity. Does this latest update fix the issue?

Yes it fixes the issue! Thanks.

PR #3584, opened 2022-09-15
Build receiving workflow for shareable reading lists

cooltey:
https://phabricator.wikimedia.org/T316834

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageActivity.kt
+++ b/app/src/main/java/org/wikipedia/page/PageActivity.kt
@@ -52,6 +52,7 @@ import org.wikipedia.staticdata.UserTalkAliasData
 import org.wikipedia.suggestededits.SuggestedEditsSnackbars
 import org.wikipedia.talk.TalkTopicsActivity
 import org.wikipedia.util.*
+import org.wikipedia.util.log.L
 import org.wikipedia.views.FrameLayoutNavMenuTriggerer
 import org.wikipedia.views.ObservableWebView
 import org.wikipedia.views.ViewUtil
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/analytics/InstallReferrerListener.kt
```diff
@@ -126,6 +127,12 @@ class InstallReferrerListener : InstallReferrerStateListener {
         if (!refChannel.isNullOrEmpty()) {
             Prefs.appChannel = refChannel
         }
+
+        // TODO: get specific parameter from the installReferrer and send a bus event
+        if (refUrl?.contains("ImportReadingLists") == true) {
```
cooltey, line 132: Somehow the `InstallReferrerListener` is difficult to test (within `adb` or manually click on the Google Play store button`. If the URL format is correct, we should be able to show this dialog in the app.

cooltey: Update:
I can test the `InstallReferrer` now with the correct steps.

Test steps:
1. Uninstall the existing production app.
2. Click the [https://play.google.com/store/apps/details?id=org.wikipedia&hl=en_US&gl=US&referrer=](https://play.google.com/store/apps/details?id=org.wikipedia&hl=en_US&gl=US&referrer=referrer_url=import_reading_lists)[referrer_url=](https://play.google.com/store/apps/details?id=org.wikipedia&hl=en_US&gl=US&referrer=referrer_url=import_reading_lists)[import_reading_lists](https://play.google.com/store/apps/details?id=org.wikipedia&hl=en_US&gl=US&referrer=referrer_url=import_reading_lists) from your device and exit the Google Play store (do not install the app from the Google Play store).
3. Install the app through the Android Studio, or through the APK.

PR #3592, opened 2022-09-21
Follow-up to WebView in link preview.

dbrant:
* Improve conversion of `ColorInt` resource to CSS color, preserving opacity.


app/src/main/java/org/wikipedia/util/ResourceUtil.kt
```diff
@@ -65,6 +65,10 @@ object ResourceUtil {
         MenuItemCompat.setIconTintList(item, getThemedColorStateList(context, colorAttr))
     }
 
+    fun colorToCssString(@ColorInt color: Int): String {
+        return String.format("%08x", (color shl 8) or ((color shr 24) and 0xff))
```
cooltey, line 69: 👍 👍 👍

PR #3599, opened 2022-09-26
Add logic to de-duplicate and fix consistency of reading lists.

dbrant:
* Switch (back?) to using the equality `==` operator for comparing `PageTitle` objects, because I'm tired of being bitten by having to use the `matches()` function for some reason.
* Add some logic to de-duplicate reading list entries, in case they were duplicated because of some previous bugs. This should happen automatically and transparently to the user.

### DIFF
--- a/app/src/main/res/values/preference_keys.xml
+++ b/app/src/main/res/values/preference_keys.xml
@@ -79,7 +79,7 @@
     <string name="preference_key_reading_lists_deleted_ids">readingListsDeletedIds</string>
     <string name="preference_key_reading_list_pages_deleted_ids">readingListPagesDeletedIds</string>
     <string name="preference_key_show_reading_lists_sync_prompt">showReadingListsSyncPrompt</string>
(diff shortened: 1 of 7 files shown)


app/src/main/res/values/preference_keys.xml
```diff
@@ -79,7 +79,7 @@
     <string name="preference_key_reading_lists_deleted_ids">readingListsDeletedIds</string>
     <string name="preference_key_reading_list_pages_deleted_ids">readingListPagesDeletedIds</string>
     <string name="preference_key_show_reading_lists_sync_prompt">showReadingListsSyncPrompt</string>
-    <string name="preference_key_reading_lists_first_time_sync">readingListsFirstTimeSync</string>
+    <string name="preference_key_reading_lists_first_time_sync">readingListsFirstTimeSyncV2</string>
```
dbrant, line 82: Clever-ish hack: Resetting the name of this "preference" will cause reading lists to be fully re-synced the next time any activity is launched (since BaseActivity checks and resets the value of this one-time parameter).

PR #3604, opened 2022-09-29
Replace smoothScrollToPosition with basic scrollToPostion in talk topic

cooltey:
https://phabricator.wikimedia.org/T318829

It will still have a scrolling animation from the top to the highlighted position, but faster.


dbrant:
My favorite kind of fix.

PR #3620, opened 2022-10-05
Export multiple lists

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T319345


sharvaniharan:
@cooltey @dbrant Please start the code review in parallel, since this might have to go to beta sooner than expected. Thank you!

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -54,6 +62,11 @@ abstract class BaseActivity : AppCompatActivity() {
     private var startTouchY = 0f
     private var startTouchMillis = 0L
     private var touchSlopPx = 0
+    var callback: Callback? = null
+
+    val requestPermissionLauncher = registerForActivityResult(ActivityResultContracts.RequestPermission()) { isGranted: Boolean ->
```
cooltey, line 67: Feel like it the `isGranted: Boolean` can be simplified to `isGranted` without the type, or we can directly use the `it`.

app/src/main/java/org/wikipedia/main/MainActivity.kt
```diff
@@ -19,10 +20,19 @@ import org.wikipedia.util.FeedbackUtil
...
+
     private lateinit var binding: ActivityMainBinding
 
     private var controlNavTabInFragment = false
     private val onboardingLauncher = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { }
+    var importCallback: Callback? = null
+    val filePickerLauncher = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) {
+        val uri: Uri = it.data?.data!!
+        importCallback?.onActivityListsImportResult(uri)
+    }
```
cooltey, line 35: Would it be possible to move the `launcher` to `ReadingListsFragment`? Not sure what's the benefit of putting the launcher in the `MainActivity`.

app/src/main/java/org/wikipedia/readinglist/ReadingListItemView.kt
```diff
@@ -64,6 +72,12 @@ class ReadingListItemView : ConstraintLayout {
                         menu.menu.findItem(R.id.menu_reading_list_rename).isVisible = false
                         menu.menu.findItem(R.id.menu_reading_list_delete).isVisible = false
                     }
+                    actionMode = callback?.onLongClick()
```
cooltey, line 75: Would it be possible to set `actionMode` when creating the view instead of setting the value through the callback?

app/src/main/java/org/wikipedia/readinglist/ReadingListsExportImportHelper.kt
```diff
@@ -0,0 +1,128 @@
...
+            .setSmallIcon(R.drawable.ic_wikipedia_w)
+            .setColor(ContextCompat.getColor(context, R.color.accent50))
+            .setStyle(NotificationCompat.BigTextStyle().bigText(context.getString(R.string.reading_list_notification_detailed_text)))
+    }
+
+    fun importLists(activity: BaseActivity, jsonString: String) {
+        funnel.logImportStart()
+        var readingLists: List<ExportableReadingList>? = null
+        try {
+            readingLists = JsonUtil.decodeFromString(jsonString)!!
```
cooltey, line 87: `!!` can be ignored.

app/src/main/java/org/wikipedia/readinglist/ReadingListsExportImportHelper.kt
```diff
@@ -0,0 +1,128 @@
...
+            extractListDataToExport(activity, lists)
+        } else {
+            FeedbackUtil.showMessage(activity, R.string.reading_list_export_write_permission_rationale)
+        }
+    }
+
+    @Suppress("unused")
+    @Serializable
+    private class ExportableReadingList
+        (val name: String? = null, val description: String? = null, val pages: Map<String, String>)
```
cooltey, line 127: Can we have the format like the following:
```
private class ExportableReadingList(
        val name: String? = null,
        val description: String? = null,
        val pages: Map<String, String>
    )
}
```

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
```diff
@@ -566,6 +718,13 @@ class ReadingListsFragment : Fragment(), SortReadingListsDialog.Callback, Readin
         }
     }
 
+    override fun onActivityListsImportResult(uri: Uri) {
+        val inputStr: InputStream = activity?.contentResolver?.openInputStream(uri)!!
+        val inputString = inputStr.bufferedReader().use { it.readText() }
+        ReadingListsExportImportHelper.importLists(activity as BaseActivity, inputString)
+        inputStr.close()
```
cooltey, line 725: Would it be possible to have a proper null check logic here?
```
val inputStr = activity?.contentResolver?.openInputStream(uri)
        inputStr?.let {
            val inputString = it.bufferedReader().use { bf -> bf.readText() }
            ReadingListsExportImportHelper.importLists(activity as BaseActivity, inputString)
            it.close()
        }
```

cooltey (changes requested):

app/src/main/java/org/wikipedia/main/MainActivity.kt
sharvaniharan: Good suggestion @cooltey and done.

app/src/main/java/org/wikipedia/readinglist/ReadingListItemView.kt
sharvaniharan: We need to check at run time, at the time when the user long presses, to see if the actionmode is Search, and decide what to do.. so if i create the view with the actionmode at the time of creation, it will not work. However, with the design changes, I might not need to hide it during search. Will confirm this with Robin and remove this logic altogether. Thank you for flagging.

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
```diff
@@ -60,12 +69,22 @@ class ReadingListsFragment : Fragment(), SortReadingListsDialog.Callback, Readin
...
     private val overflowCallback = OverflowCallback()
     private var currentSearchQuery: String? = null
     private var recentImportedReadingList: ReadingList? = null
+    private var selectMode: Boolean = false
+    private var importMode: Boolean = false
     private var shouldShowImportedSnackbar = false
 
+    val filePickerLauncher = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) {
+            if (it.resultCode == AppCompatActivity.RESULT_OK) {
+                val uri: Uri = it.data?.data!!
```
cooltey, line 83: Type specification can be ignored and it would be safe to use `it.data?.data?.let {}` here.

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
```diff
@@ -262,9 +303,24 @@ class ReadingListsFragment : Fragment(), SortReadingListsDialog.Callback, Readin
...
+            return
+        }
+        lists.forEach {
+            it.pages.forEach { page ->
+                if (page.sizeBytes == 0L) {
+                    return
+                }
+            }
+        }
+        importMode = false
```
cooltey, line 321: Do you think it is possible to replace these into this:
```
importMode = !lists.any { it.pages.any { pages -> pages.sizeBytes == 0L} }
```

cooltey (changes requested):

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
```diff
@@ -486,6 +580,108 @@ class ReadingListsFragment : Fragment(), SortReadingListsDialog.Callback, Readin
...
+                    unselectAllLists()
+                    mode.menu.findItem(R.id.menu_uncheck_all).isVisible = false
+                    mode.menu.findItem(R.id.menu_check_all).isVisible = true
+                }
+            }
+            return false
+        }
+
+        override fun onDeleteSelected() {
+            selectedLists.let {
```
cooltey, line 648: Please add logic checks that if:
1. more than 1 reading list selected.
2. "Saved" reading list is not selected. Or, you may disable the item from checkable in multiple selection mode.

app/src/main/res/values/strings.xml
```diff
@@ -548,6 +556,11 @@
     <string name="reading_list_articles_added_to_named">Added %1$d articles to %2$s</string>
     <string name="reading_list_articles_moved_to_named">Moved %1$d articles to %2$s</string>
     <string name="reading_list_delete_confirm">Are you sure you want to delete %s?</string>
+    <string name="reading_list_delete_lists_confirm_dialog_title">Delete selected list(s)?</string>
```
cooltey, line 559: Would it be better if we have this title supports plurals?

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
```diff
@@ -486,6 +580,108 @@ class ReadingListsFragment : Fragment(), SortReadingListsDialog.Callback, Readin
...
+                R.id.menu_uncheck_all -> {
+                    unselectAllLists()
+                    mode.menu.findItem(R.id.menu_uncheck_all).isVisible = false
+                    mode.menu.findItem(R.id.menu_check_all).isVisible = true
+                }
+            }
+            return false
+        }
+
+        override fun onDeleteSelected() {
```
cooltey, line 647: The currently selected items will be deselected and exit the action mode after the app shows the "Delete selected list(s)?" dialog. Would it be possible to keep the action mode and only exit the action mode after users manually close it?

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
```diff
@@ -171,12 +190,34 @@ class ReadingListsFragment : Fragment(), SortReadingListsDialog.Callback, Readin
...
         }
 
+        override fun importNewList() {
+            var filePickerIntent = Intent(Intent.ACTION_GET_CONTENT)
+            filePickerIntent.type = "application/json"
+            filePickerIntent = Intent.createChooser(filePickerIntent, getString(R.string.reading_lists_import_file_picker_title))
+            filePickerLauncher.launch(filePickerIntent)
+        }
+
+        override fun editList() {
```
cooltey, line 200: The name "edit list" does not make sense to me, because it is actually starting the multiselection mode. Would it be possible to change the name to "Select", just like what we have done in the `ReadingListFragment`?

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
```diff
@@ -653,6 +849,16 @@ class ReadingListsFragment : Fragment(), SortReadingListsDialog.Callback, Readin
         }
     }
 
+    fun onListsImportResult(uri: Uri) {
```
cooltey, line 852: Would it be possible to show a progress bar or the refresh animation while importing? (and probably the "new" indicator on the imported item).

And I also noticed that if we import the list that has the same name as the existing list, the action will not be able to complete.

cooltey (changes requested):

app/src/main/res/values/strings.xml
sharvaniharan: This was initially a plural.. then we changed to this ..I think Robin is trying to give strings like these to avoid plurals in some places where plurals aren't too important..

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
```diff
@@ -486,6 +580,108 @@ class ReadingListsFragment : Fragment(), SortReadingListsDialog.Callback, Readin
...
+            return displayedLists.let {
+                displayedLists.filterIsInstance<ReadingList>()
+                    .filter { it.selected }
+                    .onEach { it.selected = false }
+            }
+        }
+
+    private fun updateMenuItemsAlpha(alpha: Int, menu: Menu) {
+        val deleteItem = menu.findItem(R.id.menu_delete_selected)
+        val exportItem = menu.findItem(R.id.menu_export_selected)
```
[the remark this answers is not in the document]
sharvaniharan: Good catch!

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
sharvaniharan: Because we need this logic only for deletion and not for exporting, it has been coded this way. However, as we discussed on call, I will let Robin know to give copy for a toast message in cases where the "Saved" list was selected for deletion. Thank you for the catch.

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
sharvaniharan: Will discuss this with Robin as well, as he has named the menu item as "Edit" 😅 . Will change both if he agrees.

sharvaniharan:
@cooltey Updated the deletion and naming issues according to Robin's suggestion on the [task](https://phabricator.wikimedia.org/T319345#8512053).

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -44,13 +45,26 @@ import org.wikipedia.util.ResourceUtil
...
     private val networkStateReceiver = NetworkStateReceiver()
     private var previousNetworkState = WikipediaApp.instance.isOnline
     private val disposables = CompositeDisposable()
     private var currentTooltip: Balloon? = null
     private var imageZoomHelper: ImageZoomHelper? = null
 
+    private var startTouchX = 0f
+    private var startTouchY = 0f
+    private var startTouchMillis = 0L
+    private var touchSlopPx = 0
```
cooltey, line 61: Are these variables used anywhere?

app/src/main/java/org/wikipedia/readinglist/ReadingListBehaviorsUtil.kt
```diff
@@ -102,6 +102,26 @@ object ReadingListBehaviorsUtil {
...
+    fun deleteReadingLists(activity: Activity, readingLists: List<ReadingList>?, callback: Callback) {
+        if (readingLists == null) {
+            return
+        }
+        AlertDialog.Builder(activity)
+            .setTitle(R.string.reading_list_delete_lists_confirm_dialog_title)
+            .setMessage(activity.resources.getQuantityString(R.plurals.reading_list_delete_lists_confirm_dialog_message, readingLists.size, readingLists.size))
+            .setPositiveButton(R.string.reading_list_delete_lists_dialog_delete_button_text) { _, _ ->
+                readingLists.forEach {
+                    if (!it.isDefault) {
```
cooltey, line 114: This can be replaced with `.filterNot { it.isDefault }`

app/src/main/java/org/wikipedia/readinglist/ReadingListBehaviorsUtil.kt
```diff
@@ -205,6 +225,41 @@ object ReadingListBehaviorsUtil {
...
 
+    fun showDeleteListsUndoSnackbar(activity: Activity, readingLists: List<ReadingList>?, callback: SnackbarCallback) {
+        if (readingLists == null) {
+            return
+        }
+        val snackBar = FeedbackUtil.makeSnackbar(activity, getDeleteListMessage(activity, readingLists))
+        if (!(readingLists.size == 1 && readingLists[0].isDefault)) {
+            snackBar.setAction(R.string.reading_list_item_delete_undo) {
+                readingLists.forEach {
+                    if (!it.isDefault) {
```
cooltey, line 236: This can be replaced with `.filterNot { it.isDefault }`

app/src/main/java/org/wikipedia/readinglist/ReadingListBehaviorsUtil.kt
```diff
@@ -205,6 +225,41 @@ object ReadingListBehaviorsUtil {
             .show()
     }
 
+    fun showDeleteListsUndoSnackbar(activity: Activity, readingLists: List<ReadingList>?, callback: SnackbarCallback) {
+        if (readingLists == null) {
+            return
+        }
+        val snackBar = FeedbackUtil.makeSnackbar(activity, getDeleteListMessage(activity, readingLists))
+        if (!(readingLists.size == 1 && readingLists[0].isDefault)) {
```
cooltey, line 233: Would it be possible to filter the default list so that the remaining logic can be skipped?

app/src/main/java/org/wikipedia/readinglist/ReadingListItemView.kt
```diff
@@ -194,6 +210,15 @@ class ReadingListItemView : ConstraintLayout {
                     list?.let { callback?.onRemoveAllOffline(it) }
                     return true
                 }
+                R.id.menu_reading_list_export -> {
+                    list?.let { ReadingListsExportImportHelper
+                        .exportLists(context as BaseActivity, Collections.singletonList(it)) }
```
cooltey, line 215: `Collections.singletonList(it)` can be replaced with `listOf(it)`.

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
```diff
@@ -60,12 +69,22 @@ class ReadingListsFragment : Fragment(), SortReadingListsDialog.Callback, Readin
...
     private var recentImportedReadingList: ReadingList? = null
+    private var selectMode: Boolean = false
+    private var importMode: Boolean = false
     private var shouldShowImportedSnackbar = false
 
+    val filePickerLauncher = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { activityResult ->
+        if (activityResult.resultCode == AppCompatActivity.RESULT_OK) {
+            val uri = activityResult.data?.data
+            uri?.let { onListsImportResult(it) }
+        }
```
cooltey, line 85: Feels like the `activityResult` can be ignored since we are using `it` everywhere else in the app.

```
if (it.resultCode == AppCompatActivity.RESULT_OK) {
    it.data?.data?.let { uri ->
        onListsImportResult(uri)
    }
}
```

app/src/main/java/org/wikipedia/util/FileUtil.kt
```diff
@@ -19,6 +25,29 @@ object FileUtil {
...
+                contentValues.put(MediaStore.MediaColumns.DISPLAY_NAME, filename)
+                contentValues.put(MediaStore.MediaColumns.MIME_TYPE, "application/json")
+                contentResolver.insert(MediaStore.Downloads.getContentUri(MediaStore.VOLUME_EXTERNAL_PRIMARY), contentValues)?.let { uri ->
+                    contentResolver.openOutputStream(uri)?.use { it.write(data?.toByteArray()) }
+                }
+            } else {
+                val downloadsFolder = Environment.getExternalStoragePublicDirectory(Environment.DIRECTORY_DOWNLOADS)
+                val exportFile = File(downloadsFolder, filename)
+                exportFile.delete() // To overwrite when file exists
+                FileOutputStream(exportFile, true).bufferedWriter().use { writer ->
```
cooltey, line 42: Can we simplify it to `.use { it.write(data) }`?

app/src/main/res/values-qq/strings.xml
```diff
@@ -1288,17 +1288,58 @@
...
+    <item quantity="other">Toast message shown when multiple reading lists are deleted. The \"%d\" symbol is replaced with the number of the lists deleted.</item>
+  </plurals>
   <string name="shareable_reading_lists_import_dialog_title">Title text for the import reading list dialog.</string>
   <string name="shareable_reading_lists_import_dialog_content">Content text for the import reading list dialog.</string>
   <string name="shareable_reading_lists_new_indicator">Label indicating that the imported list in the list view.</string>
   <string name="shareable_reading_lists_new_install_dialog_title">Title text for the onboarding import reading list dialog.</string>
   <string name="shareable_reading_lists_new_install_dialog_content">Content text for the onboarding import reading list dialog.</string>
   <string name="shareable_reading_lists_new_install_dialog_got_it">Button label of the onboarding import reading list dialog that dismisses the dialog.</string>
+  <string name="shareable_reading_lists_new_imported_list_title">Item title of the newly imported reading list.</string>
```
cooltey, line 1329: Can we remove this one?

app/src/main/res/values-qq/strings.xml
```diff
@@ -1288,17 +1288,58 @@
...
   <string name="reading_list_share_message_v2">Message that is sent to a recipient when sharing a reading list.</string>
   <string name="reading_list_share_survey_title">Title for dialog box that leads to a survey about sharing reading lists.</string>
   <string name="reading_list_share_survey_body">Message body for dialog box that leads to a survey about sharing reading lists.</string>
+  <string name="reading_lists_import_conflict_dialog_title">Title text for the dialog shown when there is already a list with the same name as imported list.</string>
+  <string name="reading_lists_import_conflict_dialog_message">Message text for the dialog shown when there is already a list with the same name as imported list.</string>
+  <string name="reading_lists_import_conflict_dialog_primary_action_text">Primary action text for import conflict dialog.</string>
+  <string name="reading_lists_import_conflict_dialog_secondary_action_text">Secondary action text for import conflict dialog.</string>
+  <string name="reading_lists_import_conflict_dialog_neutral_action_text">Neutral action text for import conflict dialog.</string>
```
cooltey, line 1341: Would it be possible to remove these unused strings?

cooltey (changes requested):

app/src/main/java/org/wikipedia/readinglist/ReadingListBehaviorsUtil.kt
sharvaniharan: The logic here is, do not show an undo button if -> default is the only button selected for deletion. However, if it is just one of many, then undo is shown, but upon clicking undo, default should not get created again. So a general filter might not work. Lmk what you feel about it.

cooltey (changes requested):
Found an issue when entering the multiple-select mode. 

Looks like the multiple-select icon logic has some conflict behavior with single toggle behavior. 

https://www.youtube.com/shorts/7uVzgLoRu20

sharvaniharan:
Done @cooltey Lmk if it looks ok now.

cooltey:
Looks good to me now. @dbrant Please review and merge if it looks good to you too.

PR #3625, opened 2022-10-06
Fix lint warnings from non-positional parameters in strings.

dbrant:
One of our strings (`revision_diff_line_nums`) was incorrectly using ambiguous non-positional parameters. When we have more than one parameter in a string, they must use the style of `%1$d`, `%2$d`, etc.
I updated a few of the translated strings, too, and hopefully others will be updated via TWN later.


app/src/main/res/values/strings.xml
```diff
@@ -1279,6 +1279,7 @@
     <string name="revision_diff_to">To:</string>
     <string name="revision_diff_line_num">Line %d</string>
     <string name="revision_diff_line_nums">Lines %d–%d</string>
```
dbrant, line 1281: ^^ incorrect (lint warning) usage of non-positional parameters.

PR #3649, opened 2022-10-25
Correctly handle links to Special:Contributions pages.

dbrant:
A large-looking PR, but much simpler than it seems:
This creates another static map of alias keywords that represent the `Contributions` special page in all languages, and wires it up to the logic of handling intents to open pages.
This also necessarily updates the language list used in the app.

https://phabricator.wikimedia.org/T321485


app/src/main/java/org/wikipedia/page/PageTitle.kt
```diff
@@ -75,6 +76,12 @@ data class PageTitle(
             return mainPageTitle == displayText
         }
 
+    val isContributions: Boolean
+        get() {
+            val title = text.split('/')[0]
```
cooltey, line 81: Would it be better if we change the `[0]` to `.firstOrNull()` for null-safety purpose?

dbrant: Correct!

PR #3665, opened 2022-11-01
Survey for sharing reading lists.

dbrant:
https://phabricator.wikimedia.org/T322088


dbrant:
(merging into feature branch, so that everything can be revied together)

PR #3667, opened 2022-11-02
Hygiene: use getOrNull() for tabs in PageFragmentLoadState.kt

cooltey:
A really minor PR of simplifying the code.


app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
```diff
@@ -253,13 +253,11 @@ class PageFragmentLoadState(private var model: PageViewModel,
...
             // Update our tab list to prevent ZH variants issue.
-            if (app.tabList[app.tabCount - 1] != null) {
-                app.tabList[app.tabCount - 1].setBackStackPositionTitle(title)
-            }
+            app.tabList.getOrNull(app.tabCount - 1)?.setBackStackPositionTitle(title)
 
             // Save the thumbnail URL to the DB
             val pageImage = PageImage(title, pageSummary?.thumbnailUrl)
-            Completable.fromAction { AppDatabase.instance.pageImagesDao().insertPageImage(pageImage) }.subscribeOn(Schedulers.io()).subscribe()
+            disposables.add(Completable.fromAction { AppDatabase.instance.pageImagesDao().insertPageImage(pageImage) }.subscribeOn(Schedulers.io()).subscribe())
```
dbrant, line 260: I'm not sure this should be put into `disposables`, because it's totally independent of the UI.  And since it's a database write operation, it should be allowed to complete.

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
cooltey: Makes sense to me. Removed.

PR #3675, opened 2022-11-04
Update pageLoadLatency logic for upcoming network latency measurement

cooltey:
Per the discussion in the engineers' sync, we can measure the `mobile-html` load latency first and then measure the remaining API load latency later. 

Please note the logic of checking `PAGE_HTML_ENDPOINT` string because in the `mobile-html`, it will make a `page/related` API request and the `WebViewClient` will receive an event in `shouldInterceptRequest` but no event will be received in `onPageFinished`.

https://phabricator.wikimedia.org/T322142


app/src/main/java/org/wikipedia/dataclient/okhttp/OkHttpWebViewClient.kt
```diff
@@ -47,6 +47,9 @@ abstract class OkHttpWebViewClient : WebViewClient() {
         var response: WebResourceResponse
         try {
             val rsp = request(request)
+            if (rsp.networkResponse != null && rsp.cacheResponse == null && request.url.encodedPath?.contains(RestService.PAGE_HTML_ENDPOINT) == true) {
```
dbrant, line 50: I'm not sure I see how this is working. Why are we calling `pageFetchStart()` after the request has already returned?

We should call `pageFetchStart()` before the `request()` is made (1 line above), and then `pageFetchEnd()` after it's finished, i.e. right here. That's all that is needed. We shouldn't need to keep track of the URL inside the schema, or need to call `pageFetchEnd` when the page is finished _painting_ (that's not what we're measuring).

dbrant (changes requested):

app/src/main/java/org/wikipedia/dataclient/okhttp/OkHttpWebViewClient.kt
cooltey: Got it and done!

PR #3680, opened 2022-11-09
Use ViewModel and PagingSource in SearchResultsFragment

cooltey:
https://phabricator.wikimedia.org/T303933
- Use `PagingSource` in `ViewModel`.
- Remove `LRUCache` and use build-in cache in `PagingSource`.
- Simplify and optimize view loading.


app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
```diff
@@ -0,0 +1,171 @@
...
+                            }
+                        }.await()
+                    }
+                    response = ServiceFactory.get(wikiSite).prefixSearch(searchTerm, params.loadSize, params.key?.gpsoffset)
+                    prefixSearch = false
+                }
+
+                if (response?.query?.pages == null) {
+                    // Prevent using continuation string from prefix search
+                    val continuation = if (params.key?.continuation?.contains("description") == true) null else params.key?.continuation
```
dbrant, line 87: When does the continuation string contain `"description"`?

app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
```diff
@@ -0,0 +1,171 @@
...
+                    response = ServiceFactory.get(wikiSite)
+                        .fullTextSearch(searchTerm, params.key?.gsroffset?.toString(), params.loadSize, continuation)
+                }
+
+                val resultList = mutableListOf<SearchResult>()
+                addSearchResultsFromTabs(searchTerm, resultList)
```
dbrant, line 93: Adding search results from Tabs should be done only on the first call, but it looks like this is being done on every call now.

app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
```diff
@@ -0,0 +1,171 @@
...
+                    return LoadResult.Page(emptyList(), null, null)
+                }
+
+                val wikiSite = WikiSite.forLanguageCode(languageCode)
+                var response: MwQueryResponse? = null
+                var readingListSearch = SearchResults()
+                var historySearch = SearchResults()
+                if (prefixSearch) {
+                    if (searchTerm.length > 2) {
+                        readingListSearch = withContext(Dispatchers.IO) {
```
dbrant, line 69: It looks like we're doing `readingListSearch` and `historySearch` sequentially, not concurrently. Can we perform them both (along with prefixSearch) concurrently?

dbrant (changes requested):

app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
```diff
@@ -0,0 +1,171 @@
...
+            var resultsCount: MutableList<Int>?,
+            var totalResults: MutableList<SearchResult>?
+    ) : PagingSource<MwQueryResponse.Continuation, SearchResult>() {
+
+        private val batchSize = 20
+        private var prefixSearch = true
+
+        override suspend fun load(params: LoadParams<MwQueryResponse.Continuation>): LoadResult<MwQueryResponse.Continuation, SearchResult> {
+            return try {
+                // The default offset is 0 but we send the initial offset from 1 to prevent showing the same talk page from the results.
```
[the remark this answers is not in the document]
cooltey: Removed. Thanks!

app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
cooltey: Updated the comment.

The `description` string came from the `prefixsearch` API request, which happens in the first round of the `PagingSource`. It will be sent as the `nextKey`, but the next round of `PagingSource` is a `fulltextsearch`, and the `continuation` should be empty.

app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
cooltey: Makes sense. Done.

app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    class SearchResultsPagingSource(
+            val searchTerm: String?,
+            val languageCode: String?,
+            var resultsCount: MutableList<Int>?,
+            var totalResults: MutableList<SearchResult>?
+    ) : PagingSource<MwQueryResponse.Continuation, SearchResult>() {
+
+        private val batchSize = 20
```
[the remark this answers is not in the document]
cooltey: Yes!

app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
cooltey: Done!

dbrant (changes requested):
Found one issue: Search results disappear (become empty) when the device is rotated.

cooltey:
> Found one issue: Search results disappear (become empty) when the device is rotated.

Fixed!

PR #3690, opened 2022-11-15
Strip html styles when inserting links in editor.

dbrant:


dbrant:
self-merging since it's a super minor tweak.

PR #3695, opened 2022-11-19
Use DateTimeFormatter.RFC_1123_DATE_TIME.

Isira-Seneviratne:


Isira-Seneviratne:
This PR was made redundant with #3848, so closing.

PR #3698, opened 2022-11-22
Split off Reading Lists surveys into two independent surveys.

dbrant:
There is a little duplication of logic, but I believe this will make it simpler to remove when we're done.
https://phabricator.wikimedia.org/T322088

### DIFF
--- a/app/src/main/java/org/wikipedia/activity/BaseActivity.kt
+++ b/app/src/main/java/org/wikipedia/activity/BaseActivity.kt
@@ -31,6 +31,8 @@ import org.wikipedia.login.LoginActivity
 import org.wikipedia.main.MainActivity
 import org.wikipedia.page.LinkMovementMethodExt
 import org.wikipedia.readinglist.ReadingListSyncBehaviorDialogs
+import org.wikipedia.readinglist.ReadingListsReceiveSurveyHelper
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -92,7 +93,8 @@ abstract class BaseActivity : AppCompatActivity() {
         setNavigationBarColor(ResourceUtil.getThemedColor(this, R.attr.paper_color))
         maybeShowLoggedOutInBackgroundDialog()
 
-        ReadingListsSurveyHelper.maybeShowSurvey(this)
+        ReadingListsShareSurveyHelper.maybeShowSurvey(this)
```
cooltey, line 96: Somehow the dialog will show twice. 

Not sure if you are able to reproduce it, but the dialog will also show when I open an article.

cooltey (changes requested):

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
dbrant: Yeah, good observation. The logic should be all fixed now.

PR #3703, opened 2022-11-27
Update shareable reading list strings for translation

cooltey:
For the upcoming tweak PR for shareable reading list


cooltey:
self-merged PR for translation.

PR #3713, opened 2022-12-04
Improve Russian translation in reading list item statistical description

sryze:
1. Move the number of articles to the front of the text so that it may not be confused with the article size. As commas are used as decimal separators in Russian, it was not clear before if the number such as "1, 1,48 MB" indicated the size of the article (over 1 thousand megabytes) or not.
2. Translate size units to Russian as well: MB -> МБ, which stand for "МегаБайт".

### DIFF
--- a/app/src/main/res/values-ru/strings.xml
+++ b/app/src/main/res/values-ru/strings.xml
@@ -528,16 +528,16 @@
   <string name="reading_list_added_view_button">Посмотреть список</string>
   <string name="reading_list_add_to_list_button">Добавить в список</string>
   <plurals name="format_reading_list_statistical_summary">
-    <item quantity="one">Статей: %1$d, %2$.2f MB</item>
-    <item quantity="few">Статей: %1$d, %2$.2f MB</item>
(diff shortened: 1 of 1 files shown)


dbrant:
Thanks for this; However, these translated resources are auto-generated. The actual translated strings live in our translation system called TranslateWiki.net, and are periodically synchronized to the app's xml files.

Every translatable string resource is basically a wiki page that you can edit. Here are the two resources that you should modify:

https://translatewiki.net/wiki/Wikimedia:Wikipedia-android-strings-format_reading_list_statistical_summary/ru
https://translatewiki.net/wiki/Wikimedia:Wikipedia-android-strings-format_reading_list_statistical_detail/ru

After those pages are modified, they will be updated in the app's xml files the next time all the resources are synced.

sryze:
Unfortunately, I cannot modify those pages. It looks like I need to file some request, generate some SSH key, and so on - I'm sorry, but I don't have time to do that. I just wanted to submit a quick fix for the translation. You guys make it way too complex.

sryze:
Done

PR #3724, opened 2022-12-12
Show loud warning messages upon MEP errors (dev build only).

dbrant:
The original intention was for the app to crash when the MEP service returned an error. However, because the exceptions happen within an RxJava subscription, they are consumed without crashing.

Instead of crashing, let's show a large toast with the error returned by MEP, to serve as a signal that one or more MEP issues need to be investigated.


dbrant:
> Could we also keep queue size to 1 or max 10 on dev builds as well? Actually helps during development too.

Good idea -- done.

PR #3732, opened 2022-12-19
Deprecate MobileWikiAppLanguageSearching

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T324167


app/src/main/java/org/wikipedia/settings/languages/WikipediaLanguagesFragment.kt
```diff
@@ -298,6 +299,7 @@ class WikipediaLanguagesFragment : Fragment(), MenuProvider, WikipediaLanguagesI
 
     private fun deleteSelectedLanguages() {
         app.languageState.removeAppLanguageCodes(selectedCodes)
+        BreadCrumbCustomLogHelper.logAppLanguageDelete(requireActivity(), selectedCodes)
```
dbrant, line 302: Do we really need a special method for this? Won't this be tracked automatically by breadcrumbs? (`longclick` on a list item, then `click` the delete button?)

app/src/main/java/org/wikipedia/analytics/eventplatform/BreadCrumbCustomLogHelper.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    fun getCustomNameForLanguageListItem(view: View): String {
+        return WikipediaApp.instance.languageState.appLanguageCodes.toString() + "language_added:" +
+                view.findViewById<TextView>(R.id.language_subtitle).text.toString()
+    }
+
+    fun logAppLanguageDelete(context: Context, languages: List<String>) {
+        EventPlatformClient.submit(BreadCrumbLogEvent(BreadCrumbViewUtil.getReadableScreenName(context),
+            WikipediaApp.instance.languageState.appLanguageCodes.toString() + "languages_deleted:" + languages.toString()))
+    }
```
dbrant, line 34: It feels like this is getting way too specific for Breadcrumbs. The goal of breadcrumbs is to know _nothing_ about the context of the interaction that is being logged.

It looks like the old `AppLanguageSearchingFunnel` wasn't even logging deletion of languages. If we don't explicitly need this data, can we just get rid of it?

dbrant (changes requested):

app/src/main/java/org/wikipedia/settings/languages/WikipediaLanguagesFragment.kt
sharvaniharan: The delete button`menu_delete_selected` just opens the dialog... and the dialog clicks don't get caught..

app/src/main/java/org/wikipedia/analytics/eventplatform/BreadCrumbCustomLogHelper.kt
sharvaniharan: that's correct. The old schema does not log removed languages. and Done.

app/src/main/java/org/wikipedia/analytics/eventplatform/BreadCrumbCustomLogHelper.kt
```diff
@@ -0,0 +1,29 @@
...
+                // ListItemView is not in a CardView
+                return BreadCrumbViewUtil.getReadableNameForView(view.parent as RecyclerView) + "." + position
+            }
+        }
+        return currentParent.javaClass.simpleName + "." + position
+    }
+
+    fun getCustomNameForLanguageListItem(view: View): String {
+        return WikipediaApp.instance.languageState.appLanguageCodes.toString() + "language_added:" +
+                view.findViewById<TextView>(R.id.language_subtitle).text.toString()
```
dbrant, line 27: Apologies, I meant to say that _any_ reference to "language list" within Breadcrumbs is too much, and should ideally be abstracted away.

In this instance, it looks like we're trying to get the text of the list item that corresponds to a language. Is there a way for us to generalize this for getting the text of _any_ list item, in any list?

Otherwise, if it's too difficult to get arbitrary text of any list item, then do we really need this data? i.e. do we actually need the names of the specific languages that were clicked?

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/BreadCrumbCustomLogHelper.kt
sharvaniharan: The click on the language item is the only way to add the language to app language list, and that needs to be measured... and i believe what the app languages list was before adding and the code of the language added are the only important things for the schema... 
Since not all recyclerview items might have a textview in them, I thought a more custom treatment can be applied here?
One other way we can do that is to directly send a message with this info when the event happens... as in OnClick() on the item..

sharvaniharan:
Removed all the unwanted wiring like we discussed @dbrant . This is now just deprecation of the old schema.

PR #3755, opened 2023-01-12
Exclude some screens from click events

sharvaniharan:
Excluding the login screens from any event collection other than "shown"


dbrant (changes requested):
@sharvaniharan It might be a bit excessive to exclude *all* events from these activities. As we discussed, the problem is the logging of the text contents of TextViews. So then, couldn't we just no longer log the contents of a `TextView` if its inputType contains the PASSWORD flag?

sharvaniharan:
@dbrant Done.

PR #3761, opened 2023-01-14
Remove more obsolete schemas that are already in MEP

sharvaniharan:
**Phab**: https://phabricator.wikimedia.org/T324167

More details: https://docs.google.com/spreadsheets/d/1WTewfNx7zeUHrn0Dugv___O2TQ_u6IRpXgP2F0P0vb4/edit?usp=sharing


app/src/main/java/org/wikipedia/page/PageActivity.kt
```diff
@@ -568,7 +568,9 @@ class PageActivity : BaseActivity(), PageFragment.Callback, LinkPreviewDialog.Ca
             wasTransitionShown = true
         }
         if (entry.source != HistoryEntry.SOURCE_INTERNAL_LINK || !Prefs.isLinkPreviewEnabled) {
-            LinkPreviewFunnel(app, entry.source).logNavigate()
+            val articleLinkPreviewInteractionEvent = ArticleLinkPreviewInteractionEvent(pageTitle.wikiSite.dbName(),
+                pageFragment.page?.pageProperties?.pageId!!.or(0), entry.source)
```
dbrant, line 572: Incorrect usage of `or()`. The `or()` function is really just the bitwise OR operator, unrelated to null conditionals.
The intention here is probably to use the elvis operator `?:`.

dbrant (changes requested):

PR #3776, opened 2023-01-23
Add PendingIntentCompat helper class.

Isira-Seneviratne:
Add `PendingIntentCompat` helper class similar to [the one](https://github.com/androidx/androidx/blob/androidx-main/core/core/src/main/java/androidx/core/app/PendingIntentCompat.java) in AndroidX (which is not yet available).


dbrant:
We will be happy to adopt the `PendingIntentCompat` class from AndroidX, once it becomes available.

PR #3780, opened 2023-01-24
Use Java 10 URLEncoder and URLDecoder methods.

Isira-Seneviratne:
Use the new [`URLEncoder.encode(String, Charset)`](https://developer.android.com/reference/java/net/URLEncoder#encode(java.lang.String,%20java.nio.charset.Charset)) and [`URLDecoder.decode(String, Charset)`](https://developer.android.com/reference/java/net/URLDecoder#decode(java.lang.String,%20java.nio.charset.Charset)) methods introduced in Java 10 / Android API level 33.


dbrant:
This does not seem to be necessary, nor a sufficient reason to use Java 11.

PR #3781, opened 2023-01-24
Make talk page available for offline reading

cooltey:
https://phabricator.wikimedia.org/T327185

This PR also removed the unused variables and API calls, and used a proper error handler for watch and subscribe actions.


sharvaniharan:
Works very well @cooltey. Just one question. Do we have an upper limit on the offline storage for calls? Is it internal to okhttp?

sharvaniharan:
Also I am getting an error when title has a special character like this: `X-Offline-Title`seems to be the issue:
error

sharvaniharan (changes requested):
It is very smooth how it works when the user hasn't even visited the topics within a psge, but can view them offline as the calls have been saved!

cooltey:
> Works very well @cooltey. Just one question. Do we have an upper limit on the offline storage for calls? Is it internal to okhttp?

We do not have a logic for offline storage limitation since we are using the same logic for articles and reading lists offline.

cooltey:
> Also I am getting an error when title has a special character like this: `X-Offline-Title`seems to be the issue: ![error]
Fixed.

sharvaniharan:
> > Works very well @cooltey. Just one question. Do we have an upper limit on the offline storage for calls? Is it internal to okhttp?
> 
> We do not have a logic for offline storage limitation since we are using the same logic for articles and reading lists offline.

@dbrant Should we do something about this?

Other than that Lgtm, please feel free to merge.

PR #3782, opened 2023-01-24
Make layout designer work correctly with our layouts and custom views.

dbrant:
Have you noticed that the Layout Designer in Android Studio hasn't been working correctly for a while now? Me too!

This is an attempt to get our layouts working again with the Layout Designer, while also making everything clearer and a bit more efficient.

Quick summary of the relevant changes:

* Moved all the styles from `styles_light.xml` into the main `styles.xml`, as part of the main `AppTheme` style. Since the "light" theme is the default theme, it does not need to be separated into an "overlay" style, and can be prepopulated with all the default styles and colors. This will improve the efficiency of using the Light (default) theme, as well as fix the Layout Designer.
* Certain custom Views are also breaking the Layout Designer, but for slightly different reasons, e.g. initializing objects that are dependent on running on a live Android system. To make these components work in the designer, we should guard the dependent bits by using the `isInEditMode()` function.


sharvaniharan:
Looks good to me... Will leave it for @cooltey  to take a look at the stylesheet changes.

cooltey:
Noticed that for some layouts like `fragment_search.xml` or `fragment_watchlist.xml`, the layout design tool still cannot show the layout properly.

cooltey:
> > Noticed that for some layouts like `fragment_search.xml` or `fragment_watchlist.xml`, the layout design tool still cannot show the layout properly.
> 
> @cooltey Make sure to Clean & Rebuild the project (and re-sync gradle), just to make sure you're not seeing anything from cached builds. I'm able to see those fragments you mentioned correctly.

Just cleaned and rebuilt and it finally showed the design. It worked!

PR #3784, opened 2023-01-27
MEP: Wiring for App Session Event [app_session_event]

sharvaniharan:
**Phab:** : https://phabricator.wikimedia.org/T324167

- Removed the legacy schema wiring for MobileWikiAppSessions
- Added wiring for the newly created MEP schema for the same
- Updated `SessionData` to have field names matching the mep schema object


app/src/main/java/org/wikipedia/analytics/SessionData.kt
```diff
@@ -10,18 +11,24 @@ import java.util.concurrent.TimeUnit
...
-    var pagesFromRandom = 0
-    var pagesFromLangLink = 0
-    var pagesFromInternal = 0
-    var pagesFromExternal = 0
-    var pagesFromHistory = 0
-    var pagesFromReadingList = 0
-    var pagesFromBack = 0
-    var pagesWithNoDescription = 0
-    var pagesFromSuggestedEdits = 0
+    @SerialName("total_pages")
```
dbrant, line 14: The actual serialized data that gets sent seems to be missing several important things, notably `total_pages` and `page_load_latency_ms`.
Hint: we've seen in the past that the serializer for these events has some quirks, and it seems like it prefers fields to be `var` instead of `val`, even if there are IDE warnings about it.

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/SessionData.kt
sharvaniharan: Thank you for the hint @dbrant. done.

PR #3790, opened 2023-01-31
Deprecate MobileWikiAppReadingLists

sharvaniharan:
Phab : https://phabricator.wikimedia.org/T324167

Removed the legacy schema wiring for MobileWikiAppReadingLists


app/src/main/java/org/wikipedia/readinglist/AddToReadingListDialog.kt
```diff
@@ -134,8 +131,8 @@ open class AddToReadingListDialog : ExtendedBottomSheetDialogFragment() {
     }
 
     open fun logClick(savedInstanceState: Bundle?) {
-        if (savedInstanceState == null) {
-            ReadingListsFunnel().logAddClick(invokeSource)
+        if (savedInstanceState == null && activity != null) {
```
dbrant, line 134: This function is no longer called from anywhere. Do we actually still need it?

app/src/main/java/org/wikipedia/readinglist/ReadingListItemView.kt
```diff
@@ -199,6 +199,7 @@ class ReadingListItemView : ConstraintLayout {
 
     private inner class OverflowMenuClickListener constructor(private val list: ReadingList?) : PopupMenu.OnMenuItemClickListener {
         override fun onMenuItemClick(item: MenuItem): Boolean {
+            BreadCrumbLogEvent.logMenuItemSelection(context, item.title.toString())
```
dbrant, line 202: The `item.title` will give you the translated text of the menu item, which we don't want. We want the resource name of the menu item, which you can obtain by using `resources.getResourceEntryName(item.itemId)`.

app/src/main/java/org/wikipedia/views/ReadingListsOverflowView.kt
```diff
@@ -34,10 +35,12 @@ class ReadingListsOverflowView(context: Context) : FrameLayout(context) {
     init {
         binding.readingListsOverflowSelect.isVisible = ReleaseUtil.isPreBetaRelease
         binding.readingListsOverflowSortBy.setOnClickListener {
+            BreadCrumbLogEvent.logMenuItemSelection(context, "sort_lists_click")
```
dbrant, line 38: Similarly, in these events, there's no need to hard-code any names, since we know which View was clicked (`it`).
In fact, is there any reason we can't just say `BreadCrumbLogEvent.logClick(context, it)`?

app/src/main/java/org/wikipedia/analytics/eventplatform/BreadCrumbLogEvent.kt
```diff
@@ -71,5 +71,9 @@ class BreadCrumbLogEvent(
             val str = "$viewReadableName." + (view as TextView).text
             EventPlatformClient.submit(BreadCrumbLogEvent(BreadCrumbViewUtil.getReadableScreenName(context), str))
         }
+
+        fun logMenuItemSelection(context: Context, actionStr: String) {
```
dbrant, line 75: Based on earlier comments, can this become `logMenuItemSelection(context: Context, item: MenuItem)`?

dbrant (changes requested):
Needs rebase, and several comments inline.

app/src/main/java/org/wikipedia/analytics/eventplatform/BreadCrumbLogEvent.kt
sharvaniharan: Good idea @dbrant. Addressed it with a little alteration. Lmk if it looks good now.

PR #3795, opened 2023-02-03
Deprecate MobileWikiAppEdit schema

sharvaniharan:


app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -139,11 +130,9 @@ class DescriptionEditFragment : Fragment() {
 
         binding.fragmentDescriptionEditView.setLoginCallback {
             val loginIntent = LoginActivity.newIntent(requireActivity(),
-                    LoginActivity.SOURCE_EDIT, funnel.sessionToken)
+                    LoginActivity.SOURCE_EDIT, UUID.randomUUID().toString())
```
cooltey, line 133: It looks like the `EDIT_SESSION_TOKEN` in the `LoginActivity` is not actually used. Would it be possible to delete it from `LoginActivity.newIntent()`?

cooltey (changes requested):

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
sharvaniharan: Done

PR #3826, opened 2023-02-13
Shared RL: Optimize timing and display of pop-ups

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T327912


sharvaniharan:
@cooltey @dbrant This is ready for review.

app/src/main/res/values/preference_keys.xml
```diff
@@ -150,10 +150,10 @@
     <string name="preference_key_edit_line_numbers">editLineNumbers</string>
     <string name="preference_key_edit_typing_suggestions">editTypingSuggestions</string>
     <string name="preference_key_reading_lists_share_url_shorten">useUrlShortenerForSharing</string>
-    <string name="preference_key_reading_lists_share_survey_attempts">readingListShareSurveyAttempts</string>
+    <string name="preference_key_reading_lists_share_survey_dialog_shown">readingListShareSurveyAttempts</string>
```
cooltey, line 153: Should we update the string value to `readingListShareSurveyDialogShown`?

cooltey: > Should we update the string value to `readingListShareSurveyDialogShown`?

Just verified that the value should be updated since the type is different and it will cause a crash when upgrading to new version.

cooltey (changes requested):

app/src/main/res/values/preference_keys.xml
sharvaniharan: Thank you Cooltey... good catch!

PR #3830, opened 2023-02-14
API groundwork for machine-generated descriptions.

dbrant:
@sharvaniharan This adds the API layer for getting suggested descriptions.

* Notice the `requestSuggestion()` function, which makes the API request and then provides a TODO for doing something with the received suggestion(s).
* Also notice the response contains a `blp` field, which specifies whether the given article is a BLP (biography of living person). The idea will be: showing machine-suggested descriptions for BLP articles will only be available to editors with 50+ edits.

(This is mergeable now, since it's feature-flagged for pre-beta)

### DIFF
--- a/app/src/main/java/org/wikipedia/descriptions/DescriptionSuggestionService.kt
+++ b/app/src/main/java/org/wikipedia/descriptions/DescriptionSuggestionService.kt
@@ -1,16 +1,15 @@
 package org.wikipedia.descriptions
 
-import io.reactivex.rxjava3.core.Observable
 import retrofit2.http.GET
 import retrofit2.http.Query
 
 interface DescriptionSuggestionService {
     @GET("article")
-    fun getSuggestion(
+    suspend fun getSuggestion(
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/descriptions/DescriptionSuggestionService.kt
```diff
@@ -0,0 +1,18 @@
...
+import retrofit2.http.GET
+import retrofit2.http.Query
+
+interface DescriptionSuggestionService {
+    @GET("article")
+    fun getSuggestion(
+        @Query("lang") lang: String,
+        @Query("title") title: String,
+        @Query("num_beams") count: Int
+    ): Observable<DescriptionSuggestionResponse>
```
cooltey, line 13: Would it be possible to use `conroutine` instead of `RxJava`?

dbrant: yes!

PR #3831, opened 2023-02-14
Introduce periodic task to clean up stale offline Talk pages.

dbrant:
This creates a new periodic task (to run every 7 days) to clean up saved offline talk pages that are older than 7 days. It works very simply by querying the DB for offline Talk responses and cross-referencing them with the state of the filesystem.
Also includes some light refactoring of the base RecurringTask class.

https://phabricator.wikimedia.org/T328677

### DIFF
--- a/app/src/main/java/org/wikipedia/recurring/TalkOfflineCleanupTask.kt
+++ b/app/src/main/java/org/wikipedia/recurring/TalkOfflineCleanupTask.kt
@@ -1,0 +1,32 @@
+package org.wikipedia.recurring
+
+import android.content.Context
+import org.wikipedia.R
+import org.wikipedia.database.AppDatabase
+import java.io.File
+import java.util.*
+import java.util.concurrent.TimeUnit
+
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/wikipedia/recurring/TalkOfflineCleanupTask.kt
```diff
@@ -0,0 +1,32 @@
...
+            .filter {
+                (absoluteTime - File(it.path + ".0").lastModified()) > CLEANUP_MAX_AGE_MILLIS
+            }.forEach {
+                AppDatabase.instance.offlineObjectDao().deleteOfflineObject(it)
+                AppDatabase.instance.offlineObjectDao().deleteFilesForObject(it)
+            }
+    }
+
+    companion object {
+        private val CLEANUP_URL_SEARCH_KEY = "action=discussiontoolspageinfo"
```
cooltey, line 29: This could be a `const` `val`?

dbrant: yes!

PR #3832, opened 2023-02-14
Machine Generated Article Descriptions

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T327188

### DIFF
--- a/app/src/main/res/layout/view_description_edit.xml
+++ b/app/src/main/res/layout/view_description_edit.xml
@@ -180,6 +180,16 @@
                         </com.google.android.material.textfield.TextInputLayout>
                     </FrameLayout>
 
+                    <com.google.android.material.chip.Chip
+                        android:id="@+id/suggestedDescButton"
+                        style="@style/CustomChipWithIconStyle"
+                        android:layout_width="wrap_content"
+                        android:layout_height="wrap_content"
+                        android:layout_marginStart="18dp"
+                        android:layout_marginEnd="12dp"
+                        android:text="@string/description_edit_suggested_description_button_text"
+                        app:chipIcon="@drawable/ic_robot_24" />
+
                     <com.google.android.material.button.MaterialButton
                         android:id="@+id/learnMoreButton"
                         style="@style/FlatButton.Small.Transparent"
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -868,6 +868,7 @@
     <string name="description_starts_with_lowercase">Avoid starting with a lowercase letter.</string>
     <string name="description_is_in_different_language">The text appears to be in an unexpected language. The language should be %s.</string>
     <string name="description_verification_notice">Please verify your description is in %1$s. Feel free to dismiss this notification if the description you\'ve written is in %2$s.</string>
+    <string name="description_edit_suggested_description_button_text">Show suggested descriptions</string>
     <!-- /Description editing -->
 
     <!-- Description editing success -->
--- a/app/src/main/res/values-qq/strings.xml
+++ b/app/src/main/res/values-qq/strings.xml
@@ -1343,4 +1343,5 @@
   <string name="reading_list_share_survey_body">Message body for dialog box that leads to a survey about sharing reading lists.</string>
   <string name="data_opt_in_accept">Button label for accepting anonymous data collection.</string>
   <string name="data_opt_in_reject">Button label for rejecting anonymous data collection.</string>
+  <string name="description_edit_suggested_description_button_text">Suggested edits article descriptions suggestions button label.</string>
 </resources>
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -620,6 +620,25 @@
         <item name="chipMinTouchTargetSize">48dp</item>
     </style>
 
+    <style name="CustomChipWithIconStyle">
+        <item name="android:textSize">14sp</item>
+        <item name="android:fontFamily">sans-serif</item>
+        <item name="android:textColor">?attr/chart_shade1</item>
+        <item name="android:lineSpacingExtra">2sp</item>
+        <item name="ensureMinTouchTargetSize">true</item>
+        <item name="chipMinTouchTargetSize">48dp</item>
+        <item name="textAllCaps">false</item>
+        <item name="android:textStyle">bold</item>
+        <item name="chipBackgroundColor">?attr/suggestions_background_color</item>
+        <item name="chipCornerRadius">8dp</item>
+        <item name="chipIconSize">14dp</item>
+        <item name="chipIconTint">?attr/chart_shade1</item>
+        <item name="chipIconVisible">true</item>
+        <item name="iconStartPadding">12dp</item>
+        <item name="iconGravity">start</item>
+        <item name="android:elevation">1dp</item>
+    </style>
+
     <style name="AlertDialogStyle" parent="ThemeOverlay.MaterialComponents.Dialog.Alert">
         <item name="android:colorBackground">?attr/paper_color</item>
         <item name="android:textColorPrimary">?attr/material_theme_secondary_color</item>


dbrant (changes requested):
(see additional inline comments)
And a couple of design observations:

Button appears incorrect in dark mode:
image

The "X" icon appears too large (should be 24dp)
image

app/src/main/java/org/wikipedia/views/ArticleSuggestionReportDialog.kt
```diff
@@ -0,0 +1,40 @@
...
+import android.graphics.drawable.ColorDrawable
+import androidx.appcompat.app.AlertDialog
+import org.wikipedia.R
+import org.wikipedia.databinding.DialogDescriptionSuggestionReportBinding
+import org.wikipedia.util.FeedbackUtil
+import org.wikipedia.views.ArticleDescriptionsDialog.Callback
+
+class ArticleSuggestionReportDialog(context: Context, suggestion: String, callback: Callback) : AlertDialog(context) {
+
+    interface Callback {
```
cooltey, line 15: We can use `fun interface Callback` here.

app/src/main/res/values/attrs.xml
```diff
@@ -59,6 +59,8 @@
     <attr name="color_group_69" format="color"/>
     <attr name="color_group_70" format="color" />
     <attr name="action_mode_green_background" format="color"/>
+    <attr name="border" format="color"/>
```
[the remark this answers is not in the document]
sharvaniharan: This is a tiny overlap from the components work... Robin wanted me to use the new color combo for `border`, which I believe will be used by all and any borders in the app.

app/src/main/res/values/attrs.xml
```diff
@@ -59,6 +59,8 @@
...
+    <attr name="chip_background" format="color"/>
```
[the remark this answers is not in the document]
sharvaniharan: This is yet another overlap from the components work. We are trying to use the common `background` color here, but it interfered with the system `background` naming... so will change this to `app_widgets_background` - open to suggestions

sharvaniharan:
Have left a note for Robin about the dark mode colors.

app/src/main/res/values/attrs.xml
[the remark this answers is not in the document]
cooltey: I set the new color group name as `border_color`, so maybe you can set the same name and then we can merge the conflict later.

app/src/main/res/layout/view_description_edit.xml
```diff
@@ -165,6 +163,18 @@
...
                             app:counterOverflowTextAppearance="@style/CounterOverflowTextAppearance">
 
+                            <com.google.android.material.chip.Chip
+                                android:id="@+id/suggestedDescButton"
+                                style="@style/CustomChipWithIconStyle"
+                                android:layout_width="wrap_content"
+                                android:layout_height="wrap_content"
+                                android:layout_marginTop="-20dp"
+                                android:text="@string/description_edit_suggested_description_button_text"
+                                android:visibility="visible"
```
cooltey, line 173: This should be `gone` by default.

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -208,7 +209,9 @@ class DescriptionEditFragment : Fragment() {
         binding.fragmentDescriptionEditView.setEditAllowed(editingAllowed)
         binding.fragmentDescriptionEditView.updateInfoText()
 
-        if (ReleaseUtil.isPreBetaRelease && pageTitle.description.isNullOrEmpty()) {
+        if (ReleaseUtil.isPreBetaRelease && SuggestedArticleDescriptionsDialog.availableLanguages().contains(pageTitle.wikiSite.languageCode) &&
+            action == DescriptionEditActivity.Action.ADD_DESCRIPTION && pageTitle.description.isNullOrEmpty()) {
+            binding.fragmentDescriptionEditView.showSuggestedDescriptionsLoadingProgress()
```
cooltey, line 214: Would it be possible to take this logic out as a value:
```
val isSuggestionButtonEnabled= ReleaseUtil.isPreBetaRelease && SuggestedArticleDescriptionsDialog.availableLanguages().contains(pageTitle.wikiSite.languageCode) &&
            action == DescriptionEditActivity.Action.ADD_DESCRIPTION && pageTitle.description.isNullOrEmpty()
```

and use it in `updateSuggestedDescriptionsButtonVisibility()` in `DescriptionEditView,kt`? For now, the suggestion button will always be visible even if we access the article description editing from an article page.

cooltey (changes requested):

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
sharvaniharan: Done.. and we need to show suggestions even when user is coming from article .. however showing them for languages that do not have it enabled is taken care of

app/src/main/res/xml/developer_preferences.xml
```diff
@@ -392,6 +392,10 @@
             android:key="@string/preference_key_show_sequential_user_stats_tooltip"
             android:title="@string/preference_key_show_sequential_user_stats_tooltip" />
 
+        <SwitchPreferenceCompat
+            android:key="@string/preference_key_se_machine_generated_descriptions_tooltip_shown"
+            android:title="seMachineGeneratedDescriptionsTooltipShown" />
+
```
dbrant, line 398: Let's also add a developer preference for the ABC group that the user is in, since it will be very useful for testing and QA.

app/src/main/java/org/wikipedia/analytics/eventplatform/ABTest.kt
```diff
@@ -0,0 +1,66 @@
...
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
+import org.wikipedia.Constants
+import org.wikipedia.WikipediaApp
+import org.wikipedia.auth.AccountUtil
+import org.wikipedia.dataclient.ServiceFactory
+import org.wikipedia.settings.PrefsIoUtil
+import kotlin.random.Random
+
+class ABTest(private val abTestName: String, private val abTestGroupCount: Int) {
```
dbrant, line 14: I can appreciate the intention of making this a general "a/b test" class, to be reused in future tests, but most of the code here is very specific to the Machine Suggestions ABC test. Therefore, can this just be rolled into the `MachineGeneratedArticleDescriptionsAnalyticsHelper` class?

i.e. everything here can be moved to the Companion object of the `MachineGeneratedArticleDescriptionsAnalyticsHelper` class. That way, we don't even need the `machineGeneratedDescriptionsABTest` variable in the WikipediaApp class.

dbrant (changes requested):

app/src/main/java/org/wikipedia/descriptions/DescriptionEditView.kt
```diff
@@ -9,16 +10,21 @@ import android.view.LayoutInflater
...
 import org.wikipedia.language.LanguageUtil
 import org.wikipedia.mlkit.MlKitLanguageDetector
 import org.wikipedia.page.PageTitle
+import org.wikipedia.settings.Prefs
 import org.wikipedia.suggestededits.PageSummaryForEdit
 import org.wikipedia.util.*
+import org.wikipedia.views.SuggestedArticleDescriptionsDialog
 import java.util.*
 
 class DescriptionEditView : LinearLayout, MlKitLanguageDetector.Callback {
```
dbrant, line 30: Button still appears incorrect in dark mode:

image

app/src/main/java/org/wikipedia/views/SuggestedArticleDescriptionsReportDialog.kt
```diff
@@ -0,0 +1,68 @@
...
+                    suggestion, getReportReasons())
+            }
+        }
+    }
+
+    private fun getReportReasons(): List<String> {
+        val responses = mutableListOf<String>()
+
+        if (binding.notEnoughInfoCheckbox.isChecked) {
+            responses.add(binding.notEnoughInfoCheckbox.text.toString())
```
dbrant, line 49: The "report reasons" should probably be the resource names of the buttons, not the localized text. (except the "other" reason)

app/src/main/res/values/strings.xml
```diff
@@ -1064,6 +1066,15 @@
     <string name="suggested_edits_contribution_type_image_tag">Image tag</string>
     <string name="suggested_edits_contribution_current_revision_text">Your edit is currently live on Wikipedia</string>
     <string name="suggested_card_more_edits">More suggested edits</string>
+    <string name="suggested_edits_article_descriptions_dialog_title">Suggested descriptions</string>
+    <string name="suggested_edits_article_descriptions_dialog_close">Close suggested article descriptions suggestions dialog</string>
```
dbrant, line 1070: Can this just be "Close"?

app/src/main/java/org/wikipedia/dataclient/Service.kt
```diff
@@ -329,13 +329,13 @@ interface Service {
         @Query("uccontinue") uccontinue: String?
     ): Observable<MwQueryResponse>
 
-    @GET(MW_API_PREFIX + "action=query&list=usercontribs&ucprop=ids|title|timestamp|comment|size|flags|sizediff|tags")
+    @GET(MW_API_PREFIX + "action=query&list=usercontribs&ucprop=ids|title|timestamp|comment|size|flags|sizediff|tags&meta=userinfo&uiprop=groups|blockinfo|editcount|latestcontrib")
```
dbrant, line 332: This function is used for fetching lists of User Contributions, and should not be overloaded with the extra request of getting User Info in addition. We can break this out into a separate function, if there isn't an existing `userInfo` function that we can use.

dbrant (changes requested):

PR #3836, opened 2023-02-16
Fix: Overwriting of pages during RL export

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T329805

- Fixed the json structure
- Added namespace into the mix


tacsipacsi:
Non-Wikipedia pages still won’t roundtrip correctly. It’s quite an edge case, though, so it’s probably okay not to fix this; the app doesn’t even allow adding non-Wikipedia pages to reading lists, but they can get there by other means (e.g. using [Jon Robson’s gadget](https://en.wikipedia.org/wiki/User:Jdlrobson/readingList.js)).

PR #3851, opened 2023-02-23
Update undo flow from diff view in Android app

cooltey:
https://phabricator.wikimedia.org/T327232


cooltey:
The task is not feasible due to lake of API support now.

PR #3854, opened 2023-02-23
Use NetworkCallback to get the correct network connection status

cooltey:
https://phabricator.wikimedia.org/T327185#8639337


dbrant:
Superseded by (and was inspiration for) #3946.

PR #3856, opened 2023-02-25
Use view binding helper extensions in editing-related fragments.

Isira-Seneviratne:


dbrant:
Unnecessary third-party dependency.

PR #3857, opened 2023-02-25
Use Month.getDisplayName().

Isira-Seneviratne:


app/src/main/java/org/wikipedia/util/DateUtil.kt
```diff
@@ -72,8 +74,9 @@ object DateUtil {
         return getDateStringWithSkeletonPattern(date, "MMMM d")
     }
 
-    fun getMonthOnlyWithoutDayDateString(date: Date): String {
-        return getDateStringWithSkeletonPattern(date, "MMMM")
+    fun getMonthOnlyWithoutDayDateString(calendar: Calendar): String {
+        val localDateTime = LocalDateTime.ofInstant(calendar.toInstant(), ZoneId.systemDefault())
+        return localDateTime.month.getDisplayName(TextStyle.FULL, Locale.getDefault())
```
cooltey, line 79: Please explain the advantage of using these codes instead of the original ones — for example, efficiency or solving timezone issues.

cooltey (changes requested):

app/src/main/java/org/wikipedia/util/DateUtil.kt
Isira-Seneviratne: I updated the PR with some relevant changes, such as using [`MonthDay`](https://docs.oracle.com/javase/8/docs/api/java/time/MonthDay.html) in the date picker.

PR #3872, opened 2023-03-07
Take full control of `fromHtml()` transformation, and show red links in TextViews.

dbrant:
This stemmed from a potential new requirement of displaying "red links" (links to nonexistent articles) in a different color than normal links, inside a native TextView component.

But how can this be done? The platform offers a basic `fromHtml()` function which transforms a `String` into a `Spanned` object with different spans (which we use everywhere), but it doesn't allow additional customizations like:
* Processing different types of tags, beyond what the platform offers.
* Processing custom tag attributes and styles, and conditionally modifying Spans based on styles.

Well... **now it does**! 🪄


app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
```diff
@@ -0,0 +1,147 @@
...
+
+            if ("<" !in sourceStr && "&" !in sourceStr) {
+                // If the string doesn't contain any hints of HTML entities, then skip the expensive
+                // processing that fromHtml() performs.
+                return sourceStr.toSpanned()
+            }
+
+            sourceStr = sourceStr.replace("&#8206;", "\u200E")
+                .replace("&#8207;", "\u200F")
+                .replace("&amp;", "&")
```
cooltey, line 129: Would it be possible to add comments for those encoded strings `&#8206;`, `&#8207;` and `&amp`?

app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
```diff
@@ -0,0 +1,147 @@
...
+
+            // TODO: Investigate if it's necessary to inject a dummy tag at the beginning of the
+            // text, since there are reports that XmlReader ignores the first tag by default?
+            // This would become something like "<inject/>$sourceStr".parseAsHtml(...)
+            return sourceStr.parseAsHtml(HtmlCompat.FROM_HTML_MODE_LEGACY, null,
+                CustomHtmlParser(CustomTagHandler(view)))
+        }
+
+        private fun getValue(attributes: Attributes?, name: String): String? {
```
cooltey, line 138: Should we add `@Suppress("SameParameterValue")` for the `name`

app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
```diff
@@ -0,0 +1,147 @@
...
+            if (attributes != null) {
+                for (i in 0 until attributes.length) {
+                    if (name == attributes.getLocalName(i)) { return attributes.getValue(i) }
```
cooltey, line 141: The `{ }` parentheses can be removed?

cooltey (changes requested):

app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
dbrant: sure!

app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
dbrant: In the next PR, this will be used with other parameters.

PR #3874, opened 2023-03-08
Use Files methods in OfflineCacheInterceptor.

Isira-Seneviratne:
* Switch to [`desugar_jdk_libs_nio`](https://developer.android.com/studio/write/java11-nio-support-table).
* Simplify `OfflineCacheInterceptor` using the [`Files`](https://developer.android.com/reference/java/nio/file/Files) methods, which are used in the [`Path`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io.path/java.nio.file.-path/) extensions.


dbrant:
Bringing in a new or different dependency doesn't come at zero cost. What are the costs of using the `nio` version of desugaring? 
Why are there different versions of it? Does it impact performance? Does it impact final APK size? By how much?

Isira-Seneviratne:
> Why are there different versions of it?

That seems to be to limit the APIs desugared depending on the requirements, e.g. the `minimal` config limits it to the t`java.util.function` interfaces and `Optional` class.

> Does it impact performance? Does it impact final APK size? By how much?

The APK size is increased by 34 KB.

PR #3877, opened 2023-03-09
Data wiring for machine generated article descriptions

sharvaniharan:
**Phab:**https://phabricator.wikimedia.org/T329904


app/src/main/java/org/wikipedia/analytics/eventplatform/MachineGeneratedArticleDescriptionsAnalyticsHelper.kt
```diff
@@ -0,0 +1,84 @@
...
+                "ArticleDescriptionEditing.end"
+            )
+        )
+    }
+
+    fun suggestedDescriptionsButtonShown(context: Context) {
+        EventPlatformClient.submit(
+            BreadCrumbLogEvent(
+                BreadCrumbViewUtil.getReadableScreenName(context),
+                "MachineGeneratedArticleSuggestion.suggestedDescriptionsButton.shown"
```
cooltey, line 29: Not sure but feels like we can at least make the `MachineGeneratedArticleSuggestion` and `MachineGeneratedArticleSuggestionsApiResponseDetails` as `const` values so we can reduce possible issues in the future.

cooltey (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/MachineGeneratedArticleDescriptionsAnalyticsHelper.kt
sharvaniharan: @cooltey this will be the value sent to the mep schema... and will not be reused anywhere. So is it ok if we just leave it this way?

cooltey: Since the same string is used many times in this file, I think it should be better if we can set the string as `const` values so that you have a better way to control it in the future.

sharvaniharan: Done @cooltey

app/src/main/java/org/wikipedia/views/SuggestedArticleDescriptionsDialog.kt
```diff
@@ -42,6 +48,12 @@ class SuggestedArticleDescriptionsDialog(
         binding.secondSuggestionFlag.setOnClickListener {
             SuggestedArticleDescriptionsReportDialog(context, binding.firstSuggestion.text.toString()) { dismiss() }.show()
         }
+
+        setOnDismissListener {
+            if (!suggestionChosen) {
+                MachineGeneratedArticleDescriptionsAnalyticsHelper.machineGeneratedSuggestionsDialogDismissed(context)
```
cooltey, line 54: Is this expected that send breadcrumb data only if no suggestion was chosen? Feels like this `dismiss` action should be sent whenever the suggestion is selected or not.

app/src/main/java/org/wikipedia/analytics/eventplatform/MachineGeneratedArticleDescriptionsAnalyticsHelper.kt
```diff
@@ -21,4 +23,64 @@ object MachineGeneratedArticleDescriptionsAnalyticsHelper {
...
+        )
+    }
+
+    fun machineGeneratedSuggestionsDetailsLogged(context: Context, articleName: String,
+                                                 suggestionsList: List<String>, isBlp: Boolean) {
+        EventPlatformClient.submit(
+            BreadCrumbLogEvent(
+                BreadCrumbViewUtil.getReadableScreenName(context),
+                "$MACHINE_GEN_DESC_SUGGESTIONS.ApiResponseDetails.articleName:$articleName.isBlp:$isBlp.NumberOfSuggestionsOffered:${suggestionsList.size}.Suggestions:${
+                    suggestionsList.joinToString(",")
```
cooltey, line 42: Feels like we can combine the list items to a string before submitting the event.

Would it be possible to set a `val suggestions = suggestionsList.joinToString(",")` before `submit()`?

app/src/main/java/org/wikipedia/views/SuggestedArticleDescriptionsDialog.kt
sharvaniharan: The `Dismiss event needs to be sent only when the dialog is `cancelled` or dismissed because of touch outside the dislog window... but the callback is triggered even when the user chooses a suggestion, which it shouldn't so excluding only that condition from the callback

cooltey: I see, thanks! I think the word "DialogDismissed" looks a little bit confusing to me, not sure if there has any other better word to describe it?

sharvaniharan: `OptedOut` it is !

PR #3879, opened 2023-03-10
Analytics wiring for ReadingListSharing

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T331567


app/src/main/java/org/wikipedia/readinglist/ReadingListActivity.kt
```diff
@@ -29,6 +32,13 @@ class ReadingListActivity : SingleFragmentActivity<ReadingListFragment>() {
         setNavigationBarColor(getThemedColor(this, R.attr.paper_color))
     }
 
+    override fun onBackPressed() {
+        super.onBackPressed()
+        if (isPreview) {
```
dbrant, line 37: You can do the check for `intent.getBooleanExtra(...)` right here, and not need to keep it in a class field.

dbrant (changes requested):
Looks just fine; one minor nit.

app/src/main/java/org/wikipedia/readinglist/ReadingListActivity.kt
sharvaniharan: Done @dbrant

PR #3893, opened 2023-03-18
ABC test logic for machine generated article descriptions

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T329905

Test name : mBart25
Number of groups: 3
Random but not equal test groups:
Group 1: Of the people who qualify to see [Wikis falling into the available languages], This group will not be shown the suggestions
Group2: They will be shown suggestions only if they are not articles that are biographies of living persons [BLPs]
Group3: They will see both BLP and non-BLP suggestions.


sharvaniharan:
On hold just to clarify some suggestions from Jaz. Please feel free to do a cursory review.

sharvaniharan:
Jazmin has confirmed that the group2 users will remain in that group even if they happen to cross the 50 edit mark during the experiment and will not be bumped up

app/src/main/java/org/wikipedia/analytics/eventplatform/ABTest.kt
```diff
@@ -0,0 +1,67 @@
...
+                val homeSiteResponse = async { ServiceFactory.get(WikipediaApp.instance.wikiSite)
+                    .getUserContrib(AccountUtil.userName!!, 10) }
+                val commonsResponse = async { ServiceFactory.get(Constants.commonsWikiSite)
+                    .getUserContrib(AccountUtil.userName!!, 10) }
+                val wikidataResponse = async { ServiceFactory.get(Constants.wikidataWikiSite)
+                    .getUserContrib(AccountUtil.userName!!, 10) }
+
+                    totalContributions += homeSiteResponse.await().query?.userInfo!!.editCount
+                    totalContributions += commonsResponse.await().query?.userInfo!!.editCount
+                    totalContributions += wikidataResponse.await().query?.userInfo!!.editCount
```
cooltey, line 47: Can we have a `null`-safe condition here? `.query?.userInfo?.editCount ?: 0`

app/src/main/java/org/wikipedia/analytics/eventplatform/ABTest.kt
```diff
@@ -0,0 +1,67 @@
...
+                }
+            }
+        }
+    }
+
+    companion object {
+        private const val AB_TEST_KEY_PREFIX = "ab_test_"
+        const val MACHINE_GEN_DESC = "mBART25"
+        const val EXP_CONTRIBUTOR_REQ = 50
+        const val GROUP_SIZE_2 = 2
```
cooltey, line 61: Unused value?

app/src/main/java/org/wikipedia/analytics/eventplatform/ABTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        const val GROUP_SIZE_3 = 3
+        const val GROUP_1 = 0
+        const val GROUP_2 = 1
```
cooltey, line 64: Unused value?

cooltey (changes requested):

cooltey (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/ABTest.kt
```diff
@@ -0,0 +1,66 @@
...
+                    }.also {
+                        if (it) {
+                            group = GROUP_3
+                        }
+                        PrefsIoUtil.setInt(AB_TEST_KEY_PREFIX + abTestName, group)
+                        return group
+                    }
+                }
+            }
+            PrefsIoUtil.setInt(AB_TEST_KEY_PREFIX + abTestName, group)
```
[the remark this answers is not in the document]
sharvaniharan: Good catch .. thanks 😅

PR #3895, opened 2023-03-20
Improve relative time formatting in the notification activity.

Isira-Seneviratne:
Improve the relative time formatting used in the notification activity for when the time difference is 1 week or greater.

**Screenshots**

Before
Screenshot_20230320_084247_Wikipedia

After
Screenshot_20230320_084923_Wikipedia Beta


dbrant (changes requested):
This is an improvement, but might actually be too much of an improvement.
I'm not sure the "years ago" and "months ago" cases are really needed.
* If a notification was received "1 year ago", does it mean "1 year + 1 day", or does it mean "1 year + 364 days" ago? These are huge differences, and therefore no longer very useful.
* The "months" case will never be totally accurate because not all months have the same number of days.

Let's keep the maximum granularity as `weeks`, and if the time is greater than 10 weeks in the past, keep displaying it as the regular formatted date (as before).
Also, even in the `weeks` case, the `DateUtils.getRelativeTimeSpanString()` function can format it for us automatically, using the `WEEK_IN_MILLIS` constant, so we shouldn't need any new plurals resources.

PR #3900, opened 2023-03-22
Onboarding : Machine-assisted article descriptions

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T331272

- Open up the onboarding screens to **all** desc editing, not just when entering via an article.
- The 3rd screen on Description editing onboarding is shown only to users who will see the suggestions among the AB groups.


sharvaniharan:
This is ready @cooltey @dbrant

app/src/main/java/org/wikipedia/descriptions/DescriptionEditActivity.kt
```diff
@@ -2,29 +2,43 @@ package org.wikipedia.descriptions
...
+        val action = intent.getSerializableExtra(Constants.INTENT_EXTRA_ACTION) as Action
+        if (action == Action.ADD_DESCRIPTION && Prefs.isDescriptionEditTutorialEnabled) {
+            Prefs.isDescriptionEditTutorialEnabled = false
+            val pageTitle = intent.getParcelableExtra<PageTitle>(EXTRA_TITLE)!!
+            val shouldShowAIOnBoarding = ReleaseUtil.isPreBetaRelease && SuggestedArticleDescriptionsDialog.availableLanguages()
+                .contains(pageTitle.wikiSite.languageCode) && pageTitle.description.isNullOrEmpty() &&
+                    WikipediaApp.instance.machineGeneratedDescriptionsABTest.aBTestGroup != GROUP_1
+            startActivity(DescriptionEditTutorialActivity.newIntent(this, shouldShowAIOnBoarding))
+        }
+    }
```
cooltey, line 41: Please add an empty line here.

app/src/main/java/org/wikipedia/descriptions/DescriptionEditTutorialFragment.kt
```diff
@@ -42,9 +43,10 @@ class DescriptionEditTutorialFragment : OnboardingFragment() {
     companion object {
         const val ARG_POSITION = "position"
         val pages = arrayOf(R.layout.inflate_description_edit_tutorial_page_one, R.layout.inflate_description_edit_tutorial_page_two)
+        val AiIncludedPages = arrayOf(R.layout.inflate_description_edit_tutorial_page_one, R.layout.inflate_description_edit_tutorial_page_two, R.layout.inflate_description_edit_tutorial_page_three)
```
cooltey, line 46: Would it be possible to make the `pages` includes all layouts so we can handle the logic only in `getItemCount()`?

```
       override fun getItemCount(): Int {
            return if (showAIOnBoarding) pages.size else pages.size - 1
        }
```

cooltey (changes requested):

app/src/main/java/org/wikipedia/descriptions/DescriptionEditTutorialFragment.kt
sharvaniharan: Good idea!

cooltey (changes requested):
Looks like the size of the image is not correct compares to the one in ticket.

cooltey (changes requested):
Found another issue: if the user change device orientation while view onboarding screen, the app will crash.

sharvaniharan:
Done @cooltey

PR #3902, opened 2023-03-23
Follow-up: Machine generate article description ABC test

cooltey:
Initialize the test group while creating the `ABTest()` instead of dynamically fetching data when calling the `aBTestGroup` variable.


sharvaniharan:
@cooltey This changes the experiment from "Of the people who access this mBart25 [25 wikis only] feature particularly to add descriptions, divide them to 3 groups" to " Divide anyone who accesses the app to 3 groups",  and will alter the experiment. Also, those who are not logged in or have freshly downloaded the app, will never get put into the third group. So it has to be dynamic.. 
However, I like that running the entire thing blockingly makes it look cleaner.. so may be just make that change and let it be Dynamic?

cooltey:
> @cooltey This changes the experiment from "Of the people who access this mBart25 [25 wikis only] feature particularly to add descriptions, divide them to 3 groups" to " Divide anyone who accesses the app to 3 groups", and will alter the experiment. Also, those who are not logged in or have freshly downloaded the app, will never get put into the third group. So it has to be dynamic.. However, I like that running the entire thing blockingly makes it look cleaner.. so may be just make that change and let it be Dynamic?

Ah! You are correct! I'll close this PR. Thanks!

PR #3905, opened 2023-03-24
Onboarding data wiring and clean-up

sharvaniharan:
-Additional data wiring


app/src/main/java/org/wikipedia/analytics/eventplatform/MachineGeneratedArticleDescriptionsAnalyticsHelper.kt
```diff
@@ -24,11 +25,20 @@ object MachineGeneratedArticleDescriptionsAnalyticsHelper {
         )
     }
 
+    fun logActualPublishedDescription(context: Context, publishedDescription: String, wikidataEdit: Boolean = false) {
+        EventPlatformClient.submit(
+            BreadCrumbLogEvent(
+                BreadCrumbViewUtil.getReadableScreenName(context),
+                "$MACHINE_GEN_DESC_SUGGESTIONS.publishedTo.${if (wikidataEdit) "wikidata" else "articleOnWiki"}publishedDescription.$publishedDescription.published"
```
dbrant, line 32: A couple of questions:
* Instead of logging "wikidata" versus "articleOnWiki", can we just log the wiki language code of the article? We know that only `en` articles are written to the wiki, and we'll get the bonus data of the language code.
* In this string I see three separate mentions of `publishedTo`, `publishedDescription`, and `published`. Do we need all three?
* Any reason not to include the article title in the event?

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -321,6 +321,7 @@ class DescriptionEditFragment : Fragment() {
                                     AnonymousNotificationHelper.onEditSubmitted()
                                     waitForUpdatedRevision(newRevId)
                                     EditAttemptStepEvent.logSaveSuccess(pageTitle, EditAttemptStepEvent.INTERFACE_OTHER)
+                                    MachineGeneratedArticleDescriptionsAnalyticsHelper.logActualPublishedDescription(requireContext(), binding.fragmentDescriptionEditView.description.orEmpty())
```
dbrant, line 324: It looks like these events will be sent in all cases of description editing, not just adding machine-generated descriptions. Is that intentional?

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/MachineGeneratedArticleDescriptionsAnalyticsHelper.kt
sharvaniharan: Good idea @dbrant .. which also made me look into the fact that Shay will not be receiving article wiki anywhere in the event, just the primary wiki of the app. So added that information as well into the events that needed the info. Also named the `publish` related infos better to remove confusion.

PR #3906, opened 2023-03-25
Reduce dev data queue size

sharvaniharan:
I find it easier to check events as they come in during data wiring. Sometimes queueing up 4 of similar events is confusing. So reducing it to 1 to see it as they come.


app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
```diff
@@ -111,7 +111,7 @@ object EventPlatformClient {
          */
         private const val WAIT_MS = 30000L
         private const val TOKEN = "sendScheduled"
-        private val MAX_QUEUE_SIZE = if (ReleaseUtil.isDevRelease) 4 else 128
+        private val MAX_QUEUE_SIZE = if (ReleaseUtil.isDevRelease) 1 else 128
```
dbrant, line 114: What do you think about making this number into a Preference (defaulted to 128), and exposing it in Developer prefs, so that we could set it to anything we like?

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
sharvaniharan: Good idea.. and done

PR #3914, opened 2023-03-29
Onboarding: Machine generated article descriptions

sharvaniharan:
**Phab**: https://phabricator.wikimedia.org/T331272

Open up the onboarding screens to all desc editing, not just when entering via an article.
The 3rd screen on Description editing onboarding is shown only to users who will see the suggestions among the AB groups.


dbrant:
👌  will follow up shortly.

PR #3916, opened 2023-03-29
Follow-up to Machine Suggestions onboarding.

dbrant:
* Greatly simplify API call for getting total user edits (by making a call to `globalUserInfo`, which already exists in our Service class!)
* Clean up just a few bits.


sharvaniharan:
API change is great! Thank you!

PR #3919, opened 2023-03-30
Randomize the suggestions display order

sharvaniharan:
Api doesn't scramble the suggestion order as required by the experiment, so randomizing the display order here.


dbrant:
It looks like this is already part of #3905, so closing.

PR #3931, opened 2023-04-03
Adding timeSpent to end event

sharvaniharan:
Adding this back just in case @JTannerWMF  wants to get an estimate on timeSpent for all edits: published and unpublished.

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/MachineGeneratedArticleDescriptionsAnalyticsHelper.kt
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/MachineGeneratedArticleDescriptionsAnalyticsHelper.kt
@@ -23,7 +23,7 @@ class MachineGeneratedArticleDescriptionsAnalyticsHelper {
     }
 
     fun articleDescriptionEditingEnd(context: Context) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/analytics/eventplatform/MachineGeneratedArticleDescriptionsAnalyticsHelper.kt
```diff
@@ -23,7 +23,7 @@ class MachineGeneratedArticleDescriptionsAnalyticsHelper {
     }
 
     fun articleDescriptionEditingEnd(context: Context) {
-        log(context, composeGroupString() + ".end")
+        log(context, composeGroupString() + ".timeSpentMs:${timer.elapsedMillis}.end")
```
dbrant, line 26: Can `timeSpentMs` come at the end of the string, so it's consistent with the other places it's sent?

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/MachineGeneratedArticleDescriptionsAnalyticsHelper.kt
sharvaniharan: Done @dbrant

PR #3932, opened 2023-04-03
Only logged in users should be included in the experiment

sharvaniharan:
For the below scenario:

1. User is not logged in
2. Stumbles upon an article without a description
3. Enters the edit description screen

User will get put into group 0 - no suggestions or 1 - non-blp - and never in 2, because without them being logged in, we cannot find their edit count.

Since this alters the experiment, @JTannerWMF wants the gate to be tightened to logged in users.

### DIFF
--- a/app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
+++ b/app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
@@ -44,7 +44,6 @@ import org.wikipedia.suggestededits.SuggestedEditsSurvey
 import org.wikipedia.suggestededits.SuggestionsActivity
 import org.wikipedia.util.*
 import org.wikipedia.util.log.L
-import org.wikipedia.views.SuggestedArticleDescriptionsDialog
 import java.io.IOException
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -221,7 +221,7 @@ class DescriptionEditFragment : Fragment() {
         binding.fragmentDescriptionEditView.updateInfoText()
 
         binding.fragmentDescriptionEditView.isSuggestionButtonEnabled =
-                SuggestedArticleDescriptionsDialog.availableLanguages.contains(pageTitle.wikiSite.languageCode) &&
+                AccountUtil.isLoggedIn && SuggestedArticleDescriptionsDialog.availableLanguages.contains(pageTitle.wikiSite.languageCode) &&
```
dbrant, line 224: Couldn't all these conditions be replaced with `MachineGeneratedArticleDescriptionsAnalyticsHelper.isUserInExperiment`?

dbrant (changes requested):

PR #3933, opened 2023-04-03
Fix: Show promise text at the correct place

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T331272#8751536

### DIFF
--- a/app/src/main/java/org/wikipedia/descriptions/DescriptionEditTutorialFragment.kt
+++ b/app/src/main/java/org/wikipedia/descriptions/DescriptionEditTutorialFragment.kt
@@ -11,6 +11,7 @@ import org.wikipedia.R
 import org.wikipedia.analytics.eventplatform.MachineGeneratedArticleDescriptionsAnalyticsHelper
 import org.wikipedia.onboarding.OnboardingFragment
 import org.wikipedia.onboarding.OnboardingPageView
+import org.wikipedia.views.GoneIfEmptyTextView
 
 class DescriptionEditTutorialFragment : OnboardingFragment() {
     override val doneButtonText = R.string.description_edit_tutorial_button_label_start_editing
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/onboarding/OnboardingPageView.kt
```diff
@@ -81,6 +81,10 @@ class OnboardingPageView constructor(context: Context, attrs: AttributeSet? = nu
         }
     }
 
+    fun hideTertiaryTextView(isVisible: Boolean) {
```
dbrant, line 84: This function seems misnamed. I would expect `hideTertiaryTextView(true)` to hide the view, not show it. It should probably be something like `setTertiaryTextViewVisible(...)`

app/src/main/java/org/wikipedia/descriptions/DescriptionEditTutorialFragment.kt
```diff
@@ -32,15 +32,20 @@ class DescriptionEditTutorialFragment : OnboardingFragment() {
         }
 
         override fun createFragment(position: Int): Fragment {
-            return ItemFragment().apply { arguments = bundleOf(ARG_POSITION to position) }
+            return ItemFragment(showAIOnBoarding).apply { arguments = bundleOf(ARG_POSITION to position) }
         }
     }
 
-    class ItemFragment : Fragment() {
+    class ItemFragment(private val showAIOnBoarding: Boolean) : Fragment() {
```
dbrant, line 39: Incorrect usage of fragment parameters. Parameters must be passed into a fragment using "arguments" which are given to a fragment using a "Bundle". (similar to how `position` is passed into the fragment, above)

If you do not provide a no-argument constructor, the app will crash when recreating the fragment, e.g. upon screen rotation, as you might have observed while testing.

dbrant (changes requested):

app/src/main/java/org/wikipedia/descriptions/DescriptionEditTutorialFragment.kt
sharvaniharan: Used the parent fragment argument, since it is not position specific..

PR #3949, opened 2023-04-11
Fix/update snackbar styles.

dbrant:
This is to follow up on the big color consolidation, and it looks like Snackbars got a bit broken, showing slightly incorrect colors and typeface.
This brings our Snackbars back to their correct state before the consolidation.


app/src/main/res/values/styles.xml
```diff
@@ -186,6 +176,16 @@
         <item name="android:textColor">?attr/primary_color</item>
     </style>
 
+    <style name="SnackbarTextView">
+        <item name="android:textColor">@color/gray200</item>
```
dbrant, line 180: ^ note that we need to use a specific color here (instead of a styled color definition) because Snackbars always have a dark background, regardless of light or dark theme.

PR #3952, opened 2023-04-11
Greatly simplify and optimize Toolbar layouts.

dbrant:
(This is following up on the consolidation of colors and typeface styles).

This refactors our usage of Toolbars to be the way they were meant to be used: by feeding into the activity's default ActionBar and setting the Title through that mechanism, instead of using a custom TextView for the title.
Observe **how much unnecessary code** was removed from the layouts.


app/src/main/java/org/wikipedia/page/edithistory/EditHistoryListActivity.kt
```diff
@@ -72,8 +72,8 @@ class EditHistoryListActivity : BaseActivity() {
         setContentView(binding.root)
         setSupportActionBar(binding.toolbar)
 
-        binding.articleTitleView.isVisible = false
```
sharvaniharan, line 75: This is brilliant!

PR #3953, opened 2023-04-11
Adding feature specific translations

sharvaniharan:
This is still `wip` as we are waiting for some translations. However, please feel free to start review on the ones updated. Use the translation doc for verification.


sharvaniharan:
@dbrant Please verify that the correct lang codes have been used. I have added simplified Chinese strings to `zh`, since we do not have `zh-hans`

PR #3972, opened 2023-04-20
Revert "Use Files methods in OfflineCacheInterceptor. (#3874)"

dbrant:
This reverts #3874 by @Isira-Seneviratne, since it's causing sporadic crashes on a subset of devices in our Beta track.
The `desugar_jdk_libs_nio` library is _not mature enough_ for us to use. We do not need it.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -182,7 +182,7 @@ dependencies {
     String espressoVersion = '3.5.1'
     String serialization_version = '1.4.0'
 
-    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs_nio:2.0.2'
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:2.0.3'
 
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinCoroutinesVersion"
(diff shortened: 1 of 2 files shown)


Isira-Seneviratne:
That's strange, I tested the changeset on a couple of devices (Android 5 and 13) and it worked without any issues.

dbrant:
Here are a couple of the relevant stack traces:

```
Exception java.lang.NoSuchFieldError:
  at j$.desugar.sun.nio.fs.DesugarLinuxFileSystemProvider.createDirectory (SourceFile)
  at j$.nio.file.Files.createDirectory (SourceFile)
  at j$.nio.file.Files.createDirectories (SourceFile)
  at org.wikipedia.dataclient.okhttp.OfflineCacheInterceptor.getCacheWritingResponse (OfflineCacheInterceptor.kt:99)
  ...
```

```
Exception java.lang.NoSuchFieldError:
  at j$.desugar.sun.nio.fs.DesugarLinuxFileSystemProvider.checkAccess (SourceFile)
  at j$.nio.file.Files.exists (SourceFile)
  at org.wikipedia.dataclient.okhttp.OfflineCacheInterceptor.intercept (OfflineCacheInterceptor.kt:65)
  ...
```

Isira-Seneviratne:
Thank you for the stack traces. I did a bit of digging and it seems a field in [`NoSuchFileException`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/NoSuchFileException.html) is being removed by R8 in full mode, as the issue did not occur on AGP 7.4.2.

PR #3977, opened 2023-04-25
Watchlist add/remove events

sharvaniharan:


app/src/main/java/org/wikipedia/analytics/eventplatform/WatchlistAnalyticsHelper.kt
```diff
@@ -0,0 +1,51 @@
+package org.wikipedia.analytics.eventplatform
+
+import android.content.Context
+import org.wikipedia.WikipediaApp
+
+object WatchlistAnalyticsHelper {
+    fun logAddedToWatchlist(context: Context, itemWiki: String) {
```
dbrant, line 7: The `context` parameter for most of these calls is always `WikipediaApp.instance`, so it can become an optional parameter.

app/src/main/java/org/wikipedia/analytics/eventplatform/WatchlistAnalyticsHelper.kt
```diff
@@ -0,0 +1,51 @@
...
+        EventPlatformClient.submit(
+            BreadCrumbLogEvent(
+                BreadCrumbViewUtil.getReadableScreenName(context),
+                "Watchlist.itemWiki:$itemWiki.added"
+            )
+        )
+    }
+
+    fun logRemovedFromWatchlist(context: Context, itemWiki: String) {
```
dbrant, line 16: Instead of an `itemWiki: String` parameter for all of these functions, a slightly cleaner way is to take a `title: PageTitle?` parameter. The caller doesn't need to know which parts of `PageTitle` will be used here.

app/src/main/java/org/wikipedia/analytics/eventplatform/WatchlistAnalyticsHelper.kt
```diff
@@ -0,0 +1,51 @@
...
+    fun logRemovedFromWatchlistSuccess(context: Context, itemWiki: String) {
+        EventPlatformClient.submit(
+            BreadCrumbLogEvent(
+                BreadCrumbViewUtil.getReadableScreenName(context),
+                "Watchlist.itemWiki:$itemWiki.removeSuccess"
+            )
+        )
+    }
+
+    fun logFilterSelection(context: Context, excludedWikiCodes: MutableSet<String>, includedTypeCodes: MutableSet<String>) {
```
dbrant, line 43: These can be `Set` instead of `MutableSet`.

dbrant (changes requested):
OK, we're _really_ starting to abuse Breadcrumbs 😆 .
Can this be the last time we wire it up this way, until we either migrate to MP, or use individual schemas?

app/src/main/java/org/wikipedia/analytics/eventplatform/WatchlistAnalyticsHelper.kt
sharvaniharan: The dynamic context parameter catches the screen name correctly in case of activity from PageFragment @dbrant .. this will help distinguish between `add` actions

PR #3980, opened 2023-04-26
Change label of reply button from "Publish" to "Reply"

edg2s:
**phab**: https://phabricator.wikimedia.org/T334356


cooltey:
This is no longer valid since we have changed the flow of replying/publishing a new reply/topic.

edg2s:
There is now a "Next" button, followed by "Publish"....

edg2s:
Declined by https://phabricator.wikimedia.org/T334356#9775299

PR #3982, opened 2023-04-27
Integrate Metrics Platform client library

cjming:
This PR proposes to integrate the [Metrics Platform Java Client](https://github.com/wikimedia/mediawiki-libs-metrics-platform/tree/master/java/src/main/java/org/wikimedia/metrics_platform) library for submitting analytics events.

As a preliminary test to see if events submitted to the [Modern Event Platform](https://wikitech.wikimedia.org/wiki/Event_Platform) match events submitted to the [Metrics Platform](https://wikitech.wikimedia.org/wiki/Metrics_Platform), initial candidates for Metrics Platform events include article related events:

- ArticleFindInPageInteractionEvent
- ArticleInteractionEvent
- ArticleLinkPreviewInteractionEvent
- ArticleTocInteractionEvent

Bug: [T330355](https://phabricator.wikimedia.org/T330355)


app/src/main/java/org/wikipedia/analytics/metricsplatform/ArticleEvent.kt
```diff
@@ -0,0 +1,238 @@
...
+        }
+    }
+
+    private fun submitEvent(
+        eventName: String,
+        fragment: PageFragment,
+        customData: Map<String, Any>
+    ) {
+        val clientMetadata = AndroidPageClientMetadata(fragment)
+        val metricsPlatformClient = MetricsPlatformClientPage(clientMetadata)
```
dbrant, line 234: It seems like a `MetricsClient` object is something that should persist for the lifetime of the app process, instead of being recreated upon sending each event. Should this object live as a field in the `WikipediaApp` class?

dbrant (changes requested):
This is terrific! Thanks for getting this started. I know it's still WIP, but I thought I'd note a few random thoughts while looking through the code. Looking forward to discussing further soon.

app/src/main/java/org/wikipedia/analytics/metricsplatform/ArticleEvent.kt
cjming: The issue is that page-specific metadata needs to be passed into the `MetricsClient` constructor which is why you'll notice that there are 2 ClientMetadata classes (the page one inherits from the base one). Given that page metadata is dynamic/variable, this was one approach @sharvaniharan and I talked through -- one of the topics i'm hoping to broach in our discussion

build.gradle
```diff
@@ -19,5 +19,6 @@ allprojects {
         mavenCentral()
         mavenLocal()
         maven { url "https://jitpack.io" }
+        maven { url 'https://repo1.maven.org/maven2' }
```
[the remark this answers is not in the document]
cjming: you're right! removing on next commit

app/src/main/java/org/wikipedia/analytics/metricsplatform/ArticleEvent.kt
cjming: one thing i'm considering is making a provision in the MP library to update stuff like page metadata -- that might be the better solution

cjming: the more i think about it, the more it makes sense to provide a mechanism in the library to update metadata -- and keep a single `MetricsClient` object for the lifetime of an app session -- that will simplify things greatly on this end

app/src/main/java/org/wikipedia/analytics/metricsplatform/ApplicationData.kt
```diff
@@ -0,0 +1,15 @@
+package org.wikipedia.analytics.metricsplatform
+
+import org.wikipedia.WikipediaApp
+
+object ApplicationData {
+
+    val data = mapOf(
```
dbrant, line 7: This field is statically initialized with the values of the app's theme and language codes at the time of initialization, when in fact those values can change throughout the lifetime of the app. (meaning that this field won't always contain the current values.)

To make this field _dynamically_ return a fresh map every time, we can use the Kotlin syntax of a getter method:
```
val data get() = mapOf(...
```

app/src/main/java/org/wikipedia/analytics/metricsplatform/ApplicationData.kt
```diff
@@ -0,0 +1,15 @@
...
+        "app_theme" to WikipediaApp.instance.currentTheme,
+        "language_groups" to WikipediaApp.instance.languageState.appLanguageCodes
+    )
+
+    fun getApplicationData(): Map<String, Any> {
```
dbrant, line 12: In Kotlin it's no longer a convention to have a separate getter method that just returns a field. The `data` field itself can be referenced as `ApplicationData.data`.

app/src/main/java/org/wikipedia/analytics/metricsplatform/ArticleEvent.kt
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    /**
+     * Creates a new read-only map by replacing or adding entries to this map from another map.
+     *
+     * The returned map preserves the entry iteration order of the original map.
+     * Those entries of another map that are missing in this map are iterated
+     * in the end in the order of that map.
+     */
+    public operator fun <K, V> Map<out K, V>.plus(map: Map<out K, V>): Map<K, V> =
```
dbrant, line 314: The plus operator for two maps already exists in the standard library. We don't need to define it ourselves.

app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
```diff
@@ -0,0 +1,133 @@
...
+        null,
+        WikipediaApp.instance.versionCode.toString(),
+        ReleaseUtil.isProdRelease,
+        ReleaseUtil.isDevRelease,
+        WikipediaApp.instance.wikiSite.dbName(),
+        WikipediaApp.instance.languageState.systemLanguageCode,
+        WikipediaApp.instance.languageState.appLanguageLocalizedNames
+    )
+
+    private val performerData: PerformerData = PerformerData(
```
dbrant, line 31: The `performerData` and `mediawikiData` objects contain items that can easily change at runtime, such as `userName`, `isLoggedIn`, `wikiSite`, `appLanguageCode`, etc.

Ideally these should not be constructed statically for the lifetime of the process, and instead should be retrieved via callbacks on demand from within the library.

dbrant (changes requested):
Coming along nicely! Some further (unsolicited) light comments inline.

app/src/main/java/org/wikipedia/analytics/metricsplatform/ApplicationData.kt
cjming: cool

app/src/main/java/org/wikipedia/analytics/metricsplatform/ApplicationData.kt
cjming: nice -- assuming then by leveraging the getter method, everything included here will be fetched dynamically

app/src/main/java/org/wikipedia/analytics/metricsplatform/ArticleEvent.kt
cjming: gtk! removed

app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
cjming: per our convo, nulling out these property groups (performer, mediawiki) from the MetricsClient constructor and passing in ApplicationData as part of every event submission - hopefully latest commit addresses correctly

app/src/main/java/org/wikipedia/analytics/metricsplatform/ArticleEvent.kt
cjming: new release published for addressing dynamic data

app/src/main/java/org/wikipedia/analytics/metricsplatform/Event.kt
```diff
@@ -0,0 +1,23 @@
+package org.wikipedia.analytics.metricsplatform
+
+abstract class Event {
+    private var startTime = System.currentTimeMillis()
```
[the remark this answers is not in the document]
cjming: incorporated

app/proguard-rules.pro
cjming, line 1: idk that this is the answer (i could have sworn this worked w/o this before gradle + kotlin were upgraded a few weeks ago

cjming: when I was experimenting with `-keep` lines, it didn't solve the build errors so erring on `-dontwarn`

please advise what best practice is here

dbrant: Yep, the `-dontwarn` lines should be fine. If it works, it works.

Nevertheless, this is not a 100% "satisfying" solution, and we should keep it in the back of our mind to solve properly later. I know very little about "lombok", but it seems to be a compile-time code generator, so ideally I wouldn't expect there to be any lombok-specific classes or annotations left over in the compiled library.

app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
```diff
@@ -0,0 +1,23 @@
...
+    )
+
+    private val clientData: ClientData = ClientData(
+        agentData,
+        null,
+        null,
+        WikipediaApp.instance.wikiSite.authority()
+    )
+
+    val client: MetricsClient = MetricsClient.builder(clientData).build()
```
sharvaniharan, line 22: Looks like we are building a new MP client for each event submitted. @dbrant how do you feel about having it built only once in the WikipediaApp class?

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
dbrant: I'm not seeing it built more than once here. Can you elaborate?

sharvaniharan: My bad... i thought it was a get()

app/src/main/java/org/wikipedia/analytics/metricsplatform/ArticleEvent.kt
```diff
@@ -0,0 +1,306 @@
...
+     */
+    private fun submitEvent(
+        eventName: String,
+        pageData: PageData?,
+        customData: Map<String, Any>
+    ) {
+        MetricsPlatform.client.submitMetricsEvent(eventName, pageData, mergeData(customData))
+    }
+
+    private fun getPageData(fragment: PageFragment): PageData {
```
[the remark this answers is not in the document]
cjming: sure thing
not sure if latest commit addresses your request -- but it shortcuts returning null if pageProperties is null

app/src/main/java/org/wikipedia/analytics/metricsplatform/ArticleEvent.kt
```diff
@@ -0,0 +1,306 @@
...
+
+        fun addFindNext() {
+            numFindNext++
+        }
+
+        fun addFindPrev() {
+            numFindPrev++
+        }
+
+        fun logDone() {
```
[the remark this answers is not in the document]
cjming: yes!  thanks for eagle eyes catch

app/src/main/java/org/wikipedia/Constants.kt
```diff
@@ -21,6 +21,8 @@ object Constants {
     const val ACTIVITY_REQUEST_VOICE_SEARCH = 45
     const val ACTIVITY_REQUEST_WRITE_EXTERNAL_STORAGE_PERMISSION = 44
 
+    const val ANDROID_METRICS_PLATFORM_EVENT_PREFIX = "android.metrics_platform."
+
```
cjming, line 25: is the best place for this?
I figured this prefix will be used across all the files where MP events will eventually be submitted, not just ArticleEvent

dbrant:
Merging to keep things moving, but look for follow-up PR(s).

PR #3984, opened 2023-04-30
Use IntentCompat.getParcelableExtra().

Isira-Seneviratne:
Use [compat](https://developer.android.com/reference/androidx/core/content/IntentCompat#getParcelableExtra(android.content.Intent,java.lang.String,java.lang.Class%3CT%3E)) method for retrieving parcelable extras from Intents, replacing the deprecated method.


cooltey:
Thank you for creating this PR, it is excellent!

PR #3986, opened 2023-05-01
Parse href links when reference text is empty

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T335250


app/src/main/java/org/wikipedia/page/references/ReferenceDialog.kt
```diff
@@ -91,10 +93,25 @@ class ReferenceDialog : ExtendedBottomSheetDialogFragment() {
             binding.referenceText.movementMethod = LinkMovementMethodExt(callback()?.linkHandler)
         }
 
-        fun bindItem(idText: CharSequence?, contents: CharSequence?) {
+        fun bindItem(idText: CharSequence?, reference: PageReferences.Reference) {
             binding.referenceId.text = idText
             binding.root.post {
                 if (isAdded) {
+                    val contents = StringUtil.fromHtml(StringUtil.removeCiteMarkup(StringUtil.removeStyleTags(reference.html)))
+                    if (contents.isNullOrEmpty()) {
```
dbrant, line 101: Can be `isEmpty`

app/src/main/java/org/wikipedia/page/references/ReferenceDialog.kt
```diff
@@ -91,10 +93,25 @@ class ReferenceDialog : ExtendedBottomSheetDialogFragment() {
...
+                        // Inspect html for external links
+                        val links = StringUtil.extractHrefLinks(reference.html)
+                        links.add(links[0])
```
dbrant, line 104: Not sure I understand this statement. We're adding the first element of the list to itself? Wouldn't this create an unnecessary duplicate?

app/src/main/java/org/wikipedia/page/references/ReferenceDialog.kt
```diff
@@ -91,10 +93,25 @@ class ReferenceDialog : ExtendedBottomSheetDialogFragment() {
...
+                        var tags = ""
+                        for (i in 0 until links.size) {
+                            tags = tags.plus("<a href='https:${links[i]}'>[${i + 1}]</a>")
+                        }
+                        binding.referenceText.text = StringUtil.fromHtml(tags)
+                        binding.referenceText.movementMethod = LinkMovementMethod()
```
dbrant, line 110: The `referenceText` already has a movementMethod. Why do we need to give it a new one?

app/src/main/java/org/wikipedia/page/references/ReferenceDialog.kt
```diff
@@ -91,10 +93,25 @@ class ReferenceDialog : ExtendedBottomSheetDialogFragment() {
...
+                        binding.referenceExtLink.isVisible = !tags.isNullOrEmpty()
```
dbrant, line 111: Can be `isNotEmpty`

app/src/main/java/org/wikipedia/page/references/ReferenceDialog.kt
```diff
@@ -91,10 +93,25 @@ class ReferenceDialog : ExtendedBottomSheetDialogFragment() {
...
         }
 
-        fun bindItem(idText: CharSequence?, contents: CharSequence?) {
+        fun bindItem(idText: CharSequence?, reference: PageReferences.Reference) {
             binding.referenceId.text = idText
             binding.root.post {
                 if (isAdded) {
+                    val contents = StringUtil.fromHtml(StringUtil.removeCiteMarkup(StringUtil.removeStyleTags(reference.html)))
+                    if (contents.isNullOrEmpty()) {
+                        // Inspect html for external links
```
dbrant, line 102: This comment is not exactly correct; there's no guarantee that the link(s) will be external. What we're doing is inspecting for links without any anchor text.

app/src/main/java/org/wikipedia/util/StringUtil.kt
```diff
@@ -105,6 +106,16 @@ object StringUtil {
                 == Normalizer.normalize(str2, Normalizer.Form.NFC))
     }
 
+    fun extractHrefLinks(source: String?): MutableList<String> {
```
dbrant, line 109: We don't need to create our own parser for href links. We already use a powerful library called `Jsoup` elsewhere, that you can use to extract link elements with one line of code.

app/src/main/java/org/wikipedia/page/references/ReferenceDialog.kt
```diff
@@ -91,10 +93,25 @@ class ReferenceDialog : ExtendedBottomSheetDialogFragment() {
...
+                        val links = StringUtil.extractHrefLinks(reference.html)
+                        links.add(links[0])
+                        var tags = ""
+                        for (i in 0 until links.size) {
+                            tags = tags.plus("<a href='https:${links[i]}'>[${i + 1}]</a>")
```
dbrant, line 107: We shouldn't need to add `https:` to the link.

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/references/ReferenceDialog.kt
sharvaniharan: Was testing multiple links 😅

app/src/main/java/org/wikipedia/page/references/ReferenceDialog.kt
```diff
@@ -91,10 +93,23 @@ class ReferenceDialog : ExtendedBottomSheetDialogFragment() {
...
 
-        fun bindItem(idText: CharSequence?, contents: CharSequence?) {
+        fun bindItem(idText: CharSequence?, reference: PageReferences.Reference) {
             binding.referenceId.text = idText
             binding.root.post {
                 if (isAdded) {
+                    val contents = StringUtil.fromHtml(StringUtil.removeCiteMarkup(StringUtil.removeStyleTags(reference.html)))
+                    if (contents.isEmpty()) {
+                        // Inspect html for links without anchor text
+                        val links = Jsoup.parse(reference.html).select("a[href]").map { it.attr("href") }
```
sharvaniharan, line 103: @dbrant Should we still put this in StringUtil so we can re-use?

PR #3989, opened 2023-05-03
Remove unnecessary explicit IO dispatching from coroutines.

dbrant:
When making Retrofit or Room calls from within a coroutine, it is not necessary to wrap it in an explicit `withContext(Dispatchers.IO) { }` block, since Retrofit and Room automatically provide main-safe `suspend` functions.

Pardon the size of this PR, but it's literally just removing a bunch of outer `withContext(Dispatchers.IO) {` calls that are redundant. Note the reduction in code size.


sharvaniharan:
@dbrant Ready to merge this once the conflicts are resolved.

PR #3999, opened 2023-05-09
Turn off Machine generated article descriptions in production and beta

sharvaniharan:
**Phab: ** https://phabricator.wikimedia.org/T336278


sharvaniharan:
@dbrant Spoke to Jaz about this and for now she just wants it behind beta.. when we bring this back in, nothing will change with design, just the use of an actual api for it.

dbrant:
> behind beta

So then should it be `isPreProdRelease` instead of `isPreBetaRelease`?

sharvaniharan:
I meant `Should only be in alpha` so `behind alpha` 😅

PR #4004, opened 2023-05-16
Don't show preemptive block message in Edits screen.

dbrant:
https://phabricator.wikimedia.org/T336789

This interprets Blocked messages with improved granularity, by removing access to specific Suggested Edits tasks based on which site the user is blocked (Wikidata, Commons, or the current-language Wikipedia). And this will no longer show the unnecessarily scary Blocked message preemptively.


sharvaniharan:
@dbrant is there a good way to test this?

dbrant:
> @dbrant is there a good way to test this?

Good question - I would suggest the following (in lieu of actually messing around with blocking your Test account on testwiki, etc):

Open up `SuggestedEditsTasksFragmentViewModel.kt` and on line 106, right before the ViewModel signals a Success state, insert the following line:
```
blockMessageCommons = "foo"
```

This simulates that we've received a blocked response from Commons, which should automatically disable the "Image captions" and "Image tags" tasks.

Similarly, you can change it to:
```
blockMessageWikipedia = "foo"
```
This should disable the "Article descriptions" task for English wikipedia.

sharvaniharan:
That works very well! 🎉

PR #4010, opened 2023-05-19
Add exit warning dialog when there are unpublished image tags

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T335675

- Adds a warning dialog when user tries to exit Image Tags screen without publishing chosen tags
- Works on both explore feed image tag card and SE task image tag card


app/src/main/java/org/wikipedia/suggestededits/SuggestionsActivity.kt
```diff
@@ -9,18 +9,30 @@ import org.wikipedia.Constants.INTENT_EXTRA_INVOKE_SOURCE
...
         super.onCreate(savedInstanceState)
         supportActionBar!!.setDisplayHomeAsUpEnabled(true)
         supportActionBar!!.title = getActionBarTitle(intent.getSerializableExtra(INTENT_EXTRA_ACTION) as Action)
+        suggestedEditsImageTagsFragment = supportFragmentManager.findFragmentById(R.id.imageTagFragment) as SuggestedEditsImageTagsFragment?
         setImageZoomHelper()
     }
 
+    override fun onBackPressed() {
+        if (fragment.topBaseChild() is SuggestedEditsImageTagsFragment && (fragment.topBaseChild() as SuggestedEditsImageTagsFragment).atLeastOneTagChecked()) {
+                SuggestedEditsImageTagsFragment.getExitWarningDialog(this).show()
```
dbrant, line 30: I might be misunderstanding the flow here, but why do we need to show the dialog here? The `SuggestedEditsImageTagEditActivity` is the only place where image tags get edited, afaik.

dbrant (changes requested):

app/src/main/java/org/wikipedia/suggestededits/SuggestionsActivity.kt
dbrant: nevermind! I've totally forgotten how image tags work.

PR #4015, opened 2023-05-24
Add SE tasks tests

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T336810

Tests added to :
- Assert the tooltip string on user contribution stats
- User contribution click leads to the diff screen
- Filter button on user contributions screen, and subsequently assert all options on filter screen
- Verify each SE task leads to its own correct screen
- Verify that `Translate` button click when user has only one language, leads to languages addition flow, and more than one lang, leads to appropriate translation screens.


sharvaniharan:
Consolidated here: #4020

PR #4024, opened 2023-06-02
Follow-up / light refactor of Metrics Platform integration.

dbrant:
This does a bit of refactoring (a little more than I anticipated, sorry) to make the ArticleEvent classes more object-oriented and more Kotliny.

* The base class for all events is now `MetricsEvent`, which contains helper methods for getting ApplicationData and PageData.
* This continues with a subclass called `TimedMetricsEvent` which adds a timer. (not all types of events will need a timer)
* And finally ArticleEvent classes derive from the above two classes.
* This consolidates `ArticleLinkPreviewInteraction` and `ArticleLinkPreviewDialogInteraction`, since they were identical except for how they construct their PageData.
* In addition to feature-flagging for pre-Beta only, this will now also respect the user preference for opting out of data collection.


app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsEvent.kt
```diff
@@ -0,0 +1,72 @@
...
+            pageTitle.namespace().code(),
+            Namespace.of(pageTitle.namespace().code()).toString(),
+            revisionId,
+            "",
+            pageTitle.wikiSite.languageCode,
+            null, null, null)
+    }
+
+    companion object {
+        private const val EVENT_NAME_BASE = "android.metrics_platform."
```
cjming, line 70: if the MP prefix is declared here, it should be removed from Constants?
https://github.com/wikimedia/apps-android-wikipedia/blob/main/app/src/main/java/org/wikipedia/Constants.kt#L24

dbrant: hmm, I could have sworn I did!  And... done.

cjming:
this lgtm - i don't have permissions to merge so whoever does, you have my blessing

PR #4027, opened 2023-06-05
Update Metrics Platform version

cjming:
To incorporate @dbrant's recent change:
https://gerrit.wikimedia.org/r/c/mediawiki/libs/metrics-platform/+/927225

update MP lib to 1.8


cjming:
i just published 1.8 - there is a bit of latency so hopefully it'll pass here shortly

PR #4028, opened 2023-06-05
Configure Metrics Platform parameters more explicitly.

dbrant:
This is just for us to have more explicit control of the timing and capacity of the Metrics Platform logic.
(In particular, the default Stream Config fetch delay in the library is 30 seconds, which is much too frequent.)

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
+++ b/app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
@@ -4,6 +4,7 @@ import org.wikimedia.metrics_platform.MetricsClient
 import org.wikimedia.metrics_platform.context.AgentData
 import org.wikimedia.metrics_platform.context.ClientData
 import org.wikipedia.WikipediaApp
+import java.time.Duration
 
 object MetricsPlatform {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
```diff
@@ -19,5 +20,9 @@ object MetricsPlatform {
         WikipediaApp.instance.wikiSite.authority()
     )
 
-    val client: MetricsClient = MetricsClient.builder(clientData).build()
+    val client: MetricsClient = MetricsClient.builder(clientData)
+        .eventQueueCapacity(256)
```
sharvaniharan, line 24: @dbrant 
Can we keep the queue size a bit lower? 256 is almost an average session's worth of user interactions, in case we drop it for the edge case reasons. May be stick to our previous 128?

dbrant: How about like this?

sharvaniharan: Even better! 🎉

PR #4029, opened 2023-06-06
Design tweak and simplification of Nav Bar colors.

dbrant:
https://phabricator.wikimedia.org/T336702#8905766


sharvaniharan:
While the other nav icons are highlighted only on the outline, the `Saved` icon becomes `filled` on highlight. Is there a way to have it just outlined?

dbrant:
> While the other nav icons are highlighted only on the outline, the `Saved` icon becomes `filled` on highlight. Is there a way to have it just outlined?

This is the desired behavior, based on the designs. https://phabricator.wikimedia.org/T330804

PR #4037, opened 2023-06-08
Fix: Tooltip button style

sharvaniharan:
Looks like the Material button needs to inherit its style from ` style="@style/Widget.Material3.Button"` or else it is not getting the correct width.

Before fix:
image
After:
image


dbrant:
Good catch! Although this doesn't explain exactly _why_ changing the style fixes the problem, since our style `App.Button.Secondary` also inherits from `Widget.Material3.Button`.
The root of the issue seems to be a bug in the third-party Balloon library, where it's inflating and/or measuring our custom layout incorrectly, so that the `minWidth` attribute of the button becomes the actual width. (see follow-up)

PR #4043, opened 2023-06-11
Convert login methods to suspend functions.

Isira-Seneviratne:


dbrant:
This should only be done as a larger refactor of LoginActivity etc. to use a proper ViewModel architecture.

PR #4044, opened 2023-06-11
Convert getBlockMessageHtml() to a suspend function.

Isira-Seneviratne:


dbrant:
This is not yet necessary or ready to be converted. We will convert the relevant activities to use a proper ViewModel in due course.

PR #4064, opened 2023-06-28
Update client data for metrics platform

cjming:
- Bump Metrics Platform Java library version.
- Improve logging mechanism. 
- Pass in client data into submit event method to capture dynamic data.
- Pass in prod status to Metrics Client Builder.

Bug: [T338964](https://phabricator.wikimedia.org/T338964)


app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
```diff
@@ -3,22 +3,37 @@ package org.wikipedia.analytics.metricsplatform
...
     )
 
+    val mediawikiData = MediawikiData(
+        WikipediaApp.instance.currentTheme.toString(),
+        WikipediaApp.instance.versionCode.toString(),
+        ReleaseUtil.isProdRelease,
+        ReleaseUtil.isDevRelease,
+        WikipediaApp.instance.wikiSite.dbName(),
+        WikipediaApp.instance.languageState.systemLanguageCode,
+        null
```
cjming, line 26: I err'd on the side of sending this data as part of the MW data object -- this null represents "siteContentLanguageVariant" << does it make sense to put anything here or is null appropriate?

app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
```diff
@@ -3,22 +3,37 @@ package org.wikipedia.analytics.metricsplatform
...
+
+    val domain = WikipediaApp.instance.wikiSite.authority()
+
     private val clientData = ClientData(
         agentData,
         null,
+        mediawikiData,
         null,
-        WikipediaApp.instance.wikiSite.authority()
+        domain
```
cjming, line 36: I know it's weird to have client data initiated with the MP client and then passed in with the event -- there's likelihood that the MP monoschema will change here soon so rather than revamping the functionality now, I err'd on the side of getting this out the door to start collecting data to help inform how the monoschema should change

cjming:
One thing I forgot and just remembered -- adding the `hasty` param on prod only -- I'll work on that and make adjustments here accordingly.

cjming:
> One thing I forgot and just remembered -- adding the `hasty` param on prod only -- I'll work on that and make adjustments here accordingly.

done!

dbrant:
Seeing as the monoschema itself is not yet finalized, and that we're just interested in getting some preliminary data to prove things out, I think this is fine to merge and keep testing.

PR #4072, opened 2023-07-07
Complete smoke tests

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T335160

Few things to keep in mind while reviewing or testing in the future

- Make sure you do not have the app already installed or logged in. this will fail login tests
- When the suite runs through the login, the reading lists start downloading the articles which creates a lot of background work traffic and the espresso get deprioritized - doesn't crash but goes on a very long wait for a while till it times out. Make sure you use a new account or one with no saved pages
- If you end up using new account, you might have to make 3-4 edits on it so your SE tab is enabled so we can run the related tests [There is logic to check for any disable view, but if you want to run se tests you need this]


sharvaniharan:
Marking this wip until we are able to run the test end to end in one go.

sharvaniharan:
@dbrant @cooltey I removed the `wip` so we can merge the tests to keep things moving.

cooltey:
Hi @sharvaniharan 

I did a couple of runs and still saw failed every time. Would it be possible to fix the issues before merging to the main branch? Thanks!

sharvaniharan:
@cooltey @dbrant Please test this again. I was able to run end-to-end multiple times and have left a note in the pr notes as well.

cooltey:
> @cooltey @dbrant Please test this again. I was able to run end-to-end multiple times and have left a note in the pr notes as well.

Tried a couple of times and it failed. Would you be able to fix it?

app/src/androidTest/java/org/wikipedia/main/SmokeTests.kt
```diff
@@ -516,15 +796,401 @@ class SmokeTests {
...
+
+        onView(allOf(TestUtil.withGrandparent(withId(R.id.login_password_input)), withClassName(Matchers.`is`("org.wikipedia.views.PlainPasteEditText"))))
+            .perform(replaceText(BuildConfig.TEST_LOGIN_PASSWORD), closeSoftKeyboard())
+
+        // Click the login button
+        onView(withId(R.id.login_button)).perform(scrollTo(), click())
+
+        TestUtil.delay(5)
+
+        onView(allOf(withId(android.R.id.button2), withText("No thanks"),
```
dbrant, line 865: It's all working for me, until this point. What/where is the "No thanks" button supposed to be? (what component is that part of?)

app/src/androidTest/java/org/wikipedia/main/SmokeTests.kt
```diff
@@ -516,15 +796,401 @@ class SmokeTests {
...
+
+        TestUtil.delay(2)
+
+        // Click the login button
+        onView(allOf(withId(R.id.create_account_login_button), withText("Log in"), isDisplayed()))
+            .perform(click())
+
+        TestUtil.delay(2)
+
+        // Type in an incorrect username and password
```
dbrant, line 853: It looks like all the following actions depend on a _correct_ username and password, which makes this comment confusing.

dbrant (changes requested):

app/src/androidTest/java/org/wikipedia/main/SmokeTests.kt
sharvaniharan: This is the lists sync dialog dismiss button :
image

It should appear right after login... will check again why it is failing 🤔

sharvaniharan:
> 
> Tried a couple of times and it failed. Would you be able to fix it?


This is a very common one @cooltey 🙈  This happens when it tries to test explore feed cards by saving to list.. depending on the length of a card sometimes it is not visible and it fails... Scrolling to it when its not visible is flaky. I have a few options as to how to deal with it.. we can discuss on sync Monday and will update a decision here later since you are on vacation 🏖️

sharvaniharan:
@dbrant I have added more checks for some login tests. Please test again and let me know.

PR #4106, opened 2023-08-01
Make sure to show color on the main navigation correctly

cooltey:
https://phabricator.wikimedia.org/T343177

### DIFF
--- a/app/src/main/res/layout/fragment_main.xml
+++ b/app/src/main/res/layout/fragment_main.xml
@@ -42,7 +42,8 @@
             app:labelVisibilityMode="labeled"
             app:tabGravity="fill"
             app:itemPaddingTop="8dp"
-            app:tabIndicatorColor="@android:color/transparent" />
+            app:tabIndicatorColor="@android:color/transparent"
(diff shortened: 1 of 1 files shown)


app/src/main/res/layout/fragment_main.xml
```diff
@@ -42,7 +42,8 @@
             app:labelVisibilityMode="labeled"
             app:tabGravity="fill"
             app:itemPaddingTop="8dp"
-            app:tabIndicatorColor="@android:color/transparent" />
+            app:tabIndicatorColor="@android:color/transparent"
+            app:elevation="0dp"/>
```
dbrant, line 46: Almost there. We can also remove the two references to `@android:color/transparent` in this block, since they are not necessary, and also contribute to the issue. This will make the total of -1 lines added.

dbrant (changes requested):

app/src/main/res/layout/fragment_main.xml
cooltey: Done!

PR #4107, opened 2023-08-01
Open the user page in the app instead of an external browser

cooltey:
https://phabricator.wikimedia.org/T342849

### DIFF
--- a/app/src/main/java/org/wikipedia/main/MainFragment.kt
+++ b/app/src/main/java/org/wikipedia/main/MainFragment.kt
@@ -9,7 +9,6 @@ import android.content.Intent
 import android.content.IntentFilter
 import android.content.pm.PackageManager
 import android.graphics.Bitmap
-import android.net.Uri
 import android.os.Build
 import android.os.Bundle
 import android.speech.RecognizerIntent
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -441,7 +440,7 @@ class MainFragment : Fragment(), BackPressedHandler, MenuProvider, FeedFragment.
 
     override fun usernameClick() {
         val pageTitle = PageTitle(UserAliasData.valueFor(WikipediaApp.instance.languageState.appLanguageCode) + ":" + AccountUtil.userName, WikipediaApp.instance.wikiSite)
-        UriUtil.visitInExternalBrowser(requireContext(), Uri.parse(pageTitle.uri))
+        CustomTabsUtil.openInCustomTab(requireContext(), pageTitle.uri)
```
dbrant, line 443: I think the intention is to open the user page in the native PageActivity (`PageActivity.newIntent(...)`) rather than as an external web page.

dbrant (changes requested):

app/src/main/java/org/wikipedia/main/MainFragment.kt
cooltey: Done

PR #4108, opened 2023-08-02
Fix display of non-public revisions in Diff view.

dbrant:
https://phabricator.wikimedia.org/T343044


cooltey (changes requested):
The "Thank" and "Undo" buttons will show up for a few seconds while loading the revision, even if it is a private revision. 

https://youtube.com/shorts/7dnRnXPC8V8

dbrant:
> The "Thank" and "Undo" buttons will show up for a few seconds while loading the revision, even if it is a private revision.

And.. fixed!

cooltey:
> > The "Thank" and "Undo" buttons will show up for a few seconds while loading the revision, even if it is a private revision.
> 
> And.. fixed!

That's better!

Just noticed another issue if I change the device's orientation, the buttons will show up again.

dbrant:
> Just noticed another issue if I change the device's orientation, the buttons will show up again.

Done!!!111

PR #4119, opened 2023-08-07
Fix usage of ArticleFindInPageInteraction for Metrics Platform

cjming:
After reviewing the number of events for the initial Android instruments that were integrated with Metrics Platform, the number of events for the `ArticleFindInPageInteraction` instrument returned with zero events.

This PR attempts to fix this by injecting the `FindInWebPageActionProvider` class with an MP `ArticleFindInPageInteraction` parameter.

The corresponding table that logs these events using Modern Event Platform returns 14,675 rows on 8/2/23. The expectation is that some fraction (greater than zero) of this number of events will be returned in the analogous Metrics Platform table when this change is included in the next beta release.

Bug: [T330355](https://phabricator.wikimedia.org/T330355)


cooltey (approved):
Looks good to me.

PR #4124, opened 2023-08-08
Patroller Tasks (feature branch)

dbrant:
This can serve as the branch where Patroller Tasks functionality will be merged, until it's ready to go.

https://phabricator.wikimedia.org/T343224


app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -249,12 +273,43 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback, L
...
+        binding.watchButton.setOnClickListener {
+            viewModel.watchOrUnwatch(isWatched, WatchlistExpiry.NEVER, isWatched)
+            if (isWatched) editHistoryInteractionEvent?.logUnwatchClick() else editHistoryInteractionEvent?.logWatchClick()
         }
+        updateWatchButton(isWatched, hasWatchlistExpiry)
 
-        binding.rollbackButton.setOnClickListener {
-            showRollbackDialog()
+        binding.warnButton.setOnClickListener {
+            // TODO: implement this
```
sharvaniharan, line 312: Could we please connect the warn flow here?

cooltey: This will be implemented in the PR of connecting warn template with the reply activity.

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -448,10 +575,87 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback, L
             .show()
     }
 
+    private fun showRollbackSnackbar() {
+        val snackbar = FeedbackUtil.makeSnackbar(requireActivity(), getString(R.string.revision_rollback_success))
```
sharvaniharan, line 579: Rollback feedback string seems to be different than designs. Expected: "Rollback complete" Actual: Edits reverted: https://www.figma.com/file/cJmB34iJE5p8qbqrQWnE50/Patrolling-on-Apps?type=design&node-id=1504-69602&mode=design&t=u7XLyt3NC01sAoBV-0

cooltey: Done and please see #4243

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -448,10 +575,87 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback, L
...
+            }
+        }
+
+        feedbackView.findViewById<TextView>(R.id.optionSatisfied).setOnClickListener(clickListener)
+        feedbackView.findViewById<TextView>(R.id.optionNeutral).setOnClickListener(clickListener)
+        feedbackView.findViewById<TextView>(R.id.optionUnsatisfied).setOnClickListener(clickListener)
+
+        dialog = MaterialAlertDialogBuilder(requireActivity())
+            .setTitle(R.string.patroller_diff_feedback_dialog_title)
+            .setCancelable(false)
```
sharvaniharan, line 616: Do we really want to set it to `false` for cancellable? This forces the user to choose some option even when they accidentally choose it, hence tainting data. However, if it were to be the other way they can always choose it from menu when they accidentally cancel the dialog.

sharvaniharan:
text cut-off
Text is getting cut-off at the end.

cooltey:
> text cut-off Text is getting cut-off at the end.

I cannot reproduce it on both of my devices but I can guess why it has the issue. I have updated the end margin, please see the change, thanks!

PR #4131, opened 2023-08-10
Wikitext syntax highlighting in Talk input fields.

dbrant:
https://phabricator.wikimedia.org/T343889


app/src/main/java/org/wikipedia/edit/SyntaxHighlightViewAdapter.kt
```diff
@@ -0,0 +1,131 @@
...
+            }
+        }
+    }
+
+    override fun onRequestInsertMedia() {
+        requestInsertMedia.launch(InsertMediaActivity.newIntent(activity, wikiSite, ""))
+    }
+
+    override fun onRequestInsertLink() {
+        requestLinkFromSearch.launch(SearchActivity.newIntent(activity, Constants.InvokeSource.TALK_REPLY_ACTIVITY, null, true))
```
cooltey, line 88: Do we need to add a parameter to the adapter in order to send the correct source here?

app/src/main/java/org/wikipedia/views/UserMentionEditText.kt
```diff
@@ -164,44 +146,7 @@ class UserMentionEditText : PlainPasteEditText {
...
-                val lenDiff = name.length - (pairs[i].second - pairs[i].first)
-                for (j in i + 1 until pairs.size) {
-                    pairs[j].first += lenDiff
-                    pairs[j].second += lenDiff
-                }
-            }
-        }
-        return str
+    private fun composeUserNameLink(userName: String, wikiSite: WikiSite): String {
+        return "@[[" + UserAliasData.valueFor(wikiSite.languageCode) + ":" + userName + "|" + userName + "]]"
```
cooltey, line 150: It looks like the new @ mention text format will not include the `@` in the clickable link, is that expected?

cooltey (changes requested):

app/src/main/java/org/wikipedia/edit/SyntaxHighlightViewAdapter.kt
dbrant: yes!

app/src/main/java/org/wikipedia/views/UserMentionEditText.kt
dbrant: yes, it's expected, and that's actually how it looks in desktop DiscussionTools.

cooltey (commented):
Found some format presentation issues:
1. The `#` list symbols are not displayed as numbers.
Screenshot_20230921-131456_Wikipedia Dev
2. The link to Obama's article in `zh.wiki` is invalid. But it shows correctly in the desktop version of the talk thread. 
Screenshot_20230921-131500_Wikipedia Dev

Desktop version:
Screenshot 2023-09-21 at 13-14-36 使用者討論 Cooltey - 維基百科，自由的百科全書

https://zh.wikipedia.org/wiki/User_talk:Cooltey#%E6%B8%AC%E8%A9%A66

dbrant:
> Found some format presentation issues:

@cooltey  It doesn't seem like this is an issue with this PR, is it? These presentation issues might need their own task.

The presentation of lists is simply a function of how a native `TextView` handles HTML content.

And the Chinese Obama link might actually be a bug in Parsoid (which is used by the DiscussionTools API for giving us the content). When you inserted the link, the edit was made in Traditional Chinese (`貝拉克·歐巴馬`), but the "canonical" name of the Obama article is in Simplified Chinese (`贝拉克·欧巴马`), so it looks like Parsoid is getting confused and believes that the Traditional link is a new article that should be created.

cooltey:
> > Found some format presentation issues:
> 
> @cooltey It doesn't seem like this is an issue with this PR, is it? These presentation issues might need their own task.
> 
> The presentation of lists is simply a function of how a native `TextView` handles HTML content.
> 

Got it. I'll create a ticket for it.

> And the Chinese Obama link might actually be a bug in Parsoid (which is used by the DiscussionTools API for giving us the content). When you inserted the link, the edit was made in Traditional Chinese (`貝拉克·歐巴馬`), but the "canonical" name of the Obama article is in Simplified Chinese (`贝拉克·欧巴马`), so it looks like Parsoid is getting confused and believes that the Traditional link is a new article that should be created.

I see. Maybe I can create a small PR to fix it.

PR #4144, opened 2023-08-19
Fix: make sure to reload the content after updating flow state and changing device orientation

cooltey:
Steps to reproduce:

1. Go to the user talk page
2. Click on the overflow menu and click "Watch" to watch the page
3. After seeing the watchlist snackbar, change the device orientation to landscape.
4. The content will keep loading. 

The same thing happens to the "Undo" publish function if you click the back button on the publish screen.


dbrant (changes requested):
One of the huge benefits of a ViewModel is to be able to _not reload anything_ on screen rotation, or any other configuration change.

The real problem in the TalkTopicsViewModel is that it passes all of its different states into a single  `uiState` flow.  If the different UiStates were separated into their own flows, I believe this issue could be solved without having to reload everything. Can we do it that way?

cooltey:
> One of the huge benefits of a ViewModel is to be able to _not reload anything_ on screen rotation, or any other configuration change.
> 
> The real problem in the TalkTopicsViewModel is that it passes all of its different states into a single `uiState` flow. If the different UiStates were separated into their own flows, I believe this issue could be solved without having to reload everything. Can we do it that way?

Done!

Just curious should we make the `UiState.ActionError`  has its own flow? Or should we include it in the one that has three states? (`UpdateNamespace`, `LoadTopic` and `LoadError`)?

cooltey:
🤔 Just noticed that I forgot to remove the `resetState` from the `viewModel` class. After removing it. Will fix it.

cooltey:
> 🤔 Just noticed that I forgot to remove the `resetState` from the `viewModel` class. The issue is still there after I removed it. Will fix it.

Done!

PR #4151, opened 2023-08-23
Patroller tasks: add required strings for the feature

cooltey:
https://phabricator.wikimedia.org/T343287


cooltey:
@dbrant @sharvaniharan
The strings for the filters (e.g. strings starting with `patroller_tasks_filters_`) are basically copied from the watchlist filter strings, not sure if we need to create new ones or reuse the watchlist filter strings.

PR #4157, opened 2023-08-29
Set proper background color to action bar overflow menu

cooltey:
The action bar overflow menu contains a weird background color (not a pure `paper_color`), which will be noticeable when you carefully look at it.

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -39,6 +39,7 @@
         <item name="android:textColorHint">?attr/placeholder_color</item>
 
         <item name="popupMenuStyle">@style/PopupMenuStyle</item>
+        <item name="actionOverflowMenuStyle">@style/OverflowMenuStyle</item>
 
         <item name="collapsingToolbarLayoutStyle">@style/CollapsingToolbarStyle</item>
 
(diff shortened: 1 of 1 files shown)


dbrant (changes requested):
Unfortunately this is not quite correct.
The background of an overflow menu is actually a `MaterialShapeDrawable`, which has a special shape and special dimensions/margins. If you override the whole thing with a simple color, it will no longer have the proper shape. Ideally the color should be applied as a `tint` onto the background.

I've investigated this myself, and the only solution seems to be to no longer use the system-default ActionBar. (In every Activity that has an overflow menu, we should use our own MaterialToolbar in the xml layout, in which the overflow menu will get the correct styling automatically.)

cooltey:
> Unfortunately this is not quite correct. The background of an overflow menu is actually a `MaterialShapeDrawable`, which has a special shape and special dimensions/margins. If you override the whole thing with a simple color, it will no longer have the proper shape. Ideally the color should be applied as a `tint` onto the background.
> 
> I've investigated this myself, and the only solution seems to be to no longer use the system-default ActionBar. (In every Activity that has an overflow menu, we should use our own MaterialToolbar in the xml layout, in which the overflow menu will get the correct styling automatically.)

Got it! That makes sense to me!

PR #4168, opened 2023-09-05
Update required strings for Patroller Tasks

cooltey:
https://phabricator.wikimedia.org/T343834#9142930


app/src/main/res/values/strings.xml
```diff
@@ -1562,8 +1562,8 @@
     <string name="patroller_tasks_edits_list_card_overflow_view_user_contributions">View user contributions</string>
     <string name="patroller_tasks_edits_list_card_overflow_thank">Thank</string>
     <string name="patroller_tasks_edits_list_user_information_dialog_title">User information</string>
-    <string name="patroller_tasks_edits_list_user_information_dialog_tenure">Tenure</string>
-    <string name="patroller_tasks_edits_list_user_information_dialog_edit_count">Edit count</string>
+    <string name="patroller_tasks_edits_list_user_information_dialog_tenure_text"><![CDATA[<b>Tenure: </b> %s]]></string>
```
dbrant, line 1565: Are we sure everyone knows what the word "Tenure" means? Would it be simpler if we just say "Joined Wikipedia _nnn_ years ago"?  Or even "Joined Wikipedia on _[date]_" (in the standard date format)?  This way we wouldn't need to hand-craft unnecessary strings and logic for time spans.

dbrant (changes requested):

app/src/main/res/values/strings.xml
cooltey: That makes sense to me. I have addressed this to Olga and Jazmin about this. Thanks!

cooltey: Done!

dbrant:
\o/

PR #4175, opened 2023-09-08
IR data: EditattemptStep events

sharvaniharan:
Couple more data wirings for image recs data. Simultaneously sending data to the legacy EditAttemptStep table as well.
Doc: https://docs.google.com/presentation/d/10Twu9rnHy4XFzBjxRKDQPMKDVPDJDd9hBQXYwN7nilU/edit#slide=id.g1e4cd7273df_0_29

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageRecsFragment.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageRecsFragment.kt
@@ -84,8 +84,7 @@ class SuggestedEditsImageRecsFragment : SuggestedEditsItemFragment(), MenuProvid
                 .show()
             ImageRecommendationsEvent.logAction("editsummary_success_confirm", "editsummary_dialog",
                 getActionStringForAnalytics(acceptanceState = "accepted", revisionId = revId, addTimeSpent = true), viewModel.langCode)
-            val title = PageTitle(viewModel.recommendation.titleText, WikiSite.forLanguageCode(viewModel.langCode))
-            EditAttemptStepEvent.logSaveSuccess(title)
+            EditAttemptStepEvent.logSaveSuccess(viewModel.pageTitle)
             viewModel.acceptRecommendation(null, revId)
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageRecsFragment.kt
```diff
@@ -83,6 +84,8 @@ class SuggestedEditsImageRecsFragment : SuggestedEditsItemFragment(), MenuProvid
                 .show()
             ImageRecommendationsEvent.logAction("editsummary_success_confirm", "editsummary_dialog",
                 getActionStringForAnalytics(acceptanceState = "accepted", revisionId = revId, addTimeSpent = true), viewModel.langCode)
+            val title = PageTitle(viewModel.recommendation.titleText, WikiSite.forLanguageCode(viewModel.langCode))
```
cooltey, line 87: Can we use `viewModel.pageTitle` instead of creating a new `PageTitle` here?

Please help to update the same thing in `requestEdit` launcher, thanks!

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageRecsFragment.kt
```diff
@@ -129,6 +132,8 @@ class SuggestedEditsImageRecsFragment : SuggestedEditsItemFragment(), MenuProvid
         binding.acceptButton.setOnClickListener {
             ImageRecommendationsEvent.logAction("suggestion_accept", "recommendedimagetoolbar",
                 getActionStringForAnalytics(acceptanceState = "accepted"), viewModel.langCode)
+            val title = PageTitle(viewModel.recommendation.titleText, WikiSite.forLanguageCode(viewModel.langCode))
```
cooltey, line 135: Can we use `viewModel.pageTitle` instead of creating a new `PageTitle` here?

cooltey (changes requested):

PR #4180, opened 2023-09-12
Enable R8 full mode

Isira-Seneviratne:
* Enable R8 full mode.
* Add R8 [full mode rules](https://github.com/square/retrofit/blob/master/retrofit/src/main/resources/META-INF/proguard/retrofit2.pro) for Retrofit.


dbrant:
Sorry, it's not reasonable for us to keep track of a dozen rules to satisfy a third-party dependency.
We can revisit once Retrofit ships with the correct rules built in.

PR #4182, opened 2023-09-12
Places (feature branch).

dbrant:
https://phabricator.wikimedia.org/T347201


app/src/main/java/org/wikipedia/places/PlacesFragment.kt
```diff
@@ -189,6 +244,19 @@ class PlacesFragment : Fragment(), LinkPreviewDialog.Callback, MapboxMap.OnMapCl
         }
     }
 
+    private fun updateSearchCardViews() {
```
cooltey, line 247: Hi @dbrant, thanks for making the changes to the map so I can continue working on the link preview dialog.

I just noticed one issue: the compass is overlapped by the search bar view, can you fix that?

dbrant: good catch, and Done!

dbrant:
@sharvaniharan  It looks like the Filter screen still needs several design updates:

* The background of the activity should be `background_color`, not `paper_color`.
* Each language item should have a divider in between.
* The border around the language code should be thicker. (Should use the same component as in `PlacesFragment` itself)
* The font of the languages should be regular, not bold.
* The height of each language item should be 48dp. It feels like they are a bit larger than that.
* All of the items (including "Update app languages") should have a ripple effect. (selectableItemBackground)

image image

sharvaniharan:
I noticied  a couple things while creating the demo video 😅 
https://youtube.com/shorts/awuSoZis9cM?feature=share
@dbrant I think after the prefix search change the search seems to be not recognizing some results... like in this demo.. When you search for `Bost` you can see in the results that `Bost motorsports` is present but if you keep typing further `Bost mot` results in no search.. This is happening on main and this feature branch.

@cooltey I also saw that some markers that do not have image on then seem to actually have a thumbnail associated with them when previewed. Is this expected?

Also saw that when we click on a marker on places and carry that string to searches screen, it doesnt recognize that title in results ... will look into it after the search thing is fixed, as this might be a side effect of the search issue.

dbrant:
> When you search for `Bost` you can see in the results that `Bost motorsports` is present but if you keep typing further `Bost mot` results in no search.. This is happening on main and this feature branch.

This should now be fixed as of today's earlier PRs.

> @cooltey I also saw that some markers that do not have image on then seem to actually have a thumbnail associated with them when previewed. Is this expected?

Yes, this is expected. We are using the MediaWiki API to get the list of results and their thumbnails, whereas the Link Preview dialog uses the `page/summary` endpoint. For some reason the Summary endpoint returns a thumbnail for certain pages that don't have a thumbnail via the regular API. Technically not our bug, and we should try not to rely on the Summary endpoint any more than we already are.

> Also saw that when we click on a marker on places and carry that string to searches screen, it doesnt recognize that title in results ... will look into it after the search thing is fixed, as this might be a side effect of the search issue.

Also fixed via earlier PRs.

app/src/main/res/layout/item_places_list.xml
```diff
@@ -9,50 +9,49 @@
     android:paddingVertical="12dp"
     android:paddingHorizontal="@dimen/list_item_horizontal_padding">
 
-    <TextView
-        android:id="@+id/listItemTitle"
-        style="@style/H3.Article"
+    <LinearLayout
```
cooltey, line 12: I use `LinearLayout` to wrap up items in one container so it can align to the center correctly if no descriptions or has multiple lines on article title.

PR #4184, opened 2023-09-12
Minor fix

sharvaniharan:
On edge case, when a user sees the fullscreen onboarding for IR, if they backpress, app crashes because the image hasnt been set up yet. Adding this check will prevent the crash.


sharvaniharan:
Has been fixed on feature branch

PR #4195, opened 2023-09-21
Fix the issue of inserting invalid article link in wikitext keyboard

cooltey:
This fixes the issue addressed here: https://github.com/wikimedia/apps-android-wikipedia/pull/4131#issuecomment-1730291098

### DIFF
--- a/app/src/main/java/org/wikipedia/edit/WikiTextKeyboardView.kt
+++ b/app/src/main/java/org/wikipedia/edit/WikiTextKeyboardView.kt
@@ -142,7 +142,12 @@ class WikiTextKeyboardView : FrameLayout {
             if (title.wikiSite.languageCode != baseLangCode) {
                 text += ":" + title.wikiSite.languageCode + ":"
             }
-            text += title.prefixedText + "|" + StringUtil.fromHtml(title.displayText).toString() + "]]"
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/edit/WikiTextKeyboardView.kt
```diff
@@ -142,7 +142,7 @@ class WikiTextKeyboardView : FrameLayout {
             if (title.wikiSite.languageCode != baseLangCode) {
                 text += ":" + title.wikiSite.languageCode + ":"
             }
-            text += StringUtil.fromHtml(title.displayText).toString() + "]]"
+            text += title.prefixedText + "|" + StringUtil.fromHtml(title.displayText).toString() + "]]"
```
dbrant, line 145: This will work nicely, but can we do it only if `prefixedText` and `displayText` are different?
(not including spaces/underlines, e.g. "Barack Obama" and "Barack_Obama" should be the same.)

dbrant (changes requested):

app/src/main/java/org/wikipedia/edit/WikiTextKeyboardView.kt
cooltey: It makes sense to me. Done!

PR #4196, opened 2023-09-22
Improve code using serializable extensions

Isira-Seneviratne:
Use the new serializable retrieval methods in `IntentCompat` and `BundleCompat` to fix the deprecation warnings.


dbrant:
This PR can be merged only if it is slimmed down to _only_ add the serializable extensions, and nothing else. (No moving of constants, no changes to how we pass arguments to fragments, no changes to navigation tab switching.)

dbrant:
Cleaning up old/no-longer-applicable PRs.

PR #4204, opened 2023-09-26
Support `<code>` tags in native TextViews.

dbrant:
Supporting ordered lists (`<ol>`) is an exercise for someone else.


cooltey:
This works for me!

Just have a question: the `<code></code>` tag in the desktop talk page will have a light grey background on the span, do we need to have the same style as the desktop?

dbrant:
> do we need to have the same style as the desktop?

I don't think this is necessary for now. It can be revisited in the future.

PR #4205, opened 2023-09-26
Model underpinnings for new campaign announcements.

dbrant:
https://phabricator.wikimedia.org/T346902

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/donate/CampaignCollection.kt
+++ b/app/src/main/java/org/wikipedia/dataclient/donate/CampaignCollection.kt
@@ -27,8 +27,6 @@ object CampaignCollection {
             val response = OkHttpConnectionFactory.client.newCall(request).execute()
             val json = response.body!!.string()
 
-            Thread.sleep(5000)
-
             val campaigns = JsonUtil.decodeFromString<List<JsonElement>>(json).orEmpty()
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/dataclient/donate/CampaignCollection.kt
```diff
@@ -0,0 +1,50 @@
...
+    private const val CAMPAIGNS_URL = "https://donate.wikimedia.org/wiki/MediaWiki:AppsCampaignConfig.json?action=raw"
+    private const val CAMPAIGNS_URL_DEBUG = "https://donate.wikimedia.org/wiki/MediaWiki:AppsCampaignConfigStaging.json?action=raw"
+
+    suspend fun getActiveCampaigns(): List<Campaign> {
+        val campaignList = mutableListOf<Campaign>()
+
+        withContext(Dispatchers.IO) {
+            val request = Request.Builder().url(CAMPAIGNS_URL_DEBUG).build()
+            val response = OkHttpConnectionFactory.client.newCall(request).execute()
+            val json = response.body!!.string()
```
cooltey, line 28: Can we use `null-safey` here?

dbrant: yep. (and also updated to the newest actions-array structure)

PR #4211, opened 2023-09-27
New app donation banner with new donate.wiki source

cooltey:
https://phabricator.wikimedia.org/T344026
https://phabricator.wikimedia.org/T346902


sharvaniharan:
Only 2 minor things:
1. The `donor policy` link opens in an external browser. just wanted to make sure that's ok. Since we were discussing about not leaving the app for other stuff.
policy

2. Could you please leave detailed step-by-step instructions for Anthony about how to use the developer settings and what the values mean.

cooltey:
> Only 2 minor things:
> 
>     1. The `donor policy` link opens in an external browser. just wanted to make sure that's ok. Since we were discussing about not leaving the app for other stuff.

I think it should be fine:
1. We open the policy or license links in an external browser across the app.
2. If we try to open the link in a CustomTab, it may lead the dialog to be closed.

> 
>     2. Could you please leave detailed step-by-step instructions for Anthony about how to use the developer settings and what the values mean.

Sure! I'll add it in the phab ticket.

PR #4214, opened 2023-09-29
Add required strings for donor experience dialog

cooltey:
https://phabricator.wikimedia.org/T344026


sharvaniharan:
@cooltey Wondering if we should create new strings for the button texts too?

sharvaniharan:
Will go ahead and merge this but let us think about the other strings too.

PR #4221, opened 2023-10-05
Donor experience: Data wiring

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T347730
Schema doc to check against: https://docs.google.com/presentation/d/1vZZwXb0rOGFlqvSy5PxxqRVy83ORPBAwYVHisfp4_o0/edit#slide=id.g28073e6aeeb_2_104


app/src/main/java/org/wikipedia/activity/SingleWebViewActivity.kt
```diff
@@ -115,6 +116,7 @@ class SingleWebViewActivity : BaseActivity() {
     }
 
     private fun goBack() {
+        DonorExperienceEvent.logAction("article_return_click", "webpay_processed")
```
cooltey, line 119: This should have a logic to check whether the session is for donor experience. 

Can we add a parameter to the `newIntent()` such as `newIntent(logDonorExperience: Boolean = false)`?

cooltey (changes requested):

app/src/main/java/org/wikipedia/navtab/MenuNavTabDialog.kt
```diff
@@ -70,6 +71,7 @@ class MenuNavTabDialog : ExtendedBottomSheetDialogFragment() {
         }
 
         binding.mainDrawerDonateContainer.setOnClickListener {
+            DonorExperienceEvent.logAction("donate_start_click", "article_banner")
```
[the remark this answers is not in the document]
sharvaniharan: Good catch!

app/src/main/java/org/wikipedia/activity/SingleWebViewActivity.kt
```diff
@@ -135,12 +137,15 @@ class SingleWebViewActivity : BaseActivity() {
         const val EXTRA_URL = "url"
         const val EXTRA_SHOW_BACK_BUTTON = "goBack"
         const val EXTRA_PAGE_TITLE = "pageTitle"
+        const val EXTRA_PAGE_CONTENT_INFO = "pageContentInfo"
+        const val EXTRA_DONOR_EXPERIENCE = "donorExperience"
```
cooltey, line 141: Should it be `PAGE_CONTENT_SOURCE_DONOR_EXPERIENCE` ?

cooltey (changes requested):

PR #4224, opened 2023-10-06
Translations: Donor Experience

sharvaniharan:
We have still not received a few more translations for the feature in `it` and `nl`. This adds them manually. We will monitor if fuzzyBot picks up these translations into `translateWiki`.


sharvaniharan:
Looks like we got the Dutch translations organically with the translatewiki dump.

PR #4225, opened 2023-10-08
Use TypedArray.use() extension

Isira-Seneviratne:


cooltey:
Thanks for the PR. May I know the advantage of using the extension in this case? Otherwise, it would be better if we kept the code as it is.

Isira-Seneviratne:
> Thanks for the PR. May I know the advantage of using the extension in this case? Otherwise, it would be better if we kept the code as it is.

It works much like the standard Kotlin `use` extension, by recycling the `TypedArray` after returning a value.

app/src/main/java/org/wikipedia/views/CircularProgressBar.kt
```diff
@@ -21,17 +21,18 @@ class CircularProgressBar constructor(context: Context, attrs: AttributeSet? = n
...
     init {
-        var progressColor = ResourceUtil.getThemedColor(getContext(), R.attr.progressive_color)
-        var progressBackgroundColor = ResourceUtil.getThemedColor(getContext(), R.attr.placeholder_color)
-        var progressStrokeWidth = DimenUtil.dpToPx(DEFAULT_STROKE_WIDTH_DP.toFloat()).toInt()
-        if (attrs != null) {
-            context.withStyledAttributes(attrs, R.styleable.CircularProgressBar) {
-                progressColor = getColor(R.styleable.CircularProgressBar_progressColor, progressColor)
-                progressBackgroundColor = getColor(R.styleable.CircularProgressBar_progressBackgroundColor, progressBackgroundColor)
-                progressStrokeWidth = getDimensionPixelSize(R.styleable.CircularProgressBar_progressStrokeWidth, progressStrokeWidth)
-                maxProgressValue = getInt(R.styleable.CircularProgressBar_maxProgress, MAX_PROGRESS).toDouble()
```
cooltey, line 32: Please make sure to apply `maxProgressValue` in the code.

cooltey (changes requested):

PR #4229, opened 2023-10-11
Add required strings for Patroller Tasks onboarding screen

cooltey:
https://phabricator.wikimedia.org/T343242


cooltey:
self-merge strings.

PR #4232, opened 2023-10-13
Update string for Patroller Tasks

cooltey:
Add quote for the target title.


sharvaniharan:
@cooltey since this has already been translated, wondering if we should delete the string and qq resources and then create a new string with the quotes... it feels too much lift for a simple change, but otherwise it might not trigger changes for already translated langs.

cooltey:
> @cooltey since this has already been translated, wondering if we should delete the string and qq resources and then create a new string with the quotes... it feels too much lift for a simple change, but otherwise it might not trigger changes for already translated langs.

It makes sense to me. Updated!

PR #4273, opened 2023-11-06
Improve HTML in native TextViews

cooltey:
https://phabricator.wikimedia.org/T347118


dbrant:
Closing this for the sake of tidying things up. We can explore adding more HTML features to TextViews in the future, but also balance it against adding too much highly-specific HTML support. Let's instead insist that content coming from the server shouldn't have any kind of fancy styling to begin with.

PR #4274, opened 2023-11-07
Add a bottom gradient as scroll indicator

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T350710

On small screens the user might not realize there is more content to scroll to.
Introducing a bottom gradient as s scroll indicator to resolve this.


cooltey (changes requested):
Feels like we can simply apply the following code to the scrollview.
```

            android:fadingEdge="vertical"
            android:requiresFadingEdge="vertical"
            android:fadingEdgeLength="20dp"
```

sharvaniharan:
Done @cooltey

PR #4276, opened 2023-11-08
Patroller Tasks: update talk template flows

cooltey:
https://phabricator.wikimedia.org/T350623


sharvaniharan:
**Lgtm**, but will make a PR based on this to add eventlogging where there are `//Todo:`s and then merge

sharvaniharan (changes requested):
@cooltey Just one change needed:
On an empty message templates screen, the more menu is still shown and the `Remove` option is still available even though there are no templates to remove. This gives rise to a weird state. Can we please not show the menu when there are no templates present?
remove

cooltey:
> @cooltey Just one change needed: On an empty message templates screen, the more menu is still shown and the `Remove` option is still available even though there are no templates to remove. This gives rise to a weird state. Can we please not show the menu when there are no templates present?

Done!

PR #4279, opened 2023-11-09
Fix: Final touches to PT data

sharvaniharan:
Final data changes


app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -346,6 +346,7 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback, L
         }
 
         binding.watchButton.setOnClickListener {
+            PatrollerExperienceEvent.logAction(if (isWatched) "unwatch_init" else "watch_init", "pt_toolbar")
```
cooltey, line 349: Does this need to be only in Patroller Tasks by using `sendPatrollerExperienceEvent`?

PR #4282, opened 2023-11-11
Convert time preferences to Instants

Isira-Seneviratne:
Convert the time preferences to `Instant`s and improve the time difference calculations.


cooltey:
Could you please help to provide the context of switching to `Instants` and its advantages?

Isira-Seneviratne:
I improved some of the time difference calculations, e.g. switching from `Duration.between` to `Instant.until`.

PR #4294, opened 2023-11-18
Save description in HistoryEntry

Pierre-Monier:
## Related to [T115035](https://phabricator.wikimedia.org/T115035) and [T270229](https://phabricator.wikimedia.org/T270229)

I store the `PageTitle` description in local db, also add a test to check that every persisted properties of `PageTitle` are correctly set on the `HistoryEntry` instance.

Screenshot 2023-11-18 at 22 11 24
Screenshot 2023-11-18 at 22 10 47


Pierre-Monier:
Working on a migration

Pierre-Monier:
Migration is done, adding a default value to empty string for the new description column on the HistoryEntry object

app/src/main/java/org/wikipedia/history/HistoryEntry.kt
```diff
@@ -30,10 +30,12 @@ class HistoryEntry(
     @Serializable(with = DateSerializer::class) var timestamp: Date = Date(),
     var source: Int = SOURCE_INTERNAL_LINK,
     var timeSpentSec: Int = 0,
+    var description: String = ""
 ) : Parcelable {
     constructor(title: PageTitle, source: Int, timestamp: Date = Date(), timeSpentSec: Int = 0) : this(title.wikiSite.authority(),
         title.wikiSite.languageCode, title.text, title.displayText, namespace = title.namespace,
-        timestamp = timestamp, source = source, timeSpentSec = timeSpentSec) {
+        timestamp = timestamp, source = source, timeSpentSec = timeSpentSec,
+        description = title.description ?: "") {
```
cooltey, line 38: You can use `orEmpty()` here.

app/src/main/java/org/wikipedia/history/db/HistoryEntryWithImageDao.kt
```diff
@@ -82,6 +82,10 @@ interface HistoryEntryWithImageDao {
             entryWithImage.displayTitle, 0, entryWithImage.namespace, entryWithImage.timestamp,
             entryWithImage.source, entryWithImage.timeSpentSec)
         entry.title.thumbUrl = entryWithImage.imageName
+
+        val hasDescription = entryWithImage.description?.isNotEmpty() ?: false
+        entry.title.description = if (hasDescription) entryWithImage.description else null
```
cooltey, line 87: You can simply assign the value instead.

```
entry.title.description = entryWithImage.description
```

cooltey (changes requested):

app/src/main/java/org/wikipedia/history/db/HistoryEntryWithImageDao.kt
Pierre-Monier: Ok, I wasn't sure if it was a problem :)

dbrant:
🎉 Thanks!

PR #4303, opened 2023-11-22
Optimize Gradle and Kotlin properties

kaeawc:
Now that Gradle and Kotlin are at the latest stable versions we can make some changes to `gradle.properties` file in isolation that should immediately improve local build performance. CI build performance really shouldn't change much yet. 

Just as a reminder for context, I want to work towards getting to [this end state](https://github.com/kaeawc/apps-android-wikipedia/actions/runs/6902262761/job/18779641998) where CI times are about a minute for full cache hits on leaf node changes.

I'm more than happy to talk through any of these changes in greater detail.

JVM arguments:
* UseParallelGC - JDK 9+ defaults to G1GC, but Parallel is slightly faster and the Wiki Android codebase doesn't use much JVM Heap anyway, so I'd recommend Parallel.
* SoftRefLRUPolicyMSPerMB - Defaults to `1000` which would mean `1000 seconds per GB`. At a heap size of `3GB` this would mean `50` minutes to release soft references. I've found through testing that lowering this time since these soft references are not expensive to recreate and often aren't even reused later in Gradle builds we can save on JVM heap size and overall build time. Therefore I recommend setting this to `1`.
* ReservedCodeCacheSize - Defaults to 48MB but allowing a bigger size can be beneficial to caching more native JVM bytecode. This is the only setting that I don't believe will effect build performance for this app today, I would need to do some iterative testing on this specific change to find the optimal value.
* Xmx & Xms - My research has shown that most JVM applications that know they need the full heap size should just start with it as a minimum. Also, 3GB is just about the maximum GitHub free runners can provide without OOMing, and Gradle will absolutely use all of the heap provided, so we might as well feed the daemons.
* MetaspaceSize - There is a [great set of articles](https://stuefe.de/posts/metaspace/what-is-metaspace/) on Metaspace that explain why the JVM default of unlimited MaxMetaspaceSize makes sense. In previous versions of Gradle we needed a cap on the maximum size to prevent unbounded growth due to some internal bugs. Those memory bugs seem to have been fixed in recent versions of Gradle, yet the default for MetaspaceSize is 20MB which means there can be many metaspace related GCs early on in builds. Similar to Xms, we want to avoid unnecessary and expected GCs, so increasing this to 1GB is a safe bet that helps with Gradle build performance.

Gradle
* org.gradle.caching by default is false, which means the local Gradle build cache is not used. Setting it to true should be an improvement to local developer build speed as well as eventually on CI when I setup the custom caching in GitHub Actions.
* org.gradle.parallel by default is false and therefore doesn't allow parallel module building. Setting it to true will allow this.
* Gradle Configuration Cache enabled with latest file entry workarounds from https://github.com/androidx/androidx. The official Gradle Github Action does not support caching this on CI for security vulnerability reasons, but A) it really helps with local builds and B) it will help CI builds that do not use internal build secrets. Not only does it help with configuration reuse, but also dependency resolution time. I've used this in much larger projects (Hinge) with many more annotation processors and libraries.

Kotlin
* `useClasspathSnapshot` is an incremental compilation optimization for Kotlin that has been available since Kotlin 1.7 and in my experience is pretty safe to use.

Android settings
* Experimental Build Config as Bytecode - Latest Android Studio versions support this. If Wiki Android devs use older Android Studio versions I'm happy to check whether this is supported on those versions.
* Experimental R8 multiple workers


kaeawc:
Is there anything I can clarify about why this change is helpful? Would you prefer it to be broken down into smaller changes?

cooltey:
> Is there anything I can clarify about why this change is helpful? Would you prefer it to be broken down into smaller changes?

Thanks for submitting the PR! I did a quick check on building the project, and it looks good to me. But we will need more time to discuss with engineers in the team whether should we put the customized configuration or the default one from the Android Studio.

dbrant:
@kaeawc Thanks! It has been a while since we revisited our `gradle.properties`. However, I do feel like these changes are a bit much. It's a bit mind-boggling that developers need to be aware of this many tweaks to get their environment running optimally.

I've tried these settings for a little while, and I think I see a _slightly_ appreciable improvement in local build times (Windows 11, i9, 64gb), but I'm guessing that's mostly attributable to `org.gradle.caching` and `org.gradle.parallel`. Can we start with those two, and see how it goes?

PR #4308, opened 2023-11-29
Nearby: Add View on map optionto customize toolbar

cooltey:
https://phabricator.wikimedia.org/T351394

### DIFF
--- a/app/src/main/java/org/wikipedia/page/action/PageActionItem.kt
+++ b/app/src/main/java/org/wikipedia/page/action/PageActionItem.kt
@@ -58,6 +58,11 @@ enum class PageActionItem constructor(val id: Int,
             cb.onViewEditHistorySelected()
         }
     },
+    VIEW_ON_MAP(13, R.id.page_view_on_map, R.string.action_item_view_on_map, R.drawable.baseline_location_on_24, true) {
+        override fun select(cb: Callback) {
+            cb.onViewOnMapSelected()
+        }
+    },
     NEW_TAB(9, R.id.page_new_tab, R.string.menu_new_tab, R.drawable.ic_add_gray_white_24dp) {
         override fun select(cb: Callback) {
             cb.onNewTabSelected()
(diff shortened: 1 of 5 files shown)


sharvaniharan (changes requested):
Hi @cooltey 

When I click on `View on map` I see that it just loads the map to a general locations with a zoomed-out approach like this: https://youtube.com/shorts/0usNNFKtGVg
I was expecting that we will zoom to that article on the map.. is that not the case?

cooltey:
> Hi @cooltey
> 
> When I click on `View on map` I see that it just loads the map to a general locations with a zoomed-out approach like this: https://youtube.com/shorts/0usNNFKtGVg I was expecting that we will zoom to that article on the map.. is that not the case?

Done!

sharvaniharan:
@cooltey I think the recent change opens up the bottom sheet on loading the map, with markers on the background. I was thinking of a more smoother experience like this:
- We get the location from the article
- When user clicks on `View on map` it takes them to the places fragment
- Here let us pass on that `LatLng`, and zoom slowly to that `marker` at that location, similar to the cluster experience
- And, let us not open the bottom sheet, which is actually looking a bit more disruptive to me.
I feel this experience will be what they expect. Please let me know your thoughts.

cooltey:
> @cooltey I think the recent change opens up the bottom sheet on loading the map, with markers on the background. I was thinking of a more smoother experience like this:
> 
>     * We get the location from the article
> 
>     * When user clicks on `View on map` it takes them to the places fragment
> 
>     * Here let us pass on that `LatLng`, and zoom slowly to that `marker` at that location, similar to the cluster experience
> 
>     * And, let us not open the bottom sheet, which is actually looking a bit more disruptive to me.
>       I feel this experience will be what they expect. Please let me know your thoughts.

Sounds good to me, I can remove the open-up logic from the PR. But let me hold this PR until we merge the cluster PR to the feature branch 😅

cooltey:
> @cooltey I think the recent change opens up the bottom sheet on loading the map, with markers on the background. I was thinking of a more smoother experience like this:
> 
>     * We get the location from the article
> 
>     * When user clicks on `View on map` it takes them to the places fragment
> 
>     * Here let us pass on that `LatLng`, and zoom slowly to that `marker` at that location, similar to the cluster experience
> 
>     * And, let us not open the bottom sheet, which is actually looking a bit more disruptive to me.
>       I feel this experience will be what they expect. Please let me know your thoughts.

Done. Now it will not show the bottom sheet and also will zoom in to the max level.

Bhing09 (approved):

PR #4309, opened 2023-11-29
Places: Clustering

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T351159


app/src/main/java/org/wikipedia/nearby/NearbyFragment.kt
```diff
@@ -172,6 +186,41 @@ class NearbyFragment : Fragment(), LinkPreviewDialog.Callback {
...
+            androidx.core.util.Pair(
+                0, ContextCompat.getColor(requireActivity(), ResourceUtil.getThemedAttributeId(requireContext(), R.attr.success_color))
+            )
+        )
+        val clusterOptions = ClusterOptions()
+            .withClusterRadius(60)
+            .withColorLevels(clusterColorLayers)
+            .withTextSize(literal(12f))
+            .withTextField(Expression.toString(get(POINT_COUNT)))
+            .withTextColor(Expression.color(fetchAttributeId(R.attr.paper_color)))
```
cooltey, line 200: We should use `ResourceUtil.getThemedColor(requireContext(), R.attr.paper_color)` instead and I think it should be fine if we are using the `getThemedColor` instead of creating a new method of referencing it.

app/src/main/java/org/wikipedia/nearby/NearbyFragment.kt
```diff
@@ -172,6 +186,41 @@ class NearbyFragment : Fragment(), LinkPreviewDialog.Callback {
...
+            style.getLayer(CLUSTER_TEXT_LAYER_ID)?.apply {
+                this.setProperties(
+                    textFont(MARKER_FONT_STACK),
+                    textIgnorePlacement(true),
+                    textAllowOverlap(true)
+                )
+            }
+            style.getLayer(CLUSTER_CIRCLE_LAYER_ID)?.apply {
+                this.setProperties(
+                    circleStrokeColor(fetchAttributeId(R.attr.paper_color)),
```
cooltey, line 215: Same as above. We should use `ResourceUtil.getThemedColor(requireContext(), R.attr.paper_color)` instead.

app/src/main/java/org/wikipedia/nearby/NearbyFragment.kt
sharvaniharan: Have changed this but I anticipate many dynamic colors like this by the end so had created a single place for it. If this is more readable thats fine.

PR #4311, opened 2023-11-30
Fix: make sure getting the coordinates from the page/summary endpoint

cooltey:
When checking the `page/summary` endpoint, you will notice the coordinates are from `coordinates` attribute instead of `geo`; it looks like it has broken for a while.
https://en.wikipedia.org/api/rest_v1/page/summary/Lake_Natoma

This PR also fixes the incorrect HTML title shown on the map that is sent from the app.

TODO:
The test will fail if the `Location` is not null. Cannot find the root cause between the `PageSummary` and `PageProperties`. Will discuss this in the engineers' sync.


cooltey:
Test result:
```
Expected: is <PageProperties(pageId=19004, namespace=MAIN, revisionId=1015030565, lastModified=Tue Mar 30 01:43:22 PDT 2021, displayTitle=Moscow, editProtectionStatus=, isMainPage=false, leadImageUrl=https://upload.wikimedia.org/wikipedia/commons/thumb/2/23/Spasskaya_Tower_and_the_St._Basil%27s_Cathedral.jpg/640px-Spasskaya_Tower_and_the_St._Basil%27s_Cathedral.jpg, leadImageName=Spasskaya_Tower_and_the_St._Basil's_Cathedral.jpg, leadImageWidth=320, leadImageHeight=242, geo=Location[ 55.755833,37.617222 acc=??? t=?!? et=?!?], wikiBaseItem=Q649, descriptionSource=local, canEdit=false)>
     but: was <PageProperties(pageId=19004, namespace=MAIN, revisionId=1015030565, lastModified=Tue Mar 30 01:43:22 PDT 2021, displayTitle=Moscow, editProtectionStatus=, isMainPage=false, leadImageUrl=https://upload.wikimedia.org/wikipedia/commons/thumb/2/23/Spasskaya_Tower_and_the_St._Basil%27s_Cathedral.jpg/640px-Spasskaya_Tower_and_the_St._Basil%27s_Cathedral.jpg, leadImageName=Spasskaya_Tower_and_the_St._Basil's_Cathedral.jpg, leadImageWidth=320, leadImageHeight=242, geo=Location[ 55.755833,37.617222 acc=??? t=?!? et=?!?], wikiBaseItem=Q649, descriptionSource=local, canEdit=false)>

```

The result is the same as the expected one.

app/src/test/java/org/wikipedia/test/ParcelableTest.kt
```diff
@@ -41,6 +41,8 @@ class ParcelableTest {
     fun testPagePropertiesFromSummary() {
         val json = TestFileUtil.readRawFile("rb_page_summary_geo.json")
         val summary = JsonUtil.decodeFromString<PageSummary>(json)!!
+        // FIXME: somehow the Location object is different even though the values are the same. Set null to bypass the test temporary
```
dbrant, line 44: This test fails because the `PageProperties` class contains a `Location` field, which is not a primitive type. If you parcel and then unparcel a `PageProperties` object, the actual instance of the Location field will be different in the unparcelled object than the parcelled object.

The issue is really that the test isn't smart enough to perform a "deep-equal" between the two objects, and instead it's just comparing the literal instances.

I'm actually not sure I see the point of these tests anyway. We're testing whether `Parcelable` objects can be parcelled and unparcelled... Are we testing whether the Android SDK works? Whether the Java language works? What exactly could fail here?
Anyway, I think the fix in this PR is good to merge.

PR #4313, opened 2023-12-01
Fix playback of videos in Gallery.

dbrant:
Video playback is currently broken.
This is because our logic for selecting the best possible derivative for playback was not smart enough, and was blindly selecting the "last" derivative.
The list of derivatives can change, and indeed it has changed to include an unplayable derivative as the "last" one.
This improves the logic of selecting the proper derivative.

https://phabricator.wikimedia.org/T352315

### DIFF
--- a/app/src/main/java/org/wikipedia/gallery/ImageInfo.kt
+++ b/app/src/main/java/org/wikipedia/gallery/ImageInfo.kt
@@ -46,10 +46,11 @@ class ImageInfo {
     val width = 0
     val height = 0
 
+    @Suppress("KotlinConstantConditions")
     fun getBestDerivativeForSize(widthDp: Int): Derivative? {
         var derivative: Derivative? = null
         derivatives.forEach {
-            if (it.width > 0 && it.width < widthDp) {
+            if (it.width in 1..<widthDp) {
                 if (derivative == null || it.width > derivative!!.width) {
                     derivative = it
                 }
--- a/app/src/main/java/org/wikipedia/gallery/ImageInfo.kt
+++ b/app/src/main/java/org/wikipedia/gallery/ImageInfo.kt
@@ -46,16 +46,25 @@ class ImageInfo {
     val width = 0
     val height = 0
 
-    val bestDerivative get() = derivatives.lastOrNull()
+    fun getBestDerivativeForSize(widthDp: Int): Derivative? {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/gallery/ImageInfo.kt
```diff
@@ -46,16 +46,25 @@ class ImageInfo {
     val width = 0
     val height = 0
 
-    val bestDerivative get() = derivatives.lastOrNull()
+    fun getBestDerivativeForSize(widthDp: Int): Derivative? {
```
cooltey, line 49: Can we add `@Suppress("KotlinConstantConditions")` for this function?

app/src/main/java/org/wikipedia/gallery/ImageInfo.kt
```diff
@@ -46,16 +46,25 @@ class ImageInfo {
...
+        var derivative: Derivative? = null
+        derivatives.forEach {
+            if (it.width > 0 && it.width < widthDp) {
+                if (derivative == null || it.width > derivative!!.width) {
+                    derivative = it
+                }
+            }
+        }
```
cooltey, line 57: Would it be better to use `filter` and `mayByOrNull`?
```
derivatives.filter { it.width < widthDp && it.width < widthDp }.maxByOrNull { it.width }
```

I tried a couple of times and it looks like using `filter` and `maxByOrNull` can mostly improve the efficiency
```
forEach                581.093us
filter+mayByOrNull     53.75us

forEach                62.032us
filter+mayByOrNull     67.604us

forEach                51.51us
filter+mayByOrNull     50.781us

forEach                118.489us
filter+mayByOrNull     57.239us
```

cooltey (changes requested):

app/src/main/java/org/wikipedia/gallery/ImageInfo.kt
dbrant: of course!

app/src/main/java/org/wikipedia/gallery/ImageInfo.kt
cooltey: Thanks for the information, that explains the benefit of using the original logic you put in the code! I checked the library and you are correct!

PR #4319, opened 2023-12-05
Places: Search

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T351158
**Phab:** https://phabricator.wikimedia.org/T351397

This change has:

- Multilingual search initiated from places screen
- Filter for language on filter screen
- Ui update in size for the selected marker - on Search and on visiting from article
- Updates the use of Location to LatLng


app/src/main/java/org/wikipedia/places/PlacesActivity.kt
```diff
@@ -2,25 +2,35 @@ package org.wikipedia.places
...
     public override fun createFragment(): PlacesFragment {
-        return PlacesFragment.newInstance(intent.parcelableExtra(EXTRA_WIKI)!!, intent.parcelableExtra(EXTRA_TITLE), intent.parcelableExtra(EXTRA_LOCATION))
+        return PlacesFragment.newInstance(intent.parcelableExtra(EXTRA_TITLE), intent.parcelableExtra(EXTRA_LOCATION))
     }
 
     companion object {
-        const val EXTRA_WIKI = "wiki"
         const val EXTRA_TITLE = "pageTitle"
         const val EXTRA_LOCATION = "location"
-        fun newIntent(context: Context, wiki: WikiSite, pageTitle: PageTitle? = null, location: Location? = null): Intent {
```
sharvaniharan, line 21: Persisting the code for places makes this redundant and can always fallback to app language or get it from pageTitle from special cases.

app/src/main/java/org/wikipedia/places/PlacesFragment.kt
```diff
@@ -88,20 +95,48 @@ class PlacesFragment : Fragment(), LinkPreviewDialog.Callback, MapboxMap.OnMapCl
...
+    private var searchRadius: Int = 50
+    private var zoom: Double = 15.0
+    private var magnifiedMarker: Symbol? = null
 
     private val locationPermissionRequest = registerForActivityResult(ActivityResultContracts.RequestMultiplePermissions()) { permissions ->
         when {
             permissions.getOrDefault(Manifest.permission.ACCESS_FINE_LOCATION, false) ||
             permissions.getOrDefault(Manifest.permission.ACCESS_COARSE_LOCATION, false) -> {
                 startLocationTracking()
-                goToLastKnownLocation(1000, viewModel.location, viewModel.pageTitle != null)
```
sharvaniharan, line 97: @cooltey I removed this change as the zoom was not that much different between 15.99 and 15. Instead it now enlarges the icon. Robin has mentioned 2x in the specs but I have settled for 1.75. will discuss this during design review.

sharvaniharan:
Open for review :-)

dbrant (changes requested):
This change seems to cause some weird behavior when panning through the map by small amounts. The markers seem to change from one batch to another completely different batch:

https://github.com/wikimedia/apps-android-wikipedia/assets/1682214/424ef9ed-2b2d-49fb-870d-f6410ff4f852

This does not happen in the base Places branch:

https://github.com/wikimedia/apps-android-wikipedia/assets/1682214/d4bd7e9f-d012-4abf-9f0c-417bca7af139

sharvaniharan:
@dbrant It is actually happening on both places. In the base branch video, if you see at the :20 second mark, some of the markers disappear. In this branch it is more pronounced because I have removed the "minimum distance to move before updating markers" condition. I believe the root cause is the cache size we have and the logic around it to remove the oldest marker from the AnnotationCache when we max the cache size. I feel removing the minimum distance moved logic made the marker update experience much better, so I am thinking i will keep that and try to increase cache size and check performance.

dbrant:
> @dbrant It is actually happening on both places. In the base branch video, if you see at the :20 second mark, some of the markers disappear. In this branch it is more pronounced because I have removed the "minimum distance to move before updating markers" condition. I believe the root cause is the cache size we have and the logic around it to remove the oldest marker from the AnnotationCache when we max the cache size. I feel removing the minimum distance moved logic made the marker update experience much better, so I am thinking i will keep that and try to increase cache size and check performance.

I'm not sure I agree.  I think the root cause is that the calculation of the `searchRadius` was moved to an incorrect location.  It looks like it's now calculated when the map is first initialized, which would give it an arbitrarily large radius. The radius should be calculated just before the API call is made (i.e. it should always be the current size of the viewport.)

And because you're making an API call with a huge radius, the returned items could be arbitrarily far, and in an arbitrary order, and will therefore clobber our cache of markers.

sharvaniharan:
@dbrant That is a very good catch. I have fixed that, but just that doesnt still fix the issue, as at a given zoom level and a heavy article activity area, we can have more than 64 markers. So i had the same issue after the searchRadius fix as well

https://github.com/wikimedia/apps-android-wikipedia/assets/9540770/d4da5c6e-e519-48a9-8be3-a4cacef09ec0

Increasing the cache size along with the searchRadius fix, fixes it, but we can discuss the best cache size in our sync

https://github.com/wikimedia/apps-android-wikipedia/assets/9540770/786f3aa2-bce2-473d-9872-7a23651b6d25

dbrant:
Merging to keep things moving (and so that Design can have a single APK to test with).
Please look for code review / design feedback on the feature branch.

PR #4320, opened 2023-12-06
Fix: display ordered lists and numbers in native TextViews

cooltey:
https://phabricator.wikimedia.org/T347118

This breaks out the #4273 into a smaller PR.

In this PR, it adds an option of hiding the Headings option from the wikitext keyboard since the talk page does not support them.


app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
```diff
@@ -187,9 +189,36 @@ class CustomHtmlParser constructor(private val handler: TagHandler) : TagHandler
...
+                val start = output.length - split.last().length - 1
+                val replaceStr = "$listItemCount. ${split.last()}"
+                output.replace(start - 1, output.length - 1, replaceStr)
+
+                val spans = output.getSpans<LeadingMarginSpan>(output.length)
+                if (spans.isNotEmpty()) {
+                    val span = spans.last()
+                    val startSpan = output.getSpanStart(span)
+                    output.removeSpan(span)
+                    output.setSpan(LeadingMarginSpan.Standard(50 * listParents.size), startSpan, output.length, 0)
```
dbrant, line 218: What does the number `50` mean? If it's a dimension of pixels, it should probably be expressed in `dp` units, instead of raw pixels. (probably 16dp)

dbrant (changes requested):

app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
cooltey: Thanks, it makes sense to me. Done!

PR #4321, opened 2023-12-06
Fix: allow the app shortcuts works if the app is not fully closed

cooltey:
https://phabricator.wikimedia.org/T352800


dbrant:
This change looks very small and simple, but I think it might have some unintended consequences that we need to understand.
We have a lot of activities and flows that depend on a certain order, and it's hard to believe that this change won't impact any of them.
Can we test all of our workflows and ensure that nothing is impacted in a weird way?

cooltey:
> This change looks very small and simple, but I think it might have some unintended consequences that we need to understand. We have a lot of activities and flows that depend on a certain order, and it's hard to believe that this change won't impact any of them. Can we test all of our workflows and ensure that nothing is impacted in a weird way?

That makes sense to me. I rechecked the document and [found this](https://developer.android.com/develop/ui/views/launch/shortcuts/managing-shortcuts#start-one) in the Android Developers, which solves the issue by adding `Intent.FLAG_ACTIVITY_CLEAR_TASK` to the intents.

PR #4334, opened 2023-12-13
Places: Update ‘article preview’ bottom sheet for places

cooltey:
https://phabricator.wikimedia.org/T351395


dbrant (changes requested):
There's nothing wrong that I see in the functionality that you're adding here, but I'm nervous that the scope of all these changes is blowing up a bit.

I wonder if there is a way for us to "contain" the changes inside the LinkPreviewDialog class and its ViewModel, instead of the changes spilling into a dozen other activities and components.

For a start, it looks like there's a new `onLinkPreviewViewOnMap()` callback function, which needs to be handled by _every consumer_ of link previews. Wouldn't it be simpler for LinkPreviewDialog to just call `startActivity(PlacesActivity)` on its own?

cooltey:
> There's nothing wrong that I see in the functionality that you're adding here, but I'm nervous that the scope of all these changes is blowing up a bit.
> 
> I wonder if there is a way for us to "contain" the changes inside the LinkPreviewDialog class and its ViewModel, instead of the changes spilling into a dozen other activities and components.
> 
> For a start, it looks like there's a new `onLinkPreviewViewOnMap()` callback function, which needs to be handled by _every consumer_ of link previews. Wouldn't it be simpler for LinkPreviewDialog to just call `startActivity(PlacesActivity)` on its own?

Yes. Done!

app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
```diff
@@ -173,26 +224,43 @@ class LinkPreviewDialog : ExtendedBottomSheetDialogFragment(), LinkPreviewErrorV
         showError(throwable)
     }
 
+    override fun onStart() {
+        super.onStart()
+        if (viewModel.fromPlaces) {
+            disableBackgroundDim()
+            BottomSheetBehavior.from(requireView().parent as View).state = BottomSheetBehavior.STATE_EXPANDED
```
dbrant, line 231: This makes the dialog very jumpy and glitchy when being shown. Can we remove this logic? (and tell Robin it can't work this way.) (maybe we could increase the default `bottomSheetPeekHeight` a little bit instead?)

app/src/main/java/org/wikipedia/places/PlacesFragment.kt
```diff
@@ -485,6 +509,100 @@ class PlacesFragment : Fragment(), MapboxMap.OnMapClickListener {
...
+    override fun onLinkPreviewAddToList(title: PageTitle, isInReadingList: Boolean, anchor: View?) {
+        anchor?.let {
+            onAddToReadingListClick(title, isInReadingList, it)
+        }
+    }
+    override fun onExpirySelect(expiry: WatchlistExpiry) {
+        viewModel.watchOrUnwatch(expiry, false)
+        ExclusiveBottomSheetPresenter.dismiss(childFragmentManager)
+    }
+
```
dbrant, line 605: Similar to previous comments, can all of this Watchlist and Reading List logic be self-contained in the Link Preview component?  (with some logic possibly going into WatchlistExpiryDialog)
This will make all of the logic reusable wherever Link Previews are shown.

app/src/main/res/values/strings.xml
```diff
@@ -1675,7 +1675,13 @@
     <string name="menu_single_webview_open_external">Open in system browser</string>
     <!-- Donor experience -->
 
+    <!-- Places -->
     <string name="places_title">Places</string>
     <string name="places_permissions_denied">Location permission is required to see your location on the map. Please try again.</string>
+    <string name="places_link_preview_dialog_read_button">Read</string>
```
dbrant, line 1681: These strings apply to the Link Preview component, so they shouldn't be mentioning "places" in their names.

app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
```diff
@@ -106,14 +137,12 @@ class LinkPreviewDialog : ExtendedBottomSheetDialogFragment(), LinkPreviewErrorV
...
-            PopupMenu(requireActivity(), binding.linkPreviewOverflowButton).run {
-                inflate(R.menu.menu_link_preview)
-                setOnMenuItemClickListener(menuListener)
-                show()
-            }
+            setupOverflowMenu()
         }
         L10nUtil.setConditionalLayoutDirection(binding.root, viewModel.pageTitle.wikiSite.languageCode)
 
+        binding.linkPreviewTopHandle.isVisible = viewModel.fromPlaces
```
dbrant, line 144: Don't we want to show the handle in all cases, not just Places?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
cooltey: Sounds good to me. I checked if we put a higher peek height (e.g. `400dp`), it will still have the jumpy issue, and it looks like showing the gallery causes the issue. I'll let Robin know about the issue and remove it, thanks!

app/src/main/res/values/strings.xml
cooltey: Done.

app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
cooltey: I'll ask Robin about this.

app/src/main/java/org/wikipedia/places/PlacesFragment.kt
cooltey: Please review the separate PR of moving the logic of changing the watchlist expiry here: #4365

dbrant:
@cooltey  Can we rebase this to be merged into `main` instead of the feature branch? That way we can have these updates already in `main` (which shouldn't impact the default behavior), and ready to be integrated with Places.

cooltey:
> @cooltey Can we rebase this to be merged into `main` instead of the feature branch? That way we can have these updates already in `main` (which shouldn't impact the default behavior), and ready to be integrated with Places.

Done #4382

PR #4335, opened 2023-12-14
Fix: update widget background color

cooltey:
I checked the logic of manually parsing the main page, it looks like it needs more tuning. I think we can wait until we are ready for the widget updates.


shashankiitbhu:
Can I work on this? @cooltey

dbrant:
Merging, because this makes the widget slightly less broken.

PR #4339, opened 2023-12-19
Fix incorrect logic for showing Patroller Tasks feedback dialog.

dbrant:
This fixes a condition where the feedback dialog gets shown every time a rollback is performed, because of incorrect assumptions about the state of the current child fragments.


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
```diff
@@ -106,10 +106,7 @@ class SuggestedEditsCardsFragment : Fragment(), MenuProvider, SuggestedEditsItem
         super.onViewCreated(view, savedInstanceState)
         requireActivity().addMenuProvider(this, viewLifecycleOwner, Lifecycle.State.RESUMED)
         setInitialUiState()
-        // Disable offscreenPageLimit for recentEdits can avoid the tooltips showing incorrectly.
-        if (action != VANDALISM_PATROL) {
-            binding.cardsViewPager.offscreenPageLimit = 2
-        }
```
cooltey, line 112: Hi @dbrant 

The reason why I put the logic here was trying to avoid the following issue:

dbrant: ok, so let's fix the underlying issue. Done!

PR #4340, opened 2023-12-19
Patroller tasks: improve and update algorithm for fetching new changes.

dbrant:
This improves the way we fetch new recent changes for the Patroller Tasks ViewPager, to make the experience more fluid for a user who wants to swipe through changes quickly.


app/src/main/java/org/wikipedia/suggestededits/provider/EditingSuggestionsProvider.kt
```diff
@@ -366,41 +372,55 @@ object EditingSuggestionsProvider {
...
                     while (this.coroutineContext.isActive) {
                         try {
-                            val recentChanges = SuggestedEditsRecentEditsViewModel.getRecentEditsCall(WikiSite.forLanguageCode(lang), 10, null, mutableListOf())
-                                .first.sortedByDescending { it.curRev }
-                            var maxRevId = 0L
+                            // If we have been reset, then fetch a few *older* changes, so that the user
+                            // has a few changes to flip through. Otherwise, start fetching *newer* changes,
+                            // starting from the last recorded timestamp.
+                            val triple = if (revertCandidateLastRevId == 0L)
+                                SuggestedEditsRecentEditsViewModel.getRecentEditsCall(WikiSite.forLanguageCode(lang),
```
cooltey, line 386: Would it be better if we move `WikiSite.forLanguageCode(lang)` out of the `while`?

app/src/main/java/org/wikipedia/suggestededits/provider/EditingSuggestionsProvider.kt
```diff
@@ -366,41 +372,55 @@ object EditingSuggestionsProvider {
...
+                                10, "now", "older", null, mutableListOf())
```
cooltey, line 387: Not sure but should we make this more consistent that we use `Date().toInstant().toString()`  in `SuggestedEditsRecentEditsViewModel` but use `now` here? I am wondering which one is better 🤔

cooltey (commented):
The logic looks good to me! Few comments in `EditingSuggestionsProvider`.

app/src/main/java/org/wikipedia/suggestededits/provider/EditingSuggestionsProvider.kt
dbrant: sure!

app/src/main/java/org/wikipedia/suggestededits/provider/EditingSuggestionsProvider.kt
dbrant: fair point, let's make it consistent and use Date().toInstant(),toString() in both places.

PR #4346, opened 2023-12-21
Remove even more unnecessary callbacks from LinkPreviewDialog.

dbrant:


app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
```diff
@@ -429,16 +427,6 @@ class TalkTopicActivity : BaseActivity(), LinkPreviewDialog.Callback {
...
     }
 
-    override fun onLinkPreviewLoadPage(title: PageTitle, entry: HistoryEntry, inNewTab: Boolean) {
-        startActivity(if (inNewTab) PageActivity.newIntentForNewTab(this, entry, title) else
-            PageActivity.newIntentForCurrentTab(this, entry, title, false))
-    }
-
-    override fun onLinkPreviewAddToList(title: PageTitle) {
-        ExclusiveBottomSheetPresenter.show(supportFragmentManager,
-                AddToReadingListDialog.newInstance(title, Constants.InvokeSource.TALK_TOPIC_ACTIVITY))
```
cooltey, line 439: It looks like the `InvokeSource` should be `TALK_TOPIC_ACTIVITY` but now it will be `LINK_PREVIEW_MENU`. Should we pass the `InvokeSource` to the `LinkPreviewDialog` for analytics?

app/src/main/java/org/wikipedia/talk/template/AddTemplateActivity.kt
```diff
@@ -245,16 +240,6 @@ class AddTemplateActivity : BaseActivity(), LinkPreviewDialog.Callback, UserMent
...
     }
 
-    override fun onLinkPreviewLoadPage(title: PageTitle, entry: HistoryEntry, inNewTab: Boolean) {
-        startActivity(if (inNewTab) PageActivity.newIntentForNewTab(this, entry, title) else
-            PageActivity.newIntentForCurrentTab(this, entry, title, false))
-    }
-
-    override fun onLinkPreviewAddToList(title: PageTitle) {
-        ExclusiveBottomSheetPresenter.show(supportFragmentManager,
-                AddToReadingListDialog.newInstance(title, Constants.InvokeSource.TALK_REPLY_ACTIVITY))
```
cooltey, line 255: Same here.

cooltey (changes requested):

app/src/main/java/org/wikipedia/talk/template/AddTemplateActivity.kt
dbrant: Using `TALK_REPLY_ACTIVITY` was already incorrect here. So, even if using the default `LINK_PREVIEW_MENU` is also incorrect, at least it will be incorrect with less code.

app/src/main/java/org/wikipedia/talk/TalkTopicActivity.kt
dbrant: I noticed this, but it was never consistent for all the other activities (everywhere else uses `LINK_PREVIEW_MENU`), so why not here, too? I am not aware of any analytics that depend on the invokeSource here.

cooltey: Ok, that makes sense to me.

PR #4370, opened 2024-01-08
Simplify the logic of adding/moving article to reading list

cooltey:
This PR simplified the logic of adding/moving an article to the list, which now can use either `ReadingListBehaviorsUtil.addToDefaultList` or `ReadingListBehaviorsUtil.moveToList` with an optional dialog dismiss listener.

This PR also fixed the issue that some views have incorrect invoke sources implemented.

### DIFF
--- a/app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
+++ b/app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
@@ -223,6 +223,7 @@ class LinkPreviewDialog : ExtendedBottomSheetDialogFragment(), LinkPreviewErrorV
 
     private fun doAddToList() {
         ReadingListBehaviorsUtil.addToDefaultList(requireActivity(), viewModel.pageTitle, true, Constants.InvokeSource.LINK_PREVIEW_MENU)
+        dialog?.dismiss()
     }
 
     private fun showPreview(contents: LinkPreviewContents) {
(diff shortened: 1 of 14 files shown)


cooltey:
I added `High Priority` label so this behavior can be used in #4334 after merging into the main branch.

sharvaniharan:
Lgth... will let @dbrant take a look and merge!

app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
```diff
@@ -227,14 +222,7 @@ class LinkPreviewDialog : ExtendedBottomSheetDialogFragment(), LinkPreviewErrorV
...
     private fun doAddToList() {
-        addToListCallback.let {
-            if (it != null) {
-                it.onLinkPreviewAddToList(viewModel.pageTitle)
-            } else {
-                ExclusiveBottomSheetPresenter.showAddToListDialog(requireActivity().supportFragmentManager,
-                    viewModel.pageTitle, Constants.InvokeSource.LINK_PREVIEW_MENU)
-            }
-        }
+        ReadingListBehaviorsUtil.addToDefaultList(requireActivity(), viewModel.pageTitle, true, Constants.InvokeSource.LINK_PREVIEW_MENU)
```
dbrant, line 225: Should we call `dismiss()` here, so that the LinkPreviewDialog is dismissed and the snackbar can be seen?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
cooltey: Done!

dbrant:
looks good, but will merge after Beta release, just to let this change breathe until next release.

PR #4371, opened 2024-01-09
Create multiple actionable notifications on Android 7.0 and later

Isira-Seneviratne:
Create the individual notifications and [group them together](https://developer.android.com/develop/ui/views/notifications/group) when there are more than two unread. This change allows the individual notifications to be actionable on Android 7.0 and later (the existing behaviour is preserved for older Android versions).


dbrant:
It was a specific design choice to not have more than two system-level notifications, and bring the user into the app instead.

PR #4373, opened 2024-01-09
Fix full-text search.

dbrant:
It looks like full-text search has been broken for a while (ever since switching to `SearchResultsViewModel`).
More specifically, the full-text search only happens when the prefix search returns zero items. But in fact the full-text search should happen anytime the prefix search returns fewer than the _PagingSource_ asks for.

https://phabricator.wikimedia.org/T354592


sharvaniharan:
Full text search did not solve the issue ... This PR is good to merge but I still see only 1 result 🤔

cooltey:
I can see multiple results now when trying to search `4th Avenue 9th street`

PR #4379, opened 2024-01-12
Don't create unnecessary bitmaps.

dbrant:
@cooltey  This is what I was talking about -- observe how this code no longer creates any bitmaps (or even pulls any dirty bitmaps from the pool), except the initial `markerBitmapBase`.

(to be merged into your ui-update PR)


cooltey:
This is great! Thank you!

PR #4381, opened 2024-01-12
String resources for Temporary Accounts feature.

dbrant:
While we're at it, let's add some string resources for the future IP masking work, so that these can start getting translated, too.


sharvaniharan:
@dbrant I see some long copies... are we sure they are finalized 😅  Just making sure large translations dont get discarded

dbrant:
@sharvaniharan It's ok to let these sit for a bit before merging. I'll confirm next week.

dbrant:
@cooltey  This can now be merged, so that these strings can be digested by TWN.

PR #4382, opened 2024-01-12
Update link preview dialog for Places

cooltey:
Request from [#4334#issuecomment-1889163136](https://github.com/wikimedia/apps-android-wikipedia/pull/4334#issuecomment-1889163136)


app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
```diff
@@ -135,6 +167,20 @@ class LinkPreviewDialog : ExtendedBottomSheetDialogFragment(), LinkPreviewErrorV
...
 
+    private fun setupOverflowMenu() {
+        val popupMenu = PopupMenu(requireActivity(), binding.linkPreviewOverflowButton)
+        popupMenu.inflate(R.menu.menu_link_preview)
+        popupMenu.menu.findItem(R.id.menu_link_preview_add_to_list).isVisible = !viewModel.fromPlaces
+        popupMenu.menu.findItem(R.id.menu_link_preview_share_page).isVisible = !viewModel.fromPlaces
+        popupMenu.menu.findItem(R.id.menu_link_preview_watch).isVisible = viewModel.fromPlaces
+        popupMenu.menu.findItem(R.id.menu_link_preview_watch).title = getString(if (viewModel.isWatched) R.string.menu_page_unwatch else R.string.menu_page_watch)
+        popupMenu.menu.findItem(R.id.menu_link_preview_open_in_new_tab).isVisible = viewModel.fromPlaces
+        popupMenu.menu.findItem(R.id.menu_link_preview_view_on_map).isVisible = !viewModel.fromPlaces && viewModel.location != null
```
dbrant, line 178: Should this be `viewModel.fromPlaces`? (not `!`)

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
cooltey: @dbrant This is a request from designer: https://phabricator.wikimedia.org/T351395#9410072

dbrant: @cooltey I meant to say that the behavior doesn't seem correct:
If I go to the article `Paris`, then click on any link, every link now has an option to "View on map", and they all lead to Paris. The option to "View on map" should be to go to the location of the link, not the location of the article from where the link came.

cooltey: Got it. Fixed! We can use the correct location from `pageSummary` if we do not send the location manually.

PR #4383, opened 2024-01-14
Places Filter screen: Design tweaks

sharvaniharan:


app/src/main/res/values/styles.xml
```diff
@@ -305,7 +305,6 @@
         <item name="elevation">4dp</item>
         <item name="android:outlineAmbientShadowColor" tools:targetApi="p">@color/black</item>
         <item name="android:outlineSpotShadowColor" tools:targetApi="p">@color/black</item>
-        <item name="android:background">?attr/paper_color</item>
```
dbrant, line 308: Removing this attribute introduces significant regressions in the appearance of other Toolbars. e.g. in MainActivity, upon scrolling down. This attribute should be kept, until we refactor all other Toolbars to have the same background color.

app/src/main/java/org/wikipedia/places/PlacesFilterActivity.kt
```diff
@@ -54,6 +54,9 @@ class PlacesFilterActivity : BaseActivity() {
         super.onCreate(savedInstanceState)
         binding = ActivityPlacesFiltersBinding.inflate(layoutInflater)
         setUpRecyclerView()
+        setStatusBarColor(ResourceUtil.getThemedColor(this, R.attr.background_color))
+        setNavigationBarColor(ResourceUtil.getThemedColor(this, R.attr.background_color))
+        supportActionBar?.setBackgroundDrawable(ColorDrawable(ResourceUtil.getThemedColor(this, R.attr.background_color)))
```
dbrant, line 59: To have full control of the style and color of a Toolbar, an activity should have its own explicit Toolbar in its layout (i.e. not using `style/AppTheme.ActionBar`).

app/src/main/java/org/wikipedia/places/PlacesFilterActivity.kt
```diff
@@ -155,7 +158,7 @@ class PlacesFilterActivity : BaseActivity() {
         fun bindItem(languageCode: String) {
             titleText.text = WikipediaApp.instance.languageState.getAppLanguageCanonicalName(languageCode)
             langCodeText.text = languageCode
-            radio.isVisible = languageCode == Prefs.placesWikiCode
+            radio.visibility = if (languageCode == Prefs.placesWikiCode) View.VISIBLE else View.INVISIBLE
```
dbrant, line 161: Was there an issue with using `isVisible`? It seems a bit more clean.

app/src/main/res/layout/view_places_filter_item.xml
```diff
@@ -3,42 +3,65 @@
...
-        android:layout_marginStart="24dp"
-        android:gravity="center_vertical"
-        android:textColor="?attr/primary_color" />
+        <LinearLayout
+            android:id="@+id/placesFilterLangContainer"
+            android:layout_width="wrap_content"
+            android:layout_height="match_parent"
+            android:paddingHorizontal="10dp"
+            android:layout_marginStart="14dp"
+            android:background="?attr/selectableItemBackground"
```
dbrant, line 22: This view doesn't need its own `selectableItemBackground`, since the parent view already has it.

dbrant (changes requested):

sharvaniharan:
Done @dbrant. Thank you.

PR #4384, opened 2024-01-14
Places: Data wiring

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T351161

Schema doc: https://docs.google.com/presentation/d/1lBfwEhHZe8hck0owCwEuqgX6JIyHZYmjkZM71S4c4NU/edit#slide=id.g28e4bf12787_0_0

- Data wiring added for Places feature
- Generic events added for Survey dialog


sharvaniharan:
@cooltey @dbrant Passed review with Shay. So moving to code review now.

app/src/main/java/org/wikipedia/search/SearchResultsFragment.kt
```diff
@@ -83,7 +84,7 @@ class SearchResultsFragment : Fragment() {
                 }
             }
         }
-
+        PlacesEvent.logAction("search_results_impression", "search_view")
```
dbrant, line 87: This will be shown when the Search activity is displayed, regardless of whether there are results or not. Is that desired?

dbrant (changes requested):
This all looks ok and fair (see minor inline stuff), but I just have some general observations. I'm seeing a huge amount of duplication of events, and I wonder if we could think about trimming things down where possible. Maybe not necessarily for this feature/iteration, but for the future. Here are a couple of examples:

When tapping on a map marker to see a link preview, there are **FIVE** events sent:

* Breadcrumbs `click` event due to clicking on the map itself.
* Breadcrumbs `show` event for showing the Link Preview dialog.
* Places `marker_click` event.
* Places `impression` event for `detail_view`, which I presume is another name for Link Preview? (can we call it `link_preview`?)
* Link preview interaction event, with a `linkclick` action (which contains the `source` where it came from)

When tapping on the Search bar, there are **FIVE** events sent:

* Breadcrumbs `click` event on the Places search bar.
* Breadcrumbs `show` event for showing the Search activity (which contains the `source` where it came from).
* Places `search_view_click` event.
* Places `impression` event for `search_view`
* Places `search_results_impression` event.

This seems like a trend that we should be mindful of, and think about actively reducing or consolidating.

app/src/main/java/org/wikipedia/search/SearchResultsFragment.kt
sharvaniharan: I have removed this for now and will ask shay if we really need an impression event for this since its so evolving, and we also have a result choice event. I will make a separate pr if we need to add it back.

sharvaniharan:
> This all looks ok and fair (see minor inline stuff), but I just have some general observations. I'm seeing a huge amount of duplication of events, and I wonder if we could think about trimming things down where possible. Maybe not necessarily for this feature/iteration, but for the future. Here are a couple of examples:
>
I agree @dbrant. From many data related discussions, it seems like the only comfortable choice from the analyst's point has been to create feature-specific streams each time, and `not` have to deal with breadcrumbs at all. Hence the multiple wirings at common traffic places.

PR #4390, opened 2024-01-17
Fix Commons search when inserting images.

dbrant:
When searching for images for inserting into the Editing activity, we were incorrectly handing the continuation for the API call. When there are too few search results (i.e. with no further continuation) the API call was getting repeated over and over.

https://phabricator.wikimedia.org/T354175


app/src/main/java/org/wikipedia/edit/insertmedia/InsertMediaViewModel.kt
```diff
@@ -63,20 +60,22 @@ class InsertMediaViewModel(bundle: Bundle) : ViewModel() {
                         }
                         pageTitle
                     }
-                    LoadResult.Page(results, null, response.continuation)
+                    LoadResult.Page(results, params.key, if (response.continuation?.gsroffset != 0)
+                        response.continuation?.gsroffset else null)
```
cooltey, line 64: Not sure but would it be the same if we pass `response.continuation?.gsroffset` directly and it will output either `0` or `null` initially?

dbrant: Yep. This necessitated a few updates/improvements in the main Search logic, as well.

PR #4397, opened 2024-01-18
(Places) Correctly handle insets.

dbrant:
This enables us to handle insets correctly for applying the proper margins to the Search bar, location button, etc.
Even works upon screen rotation and other configuration changes!


cooltey:
This is great!

PR #4400, opened 2024-01-19
Save last location information to the preference

cooltey:
This PR also set `10` milliseconds for the camera animation in the `goToLocation` will reduce the time of moving to the save or current location.

### DIFF
--- a/app/src/main/java/org/wikipedia/places/PlacesFragment.kt
+++ b/app/src/main/java/org/wikipedia/places/PlacesFragment.kt
@@ -459,6 +459,7 @@ class PlacesFragment : Fragment(), LinkPreviewDialog.LoadPageCallback, MapboxMap
     }
 
     override fun onDestroyView() {
+        Prefs.placesLastLocation = lastLocation
         binding.mapView.onDestroy()
         _binding = null
 
@@ -476,7 +477,6 @@ class PlacesFragment : Fragment(), LinkPreviewDialog.LoadPageCallback, MapboxMap
             it.latitude = latLng.latitude
             it.longitude = latLng.longitude
         }
-        Prefs.placesLastLocation = lastLocation
 
         lastZoom = mapboxMap?.cameraPosition?.zoom ?: 15.0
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/settings/Prefs.kt
```diff
@@ -712,4 +713,23 @@ object Prefs {
     var placesWikiCode
         get() = PrefsIoUtil.getString(R.string.preference_key_places_wiki_code, WikipediaApp.instance.appOrSystemLanguageCode).orEmpty()
         set(value) = PrefsIoUtil.setString(R.string.preference_key_places_wiki_code, value)
+
+    var placesLastLocation: Location?
```
dbrant, line 717: Should we also save the last Zoom level?

app/src/main/java/org/wikipedia/places/PlacesFragment.kt
```diff
@@ -558,7 +559,7 @@ class PlacesFragment : Fragment(), LinkPreviewDialog.LoadPageCallback, LinkPrevi
                     val location = preferredLocation?.let { loc -> LatLng(loc.latitude, loc.longitude) }
                     val targetLocation = location ?: currentLatLngLoc
                     targetLocation?.let { target ->
-                        it.animateCamera(CameraUpdateFactory.newLatLngZoom(target, zoom), object : CancelableCallback {
+                        it.animateCamera(CameraUpdateFactory.newLatLngZoom(target, zoom), 10, object : CancelableCallback {
```
dbrant, line 562: If the goal is to no longer have an animation, then we should use `moveCamera()` instead of `animateCamera()`.

dbrant (changes requested):

app/src/main/java/org/wikipedia/settings/Prefs.kt
cooltey: Yes and done!

app/src/main/java/org/wikipedia/places/PlacesFragment.kt
cooltey: Yes, done!

PR #4402, opened 2024-01-19
(Places) remove postDelayed calls.

dbrant:
This removes all delays when navigating to a new location. We were delaying movement by 1 second... does anyone recall why?


app/src/main/java/org/wikipedia/places/PlacesFragment.kt
```diff
@@ -507,7 +507,6 @@ class PlacesFragment : Fragment(), LinkPreviewDialog.LoadPageCallback, LinkPrevi
                         .withLatLng(LatLng(it.latitude, it.longitude))
                         .withTextFont(MARKER_FONT_STACK)
                         .withIconImage(MARKER_DRAWABLE)
-                        .withIconOffset(arrayOf(0f, -32f))
```
dbrant, line 510: Btw we no longer need to offset the icon. This was done because the previous marker was an arrow that pointed at the location. But now that our marker is a circle, it should sit directly on top of the location, with no offset.
This resolves weird situations when the "enlarged" version of the marker gets an incorrect offset, and reveals other markers directly underneath it.
https://phabricator.wikimedia.org/T351158#9472534

PR #4403, opened 2024-01-19
(Places) Clean up searching, and correctly remember last search term.

dbrant:
This uses existing functionality that we already had in SearchFragment to return a selected link.
This also fixes a tiny issue where we weren't remembering the last search term, to put in Recent Searches.

### DIFF
--- a/app/src/main/java/org/wikipedia/search/SearchResult.kt
+++ b/app/src/main/java/org/wikipedia/search/SearchResult.kt
@@ -27,7 +27,8 @@ data class SearchResult(val pageTitle: PageTitle,
     val location: Location? get() {
         return if (coordinates.isNullOrEmpty()) null else
             Location("").also {
-                it.latitude = coordinates[0].lat; it.longitude = coordinates[0].lon
+                it.latitude = coordinates[0].lat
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/wikipedia/search/SearchResult.kt
```diff
@@ -22,4 +23,11 @@ data class SearchResult(val pageTitle: PageTitle,
 
     constructor(pageTitle: PageTitle, searchResultType: SearchResultType) :
             this(pageTitle, null, searchResultType)
+
+    val location: Location? get() {
+        return if (coordinates.isNullOrEmpty()) null else
+            Location("").also {
+                it.latitude = coordinates[0].lat; it.longitude = coordinates[0].lon
```
cooltey, line 30: Do we need to check the first item of the `coordinates` exist? 

And, can we have separate lines when set `latitude` and `longitude`

dbrant: It's checking `isNullOrEmpty`, so the first item will definitely exist.
(and yes re separate lines)

PR #4407, opened 2024-01-19
Places: Discovery with tooltips

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T355302

- Builds 2 tooltips: One on main nav tab and one for page more menu
- Removes Watchlist main and page tooltips and related resources completely.
- Added dev preferences for new tooltips


app/src/main/res/values/strings.xml
```diff
@@ -1686,7 +1686,7 @@
     <string name="places_toggle_buttons_map">Map</string>
     <string name="places_toggle_buttons_list">List</string>
     <string name="places_nav_tab_tooltip_message">Discover Wikipedia articles around you with Places</string>
-    <string name="places_article_menu_tooltip_message"><![CDATA[Tap <b>\'View on map\'</b> to explore this article\'s location]]></string>
+    <string name="places_article_menu_tooltip_message"><![CDATA[Tap \'View on map\' to \nexplore this article\'s location]]></string>
```
dbrant, line 1689: Do we actually need an explicit newline `\n` there? Shouldn't we allow the text to overflow normally?

dbrant (changes requested):
Logic looks good, just one tiniest nit.

app/src/main/res/values/strings.xml
sharvaniharan: It makes the tooltip very wide like this, so I had added line-break to get closer to the designs:
long_tooltip

dbrant:
more code reduction = excellent

PR #4411, opened 2024-01-23
Survey text strings for translation

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T355140

Survey text to send for translation


cooltey (commented):
I'm curious if we should create new strings for the Places survey instead of using the existing Patroller Tasks strings.

sharvaniharan:
I left the old ones because they would have collected translations by now that could be useful... but they are in the same context so it should satisfy the translatewiki terms. Because they asked us to create new ones because the context could be different for the string.

PR #4415, opened 2024-01-24
Harmonize dynamic tab and language icon in the app

cooltey:
https://phabricator.wikimedia.org/T354522


cooltey:
> Hi @cooltey Looked like this was Robin's preliminary thoughts on the language button changes... Could you pleaseplease take a look at this and either merge it here if it is useful, or close it if Robin doesnt need this anymore? #4351

It is good to merge it into this PR and I will let Robin know about this. Thanks!

sharvaniharan:
Hi @cooltey 
Icon color seems to be different on main screen 
image
Compared to other places
image
Is that to be expected

cooltey:
Yes, it is expected.

dbrant:
Harmonizing _and_ reducing code by 250 lines? The best kind or PR.

PR #4416, opened 2024-01-24
(Places) Don't use extra list along with LiveData.

dbrant:
@cooltey This is to merge into your List PR, and this probably fixes any issues with list items disappearing upon moving the map, since now the list items will be literally what's shown on the map.


cooltey:
Thanks! This is great!

PR #4419, opened 2024-01-25
Log invokeSource in breadcrumbs "show" events.

dbrant:
This automatically detects if a screen that's being shown contains an `InvokeSource` in its intent, and logs it as part of the event in a structured way, making breadcrumbs a bit more powerful.
This will automatically enable us to measure incoming clicks from widgets, without extra schemas.


app/src/main/java/org/wikipedia/analytics/eventplatform/BreadCrumbLogEvent.kt
```diff
@@ -54,7 +56,9 @@ class BreadCrumbLogEvent(
         }
 
         fun logScreenShown(context: Context, fragment: Fragment? = null) {
-            EventPlatformClient.submit(BreadCrumbLogEvent(BreadCrumbViewUtil.getReadableScreenName(context, fragment), "show"))
+            val invokeSource = (fragment?.activity?.intent ?: (context as? Activity)?.intent)?.getSerializableExtra(Constants.INTENT_EXTRA_INVOKE_SOURCE) as? Constants.InvokeSource
+            EventPlatformClient.submit(BreadCrumbLogEvent(BreadCrumbViewUtil.getReadableScreenName(context, fragment),
+                "show" + (invokeSource?.let { ".from." + it.value } ?: "")))
```
cooltey, line 61: Can we use `orEmpty()` for the `invokeSource ?: ""`

dbrant: yep!

cooltey:
@sharvaniharan Please also check the changes to the Breadcrumbs events

PR #4429, opened 2024-02-01
Update tab and language icon style for long text

cooltey:
https://phabricator.wikimedia.org/T354522#9501302

### DIFF
--- a/app/src/main/java/org/wikipedia/views/LangCodeView.kt
+++ b/app/src/main/java/org/wikipedia/views/LangCodeView.kt
@@ -10,6 +10,7 @@ import android.widget.FrameLayout
 import androidx.annotation.ColorInt
 import androidx.core.content.res.use
 import androidx.core.view.ViewCompat
+import androidx.core.widget.TextViewCompat
 import org.wikipedia.R
 import org.wikipedia.databinding.ViewLangCodeBinding
 import org.wikipedia.language.LanguageUtil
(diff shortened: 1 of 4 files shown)


dbrant (changes requested):
I don't think it looks correct on my device (Galaxsy S9, Android 10):

image image

Can we please push back on Designers very strongly when something _cannot_ be done a certain way? Fonts across various devices are not guaranteed to be exactly the same pixel width; we cannot guarantee that a box with an exact pixel size will contain a specific number of characters, because Fonts are not measured the same way that Views are. This is why font sizes are specified in `sp` units, while views are specified in `dp` units.

dbrant:
We're also forgetting that the user is able to control the **global font size** adjustment throughout the whole system (many users select a "larger" font size), which makes pixel-defined Views look even worse:

image image

Therefore it's generally not a good idea to define the size of a View in precise dp values, when expecting it to contain a specific number of characters. These kinds of Views should be variable size (perhaps with a minimum size) and scalable based on the font.

cooltey:
Thanks @dbrant 

I posted a reply to the designer to the ticket: https://phabricator.wikimedia.org/T353562#9508461

and I also tried to apply `match_parent` to the custom views but the result does not work as expected. Will let you know when the designer reviews it and passes it to code review.

cooltey:
@dbrant it's ready for review!

PR #4435, opened 2024-02-02
Fix: show the progressbar correctly in the diff view

cooltey:
The progress bar is not showing in the Diff view correctly, this PR adds a `FrameLayout` to contain the `CollapsingToolbarLayout` and move the progressbar to so it will not make the view jumps after gone.


dbrant (changes requested):
This makes the collapsing toolbar no longer scrollable or collapsing.
A `CollapsingToolbarLayout` should be a direct child of `AppBarLayout` for scrolling to work correctly.

https://github.com/wikimedia/apps-android-wikipedia/assets/1682214/02a66dda-26ea-4707-babb-40558dfcd774

cooltey:
Fixed!

PR #4447, opened 2024-02-06
Experimental: in-app updates

cooltey:
https://developer.android.com/guide/playcore/in-app-updates/kotlin-java

Follow the top-voted answer steps to test the APK locally: 
https://stackoverflow.com/questions/56087064/how-can-i-test-in-app-updates-in-android

Demo: https://youtu.be/E8SkgfyQT9g

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -37,7 +37,7 @@ android {
         applicationId 'org.wikipedia'
         minSdk 21
         targetSdk 34
-        versionCode 50468
+        versionCode 50469
         testApplicationId 'org.wikipedia.test'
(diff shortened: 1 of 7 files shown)


dbrant:
Very nice! Some initial comments:

- Obviously this won't do anything in builds released to other stores, e.g. Samsung, Huawei, etc. But we should make sure in those cases it doesn't have any unusual side effects.

- In the case of F-Droid, bundling a Google library will not be acceptable. Ideally this should be split off into a class in our `extras` folder, similar to how we handle Firebase Messaging.

cooltey:
Close for now but keep the branch for future reference.

PR #4448, opened 2024-02-06
Update deps

Jvr2022:


app/build.gradle
```diff
@@ -171,15 +171,15 @@ dependencies {
...
+    String mockitoVersion = '5.10.0'
     String leakCanaryVersion = '2.13'
     String kotlinCoroutinesVersion = '1.7.3'
     String firebaseMessagingVersion = '23.4.0'
     String mlKitVersion = '17.0.4'
     String roomVersion = "2.6.1"
     String espressoVersion = '3.5.1'
     String serialization_version = '1.6.2'
-    String metricsVersion = '2.1'
+    String metricsVersion = '4.2.25'
```
dbrant, line 182: This version does not exist. Where did this come from?

Jvr2022: https://mvnrepository.com/artifact/io.dropwizard.metrics/metrics-core

dbrant: ?? we don't use `io.dropwizard.metrics`.

PR #4453, opened 2024-02-06
Account Vanishing integration.

dbrant:
https://phabricator.wikimedia.org/T358018


cooltey (changes requested):
Found two issues that relate to orientation change:
1. When the "Vanish account" dialog is open (portrait), and go change the device's orientation, the entire content will disappear.
2. Enter the "Account vanishing request form", and change the device's orientation, the web page will stay in the loading state.

dbrant:
> 1. When the "Vanish account" dialog is open (portrait), and go change the device's orientation, the entire content will disappear.

It looks like this is just the standard MaterialDialog component squeezing the dialog text into an invisible size. I'm not sure we can do much about it, and I'm not sure it's worth building any custom components to accommodate.

> 2. Enter the "Account vanishing request form", and change the device's orientation, the web page will stay in the loading state.

Good catch, and done!

>  is this passed the design review?

This can be a design-review-in-alpha situation, since we'd like to get this out the door asap, and it is tucked away in a relatively hidden spot.

sharvaniharan:
Looks good to me.. only one minor ui issue I saw:
Not sure of it was an intentional change, but I see on the task on screenshot `1`, that the `LOG OUT` button is now in all caps and not red anymore. Since it is related to account status, please check with designers if it is an intentional change.

app/src/main/res/layout/view_preference_logout.xml
```diff
@@ -43,4 +43,17 @@
 
     </LinearLayout>
 
+    <com.google.android.material.button.MaterialButton
+        android:id="@+id/accountVanishButton"
+        style="@style/App.Button.Text"
+        android:layout_width="wrap_content"
```
cooltey, line 49: Would it be possible to have `match_parent` for the width to keep the layout consistency?

app/src/main/java/org/wikipedia/settings/LogoutPreference.kt
```diff
@@ -43,5 +45,18 @@ class LogoutPreference : Preference {
...
         }
+        holder.itemView.findViewById<Button>(R.id.accountVanishButton).setOnClickListener {
+            activity?.let {
+                MaterialAlertDialogBuilder(it, R.style.AlertDialogTheme_Icon_Delete)
+                    .setIcon(R.drawable.ic_person_remove)
+                    .setTitle(R.string.account_vanish_request_confirm_title)
+                    .setMessage(StringUtil.fromHtml(it.getString(R.string.account_vanish_request_confirm)))
+                    .setNegativeButton(android.R.string.cancel, null)
+                    .setPositiveButton(R.string.account_vanish_request_title) { _, _ ->
+                        it.finish()
```
cooltey, line 56: Not sure but is this intentional? Or should we just need to dismiss the dialog?

app/src/main/res/layout/view_preference_logout.xml
dbrant: yep!

app/src/main/java/org/wikipedia/settings/LogoutPreference.kt
dbrant: This is intentional - when the user navigates "back" from the web form (whether they submitted it or not), we want them to get back to the Main screen, not the Settings screen, and this is the easiest way to do it.

cooltey: Got it. Thanks!

cooltey (approved):
It looks good to me. Please feel free to merge it if it looks good to you too, @sharvaniharan

dbrant:
> the `LOG OUT` button is now in all caps and not red anymore.

This is an outdated design, from before we changed all our buttons to be no longer all-caps. (notice the design also has a "Send usage reports" option, which no longer exists.)

PR #4462, opened 2024-02-08
Broaden usage of android edit tags in comments

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T347431

Adding edit comment tags for analytics purposes


app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -545,6 +545,11 @@ class DescriptionEditFragment : Fragment() {
         const val SUGGESTED_EDITS_TRANSLATE_DESC_COMMENT = "#suggestededit-translate-desc $SUGGESTED_EDITS_UI_VERSION"
         const val SUGGESTED_EDITS_ADD_CAPTION_COMMENT = "#suggestededit-add-caption $SUGGESTED_EDITS_UI_VERSION"
         const val SUGGESTED_EDITS_TRANSLATE_CAPTION_COMMENT = "#suggestededit-translate-caption $SUGGESTED_EDITS_UI_VERSION"
+        const val SUGGESTED_EDITS_IMAGE_TAGS_COMMENT = "##suggestededit-add-tag $SUGGESTED_EDITS_UI_VERSION"
```
cooltey, line 548: It has two `#`, please remove one `#` from the comment.

cooltey (changes requested):

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageTagsFragment.kt
```diff
@@ -356,7 +357,7 @@ class SuggestedEditsImageTagsFragment : SuggestedEditsItemFragment(), CompoundBu
                 .subscribe({ token ->
                     val mId = "M" + page!!.pageId
                     var claimStr = "{\"claims\":["
-                    var commentStr = "/* add-depicts: "
+                    var commentStr = "/* add-depicts: " + DescriptionEditFragment.SUGGESTED_EDITS_IMAGE_TAGS_COMMENT
```
dbrant, line 360: Here we are building a list into `commentStr` of the image tags that were added. We should not be putting our additional #comment in the middle of the list. It should come at the end, after the `commentStr` is closed with `*/`

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -545,6 +545,11 @@ class DescriptionEditFragment : Fragment() {
...
+        const val WATCHLIST_UNDO_COMMENT = "#watchlist-undo"
+        const val WATCHLIST_ROLLBACK_COMMENT = "#watchlist-rollback"
+        const val DIFF_UNDO_COMMENT = "#diff-undo"
+        const val DIFF_ROLLBACK_COMMENT = "#diff-rollback"
```
dbrant, line 552: Do all of the above new comments belong in `DescriptionEditFragment.kt`, if they don't relate to description editing? Can we put them in more relevant places?

dbrant (changes requested):

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
sharvaniharan: Made this change... but was wondering if we should do the opposite and have all the comment tags at one place... makes answering analytics questions easier... even if it means a new section in `Constants` or a whole new file of its own..

PR #4463, opened 2024-02-09
Correctly handle WebView update exception.

dbrant:
We try to handle the exception that is thrown when the System WebView is being updated, but we're no longer handling it in the right place. This is probably a vestige of updating to ViewBindings.

This will now handle such exceptions correctly, by handling it in `onStart()`, which is where child fragments (`PageFragment`) are inflated.


app/src/main/java/org/wikipedia/page/PageActivity.kt
```diff
@@ -257,6 +241,24 @@ class PageActivity : BaseActivity(), PageFragment.Callback, LinkPreviewDialog.Lo
         }
     }
 
+    override fun onStart() {
+        try {
+            super.onStart()
+        } catch (e: Exception) {
+            if (e.message.orEmpty().lowercase(Locale.getDefault()).contains(EXCEPTION_MESSAGE_WEBVIEW) ||
```
cooltey, line 248: Would it be better to use `contains(..., ignoreCase = true)`?

PR #4465, opened 2024-02-09
`Edit` call to action in preview sheet for stub articles

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T356058

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -401,6 +401,7 @@
     <string name="link_preview_dialog_save_button">Save</string>
     <string name="link_preview_dialog_saved_button">Saved</string>
     <string name="link_preview_dialog_share_button">Share</string>
+    <string name="link_preview_stub_placeholder_text">This article is a stub, which means there isn\’t enough content in the article to generate a preview. However, you can change that</string>
     <string name="error_page_does_not_exist">This page does not exist</string>
     <string name="error_user_page_does_not_exist"><![CDATA[Wikipedia does not have a <a href="%1$s">user page</a> with this exact name. In general, this page should be created and edited by <b>%2$s</b>. If in doubt, please verify that \"%3$s\" exists.]]></string>
     <string name="view_wiki_error_message_offline">Cannot connect to internet</string>
--- a/app/src/main/res/values-qq/strings.xml
+++ b/app/src/main/res/values-qq/strings.xml
@@ -1617,4 +1617,5 @@
   <string name="places_survey_dialog_cancel">Negative action button text for places survey dialog</string>
   <string name="places_survey_feedback_low_satisfaction_dialog_title">Title text for the secondary dialog which is shown as a result of an action on the main dialog for places feature - shown only to users who are very unsatisfied with the feature</string>
   <string name="places_empty_list">Empty message for the list in the current map area. Please keep the anchor link around the \"on the map\".</string>
+  <string name="link_preview_stub_placeholder_text">Placeholder text shown to users in the preview dialog of an article, when there isn\'t enough content in the article to show in the preview.</string>
 </resources>


app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
```diff
@@ -374,23 +389,25 @@ class LinkPreviewDialog : ExtendedBottomSheetDialogFragment(), LinkPreviewErrorV
...
+        binding.linkPreviewExtractWebview.setBackgroundColor(Color.TRANSPARENT)
+        val colorHex = ResourceUtil.colorToCssString(
+            ResourceUtil.getThemedColor(
+                requireContext(),
+                android.R.attr.textColorPrimary
             )
-        }
+        )
+        val dir = if (L10nUtil.isLangRTL(viewModel.pageTitle.wikiSite.languageCode)) "rtl" else "ltr"
+        val editVisibility = (viewModel.fromPlaces || viewModel.fromArticle) && contents.extract.isNullOrEmpty()
```
dbrant, line 400: What's the reason for caring if it's `fromPlaces` or `fromArticle`? Don't we want to do the stub treatment in all cases?

app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
```diff
@@ -145,12 +147,25 @@ class LinkPreviewDialog : ExtendedBottomSheetDialogFragment(), LinkPreviewErrorV
...
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         _binding = DialogLinkPreviewBinding.inflate(inflater, container, false)
         binding.linkPreviewToolbar.setOnClickListener { goToLinkedPage(false) }
         binding.linkPreviewOverflowButton.setOnClickListener {
             setupOverflowMenu()
         }
+        binding.linkPreviewEditChip.setOnClickListener {
+            viewModel.pageTitle.run {
+                requestEditSectionLauncher.launch(EditSectionActivity.newIntent(requireContext(), -1, null, this, Constants.InvokeSource.LINK_PREVIEW_MENU, null))
+                dismiss()
```
dbrant, line 166: Instead of dismissing here, it can be done in the result handler.

app/src/main/java/org/wikipedia/page/linkpreview/LinkPreviewDialog.kt
```diff
@@ -145,12 +147,25 @@ class LinkPreviewDialog : ExtendedBottomSheetDialogFragment(), LinkPreviewErrorV
         }
     }
 
+    private val requestEditSectionLauncher = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) {
+        if (it.resultCode == EditHandler.RESULT_REFRESH_PAGE) {
+            FeedbackUtil.showMessage(this, R.string.edit_saved_successfully)
```
dbrant, line 152: For showing the edit-success message, the context should be the parent activity (`requireActivity()`) instead of `this`, and then the dialog can be dismissed right after.
This way we don't need to worry about reloading the stub.

app/src/main/res/values/strings.xml
```diff
@@ -403,6 +403,7 @@
     <string name="link_preview_dialog_save_button">Save</string>
     <string name="link_preview_dialog_saved_button">Saved</string>
     <string name="link_preview_dialog_share_button">Share</string>
+    <string name="link_preview_stub_placeholder_text">This article is a stub, which means there isn\’t enough content in the article to generate a preview. However, you can change that</string>
```
dbrant, line 406: Sentence is missing a period `.`

app/src/main/res/layout/dialog_link_preview.xml
```diff
@@ -117,6 +117,23 @@
                     android:layout_gravity="center_horizontal"
                     android:layout_marginTop="4dp" />
 
+                <com.google.android.material.chip.Chip
```
dbrant, line 120: What's the reason for making this a `Chip` instead of a `MaterialButton` like elsewhere in the app? We have a number of predefined button styles called `App.Button.*` for you to use, one of which is the exact style that's needed here, without any other changes.

dbrant (changes requested):

dbrant:
@sharvaniharan just making sure you saw my code review comments above. (they would have helped it pass design review.)

sharvaniharan:
@dbrant Updated it. Sorry was distracted with the other work.

PR #4478, opened 2024-02-22
Fix possible crash when tapping on map markers with location permissions disabled.

dbrant:
This (minor) crash occurs when a user declines the Location permission, then goes on to tap on map markers to see a link preview.


dbrant:
Self-merging to expedite hotfix release.

PR #4483, opened 2024-02-24
Saved messages flow for edit patrol

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T356774


app/src/main/java/org/wikipedia/talk/template/TalkTemplatesFragment.kt
```diff
@@ -95,41 +152,77 @@ class TalkTemplatesFragment : Fragment(), MenuProvider {
...
 
     override fun onDestroyView() {
         super.onDestroyView()
         _binding = null
     }
 
-    override fun onCreateMenu(menu: Menu, inflater: MenuInflater) {
-        inflater.inflate(R.menu.menu_talk_templates, menu)
+    override fun onPrepareMenu(menu: Menu) {
+        menu.findItem(R.id.menu_edit_messages).isVisible = binding.talkTemplatesTabLayout.selectedTabPosition == 0
```
cooltey, line 212: Can we hide the button when there are no messages in the list as well?

app/src/main/java/org/wikipedia/talk/template/TalkTemplatesFragment.kt
```diff
@@ -24,24 +25,38 @@ import androidx.recyclerview.widget.ItemTouchHelper
...
     private var _binding: FragmentTalkTemplatesBinding? = null
 
-    private val viewModel: TalkTemplatesViewModel by viewModels()
+    private val viewModel: TalkTemplatesViewModel by viewModels { TalkTemplatesViewModel.Factory(requireArguments()) }
     private val binding get() = _binding!!
 
-    private lateinit var itemTouchHelper: ItemTouchHelper
+    private var itemTouchHelper: ItemTouchHelper? = null
+    private var itemSwipeTouchHelper: ItemTouchHelper? = null
+    private var touchCallback: SwipeableItemTouchHelperCallback? = null
```
cooltey, line 56: Is it possible to change this to `lateinit var` so we do not need to worry about checking the `null`?

app/src/main/res/values/styles.xml
```diff
@@ -89,6 +89,11 @@
 
     </style>
 
+    <style name="AppTheme.CustomMenuText" parent="AppTheme">
+        <item name="android:actionMenuTextAppearance">@style/H3</item>
+        <item name="android:actionMenuTextColor">?attr/progressive_color</item>
+    </style>
```
cooltey, line 95: We have already used a `MaterialToolbar` in the fragment, would it be possible to add a `TextView` instead of applying a style on a menu item?

cooltey (changes requested):
- When select an example message and want to go back to the list, it will not stay at the previous "Example message" tab.
- I am still able to swipe-to-delete on the example messages, is that expected?
- After removing the example messages, and tapping on "Undo", the message will be saved as a custom message (My messages).

sharvaniharan:
Done @cooltey

app/src/main/res/values/styles.xml
cooltey, line 95: Can we remove this style?

cooltey (changes requested):

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsRecentEditsFragment.kt
```diff
@@ -173,6 +174,11 @@ class SuggestedEditsRecentEditsFragment : Fragment(), MenuProvider {
                 FeedbackUtil.showAndroidAppEditingFAQ(requireContext())
                 true
             }
+            R.id.menu_saved_messages -> {
+                    val pageTitle = PageTitle(UserAliasData.valueFor(viewModel.wikiSite.languageCode), "", viewModel.wikiSite)
```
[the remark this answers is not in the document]
sharvaniharan: Since this is the list of edits, i have added account's username.

app/src/main/java/org/wikipedia/talk/TalkReplyActivity.kt
```diff
@@ -580,14 +573,26 @@ class TalkReplyActivity : BaseActivity(), UserMentionInputView.Listener, EditPre
...
             return Intent(context, TalkReplyActivity::class.java)
                     .putExtra(Constants.ARG_TITLE, pageTitle)
                     .putExtra(EXTRA_PARENT_SUBJECT, parentSubject)
                     .putExtra(EXTRA_TOPIC, topic)
                     .putExtra(EXTRA_SUBJECT, undoSubject)
                     .putExtra(EXTRA_BODY, undoBody)
                     .putExtra(EXTRA_FROM_DIFF, fromDiff)
+                    .putExtra(EXTRA_SELECTED_TEMPLATE, selectedTemplate)
+                    .putExtra(EXTRA_TEMPLATE_MANAGEMENT, templateManagementMode)
+                    .putExtra(EXTRA_SAVED_TEMPLATE, isSavedTemplate)
```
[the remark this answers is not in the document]
sharvaniharan: Moved these. I have used them to distinguish between the modes in which we use them.

dbrant (changes requested):
A few initial major observations, before delving into the code itself:

* It doesn't actually do what it's intended to do; upon publishing a message to a talk page, the triple-brace keywords don't get filled in:
Screenshot_20240325_152115

* There are a number of design mismatches/inconsistencies, particularly in Dark mode:

The "external link" icon has incorrect tint, and appears to be too small (design specifies 24dp):
image

The popup dialog for saving the message has incorrect colors in dark mode:
image

We should ellipsize (...) the message text in the list of messages, instead of cutting it off at the end:
image

The FAB icon looks a bit too large. It looks like the SVG asset itself was imported weirdly, and doesn't have the correct size of 24dp.
image

cooltey:
Hi @dbrant, the comments you posted have been addressed. Thanks.

PR #4487, opened 2024-02-26
Places: Fix requiring permissions to go to article map marker.

dbrant:
Another issue from insufficient testing of denying location permissions 😆 
We're requiring the checking of permissions in a couple of places where it's not necessary at all.

https://phabricator.wikimedia.org/T358426


sharvaniharan:
This is a great fix! 🎉

PR #4489, opened 2024-02-26
Support SteadyView (screen stabilizer) library

Sublimis:
With a few simple changes this pull request aims at supporting the [SteadyScreen](https://github.com/Sublimis/SteadyScreen) screen stabilization service (by utilizing the [SteadyView library](https://github.com/Sublimis/SteadyView)).

It makes on-screen reading easier by softening small movements of mobile screens.

### DIFF
--- a/SteadyFixedDrawerLayout.java
+++ b/SteadyFixedDrawerLayout.java
@@ -1,0 +1,41 @@
+package org.wikipedia.views;
+
+import android.content.Context;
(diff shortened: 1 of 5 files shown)


dbrant:
Many thanks, but we try to limit our dependency on third-party libraries to a minimum, especially if it's also dependent on another third-party service running in the background.

PR #4501, opened 2024-03-02
Refactor preview list fetch mechanism

Pinaki93:
Moved the logic to fetch preview reading list to a dedicated repository instead of a helper function. This is a small portion for carrying out major refactoring tasks for T303935


app/src/main/java/org/wikipedia/readinglist/PreviewReadingListRepository.kt
```diff
@@ -0,0 +1,112 @@
...
+import org.wikipedia.util.StringUtil
+import java.util.Date
+
+class PreviewReadingListRepository(
+    private val context: Context = WikipediaApp.instance,
+    private val prefs: Prefs = Prefs
+) {
+
+    suspend fun getPreviewReadingList(): ReadingList? {
+        val encodedJson = prefs.receiveReadingListsData
```
cooltey, line 28: It would be great if use `Prefs.receiveReadingListsData` directly.

Pinaki93: I've converted it to a variable so as to support constructor injection (and easily write tests later with mocks). What do you think?

cooltey:
Hi, would it be possible to change the `PreviewReadingListRepositery`  to `ReadingListViewModel` and also introduce `ViewModel` mechanism like we recently did in https://github.com/wikimedia/apps-android-wikipedia/pull/4628?

dbrant:
Closing, in favor of larger-scale refactorings in the near future.

PR #4504, opened 2024-03-04
Refactor EditPreviewFragment to be usable from other activities.

dbrant:
This is a light refactor of the edit preview fragment which decouples it from the general editing workflow, and makes it usable from other activities where a preview will be necessary, such as TalkReplyActivity.
(No changes to existing behavior.)


sharvaniharan:
Thank you! 🙌  Was about to do this on edit [PR](https://github.com/wikimedia/apps-android-wikipedia/pull/4483)

PR #4508, opened 2024-03-06
API underpinnings for TemplateData flow

cooltey:
This can be merged to the `main` branch now since it does not affect anything.


cooltey:
@dbrant and @sharvaniharan, please help to review this PR so I can continue the implementation of the task, thanks!

app/src/main/java/org/wikipedia/dataclient/mwapi/TemplateDataResponse.kt
```diff
@@ -0,0 +1,53 @@
...
+import kotlinx.serialization.json.JsonArray
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.booleanOrNull
+import kotlinx.serialization.json.contentOrNull
+import kotlinx.serialization.json.decodeFromJsonElement
+import kotlinx.serialization.json.jsonPrimitive
+import org.wikipedia.json.JsonUtil
+
+@Serializable
+class TemplateDataResponse : MwResponse() {
```
dbrant, line 14: Would it be cleaner if this was moved into `MwQueryPage`?
The response is already an instance of Page, with parameters like `title` and `description` already there. The only new stuff is the Params class, which can be a new inner class like e.g. `Revision` or `EntityTerms` that are possible in `MwQueryPage`.

app/src/main/java/org/wikipedia/dataclient/mwapi/TemplateDataResponse.kt
```diff
@@ -0,0 +1,53 @@
...
+        val type: String = ""
+        val required: Boolean = false
+        val suggested: Boolean = false
+        @SerialName("autovalue") val autoValue: String? = null
+        @SerialName("suggestedvalues") val suggestedValues: List<String> = emptyList()
+        val aliases: List<String> = emptyList()
+        private val deprecated: JsonElement? = null
+
+        val deprecatedAsBoolean get() = deprecated?.jsonPrimitive?.booleanOrNull ?: false
+        val deprecatedAsString get() = deprecated?.jsonPrimitive?.contentOrNull
```
dbrant, line 51: Can we clarify what the possible value(s) of `deprecated` can be? (can we have a single function that tells whether it's deprecated?)

dbrant (changes requested):

app/src/main/java/org/wikipedia/dataclient/mwapi/TemplateDataResponse.kt
cooltey: Most of the time, it will return `false`. But somehow, but in some cases like `Template: Help me`, it will return a `String` with the explanation of why it is deprecated. 

But for the case when it is deprecated, we can just ignore it and not show it at all.

> can we have a single function that tells whether it's deprecated?

Yes!

cooltey: Done!

PR #4515, opened 2024-03-07
Preview screen for talk messages

sharvaniharan:
**Phab:** https://phabricator.wikimedia.org/T358898


cooltey (changes requested):
Found several things: 
1. Go to any article and edit a section, the app crashes.
2. Go to a talk thread and click "Reply", when long press on the "Preview" icon, it shows the text from the action bar, and same thing happens to the preview screen as well.
3. Having weird background color when long-press on the hyperlinks in the license text, would it be better to remove it?

dbrant:
> 1. Go to any article and edit a section, the app crashes.

👍 

> 2. Go to a talk thread and click "Reply", when long press on the "Preview" icon, it shows the text from the action bar, and same thing happens to the preview screen as well.

👍 

> 3. Having weird background color when long-press on the hyperlinks in the license text, would it be better to remove it?

This doesn't seem like a problem with this PR, and looking quickly, I'm not sure what's causing that background. Let's investigate separately.

sharvaniharan:
This has passed design review @cooltey @dbrant

cooltey (approved):
Looks good to me! Will merge this after the new production is released.

PR #4518, opened 2024-03-07
Remove PreProdRelease condition from the MetricsEvent class

cooltey:
Per discussion from the Slack channel, it is time to include it in the production.


cjming:
@cooltey there's one more on line 59
https://github.com/wikimedia/apps-android-wikipedia/blob/main/app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsEvent.kt#L59

cooltey:
> @cooltey there's one more on line 59 https://github.com/wikimedia/apps-android-wikipedia/blob/main/app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsEvent.kt#L59

Thanks, done!

PR #4521, opened 2024-03-08
Adding add template data flow for wikitext keyboard

cooltey:

Bug:T355141


cooltey:
@dbrant @sharvaniharan This is ready for review.

app/src/main/java/org/wikipedia/edit/templates/TemplatesSearchActivity.kt
```diff
@@ -0,0 +1,220 @@
...
+        templatesSearchAdapter = TemplatesSearchAdapter()
+        binding.templateRecyclerView.layoutManager = LinearLayoutManager(this)
+        binding.templateRecyclerView.adapter = templatesSearchAdapter
+
+        insertTemplateFragment = supportFragmentManager.findFragmentById(R.id.insertTemplateFragment) as InsertTemplateFragment
+
+        binding.insertTemplateButton.setOnClickListener {
+            viewModel.selectedPageTitle?.let {
+                val list = Prefs.recentUsedTemplates.toMutableList()
+                list.add(it)
```
dbrant, line 81: This doesn't check if the template already exists in the list, which leads to duplicates:

image

app/src/main/java/org/wikipedia/edit/templates/TemplatesSearchActivity.kt
```diff
@@ -0,0 +1,220 @@
...
+                Prefs.recentUsedTemplates = list
```
dbrant, line 82: Should we also set a maximum number of items in the list, e.g. 100, so that it can't grow without bounds?

dbrant (changes requested):

app/src/main/java/org/wikipedia/edit/templates/TemplatesSearchActivity.kt
cooltey: Changed from `List` to `Set`, fixed!

app/src/main/java/org/wikipedia/edit/templates/TemplatesSearchActivity.kt
cooltey: Done.

PR #4543, opened 2024-03-21
Integrate with Phabricator message bot.

dbrant:
This integrates with our Phab bot which should automatically post messages on the corresponding task.
From now on, we can use the old-style GerritBot convention of mentioning the task at the bottom of the pull request description (by saying `Bug: T...` on its own line), and it will be posted automatically to the corresponding task.

(supports multiple tasks!)
(case insensitive!)
(optional whitespace between "Bug" and T number!)

Bug: T360680

### DIFF
--- a/.github/workflows/android_phab.yml
+++ b/.github/workflows/android_phab.yml
@@ -1,0 +1,24 @@
+name: Post to Phabricator
+
+on:
+  pull_request:
+    types: [opened, closed]
+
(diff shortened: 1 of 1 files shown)


tacsipacsi:
The job failed in #4586 for some reason, could you look into it? It’s not crucial in that particular PR, but it could fail in important cases as well.

PR #4544, opened 2024-03-21
Update string order for talk warn learnMore label

cooltey:
Test the Github autolinks with the Phab ticket action

Bug:T360680
BUG:T360680
Bug: T360680


cooltey:
self merged this.

PR #4546, opened 2024-03-22
Follow-up to template flow.

dbrant:
This fixes an important thing:
When a template _does not_ have TemplateData, we should still show the parameters of the template. This fixes the deserialization logic so that the parameters are shown correctly.


cooltey (changes requested):
Could you please check the `Template: Help me` and see if the changes work properly?

https://en.wikipedia.org/w/api.php?format=json&formatversion=2&errorformat=html&errorsuselocal=1&action=templatedata&includeMissingTitles=&converttitles=&lang=zh-tw&titles=Template%3A%20Help%20me

cooltey:
(and I think we'll need to update the search result manually by searching if there are any title matches locally and shifting it to the top of the list), I can create a PR for it.
Context: search "Help me" in `enwiki`.

dbrant:
> Could you please check the `Template: Help me` and see if the changes work properly?
> 
> https://en.wikipedia.org/w/api.php?format=json&formatversion=2&errorformat=html&errorsuselocal=1&action=templatedata&includeMissingTitles=&converttitles=&lang=zh-tw&titles=Template%3A%20Help%20me

@cooltey whoops, fixed! (made a typo)

app/src/main/java/org/wikipedia/dataclient/mwapi/TemplateDataResponse.kt
```diff
@@ -56,6 +62,6 @@ class TemplateDataResponse : MwResponse() {
 
         private val deprecatedAsString get() = deprecated?.jsonPrimitive?.contentOrNull
 
-        val isDeprecated get() = !deprecatedAsString.equals("false", true)
+        val isDeprecated get() = deprecatedAsString.equals("true", true)
```
cooltey, line 65: The default value of `deprecated` is `false`, but it will become a `String` if it is deprecated.

cooltey (changes requested):

app/src/main/java/org/wikipedia/dataclient/mwapi/TemplateDataResponse.kt
dbrant: ahh of course. Should be fixed now.

PR #4554, opened 2024-03-26
Fix: crash when tapping on the "Warn" button in the Patroller Tasks

cooltey:
The app crashes when following the steps below:
1. Go to `Edit patrol`
2. Select any edits
3. Click on "Warn"

It shows `kotlin.UninitializedPropertyAccessException: lateinit property adapter has not been initialized`.


dbrant:
Ha, I actually fix this crash (indirectly) in #4552, but this is good, too.

PR #4574, opened 2024-04-02
Greatly simplify retrieval and usage of UserIDs.

dbrant:
I noticed a weird-looking bit of code which can be hugely simplified. Please follow the inline comments.


app/src/main/java/org/wikipedia/WikipediaApp.kt
```diff
@@ -79,13 +78,7 @@ class WikipediaApp : Application() {
         }
 
     val appOrSystemLanguageCode: String
-        get() {
```
dbrant, line 82: This getter is accessed from a lot of places which need to know the "default" app language code.
But notice that it also checks the User ID of the current user on the current language wiki, and if the User ID is 0, it _kicks off a network request_ to fetch the User ID. This is very inefficient, and is doing way too much in a hidden way.

app/src/main/java/org/wikipedia/WikipediaApp.kt
```diff
@@ -79,13 +78,7 @@ class WikipediaApp : Application() {
...
-            val code = languageState.appLanguageCode
-            if (AccountUtil.getUserIdForLanguage(code) == 0) {
-                getUserIdForLanguage(code)
```
dbrant, line 85: If the getter for `appOrSystemLanguageCode` is called in quick succession from multiple places, it will kick off multiple unnecessary requests.
And my next question is, where exactly do we use the UserID?

app/src/main/java/org/wikipedia/analytics/eventplatform/EditAttemptStepEvent.kt
```diff
@@ -42,10 +44,13 @@ class EditAttemptStepEvent(private val event: EditAttemptStepInteractionEvent) :
         private fun submitEditAttemptEvent(action: String, editorInterface: String, pageTitle: PageTitle) {
             EventPlatformClient.submit(EditAttemptStepEvent(EditAttemptStepInteractionEvent(action,
                 WikipediaApp.instance.appInstallID, "", editorInterface,
-                INTEGRATION_ID, "", WikipediaApp.instance.getString(R.string.device_type).lowercase(), 0,
-                    if (AccountUtil.isLoggedIn) AccountUtil.getUserIdForLanguage(pageTitle.wikiSite.languageCode) else 0,
```
dbrant, line 46: This is _literally_ the only place where the UserID is used - for analytics purposes, in a single analytics funnel. Can there be a better way of getting the UserID, possibly without any additional network calls at all?

app/src/main/java/org/wikipedia/analytics/eventplatform/EditAttemptStepEvent.kt
```diff
@@ -42,10 +44,13 @@ class EditAttemptStepEvent(private val event: EditAttemptStepInteractionEvent) :
...
+                INTEGRATION_ID, "", WikipediaApp.instance.getString(R.string.device_type).lowercase(), 0, getUserIdForWikiSite(pageTitle.wikiSite),
                 1, pageTitle.prefixedText, pageTitle.namespace().code())))
         }
+
+        private fun getUserIdForWikiSite(wikiSite: WikiSite): Int {
+            return if (AccountUtil.isLoggedIn) SharedPreferenceCookieManager.instance.getCookieByName("UserID", wikiSite.authority(), false)?.toIntOrNull() ?: 0 else 0
```
dbrant, line 52: There is a more efficient way: extract the UserID directly from the cookie set by the current wiki subdomain. That's it -- none of the previous userId management code in AccountUtils is necessary anymore.

cooltey: This is great!

PR #4579, opened 2024-04-03
update to gradle 8.7

Jvr2022:


dbrant:
Thanks - Can you specify why this is beneficial?

Jvr2022:
Here are the changes

- Support for building projects with Java 22
- Avoiding Groovy build script compilation thanks to build cache support
- Build authoring improvements
- Better API for updating collection properties
- Conventions are honored
- Empty provider values are ignored
- Error and warning reporting improvements
- Configuration cache improvements
- Kotlin DSL improvements
- Build init improvements
- IDE Integration improvements
- Full Java 21 support
- Version catalog API in precompiled scripts
- Ability to enable Kotlin metadata version check for script compilation
And some more minor changes

dbrant:
Yes, those are the release notes, but how does this benefit our build process?
It is true that we like to be aggressive in using the latest versions of _library_ dependencies for the app itself, but how does using the bleeding-edge gradle wrapper improve our builds, if nothing is broken?

Jvr2022:
I open always prs to keep anything updated but idk how it improves

dbrant:
If you can't explain why your PR is an improvement, then I'm afraid we can't accept it.

PR #4581, opened 2024-04-03
Fix language-updating scripts, and update languages.

dbrant:
Our static language files haven't been updated in a while, and it turns out our Python scripts for updating the languages were broken. There have been a few changes that have made several (smaller) language wikis become redirects, but not complete redirects, rather partial redirects.

For example, the old Norwegian language code `nb` redirects fully to `no`:
https://nb.wikipedia.org
But, Alemannic wikipedia (`gsw`) is supposed to redirect to `als`, but actually redirects through incubator wiki:
https://gsw.wikipedia.org

This is the case for a few other language codes, and needs to be explicitly updated in our scripts.


cooltey (commented):
The PR looks good! Just have a question:

I noticed the order of static data items does not follow the A to Z order. Do you think it is possible to optimize it? For example, the `zh-classical` is not at the bottom of the list in `FileAliasData`.

PR #4583, opened 2024-04-03
Use Coroutine on Feed related classes

cooltey:
This is for the upcoming Chinese language variants fixes.

- Remove the usage of `Observable<AggregatedFeedContent>`.
- Remove RxJava usage from `AggregatedFeedContentClient` and `FeedCoordinator`.


cooltey:
Possible optimization: the loading circle should appear when every time reaches the bottom.

dbrant:
👏

PR #4586, opened 2024-04-04
Update links to [[mw:Wikimedia_Apps/Android_Suggested_edits]]

tacsipacsi:
While the English page redirects to the new title, translations don’t. This change will fuzzy the translations, so translators can notice and fix the broken links.

Also take the chance to make all four links point to m.mediawiki.org, instead of one pointing to m.mediawiki.org and three to www.mediawiki.org.

Bug: T348875


tacsipacsi:
Pointed out in https://www.mediawiki.org/wiki/Topic:Y29bhtvg48cce320.

tacsipacsi:
Thanks for the merge!

PR #4587, opened 2024-04-04
Fix: load correct article and description for the featured article

cooltey:
This is a temporary fix for T355192

### DIFF
--- a/app/src/main/java/org/wikipedia/language/LanguageUtil.kt
+++ b/app/src/main/java/org/wikipedia/language/LanguageUtil.kt
@@ -4,10 +4,17 @@ import android.content.Context
 import android.os.Build
 import android.view.inputmethod.InputMethodManager
 import androidx.core.os.LocaleListCompat
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.withContext
 import org.apache.commons.lang3.StringUtils
+import org.wikipedia.Constants
 import org.wikipedia.WikipediaApp
+import org.wikipedia.dataclient.ServiceFactory
+import org.wikipedia.dataclient.WikiSite
+import org.wikipedia.dataclient.page.PageSummary
 import org.wikipedia.util.StringUtil
-import java.util.*
+import java.util.Locale
 
 object LanguageUtil {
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/language/LanguageUtil.kt
```diff
@@ -144,4 +151,28 @@ object LanguageUtil {
     fun formatLangCodeForButton(languageCode: String): String {
         return languageCode.replace("-", "-\n")
     }
+
+    // TODO: This is a temporary fix for T355192
+    suspend fun getPageSummary(pageSummary: PageSummary?, wikiSite: WikiSite): PageSummary? {
```
dbrant, line 156: I'm feeling like a `Util` class should contain completely independent bits of code, with as few dependencies as possible. This feels like it's leaking too much stuff into a Util class. Do you think we could keep this as a private function in `AggregatedFeedContentClient`, to discourage its use in other places?

dbrant (changes requested):

app/src/main/java/org/wikipedia/language/LanguageUtil.kt
cooltey: You are right! Done.

cooltey:
Will have a follow-up PR using `prop=extracts` API instead of `page/summary`, because the article content itself loses some text inside the parentheses.

dbrant:
> Will have a follow-up PR using `prop=extracts` API instead of `page/summary`, because the article content itself loses some text inside the parentheses.

Sure, actually I'd like to discuss this further at the offsite, but may I suggest:
The [TextExtracts](https://www.mediawiki.org/wiki/Extension:TextExtracts) extension (which provides `prop=extracts`) is extremely old and unmaintained, and we should probably reduce our dependency on it. If you see issues with the output of the `page/summary` endpoint, it might be better for us to take ownership and update its output so that it meets our needs.

cooltey:
> Sure, actually I'd like to discuss this further at the offsite, but may I suggest:
> The [TextExtracts](https://www.mediawiki.org/wiki/Extension:TextExtracts) extension (which provides `prop=extracts`) is extremely old and unmaintained, and we should probably reduce our dependency on it. If you see issues with the output of the `page/summary` endpoint, it might be better for us to take ownership and update its output so that it meets our needs.

Got it, it makes sense to me. I'll look into the endpoint and try to fix it.

PR #4595, opened 2024-04-15
Add Chinese and Igbo for thrid roll-out for Patroller Tasks

cooltey:
Chinese language code: `zh`
Igbo language code: `ig`

No design review is needed.

Phabricator: T350512


dbrant:
> Phabricator: [T350512](https://phabricator.wikimedia.org/T350512)

Is this another variation of specifying the Phab task? Can we agree on which one to use, so that it gets picked up by the bot?

cooltey:
I thought Phabricator with T-number also works, but it didn't.

I can use Bug: T-number for future PRs.

PR #4596, opened 2024-04-16
Fix: The Featured Article widget cannot show the content properly

cooltey:
By tracking the logs, it looks like we still need to run `sendBroadcast` in order to run into the `onUpdate()` properly, otherwise the loop will end at the first `enqueue()` 

Bug: T362585


dbrant:
This definitely fixes the issue, but I'm left a bit puzzled - how did it work at all before? 😅

PR #4603, opened 2024-04-17
Fix showing Link Preview stub edit pencil only for Main namespace.

dbrant:
The way we were checking the Namespace of the PageTitle is incorrect, and in fact we should probably check other places in the code where we might do similar checks.
This changes the logic to use the `namespace` directly from the PageSummary response, which is the source of truth.

https://phabricator.wikimedia.org/T362815


cooltey:
It looks good to me. I looked into the code a bit, and found we should update `Namespac.fromLegacyString()` method since it always return `MAIN` for other namespaces. 

Also the `getPageTitle()` in `PageSummary` should be updated as well.

dbrant:
> we should update `Namespac.fromLegacyString()` method since it always return `MAIN` for other namespaces.

That's correct - We should make sure to use `namespace()` for detecting if the namespace is one of { FILE, TALK, USER, USER_TALK, SPECIAL }, in other words only the namespaces for which we have static AliasData, and **not** use it for detecting if it's anything other than that.

PR #4604, opened 2024-04-17
Update fromLegacyString method

cooltey:
- Update the `Namespace.fromLegacyString()` and try to find the closest namespace from the entries.
- Update `PageSummary.getPageTitle()` for a proper namespace
- Use `PageSumary.getPageTitle()` for `LinkPreviewViewModel.kt`

Pending discussion:
- Performance issue when searching namespace in `entries`?
- Should we show the `LinkPreviewDialog` for pages in all namespaces?

### DIFF
--- a/app/src/main/java/org/wikipedia/page/Namespace.kt
+++ b/app/src/main/java/org/wikipedia/page/Namespace.kt
@@ -4,8 +4,11 @@ import org.wikipedia.dataclient.WikiSite
 import org.wikipedia.extensions.getByCode
 import org.wikipedia.language.AppLanguageLookUpTable
 import org.wikipedia.model.EnumCode
-import org.wikipedia.staticdata.*
-import java.util.*
+import org.wikipedia.staticdata.FileAliasData
+import org.wikipedia.staticdata.SpecialAliasData
+import org.wikipedia.staticdata.TalkAliasData
+import org.wikipedia.staticdata.UserAliasData
+import org.wikipedia.staticdata.UserTalkAliasData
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/page/Namespace.kt
```diff
@@ -190,7 +193,7 @@ enum class Namespace(private val code: Int) : EnumCode {
             return if (UserTalkAliasData.valueFor(wiki.languageCode).equals(name, true) ||
                     UserTalkAliasData.valueFor(AppLanguageLookUpTable.FALLBACK_LANGUAGE_CODE).equals(name, true)) {
                 USER_TALK
-            } else MAIN
+            } else entries.firstOrNull { it.name.equals(name, true) } ?: MAIN
```
dbrant, line 196: This will only work for English. All namespaces have different localized names on each language wiki.

I think a better long-term solution will be to get rid of `fromLegacyString()` entirely, and have individual functions for `isTalk()`, `isUser()`, etc. and eliminate all the places where we attempt to check for `namespace() == MAIN`

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/Namespace.kt
dbrant: ...and then we could also remove >100 of the actual Namespace enum definitions that we don't actually use anywhere, too.

dbrant:

PR #4605, opened 2024-04-17
Hygiene: Remove useless constructor from Page, and improve ns detection.

dbrant:
The PageProperties class has a useless constructor, and the Page class performs an incorrect check for `Namespace.MAIN` that is better done via PageProperties.


app/src/main/java/org/wikipedia/page/Page.kt
```diff
@@ -1,8 +1,8 @@
 package org.wikipedia.page
 
-class Page(var title: PageTitle, var sections: List<Section> = emptyList(), var pageProperties: PageProperties) {
+class Page(val title: PageTitle, var sections: List<Section> = emptyList(), val pageProperties: PageProperties) {
     val displayTitle = pageProperties.displayTitle
     val isMainPage = pageProperties.isMainPage
-    val isArticle = !isMainPage && title.namespace() === Namespace.MAIN
+    val isArticle = !isMainPage && pageProperties.namespace.main()
```
dbrant, line 6: The main goal, as with previous similar PRs, is to reduce dependence on the PageTitle.namespace() function, which does not do exactly what we expect, and should eventually be removed.

PR #4606, opened 2024-04-17
Fix: crashes on Explore Feed when airplane mode is activated

cooltey:
Updates the proper `Dispatchers` for `CororutineScope` and `withContext`

https://play.google.com/console/u/1/developers/6169333749249604352/app/4972092403490495074/vitals/crashes/af0633ff24caa27fd1dd23c36433361a/details?days=28&versionCode=50482&isUserPerceived=true

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/aggregated/AggregatedFeedContentClient.kt
+++ b/app/src/main/java/org/wikipedia/feed/aggregated/AggregatedFeedContentClient.kt
@@ -146,36 +146,34 @@ class AggregatedFeedContentClient {
                     cb.error(caught)
                 }
             ) {
-                withContext(Dispatchers.IO) {
-                    val cards = mutableListOf<Card>()
-                    FeedContentType.aggregatedLanguages.forEach { langCode ->
-                        val wikiSite = WikiSite.forLanguageCode(langCode)
-                        val hasParentLanguageCode = !WikipediaApp.instance.languageState.getDefaultLanguageCode(langCode).isNullOrEmpty()
-                        var feedContentResponse = ServiceFactory.getRest(wikiSite).getFeedFeatured(date.year, date.month, date.day)
+                val cards = mutableListOf<Card>()
+                FeedContentType.aggregatedLanguages.forEach { langCode ->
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/feed/aggregated/AggregatedFeedContentClient.kt
```diff
@@ -140,13 +140,13 @@ class AggregatedFeedContentClient {
...
             val date = DateUtil.getUtcRequestDateFor(age)
-            aggregatedClient.clientJob = CoroutineScope(Dispatchers.Default).launch(
+            aggregatedClient.clientJob = CoroutineScope(Dispatchers.Main).launch(
                 CoroutineExceptionHandler { _, caught ->
                     L.v(caught)
                     cb.error(caught)
                 }
             ) {
-                withContext(Dispatchers.Main) {
+                withContext(Dispatchers.IO) {
```
dbrant, line 149: This whole `withContext... ` brace is actually unnecessary. We're already inside a coroutine scope, and network calls will be automatically placed in a `IO` context.

app/src/main/java/org/wikipedia/feed/becauseyouread/BecauseYouReadClient.kt
```diff
@@ -24,13 +24,13 @@ class BecauseYouReadClient : FeedClient {
...
         clientJob?.cancel()
-        clientJob = CoroutineScope(Dispatchers.Default).launch(
+        clientJob = CoroutineScope(Dispatchers.Main).launch(
             CoroutineExceptionHandler { _, caught ->
                 L.v(caught)
                 cb.success(emptyList())
             }
         ) {
-            withContext(Dispatchers.Main) {
+            withContext(Dispatchers.IO) {
```
dbrant, line 33: Same.

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/aggregated/AggregatedFeedContentClient.kt
cooltey: Done!

PR #4609, opened 2024-04-18
Refactoring FilePage related class to use coroutine and ViewModel

cooltey:
- Use ViewModel and coroutine in `FilePage` and `ImagePreviewDialog`
- Move `ImagePreviewDialog` to `commons` package folder instead of `view`.
- Update null-safe for some places that use `PageSummaryForEdit`.


app/src/main/java/org/wikipedia/commons/FilePage.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.wikipedia.dataclient.mwapi.MwQueryPage
+
+class FilePage {
+    var thumbnailWidth = 0
+    var thumbnailHeight = 0
+    var imageFromCommons = false
+    var showEditButton = false
+    var showFilename = false
+    var page = MwQueryPage()
+    var imageTags = emptyMap<String, List<String>>()
```
dbrant, line 12: Can all of the above fields be constructor parameters?
And in the places where you create instances of `FilePage`, use the constructor instead of `.apply { }`?

dbrant (changes requested):
Great, just a minor note.

app/src/main/java/org/wikipedia/commons/FilePage.kt
cooltey: Done!

PR #4617, opened 2024-04-19
[Spike] Experiment with Jetpack Compose

cooltey:
Following the migration strategy:
https://developer.android.com/develop/ui/compose/migrate/strategy

Ideally, we should create a new screen without having an XML file, but we do not have a new feature that needs a new screen, so I picked the "Send feedback" button in  `AboutActivity`, which replaced the `<Button>` with `ComposeView`. 

The TODOs will be:
1. Create a `@Composable` function with our styles definition in the `styles.xml`.

Findings:
1. Using the latest version of BOM will cause `checkstyle` failure. 
2. A very interesting comparison between XML and Compose: https://developer.android.com/develop/ui/compose/migrate/compare-metrics
3. The `@Preview` annotation does not work on some UI elements (e.g. `AndroidView`)

Bug: T358019


cooltey:
The lint error will be fixed after Google releases new version of AGP: https://issuetracker.google.com/issues/330911660

cooltey:
Spike is completed.

PR #4624, opened 2024-04-24
Google Pay integration.

dbrant:
https://phabricator.wikimedia.org/T362698


app/src/extra/res/layout/activity_donate.xml
```diff
@@ -0,0 +1,167 @@
...
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:paddingTop="12dp"
+                    style="@style/H1"
+                    android:textAlignment="center"
+                    android:inputType="numberDecimal" />
+
+            </com.google.android.material.textfield.TextInputLayout>
+
+            <CheckBox
```
cooltey, line 84: Please apply a proper text color to prevent the text is not visible in dark mode.

app/src/extra/res/layout/activity_donate.xml
```diff
@@ -0,0 +1,167 @@
...
+                android:id="@+id/checkBoxTransactionFee"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                style="@style/P"
+                android:paddingStart="8dp"
+                android:minHeight="0dp"
+                android:text="@string/donate_gpay_check_transaction_fee"
+                />
+
+            <CheckBox
```
cooltey, line 94: Same

app/src/extra/res/layout/activity_donate.xml
```diff
@@ -0,0 +1,167 @@
...
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                style="@style/P"
+                android:layout_marginTop="16dp"
+                android:paddingStart="8dp"
+                android:minHeight="0dp"
+                android:text="@string/donate_gpay_check_recurring"
+                />
+
+            <CheckBox
```
cooltey, line 105: Same

app/src/extra/res/layout/activity_donate.xml
```diff
@@ -0,0 +1,167 @@
...
+                    app:layout_constraintStart_toStartOf="parent"
+                    app:layout_constraintTop_toTopOf="parent"
+                    app:flow_horizontalGap="16dp"
+                    app:flow_verticalGap="16dp"
+                    app:flow_horizontalStyle="packed"
+                    app:flow_wrapMode="chain"/>
+
+            </androidx.constraintlayout.widget.ConstraintLayout>
+
+            <com.google.android.material.textfield.TextInputLayout
```
cooltey, line 64: The dollar sign seems to be to slightly invisible for dark theme.

cooltey (changes requested):

app/src/extra/res/layout/activity_donate.xml
dbrant: Ah, can't believe I missed these!

app/src/extra/java/org/wikipedia/donate/GooglePayViewModel.kt
```diff
@@ -0,0 +1,150 @@
...
+                        Prefs.paymentMethodsMerchantId = it.configuration?.merchantId.orEmpty()
+                        Prefs.paymentMethodsGatewayId = it.configuration?.gatewayMerchantId.orEmpty()
+                    }
+                }
+            }
+
+            if (Prefs.paymentMethodsMerchantId.isEmpty() || Prefs.paymentMethodsGatewayId.isEmpty()) {
+                uiState.value = NoPaymentMethod()
+            } else {
+                uiState.value = Resource.Success(donationConfig!!)
```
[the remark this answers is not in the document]
dbrant: I'm actually not sure this would be safer...  If `donationConfig` is null, it will throw a NullPointerException, which will be caught by the `CoroutineExceptionHandler` at the top, and will automatically provide a `Resource.Error`.

cooltey (changes requested):
Two questions:
1. When I changed the country code to TW (Taiwan), it showed the currency correctly, but it did not show the suggested amounts, is that expected? 



2. It looks like the input field allows the user to add multiple zeros to the beginning of the amount, and it does not affect the final amount. Just wonder if we need to remove the unnecessary zero automatically.


3. When processing the payment, the activity title still showed the "Select amount" title with an empty content. Do we need to change the activity title to "Processing payment" or other proper title?

dbrant:
@cooltey Excellent observations!

> 1. When I changed the country code to TW (Taiwan), it showed the currency correctly, but it did not show the suggested amounts, is that expected?

This actually might be a bug in the fundraising API -- it's not giving us the preset amounts for the TWD currency.
I'll make a note for the FR-tech team, and I updated the PR to bounce out of the GPay workflow (fall back to the browser) if the currency presets are not available. Thanks!

> 2. It looks like the input field allows the user to add multiple zeros to the beginning of the amount, and it does not affect the final amount. Just wonder if we need to remove the unnecessary zero automatically.

Ehhh I think that's an acceptable behavior. As long as the final amount is parsed correctly, why not let people enter 000001 when they really mean 1.

> 3. When processing the payment, the activity title still showed the "Select amount" title with an empty content. Do we need to change the activity title to "Processing payment" or other proper title?

¯\\\_(ツ)\_/¯  this is how it was designed, and it's probably perfectly good for a first iteration. We will surely make refinements as we go along, and I'll make an additional note to discuss this step at the next prioritization.

cooltey:
> @cooltey Excellent observations!
> 
> > 1. When I changed the country code to TW (Taiwan), it showed the currency correctly, but it did not show the suggested amounts, is that expected?
> 
> This actually might be a bug in the fundraising API -- it's not giving us the preset amounts for the TWD currency. I'll make a note for the FR-tech team, and I updated the PR to bounce out of the GPay workflow (fall back to the browser) if the currency presets are not available. Thanks!

Great! Thanks!

> 
> > 2. It looks like the input field allows the user to add multiple zeros to the beginning of the amount, and it does not affect the final amount. Just wonder if we need to remove the unnecessary zero automatically.
> 
> Ehhh I think that's an acceptable behavior. As long as the final amount is parsed correctly, why not let people enter 000001 when they really mean 1.

That makes sense to me.

> 
> > 3. When processing the payment, the activity title still showed the "Select amount" title with an empty content. Do we need to change the activity title to "Processing payment" or other proper title?
> 
> ¯\_(ツ)_/¯ this is how it was designed, and it's probably perfectly good for a first iteration. We will surely make refinements as we go along, and I'll make an additional note to discuss this step at the next prioritization.

Got it!


---

I have one final question before merging:

From the article donate campaign dialog, when the user selects "Donate now", the GPay bottomsheet pops up. If the user accidentally closes the bottomsheet by tapping on the dimmed side, the entire donation flow will still be considered as "Completed".

Do we need to update the preference logic to make sure the user clicks one of the options in the GPay bottomsheet instead of saving the preference when tapping on the the dialog "Donate now"?

cooltey (approved):

PR #4638, opened 2024-05-02
Refactoring GalleryActivity to use Coroutine and ViewModel

cooltey:
Bug: T303936

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/Service.kt
+++ b/app/src/main/java/org/wikipedia/dataclient/Service.kt
@@ -151,9 +151,6 @@ interface Service {
     ): MwQueryResponse
 
     @GET(MW_API_PREFIX + "action=query&meta=userinfo&prop=info&inprop=protection&uiprop=groups")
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/wikipedia/dataclient/Service.kt
```diff
@@ -141,7 +141,7 @@ interface Service {
     fun getVideoInfo(
         @Query("titles") titles: String,
         @Query("viextmetadatalanguage") lang: String
-    ): Observable<MwQueryResponse>
+    ): MwQueryResponse
```
dbrant, line 144: This function should also be marked as `suspend`, otherwise it will not load videos correctly in the Gallery.

dbrant (changes requested):

app/src/main/java/org/wikipedia/dataclient/Service.kt
cooltey: Done!

PR #4642, opened 2024-05-03
Update ImagePipelineBitmapGetter to a regular helper

cooltey:
The `ImagePipelineBitmapGetter` is an abstract class, but is used as a regular helper object instead of an abstract object.

This PR removes the `abstract` type from the class and adds a `Callback` to the parameter to use `lambda` when using it.


dbrant:
👏

PR #4650, opened 2024-05-08
Fix: add additional try and catch block for loading lead images info in offline

cooltey:
In the previous RxJava one, we ignore the case when the device is offline. This PR adds logic to check if the device is online and a try-and-catch block to prevent crashes.

### DIFF
--- a/app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
+++ b/app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
@@ -97,7 +97,7 @@ class LeadImagesHandler(private val parentFragment: PageFragment,
     private fun updateCallToAction() {
         dispose()
         pageHeaderView.callToActionText = null
-        if (!AccountUtil.isLoggedIn || leadImageUrl?.contains(Service.URL_FRAGMENT_FROM_COMMONS) != true || page == null) {
+        if (!WikipediaApp.instance.isOnline || !AccountUtil.isLoggedIn || leadImageUrl?.contains(Service.URL_FRAGMENT_FROM_COMMONS) != true || page == null) {
             return
         }
         title?.let {
--- a/app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
+++ b/app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
@@ -106,7 +106,7 @@ class LeadImagesHandler(private val parentFragment: PageFragment,
                 finalizeCallToAction()
                 return
             }
-            handlerJob = parentFragment.lifecycleScope.launch(CoroutineExceptionHandler { _, throwable ->
+            handlerJob = parentFragment.viewLifecycleOwner.lifecycleScope.launch(CoroutineExceptionHandler { _, throwable ->
                 L.e(throwable)
             }) {
                 lastImageTitleForCallToAction = imageTitle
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
```diff
@@ -108,35 +109,39 @@ class LeadImagesHandler(private val parentFragment: PageFragment,
             }
             handlerJob = parentFragment.lifecycleScope.launch {
                 withContext(Dispatchers.Main) {
```
dbrant, line 111: We actually don't need a `withContext()` block here; it's already in the correct context.

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
```diff
@@ -108,35 +109,39 @@ class LeadImagesHandler(private val parentFragment: PageFragment,
...
-                        captionSourcePageTitle = PageTitle(imageTitle, WikiSite(Service.COMMONS_URL, it.wikiSite.languageCode))
-                        captionSourcePageTitle!!.description = labelMap[it.wikiSite.languageCode]
-                        if (!labelMap.containsKey(it.wikiSite.languageCode)) {
-                            imageEditType = ImageEditType.ADD_CAPTION
-                        }
-                        if (WikipediaApp.instance.languageState.appLanguageCodes.size >= Constants.MIN_LANGUAGES_TO_UNLOCK_TRANSLATION) {
-                            WikipediaApp.instance.languageState.appLanguageCodes.firstOrNull { lang -> !labelMap.containsKey(lang) }?.run {
-                                imageEditType = ImageEditType.ADD_CAPTION_TRANSLATION
-                                captionTargetPageTitle = PageTitle(imageTitle, WikiSite(Service.COMMONS_URL, this))
+                    try {
```
dbrant, line 112: Instead of a big `try...catch` block, we can put a `CoroutineExceptionHandler` into the `launch` function above. (we do this in many places, for example `CaptchaHandler:60`)

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
```diff
@@ -108,35 +109,39 @@ class LeadImagesHandler(private val parentFragment: PageFragment,
...
+                            }
+                            if (imageEditType != ImageEditType.ADD_CAPTION && depicts.isEmpty()) {
+                                imageEditType = ImageEditType.ADD_TAGS
                             }
                         }
-                        if (imageEditType != ImageEditType.ADD_CAPTION && depicts.isEmpty()) {
-                            imageEditType = ImageEditType.ADD_TAGS
-                        }
+                        finalizeCallToAction()
+                    } catch (e: Exception) {
```
dbrant, line 142: Another reason against a big `try..catch` block is that we shouldn't catch the generic `Exception` class. When a `Job` is cancelled, it actually throws a `CancellationException`, which we _shouldn't_ catch. The CancellationException must be caught by the enclosing coroutine scope, so that it knows about the cancelled job.

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
cooltey: You are correct! I also changed `parentFragment.lifecycleScope` to `parentFragment.viewLifecycleOwner.lifecycleScope` to prevent crash.

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
cooltey: Done!

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
cooltey: Thanks and done!

PR #4651, opened 2024-05-08
Hygiene: clean up code in PageAvailableOfflineHandler

cooltey:
The process only needs the `Dispatchers.IO` and there's no need to apply `withContext()` and an additional `run{}` in `CoroutineExceptionHandler` block.


app/src/main/java/org/wikipedia/page/PageAvailableOfflineHandler.kt
```diff
@@ -16,19 +18,16 @@ object PageAvailableOfflineHandler {
...
         if (WikipediaApp.instance.isOnline) {
             callback.onFinish(true)
             return
         }
-        CoroutineScope(Dispatchers.Main).launch(CoroutineExceptionHandler { _, exception ->
-            run {
-                callback.onFinish(false)
-                L.w(exception)
-            }
+        CoroutineScope(Dispatchers.IO).launch(CoroutineExceptionHandler { _, exception ->
```
dbrant, line 26: Can we find a way to pass a `coroutineScope` into this function from the caller?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/PageAvailableOfflineHandler.kt
cooltey: The `setHistoryEntry` where the `check` function is used, will be called from `Fragment` and `view` class.

It is easy to use `lifecycleScope` from a `Fragment`, but not from a `view` class. I assume that if we want to do that, we will need to update the `ListCardView` and the parent class, not sure if there is another way of calling `lifecycleScope` instead of using `CoroutineScope` 🤔

dbrant: A `View` has a `context`, which is usually the Activity that owns the view, so we should be able to do something like `(view.context as? AppCompatActivity)?.lifecycleScope ?: CoroutineScope(IO)`

cooltey: Thanks! Done.

app/src/main/java/org/wikipedia/page/PageAvailableOfflineHandler.kt
```diff
@@ -16,19 +17,16 @@ object PageAvailableOfflineHandler {
...
-            run {
-                callback.onFinish(false)
-                L.w(exception)
-            }
+        lifeCycleScope.launch(CoroutineExceptionHandler { _, exception ->
+            callback.onFinish(false)
+            L.w(exception)
         }) {
-            val readingListPage = withContext(Dispatchers.IO) { AppDatabase.instance.readingListPageDao().findPageInAnyList(pageTitle) }
+            val readingListPage = AppDatabase.instance.readingListPageDao().findPageInAnyList(pageTitle)
```
dbrant, line 29: I believe this might still need to be wrapped in `withContext(Dispatchers.IO)`. The function `findPageInAnyList()` doesn't suspend, so it will run on the default dispatcher of the scope (Main).

The way this is currently written, it doesn't need to be a coroutine (there's nothing that `suspend`s).

dbrant (changes requested):

app/src/main/java/org/wikipedia/readinglist/db/ReadingListPageDao.kt
```diff
@@ -130,7 +136,7 @@ interface ReadingListPageDao {
         updateThumbAndDescriptionByName(pageProto.lang, pageProto.apiTitle, thumbUrl, description)
     }
 
-    fun findPageInAnyList(title: PageTitle): ReadingListPage? {
+    suspend fun findPageInAnyList(title: PageTitle): ReadingListPage? {
         return getPageByParams(
```
dbrant, line 140: The `getPageByParams` function also needs to be `suspend`, since that's where Room actually does its job.

dbrant (changes requested):
one tiny thing, sorry!

PR #4657, opened 2024-05-10
Migrate to MapLibre v11.

dbrant:
MapLibre recently released version 11, and changed all their package names from `mapbox` to `maplibre`.
This updates the MapLibre package itself, and the annotation plugin.


birkskyum:
@dbrant , think this can proceed now
- https://github.com/maplibre/maplibre-plugins-android/pull/55

dbrant:
@birkskyum Thanks for the ping; Looks good!

PR #4658, opened 2024-05-10
Fix RTL awareness when inserting images into wikitext.

dbrant:
When inserting images into wikitext (whether through Image Recommendations or the regular editor), we should check if the wiki is RTL, and use the corresponding `left` or `right` alignment, instead of always defaulting to `right`.

In addition, if the user doesn't change the default alignment, then we can just omit it from the insertion, since the `thumb` parameter automatically uses the wiki default for alignment.

### DIFF
--- a/app/src/main/java/org/wikipedia/edit/insertmedia/InsertMediaViewModel.kt
+++ b/app/src/main/java/org/wikipedia/edit/insertmedia/InsertMediaViewModel.kt
@@ -182,7 +182,7 @@ class InsertMediaViewModel(bundle: Bundle) : ViewModel() {
             magicWords[imageType]?.let { type ->
                 template += "|$type"
             }
-            if (imagePos != defaultImagePositionForLang(langCode)) {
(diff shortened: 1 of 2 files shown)


cooltey:
PR looks good to me. Just have one concern (sorry I should brought this out before merging).

In the desktop visual editor on enwiki, the default position is "Right"
on hewiki, the default position is "Left"

Do we need to align with the desktop settings?

Screenshot 2024-05-10 at 12-48-14 Felix M  Warburg House - Wikipedia
Screenshot 2024-05-10 at 12-46-54 ימי החנוכה – ויקיפדיה

dbrant:
> PR looks good to me. Just have one concern (sorry I should brought this out before merging).
> In the desktop visual editor on enwiki, the default position is "Right" on hewiki, the default position is "Left"
> Do we need to align with the desktop settings?

@cooltey Yes, this is how this PR is intended to work. (Does it not?)

cooltey:
Ah! You are right 😅

PR #4663, opened 2024-05-14
Improve integration with Metrics Platform library lifecycle.

dbrant:
There is one thing I noticed that might account for discrepancies seen between MP and MEP versions of certain events:
When the app is closed (which we were determining via an `onTrimMemory()` callback) we explicitly flush any unsent events in the MEP queue, but we were not doing the same for the MP queue.

Let's update the logic so that:
* We no longer rely on the `onTrimMemory` event, which does not behave consistently across devices and OS versions.
* We instead rely on our BaseActivity class to explicitly flush events in both MP and MEP at the same time.

https://phabricator.wikimedia.org/T363610

### DIFF
--- a/app/src/main/java/org/wikipedia/ActivityLifecycleHandler.kt
+++ b/app/src/main/java/org/wikipedia/ActivityLifecycleHandler.kt
@@ -6,8 +6,6 @@ import android.content.ComponentCallbacks2
 import android.content.res.Configuration
 import android.os.Build
 import android.os.Bundle
-import org.wikipedia.analytics.eventplatform.EventPlatformClient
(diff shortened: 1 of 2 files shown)


cooltey (approved):
Looks good to me. Just have a question that do we also need to add `onAppClose`  in `BaseActivity.onDestroy`

dbrant:
> Looks good to me. Just have a question that do we also need to add `onAppClose` in `BaseActivity.onDestroy`

Ah good question: Clare and I agreed to leave this out for now, while we finalize how session management will work within the library. This should not impact the flushing of unsent events, since onPause will handle it.

PR #4672, opened 2024-05-15
Remove Places survey.

dbrant:


app/src/main/java/org/wikipedia/settings/Prefs.kt
```diff
@@ -700,10 +699,6 @@ object Prefs {
         get() = PrefsIoUtil.getString(R.string.preference_key_places_wiki_code, WikipediaApp.instance.appOrSystemLanguageCode).orEmpty()
         set(value) = PrefsIoUtil.setString(R.string.preference_key_places_wiki_code, value)
 
-    var shouldShowOneTimePlacesSurvey
-        get() = PrefsIoUtil.getInt(R.string.preference_key_places_show_one_time_survey, PlacesFragment.SURVEY_NOT_INITIALIZED)
```
cooltey, line 704: `SURVEY_NOT_INITIALIZED` in `PlacesFragment` can be removed.

cooltey (changes requested):
One minor change needed.

app/src/main/java/org/wikipedia/settings/Prefs.kt
dbrant: done!

PR #4678, opened 2024-05-16
Fix: correctly extract language code from Url in WikiSite

cooltey:
Somehow the `UriUtil.getLanguageVariantFromUri()` does not get the correct language code from a URL, and always returns an empty string.

From the existing code, we should get the "second" from `parts` instead of the first, and that's why it is always empty.

This PR simplifies the logic (with Kotlin functions) and updates the test cases so we can go through all possible conditions. 

https://etherpad.wikimedia.org/p/DeepLinkForTests

### DIFF
--- a/app/src/test/java/org/wikipedia/dataclient/WikiSiteTest.kt
+++ b/app/src/test/java/org/wikipedia/dataclient/WikiSiteTest.kt
@@ -100,7 +100,31 @@ class WikiSiteTest {
 
     @Test
     fun testCtorUriLangVariant() {
-        val subject = WikiSite("zh.wikipedia.org/zh-hant/Foo")
+        var subject = WikiSite("en.wikipedia.org/wiki/Foo")
+        MatcherAssert.assertThat(subject.authority(), Matchers.`is`("en.wikipedia.org"))
+        MatcherAssert.assertThat(subject.subdomain(), Matchers.`is`("en"))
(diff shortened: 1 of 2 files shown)


app/src/test/java/org/wikipedia/dataclient/WikiSiteTest.kt
```diff
@@ -100,7 +100,31 @@ class WikiSiteTest {
...
+        MatcherAssert.assertThat(subject.subdomain(), Matchers.`is`("zh"))
+        MatcherAssert.assertThat(subject.languageCode, Matchers.`is`("zh-cn"))
+        MatcherAssert.assertThat(subject.scheme(), Matchers.`is`("https"))
+        MatcherAssert.assertThat(subject.dbName(), Matchers.`is`("zhwiki"))
+        MatcherAssert.assertThat(subject.url(), Matchers.`is`("https://zh.wikipedia.org"))
+
+        subject = WikiSite("zh.wikipedia.org/zh-hant/Foo")
         MatcherAssert.assertThat(subject.authority(), Matchers.`is`("zh.wikipedia.org"))
         MatcherAssert.assertThat(subject.subdomain(), Matchers.`is`("zh"))
         MatcherAssert.assertThat(subject.languageCode, Matchers.`is`("zh-hant"))
```
dbrant, line 130: Goood catch. What a bizarre coincidence that this test ^ was actually passing!

PR #4682, opened 2024-05-20
Use Jsoup's stream implementation

Isira-Seneviratne:
Use Jsoup's custom stream implementation for lazy evaluation of HTML elements.


dbrant:
Could you be more specific about why this is an improvement?

Isira-Seneviratne:
> Could you be more specific about why this is an improvement?

The `select` method that was being used earlier retrieves all the HTML elements at once, while this change retrieves them one by one, which would improve performance when there are a lot of elements.

dbrant:
A bit of profiling shows no difference in performance.

PR #4687, opened 2024-05-20
Bump Metrics Platform java lib version 2.7

cjming:
- Update schema ids for article instruments using custom schemas
- Add device_family to agent data object

Bug: [T365179](https://phabricator.wikimedia.org/T365179)


app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
```diff
@@ -19,6 +20,7 @@ object MetricsPlatform {
         "WikipediaApp/" + BuildConfig.VERSION_NAME,
         "android",
         "app",
+        Build.MODEL + Build.DEVICE,
         WikipediaApp.instance.languageState.systemLanguageCode,
```
cjming, line 24: @dbrant is this the right data to populate `device_family` in the agent data object?

app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
```diff
@@ -19,6 +20,7 @@ object MetricsPlatform {
         "WikipediaApp/" + BuildConfig.VERSION_NAME,
         "android",
         "app",
+        Build.MODEL + Build.DEVICE,
```
cjming, line 23: @dbrant is this the right data to populate `device_family` in the agent data object?

cjming:
@dbrant I still have to push out the MP library release but once i do, i'll mark this PR as ready -- until then tho, a Q for you inline about populating `agent.device_family`

app/src/main/java/org/wikipedia/analytics/metricsplatform/MetricsPlatform.kt
dbrant: @cjming It depends -- I don't think I'm familiar with the purpose of the `device_family` field; What would we like it to be?

dbrant: Ah sorry, I see a [comment](https://phabricator.wikimedia.org/T357371#9746840) that basically requests the `device_family` to be the same as what we're currently sending in our User-Agent.

If we want it to be exactly the same as the current User-Agent, then it should be just `Build.MODEL`, but if we need it to be like Shay requests in the above comment, it should be `Build.BRAND + " " + Build.MODEL`.

cjming: thanks! that's perfect - updated accordingly

PR #4689, opened 2024-05-21
Follow-up to coroutines in PageFragmentLoadState.

dbrant:
* Remove duplicate (and deprecated!) error handler from WebViewClient, which was causing incorrect  generic `RuntimeException`s to be surfaced.
* Consolidate code in PageFragmentLoadState into a single `loadPage` function, which can make more intuitive use of coroutines and async.
* Updated a couple more usages of `lifecycleScope` instead of new `CoroutineScope` creation.


app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
```diff
@@ -116,80 +113,72 @@ class PageFragmentLoadState(private var model: PageViewModel,
...
-                pageLoadFromNetwork(it)
-            }
-        }
-    }
-
-    private suspend fun pageLoadFromNetwork(title: PageTitle) {
-        fragment.updateQuickActionsAndMenuOptions()
-        if (!fragment.isAdded) {
-            return
-        }
```
cooltey, line 135: Do we still need this in the `pageLoad()`?

dbrant: This is exactly the benefit of using the `lifecycleScope` from the parent fragment or activity. If the fragment is destroyed, the `lifecycleScope` will automatically cancel any running coroutines. So, if we follow this pattern, we will never again need to check if a fragment is added, or if a View is attached to a window.

cooltey: Got it, it makes sense to me. Thanks!

PR #4694, opened 2024-05-23
GPay: use correct formula for transaction fee amount.

dbrant:
The transaction fee should be 4% of the donation amount, or the "minimum" transaction fee given by the API, whichever is greater.

https://phabricator.wikimedia.org/T365307


dbrant:
superseded by #4776

PR #4695, opened 2024-05-24
Build dialogs for removing generic Chinese language variant codes

cooltey:
- Add two dialogs for primary language code and secondary language codes.
- Update test cases.
- Update `refresh()` function in `FeedFragment` to make sure the feed is refreshed correctly.
- If `Developer Settings` is on, you will still be able to see `zh-hant` and `zh-hans` language codes. 

Bug: T360585


app/src/main/java/org/wikipedia/feed/view/RegionalLanguageVariantSelectionDialog.kt
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    companion object {
+
+        private val nonRegionalLanguageVariants = listOf(
+            AppLanguageLookUpTable.TRADITIONAL_CHINESE_LANGUAGE_CODE,
+            AppLanguageLookUpTable.SIMPLIFIED_CHINESE_LANGUAGE_CODE
+        )
+
+        private val regionalLanguageVariants = listOf(
```
dbrant, line 74: Extremely minor optimization: this list can be moved to the class itself (not the companion object), so that it's not persisted beyond the lifetime of this dialog.

app/src/main/java/org/wikipedia/feed/view/RegionalLanguageVariantSelectionDialog.kt
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    private fun setPositiveButtonEnabled(enabled: Boolean) {
+        dialog?.getButton(AlertDialog.BUTTON_POSITIVE)?.isEnabled = enabled
+    }
+
+    companion object {
+
+        private val nonRegionalLanguageVariants = listOf(
```
dbrant, line 69: Another extremely minor thing: this list can be a local variable in the function where it's used, so that it's also not persisted beyond the lifetime of this dialog.

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/view/RegionalLanguageVariantSelectionDialog.kt
cooltey: Sure! Done.

app/src/main/java/org/wikipedia/feed/view/RegionalLanguageVariantSelectionDialog.kt
cooltey: Done.

PR #4701, opened 2024-05-28
Refactor Feed clients to use lifecycleScope from parent activity.

dbrant:
This removes all other remaining usages of explicit `CoroutineScope()` instances.


app/src/main/java/org/wikipedia/extensions/View.kt
```diff
@@ -0,0 +1,11 @@
+package org.wikipedia.extensions
+
+import android.view.View
+import androidx.appcompat.app.AppCompatActivity
+import androidx.lifecycle.lifecycleScope
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+
+fun View.coroutineScope(): CoroutineScope {
```
cooltey, line 9: Would it be better if we provide an optional parameter to send a `Dispatchers`? 

`View.coroutineScope(coroutineContext: CoroutineContext = Dispatchers.Main)`

cooltey (commented):
Looks good to me, just have one comment for the extension.

app/src/main/java/org/wikipedia/extensions/View.kt
dbrant: sure!

PR #4703, opened 2024-05-29
Use coroutine on EventPlatformClient and do not send while offline

cooltey:
- Use coroutine with `CoroutineScope(Dispatchers.IO)` since we are not sending `lifecycleScope` from separate events. 
- Do not process the `sendEventsForStream()` when the device is offline.

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
@@ -37,6 +37,8 @@ object EventPlatformClient {
      */
     private var ENABLED = WikipediaApp.instance.isOnline
 
+    private val coroutineScope = CoroutineScope(Dispatchers.IO)
+
     fun setStreamConfig(streamConfig: StreamConfig) {
         STREAM_CONFIGS[streamConfig.streamName] = streamConfig
     }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
```diff
@@ -159,39 +164,38 @@ object EventPlatformClient {
...
-                            } else {
-                                // Something unexpected happened.
-                                if (ReleaseUtil.isDevRelease) {
-                                    // If it's a pre-beta release, show a loud toast to signal that
-                                    // a potential issue should be investigated.
-                                    WikipediaApp.instance.mainThreadHandler.post {
-                                        Toast.makeText(WikipediaApp.instance, it.message, Toast.LENGTH_LONG).show()
-                                    }
-                                }
+            if (!WikipediaApp.instance.isOnline) {
```
dbrant, line 168: This check is technically not necessary. In the `sendAllScheduled()` function, it checks `if (ENABLED)`, and the variable `ENABLED` is set automatically by the `ConnectionStateMonitor` which detects whenever the app goes offline and online.

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
cooltey: Removed.

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
```diff
@@ -159,39 +164,38 @@ object EventPlatformClient {
...
+                return
+            }
+            CoroutineScope(Dispatchers.IO).launch(CoroutineExceptionHandler { _, caught ->
```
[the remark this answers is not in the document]
cooltey: Yes and done!

PR #4705, opened 2024-05-29
fancycaptchareload variable converted to camel case by adding Kotlinx serialization 

dinesh-thiyagarajan:
1. Updatable typo fix in Android Manifest file
2. fancycaptchareload variable converted to camel case by adding SerializedName in Captcha.kt file

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -4,10 +4,10 @@
     android:installLocation="auto"
     tools:node="strict">
     <!--
-    If the build is intended to be Google Play-updatable, the package name
+    If the build is intended to be Google Play-updateable, the package name
     should be the same as for Google Play (name depends on Beta or Stable).
-    Otherwise, if the build is not intended to be Google Play-updatable (e.g.,
-    updatable via an alternative appstore update mechanism), the package name
+    Otherwise, if the build is not intended to be Google Play-updateable (e.g.,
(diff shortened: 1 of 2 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -4,10 +4,10 @@
     android:installLocation="auto"
     tools:node="strict">
     <!--
-    If the build is intended to be Google Play-updateable, the package name
+    If the build is intended to be Google Play-updatable, the package name
```
dbrant, line 7: "Updateable" is a perfectly valid spelling, and does not need to be changed.

app/src/main/java/org/wikipedia/captcha/Captcha.kt
```diff
@@ -1,12 +1,13 @@
 package org.wikipedia.captcha
 
+import com.google.gson.annotations.SerializedName
 import kotlinx.serialization.Serializable
 import org.wikipedia.dataclient.mwapi.MwResponse
 
 @Serializable
-data class Captcha(private val fancycaptchareload: FancyCaptchaReload) : MwResponse() {
+data class Captcha(@SerializedName("fancycaptchareload") private val fancyCaptchaReload: FancyCaptchaReload) : MwResponse() {
```
dbrant, line 8: We do not use Gson for serialization, therefore the `SerializedName` annotation will have no effect. Please use the correct annotation for kotlinx.serialization.

PR #4706, opened 2024-05-29
Added kotlinx serialization for fancycaptchareload variable

dinesh-thiyagarajan:
Added kotlinx serialization for fancycaptchareload variable


dinesh-thiyagarajan:
@dbrant Have updated this PR with Kotlix serialization hope this solves the issue :)

PR #4714, opened 2024-05-31
Removed unused parameter

dinesh-thiyagarajan:
Removed unwanted context parameter for PlacesLangListFilterAdapter constructor


dbrant:
Sorry, this is not substantial enough to be a pull request.

PR #4720, opened 2024-06-04
Refactoring LoginClient to use Coroutine

cooltey:
- Updated service functions to be `suspend` functions.
- Simplified functions in `LoginClient` and used `Coroutine`.
- Separated `LoginResponse` and `LoginFailedException` out as individual classes.

### DIFF
--- a/app/src/main/java/org/wikipedia/login/LoginClient.kt
+++ b/app/src/main/java/org/wikipedia/login/LoginClient.kt
@@ -1,24 +1,21 @@
 package org.wikipedia.login
 
 import android.widget.Toast
-import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers
-import io.reactivex.rxjava3.core.Observable
-import io.reactivex.rxjava3.disposables.CompositeDisposable
-import io.reactivex.rxjava3.schedulers.Schedulers
-import kotlinx.serialization.SerialName
-import kotlinx.serialization.Serializable
+import androidx.lifecycle.LifecycleCoroutineScope
+import kotlinx.coroutines.CoroutineExceptionHandler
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/wikipedia/login/LoginClient.kt
```diff
@@ -1,24 +1,21 @@
...
 import org.wikipedia.dataclient.ServiceFactory
 import org.wikipedia.dataclient.WikiSite
-import org.wikipedia.dataclient.mwapi.MwResponse
 import org.wikipedia.util.log.L
 import java.io.IOException
 
 class LoginClient {
 
-    private val disposables = CompositeDisposable()
+    private var clientJob: Job? = null
```
cooltey, line 18: @dbrant Do we still need this `Job` for canceling the current request? It will be called when the `LoginActivity` on `onStop`. I wonder if we can let the activity lifecycle decide whether to cancel the `coroutine` process or do it explicitly.

cooltey: Thanks! Done.

app/src/main/java/org/wikipedia/login/LoginClient.kt
```diff
@@ -27,150 +24,68 @@ class LoginClient {
         fun error(caught: Throwable)
     }
 
-    fun request(wiki: WikiSite, userName: String, password: String, cb: LoginCallback) {
+    fun login(coroutineScope: LifecycleCoroutineScope, wiki: WikiSite, userName: String, password: String,
```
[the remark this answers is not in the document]
cooltey: Yes, done.

PR #4722, opened 2024-06-05
Fix tint and logic of Edit pencil in page toolbar.

dbrant:
In Dark mode, when you have the "Edit article" option configured in your toolbar (instead of the overflow menu), the tint of the icon is incorrect:

image

In addition, the logic of updating the toolbar icons was slightly incorrect, and could lead to the Edit icon being in the wrong state (with the "lock" icon), or being enabled even while the article is still loading.
This fixes the logic and the tinting.

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageFragment.kt
+++ b/app/src/main/java/org/wikipedia/page/PageFragment.kt
@@ -463,6 +463,7 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
             }
             model.page?.let { page ->
                 page.pageProperties.protection = JsonUtil.decodeFromString(value)
+                updateQuickActionsAndMenuOptions()
             }
         }
     }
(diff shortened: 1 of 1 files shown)


cooltey:
Not sure but I am still seeing the edit pencil is in the incorrect state (the article is locked but the toolbar/overflow menu is still editable)

dbrant:
> Not sure but I am still seeing the edit pencil is in the incorrect state (the article is locked but the toolbar/overflow menu is still editable)

@cooltey  This is expected (and not impacted by this PR). The edit pencil **in the article** represents whether the article is protected, but the edit pencil **in the app** represents whether you (your account) are able to edit it.

Ideally, the PCS logic in the article should adjust the pencil based on the `userGroups` that we pass during the Setup phase. If PCS isn't doing this correctly, that's probably a separate issue in the page-library layer.

cooltey:
> > Not sure but I am still seeing the edit pencil is in the incorrect state (the article is locked but the toolbar/overflow menu is still editable)
> 
> @cooltey This is expected (and not impacted by this PR). The edit pencil **in the article** represents whether the article is protected, but the edit pencil **in the app** represents whether you (your account) are able to edit it.
> 
> Ideally, the PCS logic in the article should adjust the pencil based on the `userGroups` that we pass during the Setup phase. If PCS isn't doing this correctly, that's probably a separate issue in the page-library layer.

Got it, it makes sense to me.

PR #4729, opened 2024-06-07
Refactoring NotificationDirectReplyHelper to use coroutine

cooltey:
- Use coroutine.
- Add permission check


app/src/main/java/org/wikipedia/talk/NotificationDirectReplyHelper.kt
```diff
@@ -1,91 +1,75 @@
...
-            }
-            .subscribe({
-                // revisionForUndo = it
-                Toast.makeText(context, R.string.notifications_direct_reply_success, Toast.LENGTH_LONG).show()
-            }, {
-                L.e(it)
-                fallBackToTalkPage(context, title)
-            })
+    private suspend fun waitForUpdatedRevision(context: Context, wiki: WikiSite, title: PageTitle, newRevision: Long, notificationId: Int) {
+        var talkPageResponse = ServiceFactory.getRest(wiki).getTalkPage(title.prefixedText)
```
dbrant, line 64: Can we make it so that `getTalkPage()` is called just once (inside the loop), instead of twice?

dbrant (changes requested):

app/src/main/java/org/wikipedia/talk/NotificationDirectReplyHelper.kt
```diff
@@ -1,91 +1,75 @@
...
+        var tires = 0
```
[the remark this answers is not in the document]
cooltey: Done 😓

app/src/main/java/org/wikipedia/talk/NotificationDirectReplyHelper.kt
cooltey: Yes!

PR #4732, opened 2024-06-10
Convert Push Notification registry calls to use coroutines.

dbrant:


app/src/main/java/org/wikipedia/WikipediaApp.kt
```diff
@@ -239,21 +240,24 @@ class WikipediaApp : Application() {
...
+                Prefs.pushNotificationTokenOld = ""
+
+                val token = ServiceFactory.get(wikiSite).getToken().query!!.csrfToken()
+                WikipediaFirebaseMessagingService.unsubscribePushToken(token!!, Prefs.pushNotificationToken)
+                ServiceFactory.get(wikiSite).postLogout(token)
+            } finally {
+                SharedPreferenceCookieManager.instance.clearAllCookies()
+                L.d("Logout complete.")
+            }
+        }
```
cooltey, line 260: Have another idea on this instead `try-finally` inside the `launch` block:

Adding `.invokeOnCompletion { }` and treat it as a `finally`:

```
MainScope().launch(CoroutineExceptionHandler { _, t ->
            L.e(t)
}) {
...
}.invokeOnCompletion {
   SharedPreferenceCookieManager.instance.clearAllCookies()
   L.d("Logout complete.")
}
```

dbrant: Sure!

PR #4739, opened 2024-06-11
Use displayMetrics.density instead of deprecated methods

cooltey:
`defaultDisplay.getMetrics(metrics)` is deprecated. But we still need to use `scaledDensity` for converting from `sp` to `dp`.

### DIFF
--- a/app/src/main/java/org/wikipedia/util/DimenUtil.kt
+++ b/app/src/main/java/org/wikipedia/util/DimenUtil.kt
@@ -42,7 +42,7 @@ object DimenUtil {
     }
 
     fun getFontSizeFromSp(window: Window, fontSp: Float): Float {
-        return fontSp / window.context.resources.displayMetrics.density
+        return fontSp / window.context.resources.displayMetrics.scaledDensity
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/util/DimenUtil.kt
```diff
@@ -42,9 +42,7 @@ object DimenUtil {
     }
 
     fun getFontSizeFromSp(window: Window, fontSp: Float): Float {
-        val metrics = DisplayMetrics()
-        window.windowManager.defaultDisplay.getMetrics(metrics)
-        return fontSp / metrics.scaledDensity
+        return fontSp / window.context.resources.displayMetrics.scaledDensity
```
dbrant, line 45: If we're already updating this code, can we just use the new (non-deprecated) suggested usage of `TypedValueCompat.applyDimension(...)` instead?

dbrant (changes requested):

app/src/main/java/org/wikipedia/util/DimenUtil.kt
cooltey: Thanks! Done.

PR #4757, opened 2024-06-25
Prevent possible silent errors in reading lists.

dbrant:
There are a couple of places where silent errors could be produced, potentially leading to incorrect behavior.


app/src/main/java/org/wikipedia/readinglist/AddToReadingListDialog.kt
```diff
@@ -90,7 +90,7 @@ open class AddToReadingListDialog : ExtendedBottomSheetDialogFragment() {
             displayedLists.clear()
             displayedLists.addAll(readingLists)
             if (!showDefaultList && displayedLists.isNotEmpty()) {
-                displayedLists.removeAt(0)
```
dbrant, line 93: The "first" item in the list of reading lists is not necessarily the default list. We should use a more proper way of addressing the default list, instead of assuming it exists at position 0.

app/src/main/java/org/wikipedia/readinglist/db/ReadingListDao.kt
```diff
@@ -52,7 +52,7 @@ interface ReadingListDao {
         val lists = getListsWithoutContents()
         val pages = AppDatabase.instance.readingListPageDao().getAllPagesToBeSynced()
         pages.forEach { page ->
-            lists.first { it.id == page.listId }.apply { this.pages.add(page) }
```
dbrant, line 55: If the database contains an orphaned page (not belonging to any reading list, possibly caused by a crash while syncing), then this will crash silently, and prevent syncing from continuing.

PR #4764, opened 2024-06-28
Open Special:ChangePassword in an external browser in Notifications

cooltey:
If you click on "Change password" from the secondary link in the overflow menu in Notification, it will open the `Special:ChangePassword` in the app, and it will give you a warning message about the login session if you are trying to log in.

This PR adds a logic to check the notification category and opens the `LOGIN_FAIL` links in an external browser.


app/src/main/java/org/wikipedia/notifications/NotificationLinkHandler.kt
```diff
@@ -11,7 +11,9 @@ import org.wikipedia.page.PageTitle
 import org.wikipedia.util.CustomTabsUtil
 import org.wikipedia.util.log.L
 
-class NotificationLinkHandler constructor(context: Context) : LinkHandler(context) {
+class NotificationLinkHandler(context: Context) : LinkHandler(context) {
+
+    var category: NotificationCategory? = null
```
dbrant, line 16: Can we find a way to make the `category` be part of the constructor of `NotificationLinkHandler`?

dbrant (changes requested):

app/src/main/java/org/wikipedia/notifications/NotificationLinkHandler.kt
cooltey: Yes!

PR #4772, opened 2024-07-01
Fix: open wikidata history page in a customTab in the Diff screen

cooltey:
If you check the diff page of any Wikidata edits and click on the blue link to check its history page, the app will show an error message in the `LinkPreviewDialog` instead of opening in a browser.


app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
```diff
@@ -280,6 +281,8 @@ class ArticleEditDetailsFragment : Fragment(), WatchlistExpiryDialog.Callback, M
                 startActivity(TalkTopicsActivity.newIntent(requireContext(), viewModel.pageTitle, InvokeSource.DIFF_ACTIVITY))
             } else if (viewModel.pageTitle.namespace() == Namespace.FILE) {
                 startActivity(FilePageActivity.newIntent(requireContext(), viewModel.pageTitle))
+            } else if (viewModel.pageTitle.wikiSite.dbName() == Constants.WIKIDATA_DB_NAME) {
```
dbrant, line 284: Do we want to check for `COMMONS` here, too?

dbrant (changes requested):

app/src/main/java/org/wikipedia/diff/ArticleEditDetailsFragment.kt
cooltey: Done!

PR #4775, opened 2024-07-02
Properly ask for permission to show notifications on API 33+.

dbrant:
This will correctly ask for permissions to show notifications on Android 13 and higher.

This may need a bit of design input in the future, but for now the logic for asking for the permission will be:
* Only if the user has passed initial onboarding
* Only if the user is logged in
* On every activity `onStart`, the permission will be checked, but only once per hour at the most.

### DIFF
--- a/app/src/main/java/org/wikipedia/notifications/NotificationPresenter.kt
+++ b/app/src/main/java/org/wikipedia/notifications/NotificationPresenter.kt
@@ -46,7 +46,7 @@ object NotificationPresenter {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU &&
             !Prefs.isInitialOnboardingEnabled &&
             AccountUtil.isLoggedIn &&
-            (millisSinceLastRequest > TimeUnit.HOURS.toMillis(1)) &&
+            (millisSinceLastRequest > TimeUnit.HOURS.toMillis(1) || millisSinceLastRequest < 0) &&
             ContextCompat.checkSelfPermission(context, Manifest.permission.POST_NOTIFICATIONS) != PackageManager.PERMISSION_GRANTED) {
             launcher.launch(Manifest.permission.POST_NOTIFICATIONS)
             lastPermissionRequestTime = System.currentTimeMillis()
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/wikipedia/notifications/NotificationPresenter.kt
```diff
@@ -1,37 +1,57 @@
...
+
+    fun maybeRequestPermission(context: Context, launcher: ActivityResultLauncher<String>) {
+        val millisSinceLastRequest = System.currentTimeMillis() - lastPermissionRequestTime
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU &&
+            !Prefs.isInitialOnboardingEnabled &&
+            AccountUtil.isLoggedIn &&
+            (millisSinceLastRequest > TimeUnit.HOURS.toMillis(1) || millisSinceLastRequest < 0) &&
+            ContextCompat.checkSelfPermission(context, Manifest.permission.POST_NOTIFICATIONS) != PackageManager.PERMISSION_GRANTED) {
+            launcher.launch(Manifest.permission.POST_NOTIFICATIONS)
+            lastPermissionRequestTime = System.currentTimeMillis()
```
cooltey, line 51: If I click on "Don't allow" and close the app, and next time the permission check dialog will show up again, if I click "Don't allow" again, it will not show up after the second deny. 

It looks like the system saves the preference somewhere in the device, just wondering, would it be possible to use that system preference in this logic so we do not have to show the dialog twice?

dbrant: > It looks like the system saves the preference somewhere in the device, just wondering, would it be possible to use that system preference in this logic so we do not have to show the dialog twice?

Why wouldn't we want to show the dialog twice? This would be the same as any other Permissions workflow, e.g. asking for location permissions in Places, asking for storage permissions when downloading a file. All of those can be denied, and will be asked again.

cooltey: Got it, it makes sense to me as long as it aligns with the other permission checks.

PR #4777, opened 2024-07-02
Experiment with non-modal link previews.

dbrant:
Mostly to be tested when reading articles, and in Places.
(other usages of link previews throughout the app are not expected to work properly yet.)

https://phabricator.wikimedia.org/T356246


dbrant:

PR #4779, opened 2024-07-02
Use coroutine in RecurringTask

cooltey:
This PR also updates the usage of `WikipediaApp.instance` and moves it to the `RecurringTask` class.

### DIFF
--- a/app/src/main/java/org/wikipedia/activity/BaseActivity.kt
+++ b/app/src/main/java/org/wikipedia/activity/BaseActivity.kt
@@ -97,7 +97,7 @@ abstract class BaseActivity : AppCompatActivity(), ConnectionStateMonitor.Callba
         }
 
         // Conditionally execute all recurring tasks
-        RecurringTasksExecutor(WikipediaApp.instance).run()
+        RecurringTasksExecutor(lifecycleScope).run()
         if (Prefs.isReadingListsFirstTimeSync && AccountUtil.isLoggedIn) {
             Prefs.isReadingListsFirstTimeSync = false
             Prefs.isReadingListSyncEnabled = true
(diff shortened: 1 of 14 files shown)


app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -97,7 +97,7 @@ abstract class BaseActivity : AppCompatActivity(), ConnectionStateMonitor.Callba
         }
 
         // Conditionally execute all recurring tasks
-        RecurringTasksExecutor(WikipediaApp.instance).run()
+        RecurringTasksExecutor(lifecycleScope).run()
```
dbrant, line 100: Actually I don't think the Recurring Tasks should be bound to the `lifecycleScope` of this activity (they should be allowed to run independently of any activity). Can we make them use `MainScope` instead?

app/src/main/java/org/wikipedia/recurring/DailyEventTask.kt
```diff
@@ -1,22 +1,21 @@
...
-class DailyEventTask(context: Context) : RecurringTask() {
-    override val name = context.getString(R.string.preference_key_daily_event_time_task_name)
+class DailyEventTask(private val app: WikipediaApp) : RecurringTask() {
+    override val name = app.getString(R.string.preference_key_daily_event_time_task_name)
 
     override fun shouldRun(lastRun: Date): Boolean {
         return millisSinceLastRun(lastRun) > TimeUnit.DAYS.toMillis(1)
     }
 
     override fun run(lastRun: Date) {
```
dbrant, line 17: The `run()` function and `runIfNecessary()` should probably be `suspend` functions (and run in `IO` dispatcher).

dbrant (changes requested):

app/src/main/java/org/wikipedia/recurring/DailyEventTask.kt
cooltey: Done!

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
cooltey: I see. Done.

dbrant:
👏

PR #4789, opened 2024-07-08
Show a snackbar message for places that do have articles nearby

cooltey:
Bug: T368792

This PR also adds a preference to `developer_preferences` for resetting the stored location information.


app/src/main/java/org/wikipedia/places/PlacesFragment.kt
```diff
@@ -629,6 +644,15 @@ class PlacesFragment : Fragment(), LinkPreviewDialog.LoadPageCallback, LinkPrevi
             }
         }
         binding.listRecyclerView.adapter = RecyclerViewAdapter(pages)
+
+        if (pages.isEmpty() && Prefs.placesLastLocationAndZoomLevel == null) {
+            FeedbackUtil.makeSnackbar(requireActivity(), getString(R.string.places_empty_message_snackbar)).run {
+                setAction(R.string.dialog_close_description) {
+                    dismiss()
+                }
+                show()
```
dbrant, line 653: What do you think of setting `Prefs.placesLastLocationAndZoomLevel` here explicitly, so that the snackbar doesn't appear again during the current session?
Otherwise it leads to this kind of situation:

https://github.com/user-attachments/assets/ebc13fdf-c3d7-4f9c-a9ce-c2cecaf0236b

dbrant (changes requested):

app/src/main/java/org/wikipedia/places/PlacesFragment.kt
cooltey: Thanks! It makes sense to me and fixed.

PR #4790, opened 2024-07-09
Remove tooltip and survey from Places feature

cooltey:
There is no survey code included in the Places, and it has been removed in #4672 

Bug: T368990


dbrant:
Does this really need design review?

cooltey:
> Does this really need design review?

Got it. Will let designer know about it. It is ready for review and merge.

PR #4791, opened 2024-07-09
Integrate with new account vanishing form.

dbrant:
This integrates with the new and improved vanishing form on Meta Wiki.
Aside from a new URL, this also requires a couple of tweaks on the client side:

* The vanishing form calls a javascript `confirm()` popup, which we must explicitly tell the WebView to support, by giving it a WebChromeClient.
* Slightly updated Url-overriding logic to be able to handle subpages of the target page.

https://phabricator.wikimedia.org/T369627


cooltey:
Not sure if this is ready for review. When I created a new account and tried to vanish account, I clicked on "Request vanish" and clicked on confirm on the dialog, and it popped up to an external browser for the metawiki login page. Is this an expected behavior? and the account in the app was not logged out either.

cooltey:
> Not sure if this is ready for review. When I created a new account and tried to vanish account, I clicked on "Request vanish" and clicked on confirm on the dialog, and it popped up to an external browser for the metawiki login page. Is this an expected behavior? and the account in the app was not logged out either.

After I cleared app data and cache, I can see the success screen after submitting the request.

PR #4794, opened 2024-07-09
Suppress "Watch" in Places feature when user is logged out

cooltey:
Bug: T369078


dbrant:
Does this really need design review?

cooltey:
> Does this really need design review?
OK, I'll let designer know and now it is ready for merge.

PR #4796, opened 2024-07-10
Batch 2 of edit tags.

dbrant:
https://phabricator.wikimedia.org/T366731


app/src/main/java/org/wikipedia/edit/EditSectionActivity.kt
```diff
@@ -325,6 +328,20 @@ class EditSectionActivity : BaseActivity(), ThemeChooserDialog.Callback, EditPre
         BreadCrumbLogEvent.logInputField(this, editSummaryFragment.summaryText)
     }
 
+    private fun getEditTags(): List<String> {
+        return buildList {
```
cooltey, line 332: Curious about in what condition will multiple tags be output?

dbrant: That's fair -- I updated it to produce a single tag.

PR #4768, opened 2024-07-01
Migrating to Version Catalog

dinesh-thiyagarajan:
Implementing version catalog for centralised management of dependencies.

Please check this link for read more on this
https://developer.android.com/build/migrate-to-catalogs

PR #4800, opened 2024-07-11
Exclude InstallReferrer library from F-Droid build

cooltey:
Created a new package `installreferrer` for `extra` and `fdroid` package.

Bug: T369814


linsui:
Installerreferer is added back in https://github.com/wikimedia/apps-android-wikipedia/pull/4768 by mistake.

dbrant:
> Installerreferer is added back in #4768 by mistake.

Whoops, you are correct. Opened #4826

PR #4813, opened 2024-07-12
Refactoring ConfigureFragment to use Coroutine and ViewModel

cooltey:
This also adds a `ProgressBar` for loading status and also update to use `notifyItemRangeChanged()`


app/src/main/java/org/wikipedia/feed/configure/ConfigureFragment.kt
```diff
@@ -42,35 +49,6 @@ class ConfigureFragment : Fragment(), MenuProvider, ConfigureItemView.Callback {
 
         setupRecyclerView()
 
-        disposables.add(ServiceFactory.getRest(WikiSite("wikimedia.org")).feedAvailability
-            .subscribeOn(Schedulers.io())
-            .observeOn(AndroidSchedulers.mainThread())
-            .doAfterTerminate { prepareContentTypeList() }
```
dbrant, line 48: Notice that the old logic calls `prepareContentTypeList()` even in case of an error. The new logic only logs the error and doesn't do anything else (the progressBar remains spinning). Can we also call `prepareContentTypeList()` on error?

app/src/main/java/org/wikipedia/feed/configure/ConfigureViewModel.kt
```diff
@@ -0,0 +1,64 @@
...
+    val uiState = _uiState.asStateFlow()
+
+    init {
+        loadFeedAvailability()
+    }
+
+    private fun loadFeedAvailability() {
+        _uiState.value = Resource.Loading()
+        viewModelScope.launch(handler) {
+            val result = ServiceFactory.getRest(WikiSite("wikimedia.org")).feedAvailability()
```
dbrant, line 30: Let's no longer use a hardcoded `"wikimedia.org"` here. This API call can be made to any wikimedia domain, so we can just use `WikipediaApp.instance.wikiSite`.

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/configure/ConfigureFragment.kt
cooltey: Yes! I thought I added it 😅 Done!

app/src/main/java/org/wikipedia/feed/configure/ConfigureViewModel.kt
cooltey: Sure! Done.

PR #4814, opened 2024-07-12
Refactoring History to use Coroutine and ViewModel

cooltey:
This PR also optimizes the `adapter` and moves local `lifecycleScope` to `ViewModel`.


app/src/main/java/org/wikipedia/history/HistoryViewModel.kt
```diff
@@ -0,0 +1,67 @@
...
+import kotlinx.coroutines.CoroutineExceptionHandler
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import org.wikipedia.database.AppDatabase
+import org.wikipedia.util.Resource
+
+class HistoryViewModel : ViewModel() {
```
machado001, line 14:
```suggestion
class HistoryViewModel(
    private val historyRepository: HistoryRepository = HistoryRepositoryImpl()
) : ViewModel() {

```

app/src/main/java/org/wikipedia/history/HistoryViewModel.kt
```diff
@@ -0,0 +1,67 @@
+package org.wikipedia.history
+
```
machado001, line 2: Imho the ViewModel knows too much about the database implementation. We can move these Room-related methods to a new HistoryRepository class and inject it into the ViewModel constructor.

```suggestion

//new file
          interface HistoryRepository {
    suspend fun filterHistoryItems(searchQuery: String) : List<Any>
    suspend fun deleteHistory(entry: HistoryEntry)
    suspend fun deleteAllHistory()
    suspend fun insertHistory(entries: List<HistoryEntry>)
}
```

```suggestion
//new file
class HistoryRepositoryImpl(
    private val historyDao: HistoryEntryDao = AppDatabase.instance.historyEntryDao(),
    private val historyWithImageDao: HistoryEntryWithImageDao = AppDatabase.instance.historyEntryWithImageDao(),
    private val dispatcher: CoroutineDispatcher = Dispatchers.IO
): HistoryRepository {

    override suspend fun filterHistoryItems(searchQuery: String) = withContext(dispatcher) {
        historyWithImageDao.filterHistoryItems(searchQuery)
    }
    override suspend fun deleteHistory(entry: HistoryEntry) = historyDao.delete(entry)
    override suspend fun deleteAllHistory() = historyDao.deleteAll()
    override suspend fun insertHistory(entries: List<HistoryEntry>) = historyDao.insert(entries)
}
```
By doing this, the ViewModel becomes more testable, and we have a centralized place to handle these history-database operations. This improves the unit testability of the ViewModel. Since DAOs are interfaces, we can also create fakes to unit test the repository.

app/src/main/java/org/wikipedia/history/HistoryViewModel.kt
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    fun reloadHistoryItems() {
+        viewModelScope.launch(handler) {
+            loadHistoryItems()
+        }
+    }
+
+    private suspend fun loadHistoryItems() {
+        withContext(Dispatchers.IO) {
```
machado001, line 39: I would move this Dispatcher.IO to the repository and wrap this code with viewModelScope.launch(handler). Then we could remove it from the reloadHistoryItems (even more, we could remove the realodHistoryItems fun), and remove suspend to maintain consistence with the another ViewModel methods:

```suggestion
        private fun loadHistoryItems() {
        viewModelScope.launch(handler) {
            val items = historyRepository.filterHistoryItems(searchQuery.orEmpty())
            _uiState.value = Resource.Success(items)
        }
    }
```

app/src/main/java/org/wikipedia/history/HistoryViewModel.kt
```diff
@@ -0,0 +1,67 @@
...
+            val items = AppDatabase.instance.historyEntryWithImageDao().filterHistoryItems(searchQuery.orEmpty())
```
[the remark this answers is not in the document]
dbrant: Thanks - The app contains a lot of legacy code that we are working towards modernizing very gradually and systematically.

Before making any architectural choices in a single specific component, we need to create a top-down plan that will apply to all components (viewmodels, repositories, views, etc), and we're working on such a plan now.  Then we can begin future phases of refactoring. Our current refactoring work deals with creating ViewModels where they hadn't yet existed (!), and removing any remaining usages of RxJava.

Your proposed changes are totally valid, but they are an end-state of additional architectural choices that have not yet been made.

app/src/main/java/org/wikipedia/history/HistoryFragment.kt
```diff
@@ -86,6 +77,29 @@ class HistoryFragment : Fragment(), BackPressedHandler {
...
+                        when (it) {
+                            is Resource.Success -> onLoadItemsFinished(it.data)
+                            is Resource.Error -> onError(it.throwable)
+                        }
+                    }
+                }
+                launch {
+                    viewModel.actionUiState.collect {
+                        when (it) {
+                            is Resource.Success -> onPagesDeleted()
```
dbrant, line 95: This will cause `onPagesDeleted()` to be called again on every screen rotation, even after an item was deleted. (it will keep showing a snackbar that says "0 items removed")

dbrant (changes requested):

app/src/main/java/org/wikipedia/history/HistoryFragment.kt
dbrant: (hint: `SingleLiveData`)

cooltey: Done!

PR #4815, opened 2024-07-13
Remove FLAG_ACTIVITY_CLEAR_TOP when going back to Main tab on Explore feed

cooltey:
This will help to reduce the black color blinks (from the splash theme?) when clicking on the Explore option from PageActivity.


cooltey:
Just checked this ticket as the reference: https://github.com/wikimedia/apps-android-wikipedia/pull/1126, and it seems it does not affect the current app behavior.

and I compared this PR with the current production with the following conditions, and all are working as expected:
a. Open an article from the feed -> click on "Explore"
b. Open an article from the feed -> go to device Home -> open the app from the app shortcut ("Search or Continue reading") -> click on any article -> click on either "Back", "System back button" or "Explore",

dbrant (changes requested):
This leads to the following scenario:
- launch the app on the Main screen.
- tap the Featured Article card to load that article.
- tap the overflow menu and select Explore, to go back to the Explore feed.
- now press Back, and observe that _another_ Main screen is hiding underneath it.

cooltey:
> This leads to the following scenario:
> 
>     * launch the app on the Main screen.
> 
>     * tap the Featured Article card to load that article.
> 
>     * tap the overflow menu and select Explore, to go back to the Explore feed.
> 
>     * now press Back, and observe that _another_ Main screen is hiding underneath it.

Thank you, you are correct. Was trying to remove the splash screen when calling the `goToMain()` function but it did not work well.

Closed.

PR #4816, opened 2024-07-15
Support URL sharing on recent screen

ebraminio:
Bug: [T370020](https://phabricator.wikimedia.org/T370020)


dbrant:
Good find, thanks!

ebraminio:
Thanks for the quick merge, it motivated me to do the same for the images also, https://github.com/wikimedia/apps-android-wikipedia/pull/4818

PR #4818, opened 2024-07-15
Support URL sharing on recent screen for images

ebraminio:
Similar to https://github.com/wikimedia/apps-android-wikipedia/pull/4816 but for pictures.

Unlike the name, .commonsUrl, it works for both images uploaded on Wikimedia Commons and locally uploaded images.

To test this open an article, open an image from it, hit the rectangle button above the emulator, click on the anchor icon on the recents screen.

This is the reason it's only enabled for API level 23
image


dbrant:
Sure, why not 👍

PR #4824, opened 2024-07-18
Add Explore Feed as an Entry point for Places

cooltey:
Bug: T370100


app/src/main/java/org/wikipedia/history/HistoryEntry.kt
```diff
@@ -81,6 +81,7 @@ class HistoryEntry(
         const val SOURCE_APP_SHORTCUT_RANDOM = 20
         const val SOURCE_APP_SHORTCUT_CONTINUE_READING = 21
         const val SOURCE_FEED_MOST_READ_ACTIVITY = 22
+        const val SOURCE_FEED_PLACES = 42
```
dbrant, line 84: Can we move this to the bottom of the list, so that they are in numerical order?

app/src/main/java/org/wikipedia/util/GeoUtil.kt
```diff
@@ -51,4 +51,10 @@ object GeoUtil {
             "${formatter.format(distance)} km"
         }
     }
+
+    fun isSamePlace(startLat: Double, endLat: Double, startLon: Double, endLon: Double): Boolean {
+        val df = DecimalFormat("#.#######")
```
dbrant, line 56: Instead of comparing string versions of `double` variables, can we just compare the `double`s? (within a tolerance of 0.0000001)?

dbrant (changes requested):
very minor stuff

app/src/main/java/org/wikipedia/history/HistoryEntry.kt
cooltey: Yes.

app/src/main/java/org/wikipedia/util/GeoUtil.kt
cooltey: Done!

PR #4830, opened 2024-07-19
Remove feature flag for machine-suggested descriptions.

dbrant:
This opens machine-generated description suggestions back up for all users, instead of being feature-flagged for pre-beta only.
It also makes sure that the correct tag is added when a description is added through a machine suggested selection.

https://phabricator.wikimedia.org/T360581

### DIFF
--- a/app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
+++ b/app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
@@ -53,7 +53,6 @@ import org.wikipedia.suggestededits.SuggestedEditsSurvey
 import org.wikipedia.suggestededits.SuggestionsActivity
 import org.wikipedia.util.DeviceUtil
 import org.wikipedia.util.FeedbackUtil
-import org.wikipedia.util.ReleaseUtil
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -487,6 +486,7 @@ class DescriptionEditFragment : Fragment() {
             when (action) {
                 DescriptionEditActivity.Action.ADD_DESCRIPTION -> {
                     if (binding.fragmentDescriptionEditView.wasSuggestionChosen) {
+                        tags.add(EditTags.APP_DESCRIPTION_ADD)
```
cooltey, line 489: Does a machine-generated description need its own edit tag? It used to have `#machine-suggestion` in the edit summary. 
https://phabricator.wikimedia.org/T366730#9924336

dbrant: I remember hearing in one of the data syncs that these hashtags do not need to be converted to tags. We can raise it again at some point, but that needn't block this PR.

PR #4831, opened 2024-07-19
Edit tags, final batch.

dbrant:
https://phabricator.wikimedia.org/T366732


cooltey:
I see this error message when trying to submit an image tag.


The other issue was: the "Publishing" overlaps the error message so there's no way to cancel the process so I have to press the system back button.

dbrant:
> I see this error message when trying to submit an image tag.

Good catch, fixed.

> The other issue was: the "Publishing" overlaps the error message so there's no way to cancel the process so I have to press the system back button.

Also fixed -- this was because we were using the same `CoroutineExceptionHandler` for two different state flows.

PR #4849, opened 2024-08-02
Fix proper injection and hiding of context menu over WebView.

dbrant:
This fixes a long-standing issue where users have observed that the "Translate" menu option (provided by Google Translate) in the context menu doesn't seem to work, or doesn't appear at all.
This seems to have been because we were taking control of the context menu by clearing all the items from it, then repopulating the items including our own custom items. It looks like this method erases some necessary metadata from the original menu that was needed for the Translate option (and possibly others) to work properly.

The new method will preserve the original context menu, and merely append our custom items to it.

https://phabricator.wikimedia.org/T318041
https://phabricator.wikimedia.org/T346999

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageActivity.kt
+++ b/app/src/main/java/org/wikipedia/page/PageActivity.kt
@@ -658,17 +658,16 @@ class PageActivity : BaseActivity(), PageFragment.Callback, LinkPreviewDialog.Lo
 
     private fun modifyMenu(mode: ActionMode) {
         val menu = mode.menu
-        val menuItemsList = menu.children.filter {
-            val title = it.title.toString()
-            !title.contains(getString(R.string.search_hint)) &&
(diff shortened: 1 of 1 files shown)


cooltey:
The "Map" icon does not open the map app, not sure if this is related. 


Maybe not related: I noticed the "search Wikipedia" option is missing from the context menu (for this PR and also the current production), not sure if there's a way to show it or add it back?

dbrant:
> The "Map" icon does not open the map app, not sure if this is related.

@cooltey It works for me, and if it's not working for you, it's highly unlikely to be related.

> Maybe not related: I noticed the "search Wikipedia" option is missing from the context menu (for this PR and also the current production), not sure if there's a way to show it or add it back?

The "Search Wikipedia" option is removed on purpose, because it's intended to be used from other apps. (highlighting text in Gmail for example, and searching Wikipedia from there.)

PR #4873, opened 2024-08-16
Present, not past

Justman100:


dbrant:
In what sense is the current string "past"?

Justman100:
> In what sense is the current string "past"?

> would

dbrant:
Sorry, the current verbiage is perfectly fine.

Justman100:
> Sorry, the current verbiage is perfectly fine.

How many more times? DON'T JUST CLOSE PRS! If it was ok like this, this PR wouldn't exist

PR #4882, opened 2024-08-20
Fix blocked message in Suggested Edits, and use coroutines.

dbrant:
There is an issue in the Suggested Edits screen, if the current account is blocked, the block message overlaps the other components on the screen, because the other components are not properly hidden. This is most prominent in Black mode; in other modes it looks more correct, but only by accident.

image

This fixes the showing/hiding of the other components in case of a block, and while fixing the issue I converted some of the block message retrieval logic to coroutines, which will simplify the work of several upcoming tasks.

### DIFF
--- a/app/src/main/java/org/wikipedia/edit/EditSectionActivity.kt
+++ b/app/src/main/java/org/wikipedia/edit/EditSectionActivity.kt
@@ -421,7 +421,7 @@ class EditSectionActivity : BaseActivity(), ThemeChooserDialog.Callback, EditPre
     private fun handleEditingException(caught: MwException) {
         val code = caught.title
         lifecycleScope.launch(CoroutineExceptionHandler { _, t ->
-            showError(caught)
+            showError(t)
         }) {
             // In the case of certain AbuseFilter responses, they are sent as a code, instead of a
             // fully parsed response. We need to make one more API call to get the parsed message:
--- a/app/src/main/java/org/wikipedia/edit/EditSectionActivity.kt
+++ b/app/src/main/java/org/wikipedia/edit/EditSectionActivity.kt
@@ -21,10 +21,13 @@ import androidx.core.os.postDelayed
 import androidx.core.view.isGone
 import androidx.core.view.isVisible
 import androidx.core.widget.doAfterTextChanged
+import androidx.lifecycle.lifecycleScope
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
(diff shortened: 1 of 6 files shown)


app/src/test/java/org/wikipedia/login/UserExtendedInfoClientTest.kt
```diff
@@ -9,27 +10,10 @@ class UserExtendedInfoClientTest : MockRetrofitTest() {
...
-            .assertValue {
-                it.query?.userInfo?.id == id &&
-                        it.query?.getUserResponse("USER")?.name == "USER"
-            }
-    }
-
-    @Test
-    @Throws(Throwable::class)
-    fun testRequestResponse404() {
-        enqueue404()
```
dbrant, line 23: I removed the rest of these tests because I no longer understand what they're testing. Are they testing if Retrofit works correctly? That's already done in Retrofit's own test suite. Are they testing if the Java language works correctly? I'm not sure that's within the scope of what we should worry about.

app/src/main/java/org/wikipedia/edit/EditSectionActivity.kt
```diff
@@ -418,23 +420,24 @@ class EditSectionActivity : BaseActivity(), ThemeChooserDialog.Callback, EditPre
...
-        // fully parsed response. We need to make one more API call to get the parsed message:
-        if (code.startsWith("abusefilter-") && caught.message.contains("abusefilter-") && caught.message.length < 100) {
-            disposables.add(ServiceFactory.get(pageTitle.wikiSite).parsePage("MediaWiki:" + StringUtil.sanitizeAbuseFilterCode(caught.message))
-                    .subscribeOn(Schedulers.io())
-                    .observeOn(AndroidSchedulers.mainThread())
-                    .subscribe({ response: MwParseResponse -> showError(MwException(MwServiceError(code, response.text))) }) { showError(it) })
-        } else if ("editconflict" == code) {
-            MaterialAlertDialogBuilder(this@EditSectionActivity)
+        lifecycleScope.launch(CoroutineExceptionHandler { _, t ->
+            showError(caught)
```
cooltey, line 424: Do we need to show `t` from `CoroutineExceptionHandler` or the `caught` from the method parameter?

dbrant: t! done!

PR #4885, opened 2024-08-21
Send fundraising banner opt-in state in app_donor_experience stream.

dbrant:
* Read the user's preferences once daily, as part of the recurring `RemoteConfigRefreshTask`.
* Send an analytics field based on the user's preference for showing fundraising banners.

https://phabricator.wikimedia.org/T371388

### DIFF
--- a/app/src/main/java/org/wikipedia/settings/RemoteConfigRefreshTask.kt
+++ b/app/src/main/java/org/wikipedia/settings/RemoteConfigRefreshTask.kt
@@ -36,8 +36,10 @@ class RemoteConfigRefreshTask : RecurringTask() {
             }
 
             val userInfo = ServiceFactory.get(WikipediaApp.instance.wikiSite).getUserInfo()
-            // This clumsy comparison is necessary because the field is an integer value when enabled, but a string when disabled.
-            Prefs.donationBannerOptIn = userInfo.query?.userInfo?.options?.fundraisingOptIn?.toString().orEmpty().replace("\"", "").isNotEmpty()
+            // This clumsy comparison is necessary because the field is an integer value when enabled, but an empty string when disabled.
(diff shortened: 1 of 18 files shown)


app/src/main/java/org/wikipedia/settings/RemoteConfigRefreshTask.kt
```diff
@@ -32,6 +35,10 @@ class RemoteConfigRefreshTask : RecurringTask() {
             } finally {
                 response?.closeQuietly()
             }
+
+            val userInfo = ServiceFactory.get(WikipediaApp.instance.wikiSite).getUserInfo()
+            // This clumsy comparison is necessary because the field is an integer value when enabled, but a string when disabled.
+            Prefs.donationBannerOptIn = userInfo.query?.userInfo?.options?.fundraisingOptIn?.jsonPrimitive.toString() == "1"
```
cooltey, line 41: Just curious, if one of the values is `null`, will the optIn preference become `false` because `null` or `""` is not equal to `1` by default?

dbrant: Yeah, fair point. We do want the default for this preference to lean towards `true`, so I modified this check to look for an explicit empty string `""`, which would definitely indicate `false`.

cooltey: Thanks. I checked the latest commit you made, if the `userInfo.query?.userInfo?.options?.fundraisingOptIn?` output `null`, and then `toString().orEmpty().replace("\"", "")` will convert it to an empty string. Is this something expected?

dbrant: It is not! And... fixed.

PR #4898, opened 2024-08-27
Send users directly to web when we do not support Native GPay 

cooltey:
https://phabricator.wikimedia.org/T373410

The idea is to use the current logic in the `GooglePayViewModel` that decides to show in the Donation dialog or CustomTab.


app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -465,7 +476,26 @@ class MainFragment : Fragment(), BackPressedHandler, MenuProvider, FeedFragment.
...
-        (requireActivity() as? BaseActivity)?.launchDonateDialog()
+        if (showDonationBottomSheet) {
+            (requireActivity() as? BaseActivity)?.launchDonateDialog()
+        } else {
+            DonateDialog.launchDonateLink(requireContext())
+        }
+    }
+
+    private fun checkDonateConfig() {
+        CoroutineScope(Dispatchers.IO).launch {
```
dbrant, line 487: This seems like a lot of highly specific business logic being placed into `MainFragment`. Can we put this into a ViewModel, or into `DonationConfigHelper`?

app/src/extra/java/org/wikipedia/donate/GooglePayViewModel.kt
```diff
@@ -85,24 +84,7 @@ class GooglePayViewModel : ViewModel() {
...
-                }
-                paymentMethodsCall.await().response?.let { response ->
-                    Prefs.paymentMethodsLastQueryTime = now
-                    response.paymentMethods.find { it.type == GooglePayComponent.PAYMENT_METHOD_NAME }?.let {
-                        Prefs.paymentMethodsMerchantId = it.configuration?.merchantId.orEmpty()
-                        Prefs.paymentMethodsGatewayId = it.configuration?.gatewayMerchantId.orEmpty()
-                    }
-                }
-            }
+            async { updatePaymentMethodsPreferences() }.await()
```
dbrant, line 87: This doesn't need to be inside `async{}.await()`

dbrant (changes requested):
Couldn't we just put all this logic in `DonateViewModel`? It's already doing most of the work for us.  (i.e. in addition to checking `isGooglePayAvailable()`, you can also check if the country code is allowed.)

app/src/extra/java/org/wikipedia/donate/GooglePayViewModel.kt
cooltey: Sure!

app/src/main/java/org/wikipedia/main/MainFragment.kt
cooltey: You are right. Moved to `DonateViewModel`.

cooltey:
> Couldn't we just put all this logic in `DonateViewModel`? It's already doing most of the work for us. (i.e. in addition to checking `isGooglePayAvailable()`, you can also check if the country code is allowed.)

I moved the logic to `DonateViewModel`, it works great but the user will see the progress bar in the bottom sheet and then automatically redirect to the donation page. 🤔

cooltey:
Thanks @dbrant

PR #4910, opened 2024-08-30
Analytics for recommended content.

dbrant:
https://phabricator.wikimedia.org/T372761

### DIFF
--- a/app/src/main/java/org/wikipedia/recommendedcontent/RecommendedContentFragment.kt
+++ b/app/src/main/java/org/wikipedia/recommendedcontent/RecommendedContentFragment.kt
@@ -17,6 +17,7 @@ import kotlinx.coroutines.launch
 import org.wikipedia.Constants
 import org.wikipedia.R
 import org.wikipedia.analytics.ABTest
+import org.wikipedia.analytics.metricsplatform.ExperimentalLinkPreviewInteraction
 import org.wikipedia.analytics.metricsplatform.RecommendedContentAnalyticsHelper
 import org.wikipedia.databinding.FragmentRecommendedContentBinding
 import org.wikipedia.databinding.ItemRecommendedContentSearchHistoryBinding
@@ -38,6 +39,8 @@ class RecommendedContentFragment : Fragment() {
     private val binding get() = _binding!!
     private val viewModel: RecommendedContentViewModel by viewModels { RecommendedContentViewModel.Factory(requireArguments()) }
 
+    private val parentSearchFragment get() = requireParentFragment().requireParentFragment() as SearchFragment
+
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         super.onCreateView(inflater, container, savedInstanceState)
 
@@ -57,7 +60,7 @@ class RecommendedContentFragment : Fragment() {
                                 buildHistoryList(it.data)
                             }
                             is Resource.Error -> {
-                                (requireParentFragment().requireParentFragment() as SearchFragment).onSearchProgressBar(false)
+                                parentSearchFragment.onSearchProgressBar(false)
                                 parentFragmentManager.popBackStack()
                                 L.d(it.throwable)
                             }
@@ -68,13 +71,13 @@ class RecommendedContentFragment : Fragment() {
                     viewModel.recommendedContentState.collect {
                         when (it) {
                             is Resource.Loading -> {
-                                (requireParentFragment().requireParentFragment() as SearchFragment).onSearchProgressBar(true)
+                                parentSearchFragment.onSearchProgressBar(true)
                             }
                             is Resource.Success -> {
                                 buildRecommendedContent(it.data)
                             }
                             is Resource.Error -> {
-                                (requireParentFragment().requireParentFragment() as SearchFragment).onSearchProgressBar(false)
+                                parentSearchFragment.onSearchProgressBar(false)
                                 parentFragmentManager.popBackStack()
                                 L.d(it.throwable)
                             }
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/wikipedia/views/SurveyDialog.kt
```diff
@@ -55,7 +62,15 @@ object SurveyDialog {
         if (source == Constants.InvokeSource.RECOMMENDED_CONTENT) {
             dialogBuilder.setPositiveButton(R.string.patroller_diff_feedback_dialog_submit) { _, _ ->
                 val feedbackInput = binding.feedbackInput.text.toString()
-                // TODO: send event
+
+                ExperimentalLinkPreviewInteraction(source = HistoryEntry.SOURCE_SEARCH,
```
cooltey, line 66: Do we need to send `action_source =  43 OR 44` for personalized or generalized sources instead of a default source `1`? Same for other places which should we set `43` or `44` if the user sees the recommended content?

app/src/main/java/org/wikipedia/views/SurveyDialog.kt
```diff
@@ -81,13 +96,13 @@ object SurveyDialog {
...
         FeedbackUtil.showMessage(activity, messageId)
-        sendAnalyticsEvent("feedback_submit_toast", "feedback_form", source)
         when (source) {
             Constants.InvokeSource.SUGGESTED_EDITS_RECENT_EDITS -> {
+                PatrollerExperienceEvent.logAction("feedback_submit_toast", "feedback_form", "")
                 activity.window.decorView.postDelayed({
                     val anchorView = activity.findViewById<View>(R.id.more_options)
                     if (!activity.isDestroyed && anchorView != null && Prefs.showOneTimeRecentEditsFeedbackForm) {
-                        sendAnalyticsEvent("tooltip_impression", "feedback_form", source)
+                        PatrollerExperienceEvent.logAction("tooltip_impression", "feedback_form", "")
```
cooltey, line 105: We can ignore `""` in the last parameter.

app/src/main/java/org/wikipedia/views/SurveyDialog.kt
```diff
@@ -81,13 +96,13 @@ object SurveyDialog {
 
     private fun showFeedbackSnackbarAndTooltip(activity: Activity, messageId: Int, source: Constants.InvokeSource) {
         FeedbackUtil.showMessage(activity, messageId)
-        sendAnalyticsEvent("feedback_submit_toast", "feedback_form", source)
         when (source) {
             Constants.InvokeSource.SUGGESTED_EDITS_RECENT_EDITS -> {
+                PatrollerExperienceEvent.logAction("feedback_submit_toast", "feedback_form", "")
```
cooltey, line 101: We can ignore `""` in the last parameter.

app/src/main/java/org/wikipedia/recommendedcontent/RecommendedContentSectionView.kt
```diff
@@ -65,6 +68,10 @@ class RecommendedContentSectionView(context: Context, attrs: AttributeSet? = nul
             }
             val entry = HistoryEntry(pageSummary.getPageTitle(WikipediaApp.instance.wikiSite), source)
             context.startActivity(PageActivity.newIntentForNewTab(context, entry, entry.title))
+            analyticsEvent?.let {
+                it.source
```
cooltey, line 72: Should we set the source?

app/src/main/java/org/wikipedia/recommendedcontent/RecommendedContentFragment.kt
```diff
@@ -119,12 +122,16 @@ class RecommendedContentFragment : Fragment() {
     }
 
     private fun buildRecommendedContent(list: List<PageSummary>) {
-        (requireParentFragment().requireParentFragment() as SearchFragment).onSearchProgressBar(false)
+        parentSearchFragment.onSearchProgressBar(false)
         if (list.isEmpty()) {
+            parentSearchFragment.analyticsEvent = ExperimentalLinkPreviewInteraction(HistoryEntry.SOURCE_SEARCH, RecommendedContentAnalyticsHelper.abcTest.getGroupName(), false)
```
cooltey, line 127: Not sure but do we need to send group-specific source here?

app/src/main/java/org/wikipedia/recommendedcontent/RecommendedContentFragment.kt
```diff
@@ -119,12 +122,16 @@ class RecommendedContentFragment : Fragment() {
...
+                .also { it.logImpression() }
             parentFragmentManager.popBackStack()
             return
         }
-        binding.recommendedContent.buildContent(list)
+        parentSearchFragment.analyticsEvent = ExperimentalLinkPreviewInteraction(HistoryEntry.SOURCE_SEARCH, RecommendedContentAnalyticsHelper.abcTest.getGroupName(), true)
```
cooltey, line 132: Not sure but do we need to send group-specific source here?

cooltey (changes requested):
Thanks for working on this!

app/src/main/java/org/wikipedia/views/SurveyDialog.kt
dbrant: OK!

app/src/main/java/org/wikipedia/recommendedcontent/RecommendedContentSectionView.kt
dbrant: Whoops!

app/src/main/java/org/wikipedia/recommendedcontent/RecommendedContentFragment.kt
dbrant: I don't believe so. I think for the `impression` event the source is not relevant.

app/src/main/java/org/wikipedia/recommendedcontent/RecommendedContentFragment.kt
dbrant: I don't believe so. I think for the `impression` event the source is not relevant.

PR #4913, opened 2024-09-02
Refactoring CreateAccountActivity to use coroutine

terofeev:
### What does this do?

- Refactored `CreateAccountActivity` to use Coroutines.
- Added `CreateAccountActivityViewModel` and moved the account creation logic into it.
- Updated test cases.

### Why is this needed?
Tech-debt, migration to use Coroutines instead of RxJava

**Phabricator:**


app/src/main/java/org/wikipedia/createaccount/CreateAccountActivity.kt
```diff
@@ -126,49 +126,53 @@ class CreateAccountActivity : BaseActivity() {
...
-                            captchaHandler.handleCaptcha(token, CaptchaResult(captchaId))
-                        } else {
-                            doCreateAccount(token)
-                        }
-                    }) { caught ->
-                        showError(caught)
-                        L.e(caught)
-                    })
+            lifecycleScope.launch {
+                try {
```
cooltey, line 130: Would it be possible to use `CoroutineExceptionHandler` instead of a `try catch`?

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivity.kt
```diff
@@ -126,49 +126,53 @@ class CreateAccountActivity : BaseActivity() {
...
-                        createAccountEvent.logError(StringUtil.removeStyleTags(response.message))
-                        throw CreateAccountException(StringUtil.removeStyleTags(response.message))
-                    }
-                }) { caught ->
-                    L.e(caught.toString())
-                    createAccountEvent.logError(caught.toString())
-                    showProgressBar(false)
-                    showError(caught)
-                })
+        lifecycleScope.launch {
```
cooltey, line 155: Would it be possible to use `CoroutineExceptionHandler` instead of a `try catch`?

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivity.kt
```diff
@@ -289,19 +292,23 @@ class CreateAccountActivity : BaseActivity() {
...
-                    .observeOn(AndroidSchedulers.mainThread())
-                    .subscribe({ response ->
-                        response.query?.getUserResponse(userName)?.let {
-                            binding.createAccountUsername.isErrorEnabled = false
-                            if (it.hasBlockError) {
-                                handleAccountCreationError(it.error)
-                            } else if (!it.canCreate) {
-                                binding.createAccountUsername.error = getString(R.string.create_account_name_unavailable, userName)
-                            }
+            lifecycleScope.launch {
```
cooltey, line 295: Would it be possible to use `CoroutineExceptionHandle`r instead of a `try catch`?

cooltey (changes requested):
Thanks for making this PR, some comments are inline and I would also suggest using `ViewModel` and putting some of the logic into a viewModel class.

terofeev:
Hello Cooltey!
Thank you for reviewing my PR. I have applied the review comments and added a ViewModel to the Activity.

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivity.kt
terofeev: Applied in cde801e500c1b1ce72193e19273fc749cc299b62

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivity.kt
terofeev: Applied in cde801e500c1b1ce72193e19273fc749cc299b62

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivity.kt
terofeev: Applied in cde801e500c1b1ce72193e19273fc749cc299b62

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivity.kt
```diff
@@ -289,19 +295,21 @@ class CreateAccountActivity : BaseActivity() {
...
+                }
+                response.query?.getUserResponse(userName)?.let {
+                    binding.createAccountUsername.isErrorEnabled = false
+                    if (it.hasBlockError) {
+                        handleAccountCreationError(it.error)
+                    } else if (!it.canCreate) {
+                        binding.createAccountUsername.error = getString(R.string.create_account_name_unavailable, userName)
+                    }
+                }
+            }
```
cooltey, line 312: Would it be possible to move this to the `ViewModel` class?

cooltey (changes requested):
Thank you for creating a `ViewModel` that contains the network calls. Just have one minor comment relate to it.

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivity.kt
terofeev: Sure! I’ve moved the logic to the ViewModel and also removed the Runnable in the latest commit.

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivity.kt
```diff
@@ -42,10 +42,10 @@ class CreateAccountActivity : BaseActivity() {
     private lateinit var binding: ActivityCreateAccountBinding
     private lateinit var captchaHandler: CaptchaHandler
     private lateinit var createAccountEvent: CreateAccountEvent
-    private val disposables = CompositeDisposable()
     private var wiki = WikipediaApp.instance.wikiSite
     private var userNameTextWatcher: TextWatcher? = null
-    private val userNameVerifyRunnable = UserNameVerifyRunnable()
+    private var verifyUserNameJob: Job? = null
```
cooltey, line 47: Can we move the `Job` to the ViewModel class and cancel the job before verifying the username?

terofeev: Yes, I've moved the job to the ViewModel, and now the `doOnTextChanged` method only calls `viewModel.verifyUserName(text)`

cooltey (approved):
Looks good to me now, thanks! Will let @dbrant do the review and merge too.

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivityViewModel.kt
```diff
@@ -0,0 +1,111 @@
...
+
+    private var verifyUserNameJob: Job? = null
+
+    fun createAccountInfo() {
+        viewModelScope.launch(CoroutineExceptionHandler { _, throwable ->
+            _createAccountInfoState.value = AccountInfoState.Error(throwable)
+        }) {
+            val response = withContext(Dispatchers.IO) {
+                ServiceFactory.get(WikipediaApp.instance.wikiSite).getAuthManagerInfo()
+            }
```
cooltey, line 39: We can ignore the `withContext()` here. See #4970

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivityViewModel.kt
```diff
@@ -0,0 +1,111 @@
...
+                _createAccountInfoState.value = AccountInfoState.DoCreateAccount(token)
+            }
+        }
+    }
+
+    fun doCreateAccount(token: String, captchaId: String, captchaWord: String, userName: String, password: String, repeat: String, email: String?) {
+        viewModelScope.launch(CoroutineExceptionHandler { _, throwable ->
+            _doCreateAccountState.value = CreateAccountState.Error(throwable)
+        }) {
+            val response = withContext(Dispatchers.IO) {
```
cooltey, line 56: We can ignore the `withContext()` here, see #4970

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivityViewModel.kt
```diff
@@ -0,0 +1,111 @@
...
+        verifyUserNameJob = viewModelScope.launch(CoroutineExceptionHandler { _, throwable ->
+            L.e(throwable)
+        }) {
+            _verifyUserNameState.emit(UserNameState.Initial)
+            if (text.isNullOrEmpty()) {
+                return@launch
+            }
+            delay(1000)
+            val userName = text.toString()
+            val response = withContext(Dispatchers.IO) {
```
cooltey, line 79: We can ignore the `withContext()` here, see #4970

app/src/main/java/org/wikipedia/createaccount/CreateAccountActivityViewModel.kt
terofeev: Got it, applied the change in c5385617f794c183ac27993cc2ac008def662515

PR #4916, opened 2024-09-03
Use style instead of deprecated method for resizing dialog window.

dbrant:
Technically this still uses `adjustResize` for the window, but this doesn't seem to be deprecated, and is simpler, and applies to all dialogs automatically.

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -537,6 +537,7 @@
         <item name="elevationOverlayEnabled">false</item>
         <item name="materialAlertDialogTitleTextStyle">@style/H2.DialogTitle</item>
         <item name="materialAlertDialogBodyTextStyle">@style/P.DialogBody</item>
+        <item name="android:windowSoftInputMode">adjustResize</item>
     </style>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -537,6 +537,7 @@
         <item name="elevationOverlayEnabled">false</item>
         <item name="materialAlertDialogTitleTextStyle">@style/H2.DialogTitle</item>
         <item name="materialAlertDialogBodyTextStyle">@style/P.DialogBody</item>
+        <item name="android:windowSoftInputMode">adjustResize</item>
```
cooltey, line 540: It looks like the "How can we improve the feature" in the Edit Patrol is not handled properly after applying it.

Would it be possible to define it in an individual style, for example: `AlertDialogTheme.Resize` or another better name so that we can use it only when we need it?

dbrant: Hmm ok then, let's create a child style, but also promise to revisit and consolidate these dialogs to use a single style in the future.

PR #4922, opened 2024-09-04
Preparations for removing recommendation-api service.

dbrant:
### What does this do?

This is a light refactor of some specific functions and classes to facilitate the deprecation of the `recommendation-api` service. No changes to existing behavior are expected.

### Why is this needed?

This will make it significantly easier to code-review the actual changes that will be made mostly in `EditingSuggestionsProvider`.

**Phabricator:**
https://phabricator.wikimedia.org/T373611


app/src/main/java/org/wikipedia/dataclient/wikidata/Entities.kt
```diff
@@ -23,10 +23,10 @@ class Entities : MwResponse() {
     @Serializable
     class Entity {
         val id: String = ""
-        val labels: Map<String, Label> = emptyMap()
-        val descriptions: Map<String, Label> = emptyMap()
-        val sitelinks: Map<String, SiteLink> = emptyMap()
-        val statements: JsonElement? = null
+        private val labels: JsonElement? = null
```
dbrant, line 26: Depending on the context, these fields that come from Wikidata (Wikibase) can actually manifest as old-style collections, where it's an empty array `[]` if there are no items, but a proper object `{...}` if it's nonempty. Therefore these need to be serialized dynamically when accessing. (or we could possibly create a custom deserializer in the future)

If we want to be 100% compatible, this is unfortunately necessary.

cooltey (changes requested):
Found an inconsistent behavior for this PR vs production:
1. Add English as the primary language, Chinese (Taiwan) as the second language, and the Test as the third language
2. Search Barack Obama in English and open it
3. In the article, you'll see "Add image caption (中文(台灣))"
4. Open the lead image, it should have "Add image caption (中文(台灣))"  instead of a regular one.
5. And in the "Add image caption" screen, it shows the "Image caption (Test)", which may be another inconsistent issue.

dbrant:
> Found an inconsistent behavior for this PR vs production:
> 5. And in the "Add image caption" screen, it shows the "Image caption (Test)", which may be another inconsistent issue.

hmm, I can't reproduce this behavior, and don't really see how it could be caused by the changes in this PR.

PR #4927, opened 2024-09-05
Optimizing the language variant logic for the feed and apply it for On This Day

cooltey:
### What does this do?
- Optimizing the way of handling language variants on the feed card, which applies `async` to improve the loading time.
- Adding support for the On This Day card and its fragment.

### Why is this needed?
- We are calling the `L10nUtil.getPagesForLanguageVariant` in a for loop in the `AggregatedFeedContentClient`, it would be better to bundle them with `async` to improve the loading time.
- On This Day card and fragment show the content without respecting the language variant.


app/src/main/java/org/wikipedia/feed/onthisday/OnThisDay.kt
```diff
@@ -17,8 +17,8 @@ class OnThisDay {
 
     @Serializable
     class Event {
```
dbrant, line 19: Can we keep this class immutable? (move vals into constructor, and re-construct if changes are needed)

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayViewModel.kt
```diff
@@ -35,14 +39,20 @@ class OnThisDayViewModel(bundle: Bundle) : ViewModel() {
         _uiState.value = Resource.Loading()
         viewModelScope.launch(handler) {
             val response = ServiceFactory.getRest(wikiSite).getOnThisDay(calendar[Calendar.MONTH] + 1, calendar[Calendar.DATE])
+            val hasParentLanguageCode = !WikipediaApp.instance.languageState.getDefaultLanguageCode(wikiSite.languageCode).isNullOrEmpty()
+            if (hasParentLanguageCode) {
+                val events = response.allEvents().filter { it.pages.isNotEmpty() }
+                val eventPages = events.map { event ->
+                    async { L10nUtil.getPagesForLanguageVariant(event.pages, wikiSite) }
```
dbrant, line 46: I'm not sure this is a good idea. This will make an API call for each page in each event, which could be **hundreds** of API calls.
This is really a case that should probably be fixed on the server side of the `/feed/onthiday` API, not patched over on the client side.

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDay.kt
cooltey: Yes, done!

app/src/main/java/org/wikipedia/feed/onthisday/OnThisDayViewModel.kt
cooltey: It makes sense to me. 😢

cooltey:
Will have a follow-up PR for simplifying the On this day card.

PR #4938, opened 2024-09-09
Simplify the logic of picking up a random event from On This Day 

cooltey:
### What does this do?
Slightly simplify the logic of picking up one event from the list by using `random()` method, and move the logic to `AggregatedFeedContentClient` instead of having the logic in the `OnThisDayCard`.

### Why is this needed?
Simplify the logic and prepare for #4927 to avoid making unnecessary API calls.

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/aggregated/AggregatedFeedContent.kt
+++ b/app/src/main/java/org/wikipedia/feed/aggregated/AggregatedFeedContent.kt
@@ -15,4 +15,6 @@ class AggregatedFeedContent(
     @SerialName("mostread") val topRead: TopRead? = null,
     @SerialName("image") val potd: FeaturedImage? = null,
     val onthisday: List<OnThisDay.Event>? = null
-)
+) {
+    var randomOnThisDay: OnThisDay.Event? = null
+}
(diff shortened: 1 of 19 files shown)


app/src/main/java/org/wikipedia/feed/aggregated/AggregatedFeedContent.kt
```diff
@@ -15,4 +15,6 @@ class AggregatedFeedContent(
     @SerialName("mostread") val topRead: TopRead? = null,
     @SerialName("image") val potd: FeaturedImage? = null,
     val onthisday: List<OnThisDay.Event>? = null
-)
+) {
+    var randomOnThisDayEvent: OnThisDay.Event? = null
```
dbrant, line 19: Since this class is `@Serializable`, but we don't actually want this new field to be serialized, can we add `@Transient` to this field?

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/aggregated/AggregatedFeedContent.kt
cooltey: Yes! Done.

PR #4946, opened 2024-09-11
Update campaign ID format sent with GPay transaction.

dbrant:
The campaign ID that is sent to the fundraising API needs to be updated to match the format used by iOS, which is: `language`+`campaignID`+`_`+`platform`.  (note no underscore between language and ID, and underscore between ID and platform)
This will enhance and simplify analytics.

**Phabricator:**
https://phabricator.wikimedia.org/T367360


dbrant:
@cooltey This should be ready. The usage of `appOrSystemLanguageCode` is identical to what we use for determining the language assets to display for a campaign.

PR #4956, opened 2024-09-13
Add is_temp field to legacy analytics, and bump schema versions.

dbrant:
This adds the `is_temp` field to our existing MEP schemas.
(Dependent on merging and deploying the corresponding [analytics MR](https://gitlab.wikimedia.org/repos/data-engineering/schemas-event-secondary/-/merge_requests/2))

**Phabricator:**
https://phabricator.wikimedia.org/T374057


app/src/main/java/org/wikipedia/analytics/eventplatform/MobileAppsEvent.kt
```diff
@@ -15,3 +15,9 @@ sealed class MobileAppsEvent(@Transient private val _streamName: String = "") :
     @SerialName("app_session_id") @Required private val sessionId = EventPlatformClient.AssociationController.sessionId
     @SerialName("app_install_id") @Required private val appInstallId = WikipediaApp.instance.appInstallID
 }
+
+@Suppress("unused")
+@Serializable
+sealed class MobileAppsEventWithTemp(@Transient private val _streamName: String = "") : MobileAppsEvent(_streamName) {
```
dbrant, line 21: This was necessary because _not all_ schemas that inherit from `MobileAppsEvent` needed to be updated with an `is_temp` parameter, so we don't want the rest of them inheriting it automatically.

PR #4958, opened 2024-09-13
Fix crash no results

cooltey:
### What does this do?
Fix a possible crash when showing results for other languages in the `SearchResultsFragment`

### Why is this needed?
We get the language code by providing a position to the language list, which may cause a crash.
```
val langCode = WikipediaApp.instance.languageState.appLanguageCodes[position]
```
This PR changes the plain list to a `Pair<>` that contains the language code with its result counts.

### The error log in console
https://play.google.com/console/u/1/developers/6169333749249604352/app/4976363884102945010/vitals/crashes/0e0fb84f88cb64408c0412acfb8550ff/details?days=28&versionCode=50501&isUserPerceived=true

### DIFF
--- a/app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
+++ b/app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
@@ -26,20 +26,20 @@ class SearchResultsViewModel : ViewModel() {
 
     private val batchSize = 20
     private val delayMillis = 200L
-    var resultsCount = mutableListOf<Int>()
+    var resultPairList = mutableListOf<Pair<String, Int>>()
     var searchTerm: String? = null
     var languageCode: String? = null
     lateinit var invokeSource: Constants.InvokeSource
 
     @OptIn(FlowPreview::class) // TODO: revisit if the debounce method changed.
     val searchResultsFlow = Pager(PagingConfig(pageSize = batchSize, initialLoadSize = batchSize)) {
-        SearchResultsPagingSource(searchTerm, languageCode, resultsCount, invokeSource)
+        SearchResultsPagingSource(searchTerm, languageCode, resultPairList, invokeSource)
     }.flow.debounce(delayMillis).cachedIn(viewModelScope)
 
     class SearchResultsPagingSource(
         private val searchTerm: String?,
         private val languageCode: String?,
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
```diff
@@ -26,20 +26,20 @@ class SearchResultsViewModel : ViewModel() {
 
     private val batchSize = 20
     private val delayMillis = 200L
-    var resultsCount = mutableListOf<Int>()
+    var resultPairList = mutableListOf<Pair<String, Int>>()
```
dbrant, line 29: Can we give this a clearer name, such as `countsPerLanguageCode`?

app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
```diff
@@ -26,20 +26,20 @@ class SearchResultsViewModel : ViewModel() {
...
     val searchResultsFlow = Pager(PagingConfig(pageSize = batchSize, initialLoadSize = batchSize)) {
-        SearchResultsPagingSource(searchTerm, languageCode, resultsCount, invokeSource)
+        SearchResultsPagingSource(searchTerm, languageCode, resultPairList, invokeSource)
     }.flow.debounce(delayMillis).cachedIn(viewModelScope)
 
     class SearchResultsPagingSource(
         private val searchTerm: String?,
         private val languageCode: String?,
-        private var resultsCount: MutableList<Int>?,
+        private var resultPairList: MutableList<Pair<String, Int>>,
```
dbrant, line 42: Can we give this a clearer name, such as `countsPerLanguageCode`?

dbrant (changes requested):
A good refactor. (The six affected users will thank us!)

app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
cooltey: Done!

app/src/main/java/org/wikipedia/search/SearchResultsViewModel.kt
cooltey: Done.

dbrant (approved):

PR #4959, opened 2024-09-15
Move description edit extras to a view model

Isira-Seneviratne:
### What does this do?
This PR moves the description edit extras to a separate view model class, and has the description edit fragment load its activity's view model using the [`activityViewModels`](https://developer.android.com/reference/kotlin/androidx/fragment/app/package-summary#(androidx.fragment.app.Fragment).activityViewModels(kotlin.Function0,kotlin.Function0)) extension.

Note: The [`SavedStateHandle`](https://developer.android.com/reference/androidx/lifecycle/SavedStateHandle) class automatically loads the activity's intent extras and is passed to the view model when using the [default view model factory](https://github.com/androidx/androidx/blob/androidx-main/activity/activity/src/main/java/androidx/activity/ComponentActivity.kt#L547). More information on `SavedStateHandle` can be found [here](https://medium.com/androiddevelopers/viewmodels-persistence-onsaveinstancestate-restoring-ui-state-and-loaders-fc7cc4a6c090).

### Why is this needed?
This change allows both the description edit activity and fragment to access the extras passed to the activity without having to pass the extras from the activity to the fragment manually.

### DIFF
--- a/app/src/main/java/org/wikipedia/Constants.kt
+++ b/app/src/main/java/org/wikipedia/Constants.kt
@@ -19,6 +19,10 @@ object Constants {
     const val ARG_WIKISITE = "wikiSite"
     const val ARG_TEXT = "text"
     const val ARG_BOOLEAN = "boolean"
+    const val ARG_HIGHLIGHT_TEXT = "highlightText"
+    const val ARG_SOURCE_SUMMARY = "sourceSummary"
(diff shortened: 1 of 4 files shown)


dbrant:
Sorry, this basically just moves things around without any real benefit, with the goal of saving a few lines of code.
The benefit of a ViewModel is to encapsulate business logic in a lifecycle-aware way, which this does not do.

PR #4965, opened 2024-09-16
Fix: possible crashes when using getItem() in PagingDataAdapter

cooltey:
### What does this do?
Add an extra check to prevent a possible `IndexOutOfBoundsException` crash in a PagingDataAdapter class.

### Why is this needed?
Multiple places where we have used `getItem(position)` for the itemViewType have similar crash reports:
[EditHistoryListActivity](https://play.google.com/console/u/1/developers/6169333749249604352/app/4976363884102945010/vitals/crashes/6dbc71596a037bfd6d2ea1c07017b84d/details?days=28&versionCode=50501&isUserPerceived=true )
[UserContribListActivity](https://play.google.com/console/u/1/developers/6169333749249604352/app/4976363884102945010/vitals/crashes/c414a99d40bf38e1bda9a3d0744742d1/details?days=28&versionCode=50501&isUserPerceived=true)


cooltey:
> This might be a band-aid for these crashes, but I'm not sure it's a "fix". Do we have any idea **why** the adapter is requesting items that are out of bounds? Having these checks will probably just move the crash to another line of code somewhere else, since the adapter will continue to try building the invalid item. This might just be an SDK issue. And to an outside observer reading this code, it's not clear why we're doing checks that shouldn't need to be done.

I see. That makes sense to me and it might be an SDK issue +1.

PR #4970, opened 2024-09-19
Use withContent(Dispatchers.IO) for the API calls in ViewModel classes

cooltey:
### What does this do?
Use `withContent(Dispatchers.IO)` for API calls in `ViewModel` classes to achieve the best practice that network or database calls should be handled on a background thread.


dbrant:
> Use `withContent(Dispatchers.IO)` for API calls in `ViewModel` classes to achieve the best practice that network or database calls should be handled on a background thread.

Can I ask where this best practice is documented? I'm not sure I see why this is necessary.
The best practice I'm familiar with is "[Suspend functions should be safe to call from the main thread](https://developer.android.com/kotlin/coroutines/coroutines-best-practices#main-safe)", meaning that `suspend` functions should take responsibility for managing their own `withContext` calls, which all of our network and database calls absolutely do, automatically.

cooltey:
> > Use `withContent(Dispatchers.IO)` for API calls in `ViewModel` classes to achieve the best practice that network or database calls should be handled on a background thread.
> 
> Can I ask where this best practice is documented? I'm not sure I see why this is necessary. The best practice I'm familiar with is "[Suspend functions should be safe to call from the main thread](https://developer.android.com/kotlin/coroutines/coroutines-best-practices#main-safe)", meaning that `suspend` functions should take responsibility for managing their own `withContext` calls, which all of our network and database calls absolutely do, automatically.



> > Use `withContent(Dispatchers.IO)` for API calls in `ViewModel` classes to achieve the best practice that network or database calls should be handled on a background thread.
> 
> Can I ask where this best practice is documented? I'm not sure I see why this is necessary. The best practice I'm familiar with is "[Suspend functions should be safe to call from the main thread](https://developer.android.com/kotlin/coroutines/coroutines-best-practices#main-safe)", meaning that `suspend` functions should take responsibility for managing their own `withContext` calls, which all of our network and database calls absolutely do, automatically.

Thanks! I think you are correct. I was reading some blog posts and threads about adding `withContext()` for API/database calls, but I guess I missed they were referring to those non-suspend functions.

PR #4976, opened 2024-09-20
Move readingListPage handling behavior to the utility class and use coroutine

cooltey:
### What does this do?
To remove the usage of `RxJava` and use Coroutine.

https://phabricator.wikimedia.org/T367082)


app/src/main/java/org/wikipedia/readinglist/ReadingListBehaviorsUtil.kt
```diff
@@ -135,6 +135,15 @@ object ReadingListBehaviorsUtil {
         }
     }
 
+    fun updateReadingListPage(activity: AppCompatActivity, item: ReadingListPage) {
+        activity.lifecycleScope.launch(exceptionHandler) {
```
dbrant, line 139: Does this really need to be bound to the lifecycleScope of the activity? This could be a case where `MainScope` is more appropriate.

dbrant (changes requested):

PR #4982, opened 2024-09-24
Fix: potential crash in the OnOffsetChangedListener

cooltey:
### What does this do?
To fix a potential crash in the `OnOffsetChangedListener`:
https://play.google.com/console/u/1/developers/6169333749249604352/app/4976363884102945010/vitals/crashes/b0c7fabc2e4b821b88b8fe42d0a7d797/details?days=28&versionCode=50501&isUserPerceived=true


app/src/main/java/org/wikipedia/feed/news/NewsFragment.kt
```diff
@@ -60,6 +60,9 @@ class NewsFragment : Fragment() {
 
         DeviceUtil.updateStatusBarTheme(requireActivity(), binding.toolbar, true)
         binding.appBarLayout.addOnOffsetChangedListener { layout, offset ->
+            if (!isAdded) {
```
dbrant, line 63: The crash you linked seems to be specific to the NewsFragment. Is there a reason to add unnecessary checks to the other fragments?

If you compare the way the OffsetChangedListener is used here, versus in ArticleEditDetailsFragment, the real reason for the crash becomes more clear.

dbrant (changes requested):

PR #4984, opened 2024-09-24
Fix: a possible crash about lateinit WikipediaApp instance has not en initialized

cooltey:
### What does this do?
Some `RuntimeException` related issues pointed out that the `WikipediaApp` instance is not initialized properly. It will be safer to assign the `instance` variable in the `onCreate` block instead of the `init { }` block.

### Why is this needed?

https://play.google.com/console/u/1/developers/6169333749249604352/app/4976363884102945010/vitals/crashes/42965d4ff897cc78885ba9659f8a975f/details?days=28&versionCode=50501&isUserPerceived=true


app/src/main/java/org/wikipedia/WikipediaApp.kt
```diff
@@ -137,6 +133,9 @@ class WikipediaApp : Application() {
     override fun onCreate() {
         super.onCreate()
 
+        // Set the instance after the Application is fully initialized.
+        instance = this
```
dbrant, line 137: Why is this safer? (What is unsafe about setting `instance = this` in the init block?)
Can you say more about why this would fix the crash?

cooltey: The `init` block is the first to be executed when the class is instantiated; the `onCreate` block is about the Android's lifecycle. By considering the crashes are related to the `Prefs` usage, which uses, `WikipediaApp.instance.resource` to get the resource from the app, I think the fix of this PR is a possible way to fix this issue.

PR #4985, opened 2024-09-24
Convert Tabs to use database

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T353656

### DIFF
--- a/app/src/main/java/org/wikipedia/page/tabs/Tab.kt
+++ b/app/src/main/java/org/wikipedia/page/tabs/Tab.kt
@@ -1,13 +1,20 @@
 package org.wikipedia.page.tabs
 
+import androidx.room.Entity
+import androidx.room.PrimaryKey
+import androidx.room.TypeConverters
 import kotlinx.serialization.Serializable
+import org.wikipedia.database.PageBackStackItemTypeConverter
 import org.wikipedia.page.PageBackStackItem
 import org.wikipedia.page.PageTitle
 
+@Entity
 @Serializable
-class Tab {
-    val backStack = mutableListOf<PageBackStackItem>()
-
+@TypeConverters(PageBackStackItemTypeConverter::class)
+class Tab(
+    @PrimaryKey(autoGenerate = true) var id: Long = 0,
+    val backStack: MutableList<PageBackStackItem> = mutableListOf()
+) {
     var backStackPosition: Int = -1
         get() = if (field < 0) backStack.size - 1 else field
 
--- a/app/src/main/java/org/wikipedia/database/AppDatabase.kt
+++ b/app/src/main/java/org/wikipedia/database/AppDatabase.kt
@@ -16,6 +16,8 @@ import org.wikipedia.notifications.db.Notification
 import org.wikipedia.notifications.db.NotificationDao
 import org.wikipedia.offline.db.OfflineObject
 import org.wikipedia.offline.db.OfflineObjectDao
+import org.wikipedia.page.tabs.Tab
+import org.wikipedia.page.tabs.TabDao
 import org.wikipedia.pageimages.db.PageImage
 import org.wikipedia.pageimages.db.PageImageDao
 import org.wikipedia.readinglist.database.ReadingList
@@ -44,7 +46,8 @@ const val DATABASE_VERSION = 26
         ReadingList::class,
         ReadingListPage::class,
         Notification::class,
-        TalkTemplate::class
+        TalkTemplate::class,
+        Tab::class
     ],
     version = DATABASE_VERSION
 )
@@ -52,7 +55,8 @@ const val DATABASE_VERSION = 26
     DateTypeConverter::class,
     WikiSiteTypeConverter::class,
     NamespaceTypeConverter::class,
-    NotificationTypeConverters::class
+    NotificationTypeConverters::class,
+    PageBackStackItemTypeConverter::class
 )
 abstract class AppDatabase : RoomDatabase() {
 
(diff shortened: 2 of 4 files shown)


cooltey:
A basic refactoring is completed. Pending discussion to see if we need to update the logic of handling tabs vs backstacks.

app/src/main/java/org/wikipedia/page/tabs/Tab.kt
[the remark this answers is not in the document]
cooltey: Since this is not related to this PR, would you be able to create a PR of it so we can review the code? Thanks!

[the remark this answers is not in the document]
Williamrai: Done, created another PR.

cooltey:
Here are the results of the comparison: 

### Getting Tab counts

Switching between screens that have the `TabCountsView` embedded.

**SharedPreference**
```
99 Tabs: org.wikipedia.page.PageActivity@8a8e6d7 : 481094 nano seconds ~= 0 milliseconds
99 Tabs: org.wikipedia.page.PageActivity@8a8e6d7 : 329219 nano seconds ~= 0 milliseconds
99 Tabs: null : 373594 nano seconds ~= 0 milliseconds
99 Tabs: null : 390260 nano seconds ~= 0 milliseconds
99 Tabs: null : 1834531 nano seconds ~= 1 milliseconds
99 Tabs: null : 594010 nano seconds ~= 0 milliseconds
99 Tabs: null : 1522084 nano seconds ~= 1 milliseconds => Removes a tab
99 Tabs: null : 486146 nano seconds ~= 0 milliseconds
99 Tabs: null : 3110781 nano seconds ~= 3 milliseconds  => Adds a new tab
99 Tabs: null : 539635 nano seconds ~= 0 milliseconds
99 Tabs: org.wikipedia.main.MainActivity@e5f6958 : 69427 nano seconds ~= 0 milliseconds
99 Tabs: null : 109271 nano seconds ~= 0 milliseconds
99 Tabs: null : 253386 nano seconds ~= 0 milliseconds
99 Tabs: null : 398698 nano seconds ~= 0 milliseconds
99 Tabs: org.wikipedia.page.tabs.TabActivity@ec351f9 : 73177 nano seconds ~= 0 milliseconds
```

**Database**
```
99 Tabs: org.wikipedia.page.tabs.TabActivity@6e96d50 : 937500 nano seconds ~= 0 milliseconds
99 Tabs: org.wikipedia.page.tabs.TabActivity@6e96d50 : 972396 nano seconds ~= 0 milliseconds
99 Tabs: null : 679375 nano seconds ~= 0 milliseconds
99 Tabs: org.wikipedia.page.PageActivity@ea24126 : 254114 nano seconds ~= 0 milliseconds
99 Tabs: org.wikipedia.page.PageActivity@ea24126 : 368906 nano seconds ~= 0 milliseconds
99 Tabs: org.wikipedia.page.PageActivity@ea24126 : 356042 nano seconds ~= 0 milliseconds
99 Tabs: null : 330886 nano seconds ~= 0 milliseconds
99 Tabs: null : 381510 nano seconds ~= 0 milliseconds
99 Tabs: null : 1926511 nano seconds ~= 1 milliseconds
99 Tabs: null : 1226771 nano seconds ~= 1 milliseconds
99 Tabs: null : 840260 nano seconds ~= 0 milliseconds
99 Tabs: null : 334792 nano seconds ~= 0 milliseconds
99 Tabs: null : 569010 nano seconds ~= 0 milliseconds
99 Tabs: null : 445104 nano seconds ~= 0 milliseconds
99 Tabs: null : 1070520 nano seconds ~= 1 milliseconds  => Adds a new tab
99 Tabs: null : 438542 nano seconds ~= 0 milliseconds
99 Tabs: null : 307656 nano seconds ~= 0 milliseconds
99 Tabs: null : 294271 nano seconds ~= 0 milliseconds
99 Tabs: null : 437292 nano seconds ~= 0 milliseconds
99 Tabs: null : 320938 nano seconds ~= 0 milliseconds
99 Tabs: null : 337396 nano seconds ~= 0 milliseconds
```
----

#### Committing Tabs

Starts with 99 tabs, and do some add/remove tabs actions:

**SharedPreference**
```
commitTabs: 4486510 nano seconds ~= 4 milliseconds
commitTabs: 2336094 nano seconds ~= 2 milliseconds
commitTabs: 4442240 nano seconds ~= 4 milliseconds
commitTabs: 1541407 nano seconds ~= 1 milliseconds
commitTabs: 7172552 nano seconds ~= 7 milliseconds
commitTabs: 1680573 nano seconds ~= 1 milliseconds
commitTabs: 2547500 nano seconds ~= 2 milliseconds
```

**Database**
```
commitTabs: 26842500 nano seconds ~= 26 milliseconds
commitTabs: 51131771 nano seconds ~= 51 milliseconds
commitTabs: 29053646 nano seconds ~= 29 milliseconds
commitTabs: 18281302 nano seconds ~= 18 milliseconds
commitTabs: 21329896 nano seconds ~= 21 milliseconds
commitTabs: 35114740 nano seconds ~= 35 milliseconds
commitTabs: 27067552 nano seconds ~= 27 milliseconds
```

In conclusion, the `SharedPreference` performs better than the database, especially when trying to commit Tabs. I think we can optimize the way of handling the delete/add action instead of saving/deleting all tabs in the `commitTabState()` function.

app/src/main/java/org/wikipedia/page/tabs/Tab.kt
```diff
@@ -1,21 +1,29 @@
...
 
+@Entity
 @Serializable
-class Tab {
-    val backStack = mutableListOf<PageBackStackItem>()
-
+@TypeConverters(PageBackStackItemTypeConverter::class)
+class Tab(
+    @PrimaryKey(autoGenerate = true) var id: Long = 0,
+    val backStack: MutableList<PageBackStackItem> = mutableListOf()
```
cooltey: Per discussion, I was thinking about using the `HistoryEntry` table and using the `id`s that can be stored as a list. 

But the question now is: the `HistoryEntryDao` contains a `deleteAll()`, which means the `id`s may not be available if users choose to delete them.  

Do we want to create another table, which might be a simpler version of `HistoryEntry`, but I'm not sure if it's really worth doing instead of storing the JSON text.

@dbrant @Williamrai

cooltey: We will create a new table to store the backstack items.

cooltey:
@dbrant @Williamrai this is ready for review.

app/src/main/java/org/wikipedia/page/tabs/TabHelper.kt
```diff
@@ -0,0 +1,194 @@
...
+
+    fun getCurrentTab(): Tab {
+        if (list.isEmpty()) {
+            list.add(Tab())
+        }
+        return list.last()
+    }
+
+    suspend fun initTabs() {
+        migrateTabsToDatabase()
```
dbrant: Shouldn't this be done during the Migration step of the database (from 31 to 32)?

app/src/main/java/org/wikipedia/page/tabs/TabHelper.kt
```diff
@@ -0,0 +1,194 @@
...
+import org.wikipedia.util.log.L
+
+object TabHelper {
+
+    val coroutineScope = CoroutineScope(Dispatchers.IO)
+    val coroutineExceptionHandler = CoroutineExceptionHandler { _, throwable ->
+        L.e(throwable)
+    }
+
+    val list = mutableListOf<Tab>()
```
dbrant: It might be possible to improve this even further architecturally.
One of the goals of storing tabs in a database is to no longer need to keep all of them in memory all the time.

All of the conveniences offered by keeping a static `list` of Tabs can now be done by making a database query. When reading an article, only the current backstack item needs to be in memory. When navigating Back or Forward, we can now make a database query to get the next or previous backstack item. For showing the total "number" of tabs in TabCountsView, that can be a database query.

This way, there would be no further need for a "helper" object, and all Tab interactions could be done through its Dao. How do you feel about this?

app/src/main/java/org/wikipedia/page/tabs/TabHelper.kt
cooltey: Yes! that's better

app/src/main/java/org/wikipedia/page/tabs/TabHelper.kt
cooltey: It sounds good to me. The only question I have is: how to manage the existing `count` (or previously `tabCount`), which has been used in a lot of places, not sure if we want to do a SQL query every time for the `count` number for various logic across the app.

cooltey: Per discussion:
- Will remove the `list` from `TabHelper` class and probably remove that helper class entirely.
- Refine the backstack logic to make it get the current tab from the `DAO` class directly in the `PageFragment`.
- Will create a `DAO` class for `TabActivity`, too.
- Keep the `count` somewhere else and update its value after committing the queries.

cooltey:
Close it now. We can resume the work when we need it.

PR #4989, opened 2024-09-27
Integrate In-App Internet censorship circumvention

Cuthead:
### What does this do?

1. Add a DNS-over-HTTPS resolver for all network request by OkHttp, fallback to system DNS if not available. Currently I use [Wikimedia DNS](https://meta.wikimedia.org/wiki/Wikimedia_DNS/Instructions).This is always enabled.
2. Add a SSLSocketFactory for removing HTTPS SNI to bypass Deep Packet Inscription, this can be enabled in settings.

### Why is this needed?

For users in suppressing countries like China.

**Phabricator:**
https://phabricator.wikimedia.org/T327286

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -204,6 +204,7 @@ dependencies {
 
     implementation libs.okhttp.tls
     implementation libs.okhttp3.logging.interceptor
+    implementation libs.okhttp3.dnsoverhttps
     implementation libs.retrofit
     implementation libs.retrofit2.adapter.rxjava3
     implementation libs.rxjava
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -14,6 +14,8 @@
 # --- OkHttp + Okio ---
 -dontwarn okhttp3.**
 -dontwarn okio.**
+-keeppackagenames okhttp3.internal.publicsuffix.*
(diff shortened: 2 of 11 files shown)


dbrant:
Thanks!
I'm not sure we're ready at a product level to have this feature be part of the app. I would suggest continuing the conversation in the Phabricator task until it's officially prioritized as a feature.

Some additional thoughts on the subject, specific to the Android case:
* China blocks not just Wikipedia, but also the Google Play Store, as well as F-Droid. Even if we build this feature into the app, how would users acquire the app to begin with?
* Android 9 and above has a system-wide setting for configuring Private DNS, which eliminates the need for individual apps to implement it themselves.

Cuthead:
> China blocks not just Wikipedia, but also the Google Play Store, as well as F-Droid. 

F-Droid client embedded many mirrors, and some Chinese university even provide their own F-Droid mirror. None of them is blocked.

> Even if we build this feature into the app, how would users acquire the app to begin with?

As I've mentioned in the phab ticket, it's about [open proxy policy](https://en.wikipedia.org/wiki/Wikipedia:Open_proxies). Even if users can download Wikipedia app  using a VPN, they cannot edit it. They have to manually apply for IPBE permission.

PR #4993, opened 2024-09-27
Facepalm: correctly get campaign id from Activity's intent.

dbrant:
In one of the most unfortunate cases of name shadowing, there were two variables called `intent`, and we were using the wrong one to get the campaign ID, resulting in a fallback to the default id.

https://phabricator.wikimedia.org/T375124

### DIFF
--- a/app/src/extra/java/org/wikipedia/donate/GooglePayActivity.kt
+++ b/app/src/extra/java/org/wikipedia/donate/GooglePayActivity.kt
@@ -271,8 +271,8 @@ class GooglePayActivity : BaseActivity() {
         if (requestCode == LOAD_PAYMENT_DATA_REQUEST_CODE) {
             when (resultCode) {
                 Activity.RESULT_OK -> {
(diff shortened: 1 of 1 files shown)


app/src/extra/java/org/wikipedia/donate/GooglePayActivity.kt
```diff
@@ -271,8 +271,8 @@ class GooglePayActivity : BaseActivity() {
         if (requestCode == LOAD_PAYMENT_DATA_REQUEST_CODE) {
             when (resultCode) {
                 Activity.RESULT_OK -> {
-                    data?.let { intent ->
-                        PaymentData.getFromIntent(intent)?.let { paymentData ->
```
dbrant, line 275: This name of `intent` is shadowing the Activity's `intent` getter (shorthand for the function `getIntent()`), so when we get the campaign ID a few lines later, by calling `intent.getStringExtra...`, we're referencing the wrong intent!

PR #4994, opened 2024-09-27
Fix FromHtml rendering of ordered lists with Paragraph spans.

dbrant:
Our handling of ordered list `<ol>` tags is kind of brittle already, and can lead to issues, because we programmatically insert text into an `Editable` inside a phase where text shouldn't really be inserted. This can cause clashes with existing Spans, especially ParagraphStyle spans.
This fixes further existing issues by explicitly removing any additional `ParagraphStyle` spans that might be interfering with an `<ol>` tag. A more permanent solution will need to be devised later.

https://phabricator.wikimedia.org/T375728

### DIFF
--- a/app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
+++ b/app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
@@ -13,6 +13,7 @@ import android.text.Html.TagHandler
 import android.text.Spannable
 import android.text.Spanned
 import android.text.style.LeadingMarginSpan
+import android.text.style.ParagraphStyle
 import android.text.style.TypefaceSpan
 import android.text.style.URLSpan
 import android.widget.TextView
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
```diff
@@ -201,9 +202,18 @@ class CustomHtmlParser(private val handler: TagHandler) : TagHandler, ContentHan
...
+                // TODO: improve this logic to no longer require explicitly inserting the count
+                // into the output text. This requires manual and fragile manipulation of any
+                // existing spans that may be present in the output text.
+                val paragraphSpans = output.getSpans<ParagraphStyle>(output.length)
+                var lastLeadingSpan: LeadingMarginSpan? = null
+                paragraphSpans.forEach {
+                    if (it !is LeadingMarginSpan)
+                        output.removeSpan(it)
+                    else
+                        lastLeadingSpan = it
```
cooltey, line 214: Can we wrap the code in the `if` statement with `{ }` for code style consistency?

dbrant: sure!

PR #4995, opened 2024-09-27
Update the item width for the customize toolbar overflow menu

cooltey:
### What does this do?
Make sure the width of the menu items have the proper width.

Before vs After

### DIFF
--- a/app/src/main/java/org/wikipedia/views/PageActionOverflowView.kt
+++ b/app/src/main/java/org/wikipedia/views/PageActionOverflowView.kt
@@ -8,6 +8,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.FrameLayout
+import android.widget.LinearLayout
 import android.widget.PopupWindow
 import androidx.core.view.doOnDetach
 import androidx.core.widget.PopupWindowCompat
@@ -22,6 +23,7 @@ import org.wikipedia.page.action.PageActionItem
 import org.wikipedia.page.customize.CustomizeToolbarActivity
 import org.wikipedia.page.tabs.Tab
 import org.wikipedia.settings.Prefs
+import org.wikipedia.util.DimenUtil
 import org.wikipedia.util.ResourceUtil
 
 class PageActionOverflowView(context: Context) : FrameLayout(context) {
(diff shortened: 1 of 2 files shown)


app/src/main/res/layout/view_page_action_overflow.xml
```diff
@@ -19,7 +19,7 @@
 
             <LinearLayout
                 android:id="@+id/overflowList"
-                android:layout_width="wrap_content"
+                android:layout_width="match_parent"
```
dbrant, line 22: This will prevent the items from being wider than the "Customize toolbar" item, since the "Customize toolbar" item is now the only one that has `wrap_content`. (And there are some languages where certain items are longer than "Customize toolbar", and will be cut off.)

The proper solution might be to leave these items as `wrap_content`, and then measure the width of the widest item at runtime, and adjust the other widths to match the maximum.

dbrant (changes requested):

app/src/main/res/layout/view_page_action_overflow.xml
cooltey: Tried to get the max width from the list items and adjust them by setting either `view.width` or `layoutParams.width`, but it will create a weird space at the bottom of the list 🤔 

I have another way to add menu items inside one single `LinearLayout` but the `border` view makes the entire view wider... any suggestions for this?

app/src/main/java/org/wikipedia/views/PageActionOverflowView.kt
```diff
@@ -47,10 +49,23 @@ class PageActionOverflowView(context: Context) : FrameLayout(context) {
...
+        view.setCompoundDrawablesRelativeWithIntrinsicBounds(R.drawable.ic_settings_black_24dp, 0, 0, 0)
+        view.setOnClickListener {
             dismissPopupWindowHost()
             context.startActivity(CustomizeToolbarActivity.newIntent(context))
         }
+        val border = View(context)
+        border.layoutParams = LinearLayout.LayoutParams(LayoutParams.WRAP_CONTENT, DimenUtil.roundedDpToPx(1f))
+        border.setBackgroundColor(ResourceUtil.getThemedColor(context, R.attr.border_color))
+        view.tag = false
+//        binding.overflowList.addView(border)
```
cooltey, line 67: If we add the border view, the overflow menu becomes wider.

cooltey:
We can visit this later.

PR #5003, opened 2024-10-01
Use CoroutineWorker in ReadingListSyncAdapter

cooltey:
### What does this do?
Since this file is still using the observable version of `getCsrfToken()`, instead of using a `CoroutineScope`, we can convert it to use `CoroutineWorker` for consistency

**Phabricator:**


app/src/main/java/org/wikipedia/readinglist/sync/ReadingListSyncAdapter.kt
```diff
@@ -24,413 +33,418 @@ import org.wikipedia.settings.RemoteConfig
...
+        inputData.keyValueMap.forEach {
+            extras.putBoolean(it.key, it.value as Boolean)
         }
-        return tokenList[0]
+        return extras
+    }
+
+    @Throws(Throwable::class)
+    private suspend fun getCsrfToken(wiki: WikiSite, token: String?): String {
+        return token ?: CsrfTokenClient.getTokenBlocking(wiki)
```
cooltey, line 447: Not sure why we were using a list to store tokens but only getting the first one, so I made a change to this function.

dbrant: I think this might have been from a time when there were plans to have _per-wiki_ reading lists, so we would need to sync with each wiki individually. But obviously this is no longer the case.

PR #5008, opened 2024-10-04
Add Places app shortcut to the app

cooltey:
### What does this do?
Add a Places app shortcut to the app, and update the logic of checking `setDynamicShortcuts()` since the `dev` app will create a `LeakCanary` shortcut to make the current shortcut list size exactly the same as the new one. 

**Phabricator:**
https://phabricator.wikimedia.org/T375367

### DIFF
--- a/app/src/main/java/org/wikipedia/appshortcuts/AppShortcuts.kt
+++ b/app/src/main/java/org/wikipedia/appshortcuts/AppShortcuts.kt
@@ -27,7 +27,7 @@ class AppShortcuts {
         fun setShortcuts(app: Context) {
             CoroutineScope(Dispatchers.Default).launch(CoroutineExceptionHandler { _, msg -> run { L.e(msg) } }) {
                 val list = listOf(searchShortcut(app), continueReadingShortcut(app), randomShortcut(app), placesShortcut(app))
-                if (ShortcutManagerCompat.getDynamicShortcuts(app).containsAll(list).not()) {
+                if (ShortcutManagerCompat.getDynamicShortcuts(app).map { it.id }.containsAll(list.map { it.id }).not()) {
                     ShortcutManagerCompat.setDynamicShortcuts(app, list)
                 } else {
                     L.d("Create dynamic shortcuts skipped.")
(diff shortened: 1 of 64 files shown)


app/src/main/java/org/wikipedia/appshortcuts/AppShortcuts.kt
```diff
@@ -22,11 +22,12 @@ class AppShortcuts {
...
         private const val APP_SHORTCUT_ID_RANDOM = "shortcutRandom"
         private const val APP_SHORTCUT_ID_SEARCH = "shortcutSearch"
+        private const val APP_SHORTCUT_ID_PLACES = "shortcutPlaces"
 
         fun setShortcuts(app: Context) {
             CoroutineScope(Dispatchers.Default).launch(CoroutineExceptionHandler { _, msg -> run { L.e(msg) } }) {
-                val list = listOf(searchShortcut(app), continueReadingShortcut(app), randomShortcut(app))
-                if (ShortcutManagerCompat.getDynamicShortcuts(app).size < list.size) {
+                val list = listOf(searchShortcut(app), continueReadingShortcut(app), randomShortcut(app), placesShortcut(app))
+                if (ShortcutManagerCompat.getDynamicShortcuts(app).containsAll(list).not()) {
```
dbrant, line 30: This check is no longer correct. The object instances in the local `list` are always different from what's returned by `getDynamicShortcuts()`, so this will always be true, and will always unnecessarily recreate the dynamic shortcuts.
If we really want to do the check correctly, we should check if the shortcut `id`s match.

dbrant (changes requested):

app/src/main/java/org/wikipedia/appshortcuts/AppShortcuts.kt
cooltey: Done. I have noticed that the app shortcut function only allows `4` shortcuts, so the `LeakCanary` app shortcut will be removed after we reset the shortcuts.

dbrant (approved):

PR #5013, opened 2024-10-07
[Feature branch] The Contributions Dashboard

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T376330
https://phabricator.wikimedia.org/T376437

### DIFF
--- a/app/src/main/java/org/wikipedia/usercontrib/ContributionsDashboardHelper.kt
+++ b/app/src/main/java/org/wikipedia/usercontrib/ContributionsDashboardHelper.kt
@@ -1,0 +1,25 @@
+package org.wikipedia.usercontrib
+
+import org.wikipedia.WikipediaApp
+import org.wikipedia.util.GeoUtil
+import org.wikipedia.util.ReleaseUtil
+import java.time.LocalDate
+
+class ContributionsDashboardHelper {
+
+    companion object {
+
+        private val enabledCountries = listOf(
+            "FR", "NL", "IT", "JP"
+        )
+
+        private val enabledLanguages = listOf(
+            "fr", "nl", "en", "it", "ja"
+        )
+
+        val contributionsDashboardEnabled get() = ReleaseUtil.isPreBetaRelease ||
+                (enabledCountries.contains(GeoUtil.geoIPCountry.orEmpty()) &&
+                        enabledLanguages.contains(WikipediaApp.instance.languageState.appLanguageCode) &&
+                        LocalDate.now() <= LocalDate.of(2024, 12, 2))
+    }
+}


app/src/main/res/values-az/strings.xml
```diff
@@ -1675,7 +1675,7 @@
   <string name="contributions_dashboard_survey_dialog_cancel">Rəy yoxdur</string>
   <string name="contributions_dashboard_survey_dialog_ok">Rəy bildir</string>
   <string name="contributions_dashboard_donation_dialog_title">Azad biliyi dəstəklədiyiniz üçün təşəkkür edirik</string>
-  <string name="contributions_dashboard_donation_dialog_message">Siz milyonların etibar etdiyi resursu dəstəkləməyi seçən 2%%-in bir hissəsisiniz. Minnətdarlığımızı nümayiş etdirmək üçün töhfələr paneliniz donor statusunuzu göstərmək üçün yeniləndi. Yeniləməni görmək istəyirsiniz?</string>
+  <string name="contributions_dashboard_donation_dialog_message">Siz milyonların etibar etdiyi resursu dəstəkləməyi seçən 2&#37;-in bir hissəsisiniz. Minnətdarlığımızı nümayiş etdirmək üçün töhfələr paneliniz donor statusunuzu göstərmək üçün yeniləndi. Yeniləməni görmək istəyirsiniz?</string>
```
dbrant, line 1678: What happened to the double-percent-encoding of this string? (in all the languages?) Can we leave it as it was?

dbrant (changes requested):

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -218,54 +245,59 @@ class SuggestedEditsTasksFragment : Fragment() {
...
+//            if (Prefs.showOneTimeSequentialUserStatsTooltip) {
+//                 showOneTimeSequentialUserStatsTooltips()
+//            }
+        }
+
+        binding.donorHistoryContainer.isVisible = true
+        if (!ContributionsDashboardHelper.contributionsDashboardEnabled) {
+            binding.donorHistoryContainer.isVisible = false
+            binding.statsDivider.isVisible = false
         }
```
Williamrai, line 281: I think we should hide `donorHistoryContainer ` by default and only display it when `ContributionsDashboardHelper.contributionsDashboardEnabled` is true. Currently we are only checking `ContributionsDashboardHelper.contributionsDashboardEnabled` when `viewModel.totalContributions != 0`. This means that if the user has not made any contributions we aren't validating this flag, which allows any regions to view the donor history container even without contributions.

cooltey: Thanks! It makes sense to me. Fixed.

app/src/main/java/org/wikipedia/views/PageActionOverflowView.kt
```diff
@@ -102,13 +102,11 @@ class PageActionOverflowView(context: Context) : FrameLayout(context), DonorBadg
     }
 
     private fun loadDonorInformation() {
-        if (ContributionsDashboardHelper.contributionsDashboardEnabled && AccountUtil.isLoggedIn.not()) {
-            binding.donorContainer.isVisible = false
-            return
+        if (ContributionsDashboardHelper.contributionsDashboardEnabled && AccountUtil.isLoggedIn) {
```
cooltey, line 105: Hi @Williamrai, I updated this logic a little bit. The previous logic will show the username item if `contributionsDashboardEnabled = false`, which is not correct.

Williamrai: okay, thank you for finding this.

Williamrai (approved):

PR #5017, opened 2024-10-08
Remove dependency on WikimediaEditorTasks extension.

dbrant:
The [WikimediaEditorTasks](https://gerrit.wikimedia.org/r/admin/repos/mediawiki/extensions/WikimediaEditorTasks,general) extension has not been maintained in years, and is no longer particularly useful for our purposes.
It was used for the following:
* To maintain a count of all the Wikidata descriptions and Commons captions added by the user.
* To maintain a count of edits (of Wikidata descriptions or Commons captions) that were rolled-back or reverted.

Both of the above things can be done with existing API calls. The only benefit of the extension was to maintain the counts over the lifetime of the account, but this marginal enhancement is not really necessary for our logic of keeping track of the "quality" of the user's edits. We can easily switch to using a regular API call to get the user's last _n_ edits, and count the number of reverts.


cooltey (commented):
It looks good to me, just noticed one thing (maybe not related to this PR):

The way to "pause" the suggested edits feature by checking the reverted edits is now more likely to be paused. The way we check the status of pause does not tie to a specific wiki, which means if you got paused in the `test` wiki, you will not be able to see the suggested edits tasks in `en` wiki, because the `suggestedEditsPauseDate` is not empty.

dbrant:
> The way to "pause" the suggested edits feature by checking the reverted edits is now more likely to be paused. The way we check the status of pause does not tie to a specific wiki, which means if you got paused in the `test` wiki, you will not be able to see the suggested edits tasks in `en` wiki, because the `suggestedEditsPauseDate` is not empty.

Good observation! I would still argue that we are very "special" users, i.e. we perform a lot of testing of reverts, testing of weird types of edits, and so on. Regular users on Wikipedia are extremely unlikely to have the same kind of pattern of edits that we have. (Regular users are also unlikely to switch their primary-language wiki as often as we do.)

I would endorse proceeding with this PR, with cautious optimism, and make tweaks to our pausing logic as we go forward. (I also foresee us removing the entire logic of pausing and blocking in the future, based on upcoming discussions.)

app/src/main/java/org/wikipedia/usercontrib/UserContribStats.kt
```diff
@@ -19,28 +21,26 @@ object UserContribStats {
...
-        }
+
+    fun verifyEditCountsAndPauseState(totalContributionsList: List<UserContribution>) {
+        totalEdits = totalContributionsList.size
+        totalReverts = totalContributionsList.count { it.ns == 0 && it.tags.contains("mw-reverted") }
+        maybePauseAndGetEndDate()
     }
 
-    suspend fun getPageViews(response: MwQueryResponse): Long {
+    suspend fun getPageViews(homeWikiContributions: List<UserContribution>, wikidataContributions: List<UserContribution>): Long {
```
cooltey, line 31: I noticed you removed the edit counts for image captions and image tags and did not include the contributions in this pageView function. Is this expected?

dbrant: Yes. Those pageviews are definitely meaningless. (It would get the pageviews of _the Commons page_ of each image, which are not normally visited by anyone.)

PR #5021, opened 2024-10-08
[Contributions Dashboard] Donor History

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T376431


app/src/main/java/org/wikipedia/donate/DonorHistoryActivity.kt
```diff
@@ -0,0 +1,170 @@
...
+        binding.lastDonationLabel.text = getString(lastDonatedText)
+        binding.lastDonationLabel.setTextColor(ResourceUtil.getThemedColorStateList(this, lastDonatedTextColor))
+        viewModel.lastDonated?.let {
+            binding.lastDonationDate.text = DateUtils.getRelativeTimeSpanString(
+                viewModel.dateTimeToMilli(it),
+                System.currentTimeMillis(),
+                DateUtils.DAY_IN_MILLIS
+            )
+        }
+        binding.recurringDonorCheckbox.isEnabled = viewModel.lastDonated != null
```
Williamrai, line 111: Are we only making recurringDonorCheckbox enabled/disabled but not recurringDonorContainer? even though checkbox is disabled we can still click recurringDonorContainer which set the checkbox.

Williamrai: also we are not updating the preference isRecurringDonor when we tap the recurringDonorCheckbox

Williamrai (changes requested):

app/src/main/res/xml/developer_preferences.xml
```diff
@@ -447,6 +447,10 @@
             android:key="@string/preference_key_donor_history_saved"
             android:title="@string/preference_key_donor_history_saved" />
 
+        <org.wikipedia.settings.SwitchPreferenceMultiLine
+            android:key="@string/preference_key_donor_history_saved"
+            android:title="@string/preference_key_donor_history_saved" />
+
```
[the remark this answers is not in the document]
cooltey: Done.

app/src/main/java/org/wikipedia/donate/DonorHistoryViewModel.kt
```diff
@@ -0,0 +1,43 @@
...
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
+import org.wikipedia.Constants
+import org.wikipedia.settings.Prefs
+import java.time.LocalDateTime
+import java.time.ZoneId
+
+class DonorHistoryViewModel(bundle: Bundle) : ViewModel() {
+
+    var completedDonation = bundle.getBoolean(Constants.ARG_BOOLEAN)
```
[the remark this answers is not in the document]
cooltey: Done!

app/src/main/java/org/wikipedia/donate/DonorHistoryActivity.kt
cooltey: Good catch! All done 👍

Williamrai (approved):

app/src/main/res/values/styles.xml
```diff
@@ -599,4 +607,8 @@
         <item name="strokeColor">?attr/paper_color</item>
     </style>
 
+    <style name="MaterialDatePickerStyle" parent="ThemeOverlay.Material3.MaterialCalendar">
```
dbrant, line 610: The date picker looks a bit incorrect in Dark mode (dark text on dark background.).

app/src/main/res/layout/activity_donor_history.xml
```diff
@@ -0,0 +1,172 @@
...
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:orientation="vertical">
+
+                <LinearLayout
+                    android:id="@+id/lastDonationContainer"
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:orientation="horizontal"
+                    android:layout_marginHorizontal="16dp"
```
dbrant, line 75: Can we use `padding` instead of `margin` for these clickable elements, so that the ripple effect can extend to the full width, instead of cutting off at the margin?

dbrant (changes requested):

app/src/main/res/values/styles.xml
```diff
@@ -430,6 +430,14 @@
         <item name="iconTint">?attr/placeholder_color</item>
     </style>
 
+    <style name="App.Button.Text.Outline" parent="Widget.Material3.Button.OutlinedButton">
```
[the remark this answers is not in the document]
cooltey: Done. Update to use the `paper_color` for its background and apply `rippleColor` for the ripple animation.

app/src/main/res/values/styles.xml
cooltey: Good catch! Updated based on [this document](https://github.com/material-components/material-components-android/blob/master/docs/components/DatePicker.md).

app/src/main/res/layout/activity_donor_history.xml
cooltey: Yes, done!

app/src/main/res/values/styles.xml
cooltey: I also updated the text in the single selection text in the dialog (Donor vs. Not a donor) to use a similar setting, based on [this document too](https://github.com/material-components/material-components-android/blob/master/docs/components/Dialog.md).

PR #5030, opened 2024-10-11
Remove RxJava from the app

cooltey:
### What does this do?
This PR removes all `RxJava` related code from the app and renames the `blocking` or `suspend` functions with more proper names. 

**Phabricator:**


dbrant:
Songs will be sung of this day.

PR #5035, opened 2024-10-15
Tabs database code review

Williamrai:
Reviewing code


app/src/main/java/org/wikipedia/page/tabs/Tab.kt
Williamrai, line 1:
```
    fun pushBackStackItem(item: PageBackStackItem) {
        // remove all backstack items past the current position
        while (backStack.size > backStackPosition + 1) {
            backStack.removeAt(backStackPosition + 1)
        }
        backStack.add(item)
        backStackPosition = backStack.size - 1
    }
```
Would it be more efficient to remove this from the end to prevent shifting?

PR #5037, opened 2024-10-16
API underpinnings for integrating with Growth Impact module.

dbrant:
This adds support for retrieving the data from the Impact module of the Growth Experiments extension.
This module provides all kinds of useful statistics about the user's editing history, which will eventually replace the statistics we show in our Edits tab.

https://phabricator.wikimedia.org/T377235

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/growthtasks/GrowthUserImpact.kt
+++ b/app/src/main/java/org/wikipedia/dataclient/growthtasks/GrowthUserImpact.kt
@@ -1,0 +1,60 @@
+package org.wikipedia.dataclient.growthtasks
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.decodeFromJsonElement
+import org.wikipedia.json.JsonUtil
+
+@Suppress("unused")
+@Serializable
+class GrowthUserImpact(
+    @SerialName("@version") val version: Int = 0,
+    val userId: Int = 0,
+    val userName: String = "",
+    val receivedThanksCount: Int = 0,
+    @SerialName("editCountByNamespace") private val mEditCountByNamespace: JsonElement? = null,
+    @SerialName("editCountByDay") private val mEditCountByDay: JsonElement? = null,
+    @SerialName("editCountByTaskType") private val mEditCountByTaskType: JsonElement? = null,
+    val totalUserEditCount: Int = 0,
+    val totalEditsCount: Int = 0,
+    val newcomerTaskEditCount: Int = 0,
+    val revertedEditCount: Int = 0,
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/dataclient/growthtasks/GrowthUserImpact.kt
```diff
@@ -0,0 +1,60 @@
...
+    val totalPageviewsCount: Long = 0,
+    @SerialName("topViewedArticles") private val mTopViewedArticles: JsonElement? = null,
+) {
+    val editCountByNamespace: Map<Int, Int> by lazy { if (mEditCountByNamespace is JsonObject) { JsonUtil.json.decodeFromJsonElement(mEditCountByNamespace) } else { emptyMap() } }
+    val editCountByDay: Map<String, Int> by lazy { if (mEditCountByDay is JsonObject) { JsonUtil.json.decodeFromJsonElement(mEditCountByDay) } else { emptyMap() } }
+    val editCountByTaskType: Map<String, Int> by lazy { if (mEditCountByTaskType is JsonObject) { JsonUtil.json.decodeFromJsonElement(mEditCountByTaskType) } else { emptyMap() } }
+    val dailyTotalViews: Map<String, Int> by lazy { if (mDailyTotalViews is JsonObject) { JsonUtil.json.decodeFromJsonElement(mDailyTotalViews) } else { emptyMap() } }
+    val topViewedArticles: Map<String, ArticleViews> by lazy { if (mTopViewedArticles is JsonObject) { JsonUtil.json.decodeFromJsonElement(mTopViewedArticles) } else { emptyMap() } }
+    val longestEditingStreak: EditStreak? by lazy { if (mLongestEditingStreak is JsonObject) { JsonUtil.json.decodeFromJsonElement(mLongestEditingStreak) } else { null } }
```
cooltey, line 35: Would it be possible to use something like `decodeJsonElementToMap` for the logic in the `lazy` blocks to make them cleaner?

Similar to this:
https://github.com/wikimedia/apps-android-wikipedia/blob/52d9272ab5c565ec89b28e097a59619c4275d62f/app/src/main/java/org/wikipedia/json/JsonUtil.kt#L18-L28

dbrant: I don't believe so, because of the way PHP serializes these types of structures, in the cases where they are empty.
I added some comments to clarify it.

We do this in plenty of other places, such as 
https://github.com/wikimedia/apps-android-wikipedia/blob/52d9272ab5c565ec89b28e097a59619c4275d62f/app/src/main/java/org/wikipedia/dataclient/mwapi/MwQueryResult.kt#L229-L231

(unless you had a different idea for making the code cleaner.)

cooltey: Got it, it makes sense to me.

PR #5039, opened 2024-10-16
Strings for Survey Dialog

Williamrai:
**Phabricator:**
https://phabricator.wikimedia.org/T376436


cooltey (changes requested):
Please update the `contributions_dashboard_survey_dialog_ok` string item from `Share feedback` to `Take survey` directly, since the string is not being used anywhere and it will be fine in this case.

Williamrai:
waiting for product to finalize the strings before making changes.

Williamrai:
This string is not required. so closing this PR.

PR #5043, opened 2024-10-17
Simple Alert Dialog

Williamrai:
### What does this do?
This adds a class called **SimpleAlertDialog**. The purpose of this class is to simplify the creation and display of alert dialog and to centralize the creation of AlertDialogs into a single reusable class. 

### Why is this needed?
It helps to avoid doing the repetitive code often needed with setting up alert dialog and reduce boilerplate code.  It provides customizable Dialog elements like title, message, icon, buttons and theme. Also adds a delay feature.


dbrant:
@Williamrai It's an interesting idea, but can you be more specific about what boilerplate this reduces? Most of our usages of `MaterialAlertDialogBuilder` are pretty concise already.

Maybe it would help if you include an example of switching a usage of `MaterialAlertDialogBuilder` for this new extension? Something like [here](https://github.com/wikimedia/apps-android-wikipedia/blob/main/app/src/main/java/org/wikipedia/page/PageActivity.kt#L687) or [here](https://github.com/wikimedia/apps-android-wikipedia/blob/main/app/src/main/java/org/wikipedia/activity/SingleWebViewActivity.kt#L121).

Williamrai:
The boilerplate I was referring to is the need to manually repeat the MaterialAlertDialogBuilder setup each time we create a dialog. On the contribution dashboard, we have dialogs with different content and some we show after some delay. I think by centralizing the dialog creation in this class, we can simply things for basic dialogs and common features like delay. This way we don't have to repeat the setup every time which makes the setup process cleaner. 

[Example](https://gist.github.com/Williamrai/68574cf5959bd0ab537aa5f0f57bda78)

Williamrai:
Closing this as it's no longer required due to the migration to Compose.

PR #5045, opened 2024-10-17
[Contributions Dashboard] Create a donor badge view

cooltey:
### What does this do?
Create a donor status view for both the overflow menu and the contribute tab.

### Why is this needed?
To be able to share the same design and logic.

### DIFF
--- a/app/src/main/res/layout/view_donor_badge.xml
+++ b/app/src/main/res/layout/view_donor_badge.xml
@@ -13,15 +13,15 @@
         android:layout_height="wrap_content"
         android:text="@string/contributions_dashboard_badge_donor"
         android:textColor="?attr/paper_color"
+        android:visibility="gone"
         app:chipBackgroundColor="?attr/progressive_color"
         app:chipCornerRadius="16dp"
         app:chipIcon="@drawable/ic_baseline_volunteer_activism_24"
         app:chipIconTint="?attr/paper_color"
         app:chipStrokeWidth="0dp"
         app:rippleColor="?attr/overlay_color"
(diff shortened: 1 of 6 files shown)


cooltey:
Hi @Williamrai, this is the PR for the donor badge, and I copied the design from your PR and did some updates to the view ids. Please check and merge it if looks good to you. Thanks!

app/src/main/res/layout/view_donor_badge.xml
```diff
@@ -0,0 +1,58 @@
...
+        android:layout_height="wrap_content"
+        android:text="@string/contributions_dashboard_badge_update"
+        android:textColor="?attr/progressive_color"
+        app:chipBackgroundColor="?attr/paper_color"
+        app:chipCornerRadius="16dp"
+        app:chipStrokeWidth="0dp"
+        app:rippleColor="?attr/overlay_color"
+        android:visibility="gone"
+        tools:text="Update donor status"
+        tools:visibility="visible"/>
```
Williamrai, line 56: chips has default padding and since the background is transparent, the text will not align with the username in the article overflow menu. Removing the start padding looks something like this. 
1. [Without start padding](https://github.com/user-attachments/assets/114ebca7-546a-4422-aa86-ebc747c2e9cd)
2. [With default padding](https://github.com/user-attachments/assets/bcd76160-20b4-44de-af4f-350f9edd721f)

should we remove start padding and does this look good visually?

cooltey: Yes, you are correct! Changed it back to use `TextView`. Thanks!

PR #5063, opened 2024-10-26
Run ReadingListSyncAdapter as expedited work

Isira-Seneviratne:
### What does this do?
This PR updates the reading list sync worker to run as [expedited work](https://developer.android.com/develop/background-work/background-tasks/persistent/getting-started/define-work#expedited).

### Why is this needed?
With this change, WorkManager configures the worker to start execution immediately. It also provides APIs to create/update the work notification.

### Demo

[Screen_recording_20241026_083731.webm](https://github.com/user-attachments/assets/f7720d28-9c25-4020-bdde-b6aa9cb5d04d)

Android 5.0


cooltey:
Thanks for the PR! 

Not sure but I don't see the notification when I start or login the app. Does this work for manually syncing the list?

Isira-Seneviratne:
That might be due to how foreground notifications work on Android 12 and later:

"Devices that run Android 12 (API level 31) or higher provide a streamlined experience for short-running foreground services. On these devices, the system waits 10 seconds before showing the notification associated with a foreground service."

cooltey:
> That might be due to how foreground notifications work on Android 12 and later:
> 
> "Devices that run Android 12 (API level 31) or higher provide a streamlined experience for short-running foreground services. On these devices, the system waits 10 seconds before showing the notification associated with a foreground service."
> 
> I'll update the PR to change the behavior.

Thanks. I am still not seeing the notification. My device is Samsung S24 with Android 14.

Isira-Seneviratne:
It might be because the syncing is usually too fast to show the notification. I set a breakpoint for the update foreground info method on my Pixel 8 with Android 15 and the notification appeared.

PR #5075, opened 2024-10-30
Separates layout on complex XML for easier maintenance

Williamrai:
### What does this do?
Separates layout on complex XML for easier maintenance

### Why is this needed?
In layouts like fragment_suggested_edits_tasks we can organize the code across multiple layouts and include them as needed. This approach will simplify the process of adding new views to the UI and help maintain the design. In Contribution dashboard we needed to add new UI elements for donor history, which increased the number of lines of code.  In the future if we add new designs this may result in a large amount code in a single file, making it challenging to maintain. By using separate layouts, I think we can make this process much more easier.


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragmentViewModel.kt
```diff
@@ -58,7 +58,7 @@ class SuggestedEditsTasksFragmentViewModel : ViewModel() {
             blockMessageWikipedia = null
             blockMessageWikidata = null
             blockMessageCommons = null
-            totalContributions = 0
+            totalContributions = 10
```
cooltey, line 61: Would it be possible to change the default value to `0`?

cooltey:
Thanks @Williamrai! I really like this idea of how it simplified the layout. Can we wait until we merge the feature branch since [I made some layout changes](https://github.com/wikimedia/apps-android-wikipedia/pull/5013/files#diff-9a453acb112a607ae60ffd27ae19e6b31017672a40148f5df96fb6c3fa52dfaaL25) by reducing the duplicate message cards?

PR #5076, opened 2024-10-31
Add strings for Contributions Dashboard

cooltey:
### What does this do?
Add newly [added strings](https://docs.google.com/spreadsheets/d/1z6EhbqeSrIfwykVVsA3QUaYWEOTmAiMEOfkMePdFeyU/edit?gid=0#gid=0) for the Contributions Dashboard feature.


Williamrai:
Could we include this string "App icon changed successfully" here. This copy would show in snackbar when user changes the icon.

cooltey:
Yes, please add the string to this PR directly and push it.

app/src/main/res/values/strings.xml
```diff
@@ -1837,6 +1837,7 @@
     <string name="contributions_dashboard_donor_history_unsaved_warning_dialog_message">If you leave this screen without clicking save, you will lose your changes.</string>
     <string name="contributions_dashboard_donor_history_unsaved_warning_dialog_cancel">Leave without saving</string>
     <string name="contributions_dashboard_donor_history_unsaved_warning_dialog_ok">Save changes</string>
+    <string-array name="contributions_dashboard_app_icon_updated">App Icon updated</string-array>
```
cooltey, line 1840: Not sure but why we are using the `string-array` here?

Williamrai: this was a mistake, it should be string. added string back.

PR #5079, opened 2024-10-31
Mailto links bug

Williamrai:
### What does this do?
Resolved a bug where clicking a URL opened in the browser instead of the email app. 

### Why is this needed?
To open the email app when URL link is clicked. 

**Phabricator:**
https://phabricator.wikimedia.org/T377421


app/src/main/java/org/wikipedia/util/FeedbackUtil.kt
```diff
@@ -110,10 +110,10 @@ object FeedbackUtil {
         UriUtil.visitInExternalBrowser(context, Uri.parse(context.getString(urlStr)))
     }
 
-    fun composeFeedbackEmail(context: Context, subject: String, body: String = "") {
+    fun composeFeedbackEmail(context: Context, email: String = context.getString(R.string.support_email), subject: String, body: String = "") {
```
cooltey, line 113: Since we are using this function for not just the feedback email, can we rename it to `composeEmail`?

app/src/main/java/org/wikipedia/page/LinkHandler.kt
```diff
@@ -31,6 +32,12 @@ abstract class LinkHandler(protected val context: Context) : JSEventListener, Ur
 
     override fun onUrlClick(url: String, titleString: String?, linkText: String) {
         var href = url
+        if (href.startsWith("mailto:")) {
+            val emailAddress = href.removePrefix("mailto:")
+            FeedbackUtil.composeFeedbackEmail(context, email = emailAddress, subject = "", body = "")
```
cooltey, line 37: We can make the `subject`'s default value to `=""`, and then we can skip assigning the empty string for both `subject` and `body`.

app/src/main/java/org/wikipedia/util/FeedbackUtil.kt
Williamrai: Done.

app/src/main/java/org/wikipedia/page/LinkHandler.kt
Williamrai: Done.

cooltey (approved):

PR #5086, opened 2024-11-04
Refining Espresso test

Williamrai:
### What does this do?
This refines the espresso smoke test. Introduces Robot pattern which separates the test into different classes based on specific screen or utility. 

### Why is this needed?
This is needed so that we can 
1. improve code organization 
2. improve test readability
3. developers can update and add test code easily  
4. developers can compose test easily  

**Phabricator:**
https://phabricator.wikimedia.org/T358021


app/src/main/java/org/wikipedia/test/MainActivity.kt
dbrant, line 1: This seems to have moved MainActivity and MainFragment to the `test` package? I'm not sure that's desired.

dbrant (changes requested):

app/src/main/java/org/wikipedia/test/MainActivity.kt
Williamrai: i updated the name for all directories instead of just the android test directories. Fixed.

dbrant:
🎉 A solid foundation for the future of our instrumented tests!

PR #5095, opened 2024-11-06
Fix: Overflow menu design changes

Williamrai:
### What does this do?
adds and changes the vertical padding for overflow menu and **Update donor status** button. 

**Phabricator:**
https://phabricator.wikimedia.org/T376422

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -385,8 +385,8 @@
     </style>
 
     <style name="OverflowMenuItem" parent="ListItem">
-        <item name="android:paddingTop">13dp</item>
-        <item name="android:paddingBottom">13dp</item>
+        <item name="android:paddingTop">10dp</item>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -385,8 +385,8 @@
     </style>
 
     <style name="OverflowMenuItem" parent="ListItem">
-        <item name="android:paddingTop">13dp</item>
-        <item name="android:paddingBottom">13dp</item>
+        <item name="android:paddingTop">10dp</item>
+        <item name="android:paddingBottom">10dp</item>
```
cooltey, line 389: Have you checked with the designer that this is a change across the app? We have used the `OverflowMenuItem` style in other places.

Williamrai: I have not confirmed this with the designer yet. Thanks for flagging that, I will reach out to confirm if this is a global change across the app.

PR #5103, opened 2024-11-07
Fix automatic re-logging-in of user in background.

dbrant:
🤦 🤦 
Our logic for automatically re-logging-in a user relies on catching a `LoginFailedException`, in case the user's credentials are no longer valid, or other exceptional cases. If this exception is caught (after a couple of retries), the logic can proceed to explicitly log the user out.

However, for some reason `LoginFailedException` inherits from `Throwable` instead of `Exception`, but our `try..catch` block catches only `Exception`, which is more specific than `Throwable`!! (This was done during the conversion from RxJava.)
Therefore the exception was no longer getting caught, and instead propagating silently to the higher-level coroutine exception handler. This could cause the app to try re-logging in the user _indefinitely_.

This is a two-pronged fix:
* Make `LoginFailedException` inherit from `Exception`
* Make these `try..catch` blocks catch everything up to `Throwable` for good measure. The logic of this whole function should probably be revisited in the future, and refactored for more structured concurrency.

**Phabricator:**
https://phabricator.wikimedia.org/T379176


app/src/main/java/org/wikipedia/csrf/CsrfTokenClient.kt
```diff
@@ -29,13 +29,13 @@ object CsrfTokenClient {
                 val service = svc ?: ServiceFactory.get(site)
                 var lastError: Throwable? = null
                 for (retry in 0 until MAX_RETRIES) {
-                    if (retry > 0) {
+                    if (retry > 0 && AccountUtil.isLoggedIn && !AccountUtil.isTemporaryAccount) {
```
dbrant, line 32: I'm also adding an explicit guard for re-logging in only if it's definitely not a temp account.

sbassett29:
LGTM

PR #5104, opened 2024-11-08
[Contributions Dashboard] Wire up the instrumentation

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T377049


app/src/main/java/org/wikipedia/analytics/eventplatform/ContributionsDashboardEvent.kt
```diff
@@ -0,0 +1,38 @@
...
+                    activeInterface,
+                    campaignId?.let { "campaign_id: ${CampaignCollection.getFormattedCampaignId(it)}, " }
+                        .orEmpty() + "donor_detected: ${Prefs.isDonor}",
+                    wikiId
+                )
+            }
+        }
+
+        fun anonSuffix(): String {
+            return if (AccountUtil.isLoggedIn) {
```
dbrant, line 31: Should this be the opposite? `!isLoggedIn`?

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -136,7 +137,10 @@ class SuggestedEditsTasksFragment : Fragment() {
         binding.swipeRefreshLayout.setOnRefreshListener { refreshContents() }
 
         binding.errorView.retryClickListener = View.OnClickListener { refreshContents() }
-        binding.errorView.loginClickListener = View.OnClickListener { requestLogin.launch(LoginActivity.newIntent(requireContext(), LoginActivity.SOURCE_SUGGESTED_EDITS)) }
+        binding.errorView.loginClickListener = View.OnClickListener {
+            ContributionsDashboardEvent.logAction("login_click", "contrib_dashboard${ContributionsDashboardEvent.anonSuffix()}")
```
dbrant, line 141: Doesn't the schema already have an automatic field for `is_anon` and `is_temp`? (Why do we need extra logic to append `_anon` here, too?)

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/ContributionsDashboardEvent.kt
cooltey: Removed.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
cooltey: Done!

PR #5109, opened 2024-11-12
Use ShapeableImageView for rounding corners.

dbrant:
(This is in preparation for Rabbit Holes work)

Right now, when we need an `ImageView` with rounded corners, we use a programmatic Glide transformation. However, the Material library already provides a `ShapeableImageView` where we can apply any shape overlay we like, including rounded corners.

This should not produce any visual differences.


app/src/main/res/values/styles.xml
```diff
@@ -480,6 +480,29 @@
         <item name="android:layout_height">28dp</item>
     </style>
 
+    <style name="ImageView.RoundedCorner" parent="">
+        <item name="cornerFamily">rounded</item>
+    </style>
+
+    <style name="ImageView.RoundedCorner.TopLeft">
```
dbrant, line 487: Notice that I'm adding styles for rounding each individual corner of a view. This will be necessary for design updates introduced in Rabbit Holes.

app/src/main/res/values/styles.xml
```diff
@@ -480,6 +480,29 @@
         <item name="android:layout_height">28dp</item>
     </style>
 
+    <style name="ImageView.RoundedCorner" parent="">
```
cooltey, line 483: Not sure but the `parent=""` is intended?

dbrant: Removed! (And renamed to reflect its true purpose)

PR #5115, opened 2024-11-12
Fix: Survey dialog design and logic changes

Williamrai:
### What does this do?
Adds design and logical changes from the design review. 

**Phabricator:**
https://phabricator.wikimedia.org/T376436#10303561

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
@@ -57,7 +57,6 @@ import org.wikipedia.views.DefaultViewHolder
 import java.time.LocalDateTime
 import java.time.ZoneId
 import java.util.Date
-import java.util.concurrent.TimeUnit
 
 class SuggestedEditsTasksFragment : Fragment() {
     private var _binding: FragmentSuggestedEditsTasksBinding? = null
@@ -171,6 +170,24 @@ class SuggestedEditsTasksFragment : Fragment() {
         }
     }
 
+    private fun showDialogOrSnackBar() {
+        when (DonorStatus.donorStatus()) {
+            DonorStatus.DONOR -> {
+                if (ContributionsDashboardHelper.shouldShowThankYouDialog) {
+                    ContributionsDashboardHelper.showThankYouDialog(requireContext())
+                    ContributionsDashboardHelper.shouldShowThankYouDialog = false
+                }
+            }
+            DonorStatus.NON_DONOR -> {
+                if (ContributionsDashboardHelper.shouldShowDonorHistorySnackbar) {
+                    FeedbackUtil.showMessage(this, R.string.donor_history_updated_message_snackbar)
+                    ContributionsDashboardHelper.shouldShowDonorHistorySnackbar = false
+                }
+            }
+            DonorStatus.UNKNOWN -> {}
+        }
+    }
+
     fun refreshContents() {
         (requireActivity() as MainActivity).onTabChanged(NavTab.EDITS)
         requireActivity().invalidateOptionsMenu()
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -463,17 +481,13 @@ class SuggestedEditsTasksFragment : Fragment() {
     }
 
     private fun maybeShowDonorHistoryUpdatedSnackbar() {
-        if (ContributionsDashboardHelper.contributionsDashboardEnabled && ContributionsDashboardHelper.shouldShowDonorHistorySnackbar) {
-            FeedbackUtil.showMessage(this, R.string.donor_history_updated_message_snackbar)
```
cooltey, line 467: Do we still need to show the `donor history updated` message **AFTER** the user seen the survey dialog/special dialog?

For example, if the user `contributionsDashboardSurveyDialogShown` is `true`, the user will not see the "donor history updated" next time when they update the donor history.

app/src/main/res/values/styles.xml
```diff
@@ -563,9 +563,24 @@
         <item name="tint">?attr/primary_color</item>
     </style>
 
+    <style name="DialogIconWithSecondaryTint" parent="MaterialAlertDialog.Material3.Title.Icon.CenterStacked">
```
cooltey, line 566: Can we do something like this:

```
<style name="DialogIconWithTint.Secondary">
    <item name="tint">?attr/secondary_color</item>
</style>
```

This automatically inherits the `DialogIconWithTint` style.

app/src/main/res/values/styles.xml
```diff
@@ -563,9 +563,24 @@
...
+        <item name="tint">?attr/secondary_color</item>
+    </style>
+
+    <style name="AlertDialogTheme.IconWithSecondaryTint" parent="AlertDialogTheme.Icon">
+        <item name="materialAlertDialogTitleIconStyle">@style/DialogIconWithSecondaryTint</item>
+    </style>
```
cooltey, line 572: Same here:

```
<style name="AlertDialogTheme.Icon.Secondary">
    <item name="materialAlertDialogTitleIconStyle">@style/DialogIconWithTint.Secondary</item>
</style>
```

app/src/main/res/values/styles.xml
```diff
@@ -563,9 +563,24 @@
...
+
     <style name="AlertDialogTheme.Icon">
         <item name="materialAlertDialogTitlePanelStyle">@style/MaterialAlertDialog.Material3.Title.Panel.CenterStacked</item>
         <item name="materialAlertDialogTitleIconStyle">@style/DialogIconWithTint</item>
+        <item name="materialAlertDialogTitleTextStyle">@style/centeredTitleText</item>
+    </style>
+
+    <style name="centeredTitleText" parent="MaterialAlertDialog.Material3.Title.Text.CenterStacked">
```
cooltey, line 580: Have you confirmed with the designer about the global change?

Also, can we make the name more dialog-specific? something like: `AlertDialogTitleTextStyle.Centered`

cooltey (changes requested):

app/src/main/res/values/styles.xml
Williamrai: I have not confirmed with design but we are making the title centered for all dialogs. I have asked design for confirmation.

Williamrai: I created a style AlertDialogTheme.Icon.CenteredText, so that we can have variation in style. If we want title centered we can use this one otherwise the other one.

app/src/main/res/values/styles.xml
Williamrai: done.

app/src/main/res/values/styles.xml
Williamrai: done.

cooltey:
Merge this PR so you can move it to get design review first, and make the change later (snackbar)

PR #5117, opened 2024-11-13
Fix: replace the %% symbol with the encoded symbol

cooltey:
### What does this do?
This replaces the double percent symbol, which was intended to show the `%` by adding another `%` to escape the symbol, which does not work in an `AlertDialog` and shows two `%` symbols in the dialog.

### Why is this needed?
To show the percent symbol in the dialog, we can use the encoded symbol.

### DIFF
--- a/app/src/main/res/values-az/strings.xml
+++ b/app/src/main/res/values-az/strings.xml
@@ -1675,7 +1675,7 @@
   <string name="contributions_dashboard_survey_dialog_cancel">Rəy yoxdur</string>
   <string name="contributions_dashboard_survey_dialog_ok">Rəy bildir</string>
   <string name="contributions_dashboard_donation_dialog_title">Azad biliyi dəstəklədiyiniz üçün təşəkkür edirik</string>
-  <string name="contributions_dashboard_donation_dialog_message">Siz milyonların etibar etdiyi resursu dəstəkləməyi seçən 2%%-in bir hissəsisiniz. Minnətdarlığımızı nümayiş etdirmək üçün töhfələr paneliniz donor statusunuzu göstərmək üçün yeniləndi. Yeniləməni görmək istəyirsiniz?</string>
+  <string name="contributions_dashboard_donation_dialog_message">Siz milyonların etibar etdiyi resursu dəstəkləməyi seçən 2&#37;-in bir hissəsisiniz. Minnətdarlığımızı nümayiş etdirmək üçün töhfələr paneliniz donor statusunuzu göstərmək üçün yeniləndi. Yeniləməni görmək istəyirsiniz?</string>
   <string name="contributions_dashboard_donation_dialog_cancel">Xeyr, təşəkkürlər</string>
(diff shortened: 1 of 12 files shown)


dbrant:
@cooltey   OK, I think I understand the confusion with this:
* If a string already has other formatting symbols (e.g. `%s` or others), then a literal percent symbol must be double-encoded `%%`.
* BUT, if a string _doesn't have any formatting symbols_, then a literal percent does not need to be encoded.

Therefore, I believe this particular string can just have a regular percent sign `%` safely.

cooltey:
> @cooltey OK, I think I understand the confusion with this:
> 
>     * If a string already has other formatting symbols (e.g. `%s` or others), then a literal percent symbol must be double-encoded `%%`.
> 
>     * BUT, if a string _doesn't have any formatting symbols_, then a literal percent does not need to be encoded.
> 
> 
> Therefore, I believe this particular string can just have a regular percent sign `%` safely.

Thanks for checking this! @dbrant 🤯 

I changed it to use a single percent sign and added an extra space after the percent sign.

cooltey:
@dbrant somehow the test failed..., the translation looks fine to me, not sure why it failed.

I think we can do either:
1. Use the Unicode character.
2. Replace the `2%` with `s%` and then pass `2%` string in the `getString(R.string.contributions_dashboard_donation_dialog_message, "2%")` 😅

dbrant:
@cooltey  After **_much_** more painstaking 😆 research, I have an updated recommendation:

* The literal percent symbol must always be encoded as a double percent `%%`.
* We should _not_ use other entities like the Unicode character, or the XML entity, since this will add to confusion for the translators.
* For strings that don't have any formatting symbols, but have a percent symbol, our code must use `String.format( getString( ... ) )` for that string.

So, to fix the issue of the AlertDialog message, you can just use an explicit `String.format(getString(R.string.contributions_dashboard_donation_dialog_message))`.  It's more verbose, but it seems to be a quirk that we will need to accept.  Does this sound reasonable?

And in this PR, I expanded the regex that catches some more incorrect percent encodings (which actually caught an incorrect string in the `uz` language!)

cooltey:
> @cooltey After **_much_** more painstaking 😆 research, I have an updated recommendation:
> 
>     * The literal percent symbol must always be encoded as a double percent `%%`.
> 
>     * We should _not_ use other entities like the Unicode character, or the XML entity, since this will add to confusion for the translators.
> 
>     * For strings that don't have any formatting symbols, but have a percent symbol, our code must use `String.format( getString( ... ) )` for that string.
> 
> 
> So, to fix the issue of the AlertDialog message, you can just use an explicit `String.format(getString(R.string.contributions_dashboard_donation_dialog_message))`. It's more verbose, but it seems to be a quirk that we will need to accept. Does this sound reasonable?
> 
> And in this PR, I expanded the regex that catches some more incorrect percent encodings (which actually caught an incorrect string in the `uz` language!)

Thanks for updating the PR! Using the `String.format()` makes sense to me! I was thinking about that too, but thought it should always be used within another `%s` 😓 

I'll update the Contributions Dashboard feature branch later, thanks!

PR #5129, opened 2024-11-15
Use a custom SwipeRefreshLayout for the style consistency

cooltey:
### What does this do?
- Rename `SwipeRefreshLayoutWithScroll` to `WikiSwipeRefreshLayout`, and make the `canChildScrollUp()` to not always return `false` if the child view is empty.
- Replace `androidx.swiperefreshlayout.widget.SwipeRefreshLayout` with `org.wikipedia.views.WikiSwipeRefreshLayout`
- Remove `setColorSchemeResources(ResourceUtil.getThemedAttributeId(requireContext(), R.attr.progressive_color))` and add it in the custom view.

### Why is this needed?
We have been using the `SwipeRefreshLayout` for a long time, and sometimes we forget to set `setColorSchemeResources`. This PR helps to make the style consistent.

For example, in the `FilePageActivity` or `NotificationsActivity`, the spinner icon is in `black` color instead of `progressive_color`.


dbrant:
Nice -- It's kind of a shame that the standard `SwipeRefreshLayout` doesn't seem to let us style it using a `styles.xml` attribute... but this is the next best thing.

PR #5143, opened 2024-11-21
Add <small> tag to support <sub> and <sup> HTML tags in TextView

cooltey:
### What does this do?
Even though we support the HTML tags in the `TextView`, the `<sup>` and `<sub>` still do not fit in the view.

### Why is this needed?
This adds three `<small>` tags to make the text smaller.


cooltey:
> This is a good observation, but I'm not sure the solution here is... "satisfying".
> 
> The TextView is clearly _trying_ to render superscripts and subscripts correctly; It would be a shame to throw that away.
> 
> I think we're capable of finding a simple solution that actually allows real superscripts and subscripts, and also accounts for proper vertical sizing.

I see. I moved the logic to the `handleTag` function, but the text is still cut off at the bottom. Do you have any better solution for this?

dbrant:
Muuuch better!
I have a couple of ideas for the future, if we want to make this a bit more efficient: (because right now this will generate three nested `RelativeSizeSpans` which is ok, but not ideal.)

* Create our own custom Span, which decreases the font size, **and** moves it slightly higher or lower, while keeping it within bounds. (And use this Span in our custom html tag processor.)
* It's possible to call `TextView.setText(text, TextView.BufferType.SPANNABLE)`, which _will_ actually adjust the height of the TextView to account for subscripts and superscripts! The downside is that we would need to change all of our calls to `setText()` to include `BufferType.SPANNABLE`, or subclass `TextView` and use it in all our layouts.

PR #5145, opened 2024-11-21
Espresso main

Williamrai:
### What does this do?
Main branch for espresso tests

Has an error log which shows exactly where the error occurred. This will help to easily find the error instead of searching for it on the espresso run log which sometimes is difficult to find. 
- Go to LogCat and filter as "EspressoError"
[Preview](https://github.com/user-attachments/assets/2cf4207a-931a-402b-ac11-29f535a5dacc)


app/src/androidTest/java/org/wikipedia/EspressoLogger.kt
```diff
@@ -0,0 +1,13 @@
+package org.wikipedia
+
+import android.util.Log
+
+class EspressoLogger {
```
dbrant, line 5: Can this just be an `object`, then? (instead of a class with companion object)

app/src/androidTest/java/org/wikipedia/TestLogRule.kt
```diff
@@ -0,0 +1,35 @@
+package org.wikipedia
+
+import android.util.Log
+import org.junit.rules.TestRule
+import org.junit.runner.Description
+import org.junit.runners.model.Statement
+
+class TestLogRule : TestRule {
+    override fun apply(base: Statement?, description: Description?): Statement {
```
dbrant, line 9: It looks like the `base` and `description` parameters don't need to be nullable here.

app/src/androidTest/java/org/wikipedia/TestLogRule.kt
```diff
@@ -0,0 +1,35 @@
...
+        return object : Statement() {
+            override fun evaluate() {
+                try {
+                    base?.evaluate()
+                } catch (t: Throwable) {
+                    Log.e("orange", "test rule")
```
dbrant, line 15: What's "orange"? Do we need this log line at all?

app/src/androidTest/java/org/wikipedia/base/ColorAssertions.kt
```diff
@@ -0,0 +1,57 @@
...
+object ColorAssertions {
+    sealed class ColorType {
+        data object TextColor : ColorType()
+        data object BackgroundColor : ColorType()
+        data object Tint : ColorType()
+        data object ShapeableImageViewColor : ColorType()
+    }
+
+    fun hasColor(
+        colorResOrAttr: Int,
```
dbrant, line 23: An overall note on color assertions:
I wonder if, instead of asserting the `R.attr` value of the expected color, we should assert the actual `R.color` expected value?
If we assert the `R.attr` value, that's basically a tautology. It's like saying "is the color I'm expecting to be Black actually expected to be Black "?
Whereas, what we should be asserting is "is the color I'm expecting to be Black **actually Black**?"

dbrant (changes requested):
Some light initial review!

app/src/androidTest/java/org/wikipedia/EspressoLogger.kt
Williamrai: done.

app/src/androidTest/java/org/wikipedia/TestLogRule.kt
Williamrai: removed.

app/src/androidTest/java/org/wikipedia/TestLogRule.kt
Williamrai: when i was testing the log i usually label it as orange, so that i can filter easily in log. Forgot to remove this. Its removed.

app/src/androidTest/java/org/wikipedia/base/ColorAssertions.kt
Williamrai: fixed.

PR #5148, opened 2024-11-21
Espresso Suggested Edit Test

Williamrai:
### What does this do?
Adds espresso test for suggested edit screen. 

In order for login to work, you would need to set the username and password on the environment variable. You would need to set this variable with the username and password.  
 TEST_LOGIN_USERNAME = ""
TEST_LOGIN_PASSWORD = "

**Phabricator:**
https://phabricator.wikimedia.org/T378655


cooltey (changes requested):
The test stops at the login screen. The username and password are incorrect.

Williamrai:
> The test stops at the login screen. The username and password are incorrect.

I forgot to mention this, we need to set the username and password on the env variable locally. We have code in the build.gradle where it would take value from the config file. 

`def TEST_LOGIN_USERNAME = System.getenv('TEST_LOGIN_USERNAME')
        def TEST_LOGIN_PASSWORD = System.getenv('TEST_LOGIN_PASSWORD')
        buildConfigField "String", "TEST_LOGIN_USERNAME", TEST_LOGIN_USERNAME != null ? "\"${TEST_LOGIN_USERNAME}\"" : '"Foo"'
        buildConfigField "String", "TEST_LOGIN_PASSWORD", TEST_LOGIN_PASSWORD != null ? "\"${TEST_LOGIN_PASSWORD}\"" : '"Bar"'`

PR #5149, opened 2024-11-23
Add fitsSystemWindows=true to activities and remove it from fragments

cooltey:
### What does this do?
Add `fitsSystemWindows=true` for all activities' XML, and remove the duplicated parameter from fragments.

**Phabricator:**
https://phabricator.wikimedia.org/T380210


cooltey:
@dbrant This is ready for review.

app/src/main/java/org/wikipedia/activity/SingleFragmentActivity.kt
```diff
@@ -29,4 +30,8 @@ abstract class SingleFragmentActivity<T : Fragment> : BaseActivity() {
     protected open fun inflateAndSetContentView() {
         setContentView(R.layout.activity_single_fragment)
     }
+
+    fun disableFitsSystemWindows() {
+        findViewById<FrameLayout>(R.id.fragment_container).fitsSystemWindows = false
```
dbrant, line 35: 👌

PR #5160, opened 2024-12-04
[Experiment] Use WebView in ReferenceDialog

cooltey:
### What does this do?
This PR replaces the `TextView` with a `WebView`, which can handle the image or the special tag for the math formula. 

**Phabricator:**
https://phabricator.wikimedia.org/T348095


dbrant:
A couple of random thoughts:
* Let's just turn the whole app into a WebView, then! Why does any of it need to be native?
* Slightly more seriously, wouldn't it be _much_ more efficient to have an in-WebView popup (a `div` inside the page) containing the reference? i.e. borrow directly from Mobile Web?

dbrant:

PR #5161, opened 2024-12-04
Fix: do not execute loadFromBackStack() twice when calling openInNewForegroundTab

cooltey:
### What does this do?
When we call `openInNewForegroundTab()`, it will call `loadFromBackStack()` two times in a certain case. This PR removes it and also adds a safe case in `PageFragmentLoadState` to set up the loading status.

**Phabricator:**
https://phabricator.wikimedia.org/T381534

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
+++ b/app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
@@ -38,6 +38,8 @@ class PageFragmentLoadState(private var model: PageViewModel,
                             private var leadImagesHandler: LeadImagesHandler,
                             private var currentTab: Tab) {
 
+    private var isLoading = false
+
     fun load(pushBackStack: Boolean) {
         if (pushBackStack && model.title != null && model.curEntry != null) {
             // update the topmost entry in the backstack, before we start overwriting things.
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
```diff
@@ -38,6 +38,8 @@ class PageFragmentLoadState(private var model: PageViewModel,
                             private var leadImagesHandler: LeadImagesHandler,
                             private var currentTab: Tab) {
 
+    private var isLoading = false
```
dbrant, line 41: I'm afraid this also adds bugs, including (but not limited to) inability to recover if an exception is thrown while loading a page.
Are we sure this is necessary?

dbrant (changes requested):

cooltey:
> If the goal is to fix the task linked in the description, recall that the "read more" items are loaded lazily when scrolling to the bottom of the page. I would recommend looking at that logic, and avoiding touching anything related to loading the actual page.

Got it. I have updated this PR, although it fixes the issue of showing duplicated items in the section, now it does not jump back to the previous position where the user entered the edit history or talk page. 🤔

PR #5172, opened 2024-12-09
Fix another potential crash throughout usages of PagingDataAdapter

dbrant:
Another minor potential crash, which happens when you:
* Go to the Edit History of any article.
* Tap the "search" bar to filter the edit history.
* Type a character and quickly backspace to remove the character. (you may need to do this more than once)

There appears to be a nasty race condition somewhere deep in the implementation of `PagingDataAdapter` that causes it to throw an exception when `submitData()` is called multiple times quickly. This PR introduces a helper class that wraps around `PagingDataAdapter` and ensures that `submitData()` is fully finished before attempting to call it again.

This also removes an incorrect "reload" mechanism that we were using for refreshing the PagingData, which was probably an early attempt at fixing this same issue. The proper way to reload the data is to call `refresh()` on the adapter itself, which simplifies and reduces code.


Williamrai:
I tested this on both emulator and real device and i did notice the same crash index out of bounds. 
[Preview](https://github.com/user-attachments/assets/141853d7-5c95-49ba-8b7a-8cc88f7f2983)

dbrant:
> I tested this on both emulator and real device and i did notice the same crash index out of bounds. [Preview](https://github.com/user-attachments/assets/141853d7-5c95-49ba-8b7a-8cc88f7f2983)

@Williamrai  Try to break it now!

Williamrai:
> > I tested this on both emulator and real device and i did notice the same crash index out of bounds. [Preview](https://github.com/user-attachments/assets/141853d7-5c95-49ba-8b7a-8cc88f7f2983)
> 
> @Williamrai Try to break it now!

tried my best, could not break it. 😅

PR #5179, opened 2024-12-13
Year-in-review Feed card.

dbrant:
https://phabricator.wikimedia.org/T381802


Williamrai (commented):
Everything looks but when i turn on the Year-in-review Feed card from the settings preference I don't see it appear on the explore feed. The shouldShow function is returning true in the YIROnboardingCard.

dbrant:
> Everything looks good but when i turn on the Year-in-review Feed card from the settings preference I don't see it appear on the explore feed. The shouldShow function is returning true in the YIROnboardingCard.

@Williamrai Ah yes, this is a little tricky. If you dismiss this card, here is how to show it again:

* In Developer settings, turn on "Year-in-review Feed Card"
* Also in Developer settings, find the setting called "feedHiddenCards" (near the middle of the list) and blank it out.
* Exit the app entirely, then relaunch it to get back to the Feed, and you should see the card.

app/src/main/res/values/strings_no_translate.xml
```diff
@@ -88,4 +88,7 @@
     <string name="diff_link_wikitext">{{{diffLink}}}</string>
     <string name="wikiText_replace_url">[[%1$s|%2$s]]</string>
 
+    <string name="year_in_review_action_positive">Take me there</string>
+    <string name="year_in_review_text"><![CDATA[<strong>Wikipedia 2024 Year in Review</strong><br/><br/>Wikipedia is the largest knowledge resource ever assembled in the history of the world, and it’s freely available to everyone everywhere. Here’s a review of 2024 through the lens of Wikipedia.]]></string>
```
dbrant, line 92: (I made these strings non-translatable on purpose, since this feature will be available for English only, therefore translators don't need to be bothered.)

Williamrai (approved):

PR #5186, opened 2024-12-19
Espresso Deep Linking Test

Williamrai:
### What does this do?
Adds Deeplinking Test (test if provided links opens the same article)


app/src/main/java/org/wikipedia/main/MainActivity.kt
```diff
@@ -159,18 +156,18 @@ class MainActivity : SingleFragmentActivity<MainFragment>(), MainFragment.Callba
...
+//            intent.data?.let {
+//                if (it.authority.orEmpty().endsWith(WikiSite.BASE_DOMAIN)) {
+//                    // Pass it right along to PageActivity
+//                    val uri = Uri.parse(it.toString().replace("wikipedia://", WikiSite.DEFAULT_SCHEME + "://"))
+//                    startActivity(Intent(this, PageActivity::class.java)
+//                            .setAction(Intent.ACTION_VIEW)
+//                            .setData(uri))
+//                }
+//            }
+//        }
```
cooltey, line 170: Is this expected?

Williamrai: forgot to remove this. will undo this comment.

PR #5197, opened 2025-01-03
Bug - Animated images stop animating after viewing the image

Williamrai:
### What does this do?
Fixes gif not animating issue when navigating back from GalleryActivity to PageFragment 

### Why this bug occurred?
The bug occurred due to the share element animation between the image and the Gallery Activity. 

**Phabricator:**
https://phabricator.wikimedia.org/T380394


app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
```diff
@@ -235,6 +235,8 @@ class LeadImagesHandler(private val parentFragment: PageFragment,
         }
     }
 
+    fun isUrlAGif(): Boolean = leadImageUrl?.endsWith(".gif") == true
```
cooltey, line 238: Would it be better if we name it `isImageAGif`

app/src/main/java/org/wikipedia/page/PageFragment.kt
```diff
@@ -300,6 +300,9 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
         updateQuickActionsAndMenuOptions()
         articleInteractionEvent?.resume()
         metricsPlatformArticleEventToolbarInteraction.resume()
+        if (leadImagesHandler.isUrlAGif()) {
```
cooltey, line 303: Very minor, could you please move this logic above the analytics code?

cooltey (changes requested):
Some very minor comments:

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
dbrant, line 238: A function like this probably belongs in a Utility class like `FileUtil`, where we have similar types of functions. And a couple of unit tests might be good, too. 😉 
Also: File extensions are not guaranteed to be lowercase.

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
Williamrai: done

app/src/main/java/org/wikipedia/page/PageFragment.kt
Williamrai: done

app/src/main/java/org/wikipedia/page/leadimages/LeadImagesHandler.kt
Williamrai: done

app/src/main/java/org/wikipedia/util/ImageUrlUtil.kt
```diff
@@ -21,4 +23,13 @@ object ImageUrlUtil {
             original
         }
     }
+
+    fun isGif(url: String?): Boolean {
+        try {
+            URL(url)
```
dbrant, line 29: I assume this is intended to validate that the string is a "valid" URL?
I think this might be doing "too much" here. Can we just stick to checking that the string ends with `.gif`?

* It's not very good practice to rely on exceptions to check the validity of something.
* We don't use `java.net.URL` anywhere else in our code, therefore whether something is valid according to `java.net.URL` is not meaningful.
* In our ecosystem we deal with `Uri`s, and something like `google.com` or even `google` is absolutely a valid Uri, and doesn't really need to be checked here.

I realize this logic is not ideal; the ideal case would be to check the MIME type of the image being loaded. But without access to the MIME type, checking just the extension should be just fine.

dbrant (changes requested):

app/src/main/java/org/wikipedia/util/ImageUrlUtil.kt
Williamrai: thanks for noting this. Done

dbrant (approved):

PR #5198, opened 2025-01-03
Compose Convert the base color styles 

Williamrai:
### What does this do?
Convert the base color styles to Compose

### Why is this needed?
To make it easier to use our color system with compose.

**Phabricator:**
https://phabricator.wikimedia.org/T381828
https://phabricator.wikimedia.org/T381910

**Testing color base style**
https://github.com/wikimedia/apps-android-wikipedia/pull/5208


app/src/main/java/org/wikipedia/theme/WikipediaColor.kt
```diff
@@ -0,0 +1,116 @@
...
+    secondaryColor = ThemeColors.Gray600,
+    primaryColor = ThemeColors.Gray700,
+    progressiveColor = ThemeColors.Blue600,
+    successColor = ThemeColors.Green700,
+    destructiveColor = ThemeColors.Red700,
+    warningColor = ThemeColors.Yellow700,
+    highlightColor = ThemeColors.Yellow500,
+    focusColor = ThemeColors.Orange500,
+    additionColor = ThemeColors.Blue300_15,
+    overlayColor = ThemeColors.Black_30
```
cooltey, line 61: Can we have the same order as the `WikipediaColor`?

app/src/main/java/org/wikipedia/theme/WikipediaColor.kt
```diff
@@ -0,0 +1,116 @@
...
+        destructiveColor = Color.Unspecified,
+        warningColor = Color.Unspecified,
+        highlightColor = Color.Unspecified,
+        focusColor = Color.Unspecified,
+        additionColor = Color.Unspecified,
+        overlayColor = Color.Unspecified,
+    )
+}
+
+val lightColors = WikipediaColor(
```
cooltey, line 46: Just curious: the naming here starts with a lowercase, is there anything special meaning for the colors?

cooltey:
Another idea: do we want to move all Compose-related classes to a `compose` package, for easier access and manage?

app/src/main/java/org/wikipedia/theme/WikipediaColor.kt
Williamrai: done.

app/src/main/java/org/wikipedia/theme/WikipediaColor.kt
Williamrai: fixed this, it was a function before converting to val that's why it had lowercase.

Williamrai:
> Another idea: do we want to move all Compose-related classes to a `compose` package, for easier access and manage?

yes, added a new package called compose

PR #5207, opened 2025-01-07
Use local language names instead of canonical names where appropriate.

dbrant:
Wherever we show a selection of languages, the names of the languages should be in the name of that language itself, instead of English.

**Phabricator:**
https://phabricator.wikimedia.org/T383150

### DIFF
--- a/app/src/main/java/org/wikipedia/language/LangLinksActivity.kt
+++ b/app/src/main/java/org/wikipedia/language/LangLinksActivity.kt
@@ -210,7 +210,7 @@ class LangLinksActivity : BaseActivity() {
             languageEntries.clear()
             for (entry in originalLanguageEntries) {
                 val languageCode = entry.wikiSite.languageCode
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/wikipedia/language/LangLinksActivity.kt
```diff
@@ -210,7 +210,7 @@ class LangLinksActivity : BaseActivity() {
             languageEntries.clear()
             for (entry in originalLanguageEntries) {
                 val languageCode = entry.wikiSite.languageCode
-                val canonicalName = app.languageState.getAppLanguageCanonicalName(languageCode).orEmpty()
+                val canonicalName = viewModel.getCanonicalName(languageCode) ?: app.languageState.getAppLanguageCanonicalName(languageCode).orEmpty()
```
dbrant, line 213: Here this also fixes an issue that existed silently: the language names were not searchable properly -- they were only searchable by the English name, instead of by the name given by the `siteinfo` call.

PR #5214, opened 2025-01-09
Refactor ReadingListFragment to use ViewModel.

terofeev:
### What does this do?

- Refactored `ReadingListFragment` to use ViewModel.
- Added `ReadingListFragmentViewModel` and moved the update list logic into it.

### Why is this needed?
Tech-debt. 
Reduce complexity by simplifying the UI layer.

**Phabricator:**
https://phabricator.wikimedia.org/T303935


app/src/main/java/org/wikipedia/readinglist/ReadingListFragmentViewModel.kt
```diff
@@ -0,0 +1,48 @@
...
+        }) {
+            val json = Prefs.suggestedReadingListsData
+            if (!json.isNullOrEmpty()) {
+                val list = ReadingListsReceiveHelper.receiveReadingLists(emptyTitle, emptyDescription, json, encoded)
+                _updateListFlow.emit(Resource.Success(list))
+            }
+        }
+    }
+
+    class ReadingListWrapper(val readingList: ReadingList?)
```
terofeev, line 47: Added wrapper class because the data has non-nullable type `T`.

```
open class Resource<T> {
    ...
    class Success<T>(val data: T) : Resource<T>()
    ...
}
```

cooltey: Instead of having a wrapper class, can we send an empty list if the list is `null`? or can you please handle the `null` list by sending an error?

cooltey (changes requested):

app/src/main/java/org/wikipedia/readinglist/ReadingListsReceiveHelper.kt
```diff
@@ -24,10 +24,18 @@ import java.util.*
 
 object ReadingListsReceiveHelper {
 
+    fun getEmptyTitle(context: Context) = context.getString(R.string.reading_lists_preview_header_title)
+
+    fun getEmptyDescription(context: Context) = DateUtil.getTimeAndDateString(context, Date())
+
     suspend fun receiveReadingLists(context: Context, json: String, encoded: Boolean): ReadingList {
```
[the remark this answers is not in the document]
terofeev: Removed the unused function but kept `getEmptyTitle` and `getEmptyDescription` so that `receiveReadingLists` can be used in ViewModels without requiring the Android Context.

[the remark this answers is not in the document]
cooltey: I see, it makes sense to me. Instead of creating separate single-use functions, can we define those strings directly as
`val` in the `updateReadingListData()`?

app/src/main/java/org/wikipedia/readinglist/ReadingListFragmentViewModel.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    fun updateListById(readingListId: Long) {
+         viewModelScope.launch(CoroutineExceptionHandler { _, throwable ->
+             viewModelScope.launch {
+                 _updateListByIdFlow.emit(Resource.Error(throwable))
+             }
+        }) {
+             val list = AppDatabase.instance.readingListDao().getListById(readingListId, true)
+             if (list == null) {
+                 _updateListByIdFlow.emit(Resource.Error(Throwable("No reading list found with id: $readingListId")))
```
cooltey, line 30: For better localization, let's move this string into `strings.xml`. And to ensure accurate translations, could you provide a clear description for it in `values-qq/strings.xml`

app/src/main/java/org/wikipedia/readinglist/ReadingListsReceiveHelper.kt
```diff
@@ -24,10 +24,14 @@ import java.util.*
 
 object ReadingListsReceiveHelper {
 
-    suspend fun receiveReadingLists(context: Context, json: String, encoded: Boolean): ReadingList {
+    fun getEmptyTitle(context: Context) = context.getString(R.string.reading_lists_preview_header_title)
+
+    fun getEmptyDescription(context: Context) = DateUtil.getTimeAndDateString(context, Date())
```
cooltey, line 29: Please see the comment in the previous thread.

cooltey (changes requested):

app/src/main/java/org/wikipedia/readinglist/ReadingListsReceiveHelper.kt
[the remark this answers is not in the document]
terofeev: Sure, I’ve moved these functions to `ReadingListFragment`

app/src/main/java/org/wikipedia/readinglist/ReadingListFragmentViewModel.kt
terofeev: Just to clarify, do we need this string? The activity finishes without showing a message, so it may not be necessary. Here's the error case:

```kotlin
is Resource.Error -> {
    // The list is no longer in the database (likely removed due to sync).
    // Nothing to do, so exit the activity.
    requireActivity().finish()
}
```

terofeev:
Hi @cooltey, could you review this PR when you get a chance? Thanks!

cooltey:
> Hi @cooltey, could you review this PR when you get a chance? Thanks!

Sorry missed the messages you've posted. I'll take a look by the end of this week, thanks for keeping this PR updated!

cooltey (approved):

app/src/main/java/org/wikipedia/readinglist/ReadingListFragmentViewModel.kt
cooltey: You are right, throwing an error message is much better :)

PR #5221, opened 2025-01-10
Bug offline article load issue

Williamrai:
### What does this do?
This provides a possible solution for bug fix for article offline load issue.  This PR also includes a test which you can run to test the implementation. If you want to check with previous implementation you can comment or uncomment the code. That is added only for testing purpose. 

**Phabricator:**
https://phabricator.wikimedia.org/T383262


Williamrai:
With this approach we will never be executing the registered callbacks whenever network state changes so i think there should be another way to resolve this issue.

PR #5230, opened 2025-01-14
WikiGames (feature branch)

dbrant:
https://phabricator.wikimedia.org/T383694


app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameViewModel.kt
```diff
@@ -0,0 +1,281 @@
...
+                // we're already done for today.
+                val totalHistory = Prefs.otdGameHistory.let { JsonUtil.decodeFromString<TotalGameHistory>(it)?.langToHistory }?.toMutableMap() ?: mutableMapOf()
+                val history = totalHistory[wikiSite.languageCode] ?: GameHistory()
+                _gameState.postValue(GameEnded(currentState, history))
+            } else if (currentState.currentQuestionState.month != currentMonth || currentState.currentQuestionState.day != currentDay &&
+                currentState.currentQuestionIndex >= currentState.totalQuestions) {
+                // we're coming back from a previous day's completed game, so start a new day's game.
+                currentState = currentState.copy(currentQuestionState = composeQuestionState(0), currentQuestionIndex = 0, answerState = List(MAX_QUESTIONS) { false })
+                _gameState.postValue(GameStarted(currentState))
+            } else {
```
Williamrai, line 130: Currently if a user is playing the previous day's game it does not reset. Should we allow them to complete it or do you think we should reset it as well? I think this is a very rare edge case.

cooltey: We should just reset it since we do not allow the user to play the previous games. @dbrant

dbrant: will do in a follow-up!

PR #5243, opened 2025-01-18
Add Wikipedia and Template namespace prompt for mobile search menu

bunnypranav:
### What does this do?
In the namespace list just above the keyboard in the mobile app search menu, `Wikipedia:` and `Template:` should be added, in addition to the `User:` `Portal:` `Help:`.

### Why is this needed?
**Wikipedia** and **Template** namespaces are highly used and visited; they would be really beneficial as an autofill entry suggestion.

**Phabricator:**
https://phabricator.wikimedia.org/TT383797


dbrant:
Thanks for the contribution!
We will need to discuss with our product managers and/or designers whether this change is necessary, and in what order to put any new namespaces.
I will bring the Phabricator task that you created to our team's next prioritization meeting.

bunnypranav:
@dbrant Any update? Thanks!

bunnypranav:
@dbrant @cooltey It's been 4 months since I created this, are you interested in reviewing this?

dbrant:
> @dbrant @cooltey It's been 4 months since I created this, are you interested in reviewing this?

Terribly sorry for the delay -- This is now OK to merge, but let's keep the original order of the namespace hints, while including the new ones at the end.

bunnypranav:
@dbrant Is it possible for you to tell when this change can be seen in the app? I'm not familiar with the deployment schedule for the mobile apps, so asking. Thanks!

PR #5246, opened 2025-01-21
Compose About Activity conversion

Williamrai:
### What does this do?
Converts About Activity to Compose

**Phabricator:**
https://phabricator.wikimedia.org/T384362

This also resolves the license screen issue. 
https://phabricator.wikimedia.org/T384526


app/src/main/java/org/wikipedia/compose/components/LicenseLinkText.kt
```diff
@@ -0,0 +1,66 @@
...
+        links.forEach { linkTextData ->
+            withLink(
+                link = LinkAnnotation.Url(
+                    url = linkTextData.url ?: "",
+                    styles = TextLinkStyles(
+                        style = SpanStyle(color = WikipediaTheme.colors.progressiveColor)
+                    ),
+                    linkInteractionListener = {
+                        if (linkTextData.asset != null) {
+                            println("orange: content ${linkTextData.text}")
```
cooltey, line 43: Please remove test related `println()` code.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
```diff
@@ -1,54 +1,418 @@
...
+                        modifier = Modifier
+                            .padding(horizontal = 16.dp),
+                        message = it.visuals.message
+                    )
+                }
+            )
+        },
+        topBar = {
+            WikiTopAppBar(
+                title = "About",
```
cooltey, line 144: Please use the string reference.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
```diff
@@ -1,54 +1,418 @@
...
+                    Snackbar(
+                        modifier = Modifier
+                            .padding(horizontal = 16.dp),
+                        message = it.visuals.message
+                    )
+                }
+            )
+        },
+        topBar = {
+            WikiTopAppBar(
```
cooltey, line 143: Something I have noticed is that it does not show the tooltip for the back button. Would it be possible to update the `WikiTopAppBar` for the back button (home button) for its content description for accessibility?

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
```diff
@@ -1,54 +1,418 @@
...
+               snackbarHostState = snackbarHostState,
+               scope = scope,
+               context = context
+           )
         }
+    )
+}
+
+@Composable
+fun AboutScreenContent(
```
cooltey, line 164: I like the way you separate the layout into different sections!

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
```diff
@@ -1,54 +1,418 @@
...
+                                onSecretCountClick(true)
+                            } else {
+                                Prefs.isShowDeveloperSettingsEnabled = true
+                                onSecretCountClick(false)
+                            }
+                        }
+                    },
+                ),
+            painter = painterResource(R.drawable.w_nav_mark),
+            contentDescription = null,
```
cooltey, line 273: Since this is a clickable icon, to improve accessibility, consider adding a new string for the content description of the globe logo.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
```diff
@@ -1,54 +1,418 @@
...
+
+    companion object {
+        const val SECRET_CLICK_LIMIT = 7
+    }
+}
 
-    private lateinit var binding: ActivityAboutBinding
+@Composable
+fun AboutWikipediaScreen(
+    modifier: Modifier = Modifier,
```
cooltey, line 120: It looks like an unused parameter, can we remove it?

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
```diff
@@ -1,54 +1,418 @@
...
+        LicenseLinkText(
+            links = credits,
+            textStyle = textStyle
+        )
+    }
+}
+
+@Preview
+@Composable
+private fun AboutWikipediaImagePreview() {
```
cooltey, line 407: Not sure but do we need to keep this one?

app/src/main/java/org/wikipedia/settings/LicenseActivity.kt
```diff
@@ -1,35 +1,83 @@
...
 
 class LicenseActivity : BaseActivity() {
-    private lateinit var binding: ActivityLicenseBinding
-
     public override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        binding = ActivityLicenseBinding.inflate(layoutInflater)
-        setContentView(binding.root)
-        setNavigationBarColor(getThemedColor(this, android.R.attr.windowBackground))
+        val asset = intent.getStringExtra(ASSET) ?: ""
```
cooltey, line 27: It is fine to use the "asset" string here.

cooltey (changes requested):

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
Williamrai: should this go in the strings.xml for translations or it can be just on code.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
Williamrai: should the content description go in the strings.xml for translations or it can be just on code.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
Williamrai: I think it is good practice to keep Preview because reviewers can easily see the preview without having to write it themselves.

app/src/main/java/org/wikipedia/settings/LicenseActivity.kt
Williamrai: I did not understand this. Are you suggesting to remove the null check?

app/src/main/java/org/wikipedia/compose/components/LicenseLinkText.kt
Williamrai: done.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
Williamrai: done.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
Williamrai: thank you.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
cooltey: It should always be put in the `strings.xml` for translations.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
cooltey: It should always be put in the `strings.xml` for translations.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
cooltey: Got it. It makes sense to me but also we can think about that do we need to keep the preview-only code in the project.

app/src/main/java/org/wikipedia/settings/LicenseActivity.kt
cooltey: Ah, I thought it was used only once. Nevermind 😅

dbrant (changes requested):
A couple of tiny things:

* Can the activity have a scrollbar when scrolling? Currently there's no visible scrollbar.
* The snackbar appears to be slightly too "low" (notice the shadow cut off at the bottom). Can it have a default bottom margin of ~8dp?
image

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
Williamrai: added a string on the xml.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
Williamrai: added string on the xml.

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
```diff
@@ -1,54 +1,417 @@
...
+    versionName: String,
+    credits: List<LinkTextData>,
+    snackbarHostState: SnackbarHostState,
+    scope: CoroutineScope,
+    context: Context
+) {
+    Column(
+        modifier = modifier
+            .verticalScroll(rememberScrollState())
+            .padding(horizontal = 16.dp),
```
cooltey, line 174: This makes the `AboutScreenBody` has `16` + `16` `dp` padding, can we change or remove either one of them?

cooltey (changes requested):

app/src/main/java/org/wikipedia/settings/AboutActivity.kt
Williamrai: removed.

cooltey (approved):

dbrant (approved):

dbrant:
For great justice

PR #5247, opened 2025-01-21
Convert animation resources to use Compose

cooltey:
### What does this do?
This PR combines the `pulse` and `tab_list_zoom_enter` resources into one `ScaleAnimation`. The `thumbnail_item` related animations can be converted by the time when we convert the `GalleryThumbnailScrollView` view to Compose.

**Phabricator:**
https://phabricator.wikimedia.org/T381833


app/src/main/java/org/wikipedia/compose/components/WikiAnimation.kt
Williamrai, line 1: I think it is more efficient if we can create this as a extension function of a Modifier instead of a composable function.  so that we can apply the animation to any component easily, something like this. 

```
fun Modifier.pulse(
    fromScale: Float = 1f,
    toScale: Float = 2.5f,
    durationMillis: Int = 1000,
    pivot: Float = 0.5f,
) : Modifier {
```
and we can apply this to any composable function on its modifier. 

```
OnThisDayDot(
modifier = Modifier.pulse()
)
```

cooltey: I think that's a great idea. But just curious about the `android:repeatCount="1"` and the `android:repeatMode="reverse"` in the original animation XML, are these something that can be done in a `Modifier`?

cooltey: @Williamrai This is updated. Created an extension for the pulse animation.

app/src/main/java/org/wikipedia/compose/extensions/Modifier.kt
Williamrai, line 1: This is really nice. 😍
Just one suggestion, what do you think about adding another animation that will have repeat count parameter?

cooltey: You are right! Totally missed the repeat count parameter. Since the `rememberInfiniteTransition` does not accept the `repeatable` AnimationSpec, I have to change it back to use the `LaunchedEffect`.

app/src/main/java/org/wikipedia/compose/extensions/Modifier.kt
```diff
@@ -0,0 +1,91 @@
...
+    pivot: Float = 0.5f,
+    repeatCount: Int = 1,
+    easing: Easing = FastOutSlowInEasing
+): Modifier = composed {
+    val scale = remember { mutableFloatStateOf(fromScale) }
+    val targetScale = rememberUpdatedState(toScale)
+
+    LaunchedEffect(repeatCount) {
+        repeat(repeatCount) {
+            animate(
```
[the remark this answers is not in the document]
cooltey: Sounds good to me. Done.

PR #5248, opened 2025-01-22
Fix: issue of showing the User_Talk metrics in Edit History

cooltey:
### What does this do?
This PR fixes the issue when requesting the `https://wikimedia.org/api/rest_v1/metrics` endpoint for a user talk page. It looks like an unsupported namespace, and we can check if the namespace is `Main` to avoid breaking the layout in the Edit History list.

### Steps to reproduce
1. Go to a user talk page.
2. Click on the Edit history

The following screenshots show the invisible icon was caused by the incompleted request in the view model.

### DIFF
--- a/app/src/main/java/org/wikipedia/page/edithistory/EditHistoryListViewModel.kt
+++ b/app/src/main/java/org/wikipedia/page/edithistory/EditHistoryListViewModel.kt
@@ -92,14 +92,20 @@ class EditHistoryListViewModel(savedStateHandle: SavedStateHandle) : ViewModel()
             val editCountsUserResponse = async { ServiceFactory.getCoreRest(pageTitle.wikiSite).getEditCount(pageTitle.prefixedText, EditCount.EDIT_TYPE_EDITORS) }
             val editCountsAnonResponse = async { ServiceFactory.getCoreRest(pageTitle.wikiSite).getEditCount(pageTitle.prefixedText, EditCount.EDIT_TYPE_ANONYMOUS) }
             val editCountsBotResponse = async { ServiceFactory.getCoreRest(pageTitle.wikiSite).getEditCount(pageTitle.prefixedText, EditCount.EDIT_TYPE_BOT) }
-            val articleMetricsResponse = async { ServiceFactory.getRest(WikiSite("wikimedia.org")).getArticleMetrics(pageTitle.wikiSite.authority(), pageTitle.prefixedText, lastYear, today) }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/page/edithistory/EditHistoryListViewModel.kt
```diff
@@ -92,14 +92,20 @@ class EditHistoryListViewModel(savedStateHandle: SavedStateHandle) : ViewModel()
             val editCountsUserResponse = async { ServiceFactory.getCoreRest(pageTitle.wikiSite).getEditCount(pageTitle.prefixedText, EditCount.EDIT_TYPE_EDITORS) }
             val editCountsAnonResponse = async { ServiceFactory.getCoreRest(pageTitle.wikiSite).getEditCount(pageTitle.prefixedText, EditCount.EDIT_TYPE_ANONYMOUS) }
             val editCountsBotResponse = async { ServiceFactory.getCoreRest(pageTitle.wikiSite).getEditCount(pageTitle.prefixedText, EditCount.EDIT_TYPE_BOT) }
-            val articleMetricsResponse = async { ServiceFactory.getRest(WikiSite("wikimedia.org")).getArticleMetrics(pageTitle.wikiSite.authority(), pageTitle.prefixedText, lastYear, today) }
+            val articleMetricsResponse = async {
+                try {
+                    ServiceFactory.getRest(WikiSite("wikimedia.org")).getArticleMetrics(pageTitle.wikiSite.authority(), pageTitle.prefixedText, lastYear, today)
+                } catch (_: Exception) {
```
dbrant, line 98: I feel like this is too blunt of a hammer for too small of a nut.
* It's generally not recommended to catch the base `Exception` anyway, and it's _really_ not recommended to catch `Exception` inside a coroutine, because then we risk catching `CancellationException` which is thrown when a coroutine is cancelled.
* If this API only works for Main-namespace articles, then can we just check that `pageTitle.namespace() == Namespace.MAIN)` here?

dbrant (changes requested):

app/src/main/java/org/wikipedia/page/edithistory/EditHistoryListViewModel.kt
cooltey: Thanks! It makes sense to me. Done.

dbrant (approved):

PR #5261, opened 2025-01-24
[Wikigames] game entry card 

Williamrai:
### What does this do?
Adds a game entry card on the Feed 

**Phabricator:**
https://phabricator.wikimedia.org/T383692


app/src/main/java/org/wikipedia/feed/FeedCoordinatorBase.kt
```diff
@@ -226,6 +227,9 @@ abstract class FeedCoordinatorBase(private val context: Context) {
     }
 
     private fun appendCard(card: Card) {
+        if (card is WikiGamesCard && age != 0) {
```
dbrant, line 230: You already have logic in `WikiGamesCardClient` to show the card only for `age == 0`, so the extra logic here is not necessary.

app/src/main/java/org/wikipedia/feed/FeedCoordinatorBase.kt
```diff
@@ -268,7 +272,7 @@ abstract class FeedCoordinatorBase(private val context: Context) {
     private fun isDailyCardType(card: Card): Boolean {
         return card is NewsCard || card is OnThisDayCard ||
                 card is TopReadListCard || card is FeaturedArticleCard ||
-                card is FeaturedImageCard
+                card is FeaturedImageCard || card is WikiGamesCard
```
dbrant, line 275: (This isn't necessary here, either)

dbrant (changes requested):
great -- just very minor stuff.

app/src/main/java/org/wikipedia/feed/FeedCoordinatorBase.kt
Williamrai: done.

PR #5288, opened 2025-02-06
Fix bottom sheet edge-to-edge issue in wikigame final screen

cooltey:
### What does this do?
This PR adds extra logic to adjust the padding of the bottomsheet to fit the Android 15 devices properly.


Williamrai:
Is this an expected behavior when we scroll the top bar changes the color to the bg color?
[Preview](https://github.com/user-attachments/assets/79f1ea9c-4acd-42f4-90e0-432036c86141)

cooltey:
> Is this an expected behavior when we scroll the top bar changes the color to the bg color? [Preview](https://github.com/user-attachments/assets/79f1ea9c-4acd-42f4-90e0-432036c86141)

I think it is not related to the bottomsheet but I can definitely look into this.

cooltey:
@Williamrai This is fixed. Thanks for sharing the feedback.

app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameActivity.kt
```diff
@@ -200,6 +207,31 @@ class OnThisDayGameActivity : BaseActivity() {
         finish()
     }
 
+    @SuppressLint("RestrictedApi")
+    private fun calculateBottomSheetTopPadding(insets: Int): Int {
```
cooltey, line 211: Still not super smooth when: dragging the bottomsheet and releasing at the middle to let it go back to the expanded state. Please share any ideas for optimization.

Williamrai: I override the onSlide function of the bottomSheet with this code and it did smooth scrolling. We also need to cache the value of statusBarHeight and toolbarHeight. We are updating the padding as the bottom sheet slides. the padding size will change between the toolbar height and the statusbar height. 
```
toolbarHeight = DimenUtil.getToolbarHeightPx(this)
 statusBarHeight = newStatusBarInsets.top
```
```
val topPadding = toolbarHeight + slideOffset * (statusBarHeight - toolbarHeight)
                bottomSheet.updatePadding(
                    top = topPadding.toInt()
                )
```

[API 35](https://github.com/user-attachments/assets/07806f49-98da-470f-af89-f91055337130)
[API 21](https://github.com/user-attachments/assets/5baf604a-a606-4dd5-91cf-7278f00a6f7b)

cooltey: Yes, it works! Thanks for providing the solution 😄

app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameActivity.kt
```diff
@@ -398,14 +435,29 @@ class OnThisDayGameActivity : BaseActivity() {
         }, 500)
     }
 
-    fun openArticleBottomSheet(pageSummary: PageSummary, callback: BottomSheetBehavior.BottomSheetCallback) {
+    fun openArticleBottomSheet(pageSummary: PageSummary, updateBookmark: () -> Unit) {
         if (viewModel.gameState.value !is OnThisDayGameViewModel.GameEnded) {
             return
         }
+        binding.root.requestApplyInsets()
```
dbrant, line 442: Calling `requestApplyInsets()` does not guarantee that insets will be applied. On some API versions, this will _not_ call the insetsListener, and your insets will remain `null`, causing a crash.

This is starting to feel really complex...

Are we sure we can't convince Design to make this component into a proper BottomSheetDialogFragment, and make it behave like a BottomSheetDialog is supposed to behave?
Something like this, where the content flows vertically, and naturally ends when it ends:

image

dbrant (changes requested):

PR #5292, opened 2025-02-08
Add margins to the talk templates to avoid text reaches the edge

cooltey:
### What does this do?
A minor tweak for the empty container.

### DIFF
--- a/app/src/main/res/layout/fragment_talk_templates.xml
+++ b/app/src/main/res/layout/fragment_talk_templates.xml
@@ -123,7 +123,7 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_gravity="center"
-            android:layout_margin="12dp"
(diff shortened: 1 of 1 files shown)


dbrant (changes requested):

app/src/main/res/layout/fragment_talk_templates.xml
```diff
@@ -122,6 +123,7 @@
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
             android:layout_gravity="center"
+            android:layout_margin="12dp"
```
dbrant, line 126: ...and here, too?

dbrant (changes requested):

app/src/main/res/layout/fragment_talk_templates.xml
cooltey: Done

dbrant (approved):

PR #5295, opened 2025-02-10
Follow up: handle close button properly for the wikigame

cooltey:
### What does this do?
Check the app to see if the `MainActivity` has been opened instead of exiting the app when clicking on the close button.


dbrant:
@cooltey  This is a bit weird-feeling navigation flow that seems to go against standard navigation guidelines. Is it really necessary to do it this way? (Was this requested by design?)

PR #5320, opened 2025-02-14
strings for about screen

Williamrai:
### What does this do?
 adds accessibility strings for about screen


app/src/main/res/values/strings.xml
```diff
@@ -1857,4 +1857,6 @@
     <string name="on_this_day_game_share_link_message">I\'m playing \"Which came first?\" a daily trivia game on the Wikipedia Android app %s</string>
     <string name="on_this_day_game_survey_q1">How satisfied were you with this game?</string>
     <string name="on_this_day_game_survey_q2">Would you be interested in playing different Wikipedia games?</string>
+    <string name="about_screen_back_button_accessibility_text">Navigate up</string>
```
dbrant, line 1860: It's a bit hard to believe that there isn't a system-provided "navigate up" string to use with the TopAppBar component...

But if we really need to do it ourselves, then can we make this a more generic-named string (e.g. `back_button_content_description`) that we'll be able to reuse for other screens?
Also, since we have an opportunity to take ownership of this string, let's provide something that makes more sense. I actually think that "navigate up" is super confusing, and most users don't know what it means. Can we say something like "Go back"?

app/src/main/res/values/strings.xml
```diff
@@ -1857,4 +1857,6 @@
...
+    <string name="about_screen_globe_puzzle_accessibility_text">Globe featuring puzzle logo</string>
```
dbrant, line 1861: This can say "Wikipedia puzzle globe logo"

dbrant (approved):

PR #5323, opened 2025-02-14
 refactors espresso base robot classes

Williamrai:
### What does this do?
refactors espresso base robot classes

### Why is this needed?
The `BaseRobot` class was becoming too large as it contained all the functions needed by various robot classes. As we will continue adding more functions, this will only grow bigger. So in order to manage this, I decided to separate the functions into different classes based on their actions:
1. ClickActions - Handles general click related functions
2. InputActions - Manages input actions like `typeTextInEditText`
3. ListActions - Contains all list related functions 
4. ScrollActions - Handles scroll events
5. SystemActions - Manage system related operations 
6. VerificationActions - Includes all verification functions 
7. WebActions - all web related operations 
Each of these will be initialized as protected variables within the `BaseRobot` class, allowing all child robot classes to access them.


Williamrai:
> Please provide a more detailed explanation of the refactoring performed and the rationale behind it.

I have added it, apologies for not adding the explanation earlier.

PR #5330, opened 2025-02-19
Bug Fix: Snackbar not appearing after bottom sheet dismissal

Williamrai:
### What does this do?
This resolves the issue of the Snackbar not appearing after the bottom sheet is dismissed. When a user adds an article to a new list, the Snackbar doesn't show because findBestView detects the bottom sheet as the anchor view, causing it to attach there and not display properly. This occurs on other similar cases as well. 

Added a boolean **ignoreBottomSheet** in the **findBestView** function to ignore finding the bottomSheet when not required.

[Bug Preview](https://github.com/user-attachments/assets/91c1f782-c0db-4fa7-84e0-64e21ac25fb1)

[Fix](https://github.com/user-attachments/assets/c8b0e8ad-d052-4a7f-a023-a15f0b56a354)


cooltey (changes requested):
If you go to Places, and click on "Watch" in the "more menu" in the article bottomsheet, and then click on "Unwatch", the "Unwatch" snackbar will not show up properly.

PR #5335, opened 2025-02-20
Fix: talk page title does not display properly when topic is missing

cooltey:
### What does this do?
Clicking a deleted notification resulted in the talk page title displaying incorrectly. This pull request resolves this by moving the `setToolbarTitle()` call to the end of the `updateOnSuccess()` method.

https://github.com/user-attachments/assets/6c2f4867-5f4c-4ad1-8fcb-03ca9e561cf0


app/src/main/java/org/wikipedia/talk/TalkTopicsActivity.kt
```diff
@@ -356,7 +354,7 @@ class TalkTopicsActivity : BaseActivity(), WatchlistExpiryDialog.Callback {
                 }
             }
             if (threadTopic != null) {
-                requestGoToTopic.launch(TalkTopicActivity.newIntent(this@TalkTopicsActivity, pageTitle, threadTopic!!.name, threadTopic!!.id, threadItem?.id, viewModel.currentSearchQuery, invokeSource))
+                requestGoToTopic.launch(TalkTopicActivity.newIntent(this@TalkTopicsActivity, pageTitle, threadTopic.name, threadTopic.id, threadItem?.id, viewModel.currentSearchQuery, invokeSource))
                 overridePendingTransition(0, 0)
                 return
             }
```
Williamrai, line 360: here we can use let this will remove that compiler warning
```
threadTopic?.let { topic ->
                requestGoToTopic.launch(TalkTopicActivity.newIntent(this@TalkTopicsActivity, pageTitle, topic.name, topic.id, threadItem?.id, viewModel.currentSearchQuery, invokeSource))
                overridePendingTransition(0, 0)
                return
            }
```

PR #5348, opened 2025-02-24
Fix wiki game uninitialized currentState

Williamrai:
### What does this do?
There were couple of crashes due to uninitialized lateinit variable currentState. Since the issue was difficult to reproduce consistently, I implemented a safeguard that checks whether the variable has been initialized. If it hasn't, the system now defaults to 'game_loading' as a fallback state.


app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameViewModel.kt
```diff
@@ -198,13 +198,11 @@ class OnThisDayGameViewModel(savedStateHandle: SavedStateHandle) : ViewModel() {
...
             "game_loading"
-        } else {
-            if (::currentState.isInitialized.not()) {
-                "game_loading"
-            } else
-                "game_play_" + (currentState.currentQuestionIndex + 1)
+        } else if (::currentState.isInitialized.not()) {
+            "game_loading"
+        } else
+            "game_play_" + (currentState.currentQuestionIndex + 1)
```
cooltey, line 204: Is it possible to combine `::currentState.isInitialized.not() || _gameState.value is Resource.Loading`?

```
fun getCurrentScreenName(): String {
      return if (_gameState.value is GameEnded) {
          "game_end"
      } else if (::currentState.isInitialized.not() || _gameState.value is Resource.Loading) {
          "game_loading"
      } else {
          "game_play_" + (currentState.currentQuestionIndex + 1)
      }
  }
```

cooltey:
It feels like we will need to handle this correctly:

These crashes relate to different code:
[OnThisDayGameFinalFragment](https://play.google.com/console/u/1/developers/6169333749249604352/app/4976363884102945010/vitals/crashes/36fd951a0e9eaa5a88e882bf7bb41922/details?days=28&versionCode=50522&isUserPerceived=true)
[OnThisDayGameActivity onBackPressed](https://play.google.com/console/u/1/developers/6169333749249604352/app/4976363884102945010/vitals/crashes/60c5d4a111e5f48cfdd770d2d995f37e/details?days=28&versionCode=50522&isUserPerceived=true)

I am not sure that we can always send "game_loading" if `currentState` is not initialized, especially for the `onBackPressed`.

cooltey:
I can reproduce the issue successfully: turn on the "Don't keep activities" in the system Developer options, and switch between the "app selection" screen and it will crash.


https://github.com/user-attachments/assets/18ebe427-dd5d-43ab-b3d3-f567357a03bf

app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameViewModel.kt
```diff
@@ -198,9 +198,10 @@ class OnThisDayGameViewModel(savedStateHandle: SavedStateHandle) : ViewModel() {
             "game_end"
         } else if (_gameState.value is Resource.Loading) {
             "game_loading"
-        } else {
+        } else if (::currentState.isInitialized.not()) {
+            "game_loading"
```
cooltey, line 202: After a discussion with Shay, please put "unknown" if the `_gameState` or `currentState` is `null`, for now.

PR #5350, opened 2025-02-25
Espresso reading list test

Williamrai:
### What does this do?
This adds reading list specific test. Per discussion with Anthony this will add
1. saved article should be in the "Saved" reading list
2. article saved to a custom list should be in the custom list in the reading list
3. can delete lists
4. unsaved article should not be in the "Saved" or "custom list" in the reading list

**Phabricator:**
https://phabricator.wikimedia.org/T378654


cooltey:
Please resolve the conflict.

PR #5352, opened 2025-02-26
Dismiss the bottomsheet onPaused to avoid possible crash

cooltey:
### What does this do?
A temporary solution for a crash when the "Don't keep activities" is on and the bottomsheet is opened, and a switch between the system app chooser will lead to a crash.


cooltey:
self merged for the beta release.

PR #5354, opened 2025-02-27
Wikigames: set initial currentState from Prefs

cooltey:
### What does this do?
This PR sets the current game state from the Prefs when the ViewModel is created.

Also moved the `impression` event to the end of `onGameEnded()` in the case of `gameState` being `null`.

https://phabricator.wikimedia.org/T387418


Williamrai:
> The changes looks good. I think with this change we can also remove this additional logic, since we are persisting the state.
> 
> ```
> else if (_gameState.value == null || ::currentState.isInitialized.not()) {
>             "unknown"
>         }
> ```

This has been resolved.

app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameViewModel.kt
```diff
@@ -236,6 +247,8 @@ class OnThisDayGameViewModel(savedStateHandle: SavedStateHandle) : ViewModel() {
         val langToState = totalState.langToState.toMutableMap()
         langToState[wikiSite.languageCode] = currentState
         Prefs.otdGameState = JsonUtil.encodeToString(TotalGameState(langToState)).orEmpty()
+        savedStateHandle[KEY_CURRENT_STATE] = JsonUtil.encodeToString(currentState)
```
dbrant, line 250: The entire game state gets saved to Prefs (on the previous line). Do we really need to duplicate it into savedStateHandle?

dbrant (changes requested):

cooltey:
The current approach seems perfectly fine, so close this PR.

PR #5366, opened 2025-03-03
Possibly (?) resolve obscure lateinit UninitializedProperty crashes.

dbrant:
This might actually fix the elusive crashes we've seen relating to uninitialized `lateinit` instance variables coming from `WikipediaApp$Companion.getInstance`.
There is a platform-specific bug related to launching the app in "restricted" mode during device backup operations. I found a Google issue that provides a clue for why this may be happening:
https://issuetracker.google.com/u/0/issues/160946170

See comments in the code for more details.

### DIFF
--- a/app/src/main/java/org/wikipedia/util/DeviceUtil.kt
+++ b/app/src/main/java/org/wikipedia/util/DeviceUtil.kt
@@ -7,6 +7,7 @@ import android.content.res.Resources
 import android.graphics.Color
 import android.net.ConnectivityManager
 import android.os.Build
+import android.os.Handler
 import android.view.KeyCharacterMap
 import android.view.KeyEvent
 import android.view.View
@@ -21,6 +22,7 @@ import androidx.core.view.WindowInsetsCompat
 import com.google.android.material.appbar.MaterialToolbar
 import org.wikipedia.R
 import org.wikipedia.WikipediaApp
+import kotlin.system.exitProcess
 
 object DeviceUtil {
     private inline val Window.insetsControllerCompat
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/util/DeviceUtil.kt
```diff
@@ -74,6 +76,28 @@ object DeviceUtil {
...
+     * instead uses a vanilla Application object, which will cause issues when other classes try
+     * to access static data from the WikipediaApp object.
+     *
+     * This is a workaround that explicitly terminates the app process if it's running in Restricted
+     * mode, to be used sparingly from places where this crash is most likely to occur.
+     */
+    fun assertAppContext(context: Context, terminateOnFail: Boolean = false): Boolean {
+        if (context.applicationContext !is WikipediaApp) {
+            if (terminateOnFail) {
+                Handler(context.mainLooper).post { exitProcess(0) }
```
cooltey, line 94: Since we're using `post`, should we implement error handling with a try-catch around the Handler?

dbrant: @cooltey What Exceptions would we be catching?
It's true that we're very careful when using `post` in an Activity, making sure it's not destroyed, but this `post` is not attached to any activity, and if the `exitProcess()` function throws an exception (?), I'd like to see it in the crash logs.

cooltey: Didn't see any crashes in testing and it makes sense to me.

PR #5370, opened 2025-03-03
[Feature branch] Year-in-review

voyagerfan:
### What does this do?
Adds Year-in-review entry point to the More menu of the Explore feed

### Why is this needed?
Supports task T387595 for EPIC 

**Phabricator:**
https://phabricator.wikimedia.org/T387034
https://phabricator.wikimedia.org/T387595

### DIFF
--- a/app/src/main/res/layout/view_main_drawer.xml
+++ b/app/src/main/res/layout/view_main_drawer.xml
@@ -201,6 +201,27 @@
 
             </LinearLayout>
 
+            <LinearLayout
+                android:id="@+id/main_drawer_yearInReview_container"
+                style="@style/SideDrawerMenuItem">
+
+                <ImageView
+                    android:layout_width="24dp"
+                    android:layout_height="24dp"
+                    android:contentDescription="@null"
+                    app:srcCompat="@drawable/baseline_event_repeat_24"
+                    app:tint="?attr/secondary_color" />
+
+                <TextView
+                    style="@android:style/TextAppearance"
+                    android:layout_width="0dp"
+                    android:layout_height="wrap_content"
+                    android:layout_marginStart="24dp"
+                    android:layout_weight="1"
+                    android:text="@string/year_in_review" />
+
+            </LinearLayout>
+
         </LinearLayout>
     </ScrollView>
 </FrameLayout>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1856,4 +1856,5 @@
     <string name="on_this_day_game_survey_q2">Would you be interested in playing different Wikipedia games?</string>
     <string name="back_button_content_description">Go back</string>
     <string name="about_screen_logo_accessibility_text">Wikipedia puzzle globe logo</string>
+    <string name="year_in_review">Year In Review</string>
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -397,6 +397,7 @@
     <string name="preference_title_collapse_tables">Collapse tables</string>
     <string name="preference_summary_collapse_tables">Automatically collapse tables in articles, such as infoboxes, references, and notes.</string>
     <string name="nav_item_donate">Donate</string>
+    <string name="year_in_review">Year In Review</string>
     <string name="error_voice_search_not_available">Sorry, voice recognition is not available.</string>
     <string name="location_service_disabled">Location services are disabled.</string>
     <string name="enable_location_service">Enable</string>
@@ -1856,5 +1857,4 @@
     <string name="on_this_day_game_survey_q2">Would you be interested in playing different Wikipedia games?</string>
     <string name="back_button_content_description">Go back</string>
     <string name="about_screen_logo_accessibility_text">Wikipedia puzzle globe logo</string>
-    <string name="year_in_review">Year In Review</string>
 </resources>
--- a/app/src/main/res/drawable/baseline_event_repeat_24.xml
+++ b/app/src/main/res/drawable/baseline_event_repeat_24.xml
@@ -1,0 +1,5 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android" android:height="24dp" android:tint="#000000" android:viewportHeight="24" android:viewportWidth="24" android:width="24dp">
+      
+    <path android:fillColor="@android:color/white" android:pathData="M21,12V6c0,-1.1 -0.9,-2 -2,-2h-1V2h-2v2H8V2H6v2H5C3.9,4 3,4.9 3,6v14c0,1.1 0.9,2 2,2h7v-2H5V10h14v2H21zM15.64,20c0.43,1.45 1.77,2.5 3.36,2.5c1.93,0 3.5,-1.57 3.5,-3.5s-1.57,-3.5 -3.5,-3.5c-0.95,0 -1.82,0.38 -2.45,1l1.45,0V18h-4v-4h1.5l0,1.43C16.4,14.55 17.64,14 19,14c2.76,0 5,2.24 5,5s-2.24,5 -5,5c-2.42,0 -4.44,-1.72 -4.9,-4L15.64,20z"/>
+    
+</vector>


app/src/main/res/values/strings.xml
```diff
@@ -397,6 +397,7 @@
     <string name="preference_title_collapse_tables">Collapse tables</string>
     <string name="preference_summary_collapse_tables">Automatically collapse tables in articles, such as infoboxes, references, and notes.</string>
     <string name="nav_item_donate">Donate</string>
+    <string name="year_in_review">Year In Review</string>
```
cooltey, line 400: Since the string item is used in the nav menu, it would make more sense to name it as `nav_item_year_in_review`.

app/src/main/res/values-qq/strings.xml
```diff
@@ -1774,4 +1775,5 @@
   <string name="on_this_day_game_survey_q2">The second question in a survey about the game feature.</string>
   <string name="back_button_content_description">Accessibility text informing users that this action naviagtes back.</string>
   <string name="about_screen_logo_accessibility_text">Accessibility text informing users that this is a logo of a wikipedia puzzle globe.</string>
+  <string name="year_in_review">Text label for the Year in Review Icon.</string>
```
cooltey, line 1778: Would it be possible to modify the description to something like: "Text label for the Year in review menu item that shows in the main navigation drawer"?

app/src/main/res/layout/view_main_drawer.xml
```diff
@@ -201,6 +201,27 @@
 
             </LinearLayout>
 
+            <LinearLayout
+                android:id="@+id/main_drawer_yearInReview_container"
```
cooltey, line 205: To ensure consistency, we must settle on either
`mainDrawerYearInReviewContainer` or `main_drawer_year_in_review_container` for the `id` in the XML. In this case, keeping lower cases with underscores is the right one. Thanks.

cooltey (changes requested):
Thanks for creating this feature branch! Just have some review comments:

app/src/main/res/values/strings.xml
voyagerfan: Thank you for the your feedback! I'll go ahead and update this per your suggestion.

app/src/main/res/values-qq/strings.xml
voyagerfan: Absolutely I can modify the description. I will work on this today. Thanks again!

app/src/main/res/layout/view_main_drawer.xml
voyagerfan: Thank you for the correction and I will make this change as well.

voyagerfan: I've completed the update

cooltey:
Please help to resolve the conflicts and make sure to update the branch frequently, thanks!

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -0,0 +1,373 @@
...
+) {
+    val scrollState = rememberScrollState()
+    val gifAspectRatio = 3f / 2f
+    Column(
+        verticalArrangement = Arrangement.Top,
+        modifier = Modifier
+            .padding(innerPadding)
+            .verticalScroll(scrollState)
+    ) {
+        Row(
```
Williamrai, line 307: Row may not be necessary here.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewEntryDialogScreen.kt
```diff
@@ -0,0 +1,107 @@
...
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.viewinterop.AndroidView
+import com.bumptech.glide.Glide
+import org.wikipedia.R
+import org.wikipedia.compose.theme.WikipediaTheme
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
```
Williamrai, line 35: @OptIn(ExperimentalMaterial3Api::class) can be removed.

app/src/main/java/org/wikipedia/yearinreview/TestScreen.kt
Williamrai, line 1: This class can be removed.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -0,0 +1,373 @@
...
+                                contentDescription = stringResource(R.string.year_in_review_navigate_right)
+                            )
+                        }
+                    }
+                }
+            }
+        )
+    }
+}
+
```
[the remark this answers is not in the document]
voyagerfan: Hey William, I was reluctant to go with the constraint layout but ended up using it because the children arent the same size and I couldn't get them to position properly. I might have tried using weight modifiers initially (I cant remember) but I'll give it another try.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -0,0 +1,373 @@
...
+) {
+    BottomAppBar(
+        containerColor = WikipediaTheme.colors.paperColor,
+        content = {
+            Row(
+                horizontalArrangement = Arrangement.SpaceBetween,
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .height(50.dp)
+                    .padding(start = 10.dp, end = 10.dp)
```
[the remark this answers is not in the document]
voyagerfan: You're right! Not sure why I did that but removed it and it looks good. Thanks for the tip

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Nice catch! it absolutely isn't necessary. Its removed now

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
```diff
@@ -0,0 +1,95 @@
...
+import androidx.compose.animation.ExitTransition
+import androidx.compose.runtime.rememberCoroutineScope
+import androidx.navigation.compose.NavHost
+import androidx.navigation.compose.composable
+import androidx.navigation.compose.rememberNavController
+import kotlinx.coroutines.launch
+import org.wikipedia.compose.theme.BaseTheme
+
+class YearInReviewActivity : ComponentActivity() {
+
```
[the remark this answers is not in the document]
voyagerfan: Understood. I've swapped out `ComponentActivity()` for `BaseActivity()`

app/src/main/java/org/wikipedia/yearinreview/TestScreen.kt
voyagerfan: Understood. I've removed it.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewEntryDialogScreen.kt
voyagerfan: Another good catch! its now removed.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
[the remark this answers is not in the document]
voyagerfan: Update: I was able to achieve the desired effect by wrapping the children in a `Box()` and pinning their position with `modifier.align()`. The position appears be stable in landscape mode as well. Let me know what you think.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -0,0 +1,373 @@
...
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun YearInReviewScreen(
+    customBottomBar: @Composable (PagerState) -> Unit,
+    screenContent: @Composable (PaddingValues, YearInReviewScreenData) -> Unit,
+    navController: NavHostController,
+    contentData: List<YearInReviewScreenData>,
+    totalPages: Int
+) {
```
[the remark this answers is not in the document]
voyagerfan: This definitely helps shave down the number of parameters for the screen! I've made the adjustment. Thanks for the tip!

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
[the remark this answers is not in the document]
voyagerfan: a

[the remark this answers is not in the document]
voyagerfan: Hi William,  I replaced the `Constraint Layout` with `Box()`. More details in the review thread.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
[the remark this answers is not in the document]
voyagerfan: I've made this change. More details in the review thread.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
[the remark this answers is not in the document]
voyagerfan: I like this idea and made the change. More details in the review thread.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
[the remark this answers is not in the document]
voyagerfan: You're right. This has been updated. More details in the review thread.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Right again! Its been removed. More details in the review thread.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewEntryDialogScreen.kt
voyagerfan: Nice catch! it is removed

app/src/main/java/org/wikipedia/yearinreview/TestScreen.kt
voyagerfan: I was thinking to keep this for some testing but I went ahead and removed it.

Williamrai:
Looks good now. Please remove the Compose ConstraintLayout library, as it's no longer needed.

voyagerfan:
@Williamrai The constraint layout dependency has been removed now now. Thanks for catching that.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
```diff
@@ -24,13 +37,24 @@ class YearInReviewActivity : ComponentActivity() {
...
-                val personalizedScreenList = listOf(
-                    readCountData, editCountData, readCountData,
-                    editCountData, readCountData
-                )
-                val getStartedList = listOf(getStartedData)
                 val coroutineScope = rememberCoroutineScope()
                 val navController = rememberNavController()
+                val canShowSurvey = viewModel.uiCanShowSurvey.collectAsState().value
+                var isSurveyVisible by remember { mutableStateOf(false) }
+
```
Williamrai, line 48: I think this could be simplified by using a simple variable in the ViewModel, which you can set to true like this:
```
if (pagerState.currentPage >= 1 && !canShowSurvey) {
     viewModel.didUserReadYirScreen = true
}
```
and I think you can do this as well 
 ```
val canShowSurvey by remember { mutableStateOf(Prefs.yirSurveyShown) }
var isSurveyVisible by remember { mutableStateOf(false) }
```

And it might be good to make sure the survey is turned off after it's shown, like here:
```
if (isSurveyVisible) {
 Prefs.yirSurveyShown = true
  YearInReviewSurvey(onSurveyButtonClick = {
      this@YearInReviewActivity.finish()
  })
}
```

File: app/src/main/java/org/wikipedia/yearinreview/YearInReviewSurvey.kt [hunk 11]
```diff
@@ -0,0 +1,267 @@
...
+                            buttonText = R.string.year_in_review_survey_cancel,
+                            onClick = {
+                                onSurveyButtonClick()
+                            }
+                        )
+                        Spacer(modifier = Modifier.width(8.dp))
+                        SurveyButton(
+                            buttonText = R.string.year_in_review_survey_submit,
+                            onClick = {
+                                PatrollerExperienceEvent.logAction(
```
Williamrai: Just a suggestion — I think it might be better to send this event from the activity itself, within the callback.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
voyagerfan: Hi @Williamrai I think this can work but I think we need to link `viewModel.didUserReadYirScreen` to `Prefs.yirSurveyShown`. Im a little concerned about having `canShowSurvey` solely dependent on `Prefs.yirSurveyShown` - this boolean needs to respond to both the pagerState AND `Prefs.yirSurveyShown`. 

not syntactically correct but...
(ex. val canShowSurvey by remember { mutableStateOf(Prefs.yirSurveyShown && pagerState.currentPage >= 1) }

I do agree with flipping  Prefs.yirSurveyShown to true once the survey is shown. @cooltey, for this prototype do we want to have the survey to deploy for each activity or just once? I can see the advantages of both and some potential disadvantages as well.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewSurvey.kt
voyagerfan: You bring up a good point and this has been some I've been thinking about from an architecture and consistency standpoint. Let me see what I can do! thanks for the suggestion.

voyagerfan: I've pushed the button click logic into the activity.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
Williamrai: https://github.com/wikimedia/apps-android-wikipedia/pull/5672

voyagerfan: merged [#5676](https://github.com/wikimedia/apps-android-wikipedia/pull/5672)

cooltey:
Hi @voyagerfan, please help resolve the errors. We have removed the custom typography.

voyagerfan:
> Hi @voyagerfan, please help resolve the errors. We have removed the custom typography.

Will do!

Williamrai (approved):

dbrant:
Merging, because:
* The only entry point into the feature is gated by a developer preference, defaulting to false for now, so very low-risk.
* Further tweaks and follow-ups can go directly to `main`.
* It looks damn good. Great work @voyagerfan.

PR #5377, opened 2025-03-05
[Compose] Convert WiktionaryDialog to use Compose

cooltey:
### What does this do?
Convert the `WiktionaryDialog ` to use Compose 

**Phabricator:**
https://phabricator.wikimedia.org/T386527


Williamrai (commented):
This looks good. Just a minor comment.  One small suggestion - maybe we could separate the compose functions from the fragment and move the string conversion logic to the viewModel. This way, we'd only pass the necessary information from viewModel to compose, keeping everything nicely separated. What do you think?

app/src/main/java/org/wikipedia/wiktionary/WiktionaryDialog.kt
```diff
@@ -30,89 +55,201 @@ class WiktionaryDialog : ExtendedBottomSheetDialogFragment() {
...
-        binding.dialogWiktionaryProgress.visibility = View.VISIBLE
-    }
+    @Composable
+    fun WiktionaryDialogContent(
+        title: String,
+        showNoDefinitions: Boolean = false,
+        showProgress: Boolean = false,
+        definitionsContent: @Composable () -> Unit
+    ) {
+        Column(
```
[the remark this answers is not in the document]
cooltey: I am kind of confused about this. If I remove the `definitionsContent: @composeable() -> Unit` parameter, how do I show the preview `WiktionaryDialogPreview` properly 🤔

cooltey:
> This looks good. Just a minor comment. One small suggestion - maybe we could separate the compose functions from the fragment and move the string conversion logic to the viewModel. This way, we'd only pass the necessary information from viewModel to compose, keeping everything nicely separated. What do you think?

Thanks and done.

app/src/main/java/org/wikipedia/wiktionary/WiktionaryDialogScreen.kt
```diff
@@ -0,0 +1,222 @@
...
+import org.wikipedia.dataclient.restbase.RbDefinition
+import org.wikipedia.util.Resource
+import org.wikipedia.util.StringUtil
+
+fun interface Callback {
+    fun onDialogLinkClick(url: String)
+}
+
+var dialogCallback: Callback? = null
+
```
Williamrai, line 44: In Jetpack Compose, it's generally best practice to have state flow down through the composable hierarchy while events flow back up. The current approach with a global callback variable steps outside this pattern and since compose recomposes this variable might behave in a different way. I think instead of a global callback we can create callback within the WiktionaryDialogScreen function and pass it down to the composable that needs it. Here it would be DefinitionWithExamples. 

Parent function 
```
@Composable
fun WiktionaryDialogScreen(
    viewModel: WiktionaryViewModel,
    onDialogLinkClick: (url: String) -> Unit,
)
```
Another parent 
```
@Composable
fun DefinitionList(
    usage: RbDefinition.Usage,
    onDialogLinkClick: (url: String) -> Unit,
) 
```

The function that requires it. and here we can pass the url back onDialogLinkClick(url)
```
@Composable
fun DefinitionWithExamples(
    definition: RbDefinition.Definition, count: Int,
    onDialogLinkClick: (url: String) -> Unit,
) 
```

cooltey: Got it, this was my original approach but changed to the callback 😅. Updated.

PR #5380, opened 2025-03-05
Compose top bar with search

Williamrai:
### What does this do?
This creates a component that can be reused where top app bar requires search functionality. This behaves similar to starting action mode as SearchActionModeCallback. This is especially required in LangLinksActivity and LanguagesListActivity.


app/src/main/java/org/wikipedia/compose/components/WikiTopAppBarWithSearch.kt
```diff
@@ -0,0 +1,120 @@
...
+                            if (searchQuery.isNotEmpty()) {
+                                IconButton(
+                                    onClick = {
+                                        onSearchQueryChange("")
+                                        currentState = TopAppBarState.NORMAL
+                                    },
+                                    content = {
+                                        Icon(
+                                            imageVector = Icons.Outlined.Clear,
+                                            contentDescription = null,
```
cooltey, line 88: Please make sure to put content descriptions for clickable icons.

app/src/main/java/org/wikipedia/compose/components/SearchTopAppBar.kt
```diff
@@ -0,0 +1,73 @@
...
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.focus.FocusRequester
+import androidx.compose.ui.focus.focusRequester
+import androidx.compose.ui.graphics.Color
+import org.wikipedia.compose.theme.WikipediaTheme
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun SearchTopAppBar(
```
cooltey, line 26: Found some issues:
1. The hint text (e.g. "Search for a language") is slightly towered to the top instead of aligning to the middle.
2. The cursor color should be in `progressColor`
3. The input text is too large, it should be `14sp`.
4. When clicking on the `X` button, it should clean up the text instead of exiting the action mode.

app/src/main/java/org/wikipedia/compose/components/WikiTopAppBarWithSearch.kt
```diff
@@ -0,0 +1,120 @@
...
+                        onNavigationClick = onBackButtonClick,
+                        actions = {
+                            IconButton(
+                                onClick = {
+                                    currentState = TopAppBarState.SEARCH
+                                },
+                                content = {
+                                    Icon(
+                                        imageVector = Icons.Outlined.Search,
+                                        contentDescription = null,
```
cooltey, line 109: Please make sure to put content descriptions for clickable icons.

cooltey (changes requested):

app/src/main/java/org/wikipedia/compose/components/WikiTopAppBarWithSearch.kt
Williamrai: added

app/src/main/java/org/wikipedia/compose/components/SearchTopAppBar.kt
Williamrai: done

app/src/main/java/org/wikipedia/compose/components/WikiTopAppBarWithSearch.kt
Williamrai: added

cooltey (approved):

PR #5384, opened 2025-03-06
Compose WikiErrorView

Williamrai:
### What does this do?
This converts the wiki error view to compose wiki error view. This is required for LangLinks Screen and other compose screen which requires error view. 

**Phabricator:**
https://phabricator.wikimedia.org/T386529

### DIFF
--- a/app/src/main/java/org/wikipedia/compose/components/error/ComposeWikiErrorView.kt
+++ b/app/src/main/java/org/wikipedia/compose/components/error/ComposeWikiErrorView.kt
@@ -1,0 +1,158 @@
+package org.wikipedia.compose.components.error
+
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.heightIn
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.layout.widthIn
+import androidx.compose.material3.Button
+import androidx.compose.material3.ButtonDefaults
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/compose/components/error/ComposeWikiErrorView.kt
```diff
@@ -0,0 +1,158 @@
...
+    ) {
+        Spacer(
+            modifier = Modifier
+                .height(16.dp)
+        )
+        Image(
+            modifier = Modifier
+                .size(72.dp),
+            painter = painterResource(errorType.icon),
+            colorFilter = ColorFilter.tint(color = ComposeColors.Gray500),
```
cooltey, line 84: I think we can update the `Gray500` to `WikipediaTheme.colors.placeholderColor` since this view is outdated.

app/src/main/java/org/wikipedia/compose/components/error/ComposeWikiErrorView.kt
```diff
@@ -0,0 +1,158 @@
...
+                            color = WikipediaTheme.colors.progressiveColor,
+                            fontSize = 14.sp
+                        )
+                    )
+                ),
+                style = TextStyle(
+                    textAlign = TextAlign.Center,
+                    fontSize = 16.sp,
+                    lineHeight = 19.2.sp,
+                    color = ComposeColors.Gray500
```
cooltey, line 107: Same

app/src/main/java/org/wikipedia/compose/components/error/ComposeWikiErrorView.kt
```diff
@@ -0,0 +1,158 @@
...
+        if (footerErrorMessage != null) {
+            Spacer(
+                modifier = Modifier
+                    .height(16.dp)
+            )
+            Text(
+                modifier = Modifier
+                    .padding(horizontal = 16.dp),
+                text = footerErrorMessage,
+                color = ComposeColors.Gray500,
```
cooltey, line 141: Same.

PR #5385, opened 2025-03-07
Create clickable navigation for the year in review entry point: T387804 

voyagerfan:
### What does this do?
This adds a functioning clickable listener that links the 'Year in Review' in MenuNavTabDialog to the YearInReviewActivty

### Why is this needed?
Supports EPIC T387034

**Phabricator:**
https://phabricator.wikimedia.org/T387804

### DIFF
--- a/app/src/main/java/org/wikipedia/main/MainFragment.kt
+++ b/app/src/main/java/org/wikipedia/main/MainFragment.kt
@@ -86,6 +86,7 @@ import org.wikipedia.util.TabUtil
 import org.wikipedia.views.NotificationButtonView
 import org.wikipedia.views.TabCountsView
 import org.wikipedia.watchlist.WatchlistActivity
+import org.wikipedia.yearinreview.YearInReviewActivity
 import java.io.File
 import java.util.concurrent.TimeUnit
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -471,6 +472,10 @@ class MainFragment : Fragment(), BackPressedHandler, MenuProvider, FeedFragment.
         (requireActivity() as? BaseActivity)?.launchDonateDialog(campaignId = campaignId)
     }
 
+    override fun yearInReviewClick() {
+        startActivity(YearInReviewActivity.newIntent(requireActivity(), AccountUtil.userName))
```
cooltey, line 476: We do not need to pass the username as a parameter since you can use the `AccountUtil.userName` directly in the `YearInReviewActivity`.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
```diff
@@ -25,4 +27,15 @@ class YearInReviewActivity : ComponentActivity() {
...
+
+    companion object {
+
+        const val INTENT_EXTRA_USER_NAME = "userName"
+
+        fun newIntent(context: Context, userName: String): Intent {
+
+            return Intent(context, YearInReviewActivity::class.java)
+                .putExtra(INTENT_EXTRA_USER_NAME, userName)
+        }
```
cooltey, line 39: Same as the comment above, we do not have to send the `username` as a parameter.

cooltey (changes requested):

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
voyagerfan: Hi Cooltey, I totally understand. I will the modifications here as well. Thanks for the input!

PR #5388, opened 2025-03-07
Refactor and simplify Watchlist messages.

dbrant:
Apologies, this is a kind of unexpected refactor based on a random rabbit hole I fell into this week, so no rush.
Our logic to add and remove articles from your Watchlist has a lot of repetitive code, which can be centralized and simplified.

But more importantly, the messages that we show after adding and removing from the Watchlist don't need to be duplicated locally in our own strings, and can just be retrieved from MediaWiki on demand, during the actual operation of adding/removing from the watchlist.


app/src/main/java/org/wikipedia/watchlist/WatchlistViewModel.kt
```diff
@@ -170,4 +179,63 @@ class WatchlistViewModel : ViewModel() {
...
+
+            val whichMessage = if (unwatch) {
+                if (isTalkPage == true) "removedwatchtext-talk" else "removedwatchtext"
+            } else {
+                if (expiry == WatchlistExpiry.NEVER) {
+                    if (isTalkPage == true) "addedwatchindefinitelytext-talk" else "addedwatchindefinitelytext"
+                } else {
+                    if (isTalkPage == true) "addedwatchexpirytext-talk" else "addedwatchexpirytext"
+                }
+            }
```
cooltey, line 202: Minor: can we have a logic below `whichMessage` for the `-talk` string and update the if statements a little bit to improve the readability?

```
if (isTalkPage == true) {
    whichMessage += "-talk"
}
```

cooltey (changes requested):
This is already great! 
1. I noticed the messages align with the account's language preference, not sure if it is possible to align either the article language or the system language?  


2. If you change the watchlist expiry to "non-permanent", the menu icon will not be changed to the "half star" icon.

app/src/main/java/org/wikipedia/watchlist/WatchlistViewModel.kt
dbrant: excellent

dbrant:
> 1. I noticed the messages align with the account's language preference, not sure if it is possible to align either the article language or the system language?

@cooltey The API call should now use the `appOrSystemLanguageCode` explicitly.

> 2. If you change the watchlist expiry to "non-permanent", the menu icon will not be changed to the "half star" icon.

Fixed!

cooltey (approved):

PR #5391, opened 2025-03-11
Prepare HistoryEntry and PageImage tables for updates.

dbrant:
This updates our database tables to accommodate future features (such as Year-in-Review), and generally improves the architecture of how we store article history and metadata in the database.

* The `HistoryEntry` table shall no longer hold the `description` or `timeSpent` fields, since we want to be able to have multiple `HistoryEntry` rows for each visitation of an article, whereas the `description` and `timeSpent` represent metadata that is specific to the article itself, instead of a single visit to the article.
* The `PageImage` table shall now hold all the other metadata columns related to an article, including the image thumbnail (the original purpose of this table) as well as `description`, `timeSpent`, and geographical `geoLat` and `geoLon`, if applicable.
* There is, however, one column being added to `HistoryEntry`, namely `prevId`, representing the database `id` of the previous HistoryEntry row from which the current HistoryEntry was clicked.

This actually fully takes care of https://phabricator.wikimedia.org/T388259, automatically saving geo coordinates of articles that are visited.
And it _halfway_ takes care of https://phabricator.wikimedia.org/T388065, by switching to recording a new HistoryEntry for every article visit, but not updating the History screen to show only the "latest" visit of each article (this will need to be done separately).

https://phabricator.wikimedia.org/T388259
https://phabricator.wikimedia.org/T388065


app/src/main/java/org/wikipedia/history/db/HistoryEntryDao.kt
```diff
@@ -32,17 +35,11 @@ interface HistoryEntryDao {
...
-    suspend fun upsertWithTimeSpent(entry: HistoryEntry, timeSpent: Int) {
-        val curEntry = findEntryBy(entry.authority, entry.lang, entry.apiTitle)
+    suspend fun upsertWithNewTitle(entry: HistoryEntry) {
+        val curEntry = findEntryBy(entry.authority, entry.lang, entry.apiTitle, entry.timestamp.time)
         if (curEntry != null) {
-            curEntry.timeSpentSec += timeSpent
-            curEntry.source = entry.source
-            curEntry.timestamp = entry.timestamp
-            curEntry.description = entry.description
+            curEntry.displayTitle = entry.displayTitle
```
Williamrai, line 41: What do you think about adding else { insertEntry(entry) } here? I believe this would allow us to remove the isRefresh check for inserting the article for the first time. Also, I think on the initial load of the article, it will insert and then upsert the same article again.

dbrant: Thanks - I looked again through the refreshing logic, and I do believe you are correct: we can just use the upsert() function in both cases. 👍  Just to reduce a bit of confusion, I added a few comments to be explicit about the logic.

Williamrai (approved):

app/src/main/java/org/wikipedia/database/AppDatabase.kt
```diff
@@ -184,10 +184,36 @@ abstract class AppDatabase : RoomDatabase() {
                 database.execSQL("ALTER TABLE HistoryEntry ADD COLUMN description TEXT NOT NULL DEFAULT ''")
             }
         }
+        val MIGRATION_26_27 = object : Migration(26, 27) {
+            override fun migrate(database: SupportSQLiteDatabase) {
+                // Rename the existing HistoryEntry table, which we're preserving for now (in case
+                // things go wrong with migrations in the field).
+                database.execSQL("ALTER TABLE HistoryEntry RENAME TO HistoryEntry_old")
```
cooltey, line 191: Not sure if we should do a `DROP HistoryEntry_old` before running `ALTER TABLE`?

I guess this only happens to us because we install old/new versions back and forth, but not sure if should we do a drop table before altering a table.

Screenshot_20250312_161339_Wikipedia Dev

cooltey: Yes it is. We can have some discussion tomorrow about this.

PR #5392, opened 2025-03-12
Math Images are now white (have full opacity) when image dimming is turned on.

expely:
Bug: T381210

### What does this do?
Math images inside of a wikipedia page are just the normal brightness instead of the dimmed amount when image dimming is selected.

### Why is this needed?
It was hard to see math images in dark mode with image dimming turned on because the math images weren't excluded from the dimming effect.

**Phabricator:**
https://phabricator.wikimedia.org/T381210

### DIFF
--- a/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
+++ b/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
@@ -161,6 +161,14 @@ object JavaScriptActionHandler {
                 "})"
     }
 
+    fun handleMathImageDimming(): String {
+        return "(function() {" +
+                "let style = document.createElement('style');" +
+                "style.innerHTML = 'img.mwe-math-fallback-image-inline { opacity: 1; } ';" +
+                "document.head.appendChild(style);" +
+                "})();"
+    }
+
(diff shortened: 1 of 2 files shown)


dbrant:
Thanks for working on this!
While this might technically solve the issue locally, I'm not sure this is the correct place for these kinds of workarounds. And in fact we'd like to work towards _reducing_ the amount of javascript and css that we inject from native code, instead of adding more.

I believe the proper place for such fixes is our [backend service](https://gerrit.wikimedia.org/g/mediawiki/services/mobileapps), which is what serves the javascript and css bundles consumed by our WebView, along with controlling the logic for image dimming. Among other things, fixing the issue at the service level would solve the display problem on both iOS and Android apps, instead of just Android.

This particular issue is actually [one](https://phabricator.wikimedia.org/T182128) [of](https://phabricator.wikimedia.org/T252904) [numerous](https://phabricator.wikimedia.org/T367485) [issues](https://phabricator.wikimedia.org/T369438) in the backend service relating to Dark mode, and there is an effort [underway](https://phabricator.wikimedia.org/T369435) to consolidate Dark mode styles between this service and the styles used by the general mobile web version of Wikipedia. This fix can probably be included as part of those larger-scale updates.

PR #5397, opened 2025-03-12
For Compose HtmlText, use our custom html TagHandler.

dbrant:
Let's not forget that when we make `Spannable` strings from html, we have a custom TagHandler that handles other bonus tags like `<code>` and `<sup>`/`<sub>` and so on. This must also be usable in Compose.

Compose provides an extension function called `AnnotatedString.fromHtml()`, but unfortunately it doesn't let us pass a custom TagHandler. This is a known feature request in Google AOSP:
https://issuetracker.google.com/issues/374066408

Until the above issue is resolved, we need to duplicate a bit of code from AOSP, so that we can make use of the same logic, except plug in our own TagHandler, or rather use the `Spanned.toAnnotatedString()` function, which is also extremely useful, but sadly `internal` to the package.

image

TODO in the future: our custom TagHandler also handles _images_ (!) to be shown as Spans in a TextView. Does Compose support this?


app/src/main/java/org/wikipedia/compose/extensions/Html.kt
```diff
@@ -0,0 +1,212 @@
...
+
+fun AnnotatedString.Companion.composeFromHtml(
+    htmlString: String,
+    linkStyles: TextLinkStyles?,
+    linkInteractionListener: LinkInteractionListener? = null
+): AnnotatedString {
+    val spanned = StringUtil.fromHtml(htmlString)
+    return spanned.toAnnotatedString(linkStyles, linkInteractionListener)
+}
+
```
dbrant, line 59: Most of this is copied over from the platform library. Move along...

app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
```diff
@@ -111,7 +111,6 @@ class CustomHtmlParser(private val handler: TagHandler) : TagHandler, ContentHan
         private var lastAClass = ""
         private var listItemCounts = Stack<Int>()
         private val listParents = mutableListOf<String>()
-        private val leadingMarginSize = DimenUtil.dpToPx(16f).toInt()
```
dbrant, line 114: This was breaking Previews in compose. Let's move this inline to where it's actually used.

cooltey:
This is great!

PR #5399, opened 2025-03-12
History Screen filter logic update

Williamrai:
### What does this do?
1. updates database query to show distinct entries for history entry 
2. saves the history database query to a list and adds in-memory filter logic for search query. Instead of hitting the database for every search query change, only query the database when you need to reload the entire list (on initial load or after item changes).
 

**Phabricator:**
https://phabricator.wikimedia.org/T388065


Williamrai:
> I checked out this change, and even did a little profiling, and my conclusion is:
> 
> The updated SQL query you made is perfectly good, and I think that's all that needs to change.
> 
> The extra in-memory caching logic might be premature optimization. When Android opens a sqlite database, it's _also_ cached in memory (for precisely these kinds of purposes), so I would argue for sticking with the simplest possible solution. (Unless you have observed a specific performance degradation?)
> 
> If you like, we could add some debouncing logic, e.g. 500ms, whenever the query changes, but that's probably it.

thank you for the review. Reverted the in-memory logic and added debounce.

dbrant (approved):

dbrant:
👌

PR #5414, opened 2025-03-19
[Bug] watchlist state synchronization between talkPage and PageFragment

Williamrai:
### What does this do?
Synchronizes the watchlist status when a user modifies their watchlist state on the talk page and returns to the PageFragment.

[Bug Preview](https://github.com/user-attachments/assets/09348eb7-0ef5-4c5b-aded-50e06b0b1a31)


cooltey (commented):
This fixes the issue when returning from the talk page, but would it be better if we can get the watchlist status when revisiting the page? e.g. refresh the watch status in `onResume`? cc @dbrant

Williamrai:
Closing this since the state be easily synced when user refresh the articles.

PR #5417, opened 2025-03-20
Fix: license links do not match the libraries

cooltey:
### What does this do?
Remove space from the `text` and append a space in the `buildAnnotatedString` but outside the `withLink` block since the space should not be a part of a clickable target.

**Phabricator:**
https://phabricator.wikimedia.org/T384362#10654307


Williamrai (approved):
Looks good. Thank you for looking into this.

PR #5419, opened 2025-03-21
Reduce ANRs: make all HistoryEntry DAO functions suspend.

dbrant:
The proper solution is to make the Room DAO functions `suspend`, which causes Room to _automatically_ put those functions into an IO dispatcher.


app/src/main/java/org/wikipedia/history/db/HistoryEntryWithImageDao.kt
```diff
@@ -20,14 +20,14 @@ interface HistoryEntryWithImageDao {
     // https://developer.android.com/topic/libraries/architecture/paging/v3-overview
     @Query("SELECT HistoryEntry.*, PageImage.imageName, PageImage.description, PageImage.geoLat, PageImage.geoLon, PageImage.timeSpentSec FROM HistoryEntry LEFT OUTER JOIN PageImage ON (HistoryEntry.namespace = PageImage.namespace AND HistoryEntry.apiTitle = PageImage.apiTitle AND HistoryEntry.lang = PageImage.lang) INNER JOIN(SELECT displayTitle, MAX(timestamp) as max_timestamp FROM HistoryEntry GROUP BY displayTitle) LatestEntries ON HistoryEntry.displayTitle = LatestEntries.displayTitle AND HistoryEntry.timestamp = LatestEntries.max_timestamp WHERE UPPER(HistoryEntry.displayTitle) LIKE UPPER(:term) ESCAPE '\\' ORDER BY timestamp DESC")
     @RewriteQueriesToDropUnusedColumns
-    fun findEntriesBySearchTerm(term: String): List<HistoryEntryWithImage>
+    suspend fun findEntriesBySearchTerm(term: String): List<HistoryEntryWithImage>
```
dbrant, line 23: If a Dao function is marked as `suspend`, then Room automatically moves it off the main thread, just like Retrofit automatically does with network calls.
Therefore there shouldn't be a reason to wrap any of these things in `Dispatchers.IO { }` manually inside our coroutines.

PR #5424, opened 2025-03-21
Introduce indexes to db tables, greatly improving performance.

dbrant:
This is a bit of a facepalm moment:
Our database tables of `HistoryEntry` and `PageImage` didn't have indexes set up! Adding indexes to tables (especially tables that will grow very large) is super important for performance when making queries or _doing migrations_.

I tested migrating a `HistoryEntry` table with 10000 entries. With our current db structure, the migration takes about **12 seconds**, which accounts for the ANRs that we have seen from a small number of users.
With this new structure with indexing, the migration (with 10000 entries) takes **100 milliseconds**.

Note that adding indexes will necessitate another db version bump (to 28), and I added two new migration paths, 27→28 and **26→28**, simply because version 27 is already in the wild (in our Beta version). But really the 26→28 migration will be the one that occurs in production, and this is the one that is now optimized for efficiency.


Williamrai:
I switched branch from the main to dbIndexes to test and then i got this this error view. 
[Image1](https://github.com/user-attachments/assets/61eca69a-b9f3-4436-81c7-11784f4e2ea7)

I did fresh install and added some history entry to have some data and then switched to this branch, then i saw this error view.
[Image2](https://github.com/user-attachments/assets/11fce2a4-e9e5-48a6-a993-4b889dcce7bf)

dbrant:
> I switched branch from the main to dbIndexes to test and then i got this this error view.
> I did fresh install and added some history entry to have some data and then switched to this branch, then i saw this error view.

Whoops, absolutely right. I removed an unnecessary index creation statement from the 27→28 migration, which should fix this issue.

app/src/main/java/org/wikipedia/history/HistoryEntry.kt
```diff
@@ -18,7 +19,9 @@ import java.util.Date
 @Serializable
 @Parcelize
 @TypeParceler<Date, DateParceler>()
-@Entity
+@Entity(
+    indices = [Index(value = ["lang", "namespace", "apiTitle"])]
```
cooltey, line 23: Not really related to this PR, but would it be possible to also add:
```
Index(value = ["authority", "lang", "apiTitle", "timestamp"])
```

for the `findEntryBy` method since this is the most used SQL query for every page when doing `createPageModel`?
```
suspend fun findEntryBy(authority: String, lang: String, apiTitle: String, timestamp: Long): HistoryEntry?
```

cooltey: That makes sense to me! Thanks.

cooltey (approved):
Looks good to me! Will let @Williamrai to merge it if it looks good to you too.

PR #5425, opened 2025-03-21
Improve performance of filtering History entries.

dbrant:
Another pretty significant performance improvement:
When there are thousands of History entries, the filtering logic in the History screen (and the initial display of the entries) can get pretty sluggish. This is partly because we perform a loop over all the entries, and generate a `getDateString()` for each entry, and compare those strings against each other (for the purpose of inserting "date" separators in the History list)
This updates the logic to use a logical `Calendar` structure, instead of `getDateString()`, to determine where to put the date separators.


app/src/main/java/org/wikipedia/history/db/HistoryEntryWithImageDao.kt
```diff
@@ -49,25 +51,28 @@ interface HistoryEntryWithImageDao {
...
+                val curDay = calendar[Calendar.YEAR] + calendar[Calendar.DAY_OF_YEAR]
+                if (curDay != prevDay) {
+                    list.add(getDateString(entry.timestamp))
+                }
+                prevDay = curDay
             }
-            list.add(toHistoryEntry(entries[i]))
+            list.add(toHistoryEntry(entry))
         }
         return list
```
Williamrai, line 77: I think we can combine this two logic prevDay == 0 || curDay != prevDay and in the if else we are doing similar logic. I think we can improve this further inside the forEach loop
```
            calendar.time = entry.timestamp
            val curDay = calendar[Calendar.YEAR] + calendar[Calendar.DAY_OF_YEAR]

            if (prevDay == 0 || curDay != prevDay) {
                list.add(getDateString(entry.timestamp))
            }
            prevDay = curDay
            list.add(toHistoryEntry(entry)) 
```

dbrant: Sure! Thanks.

PR #5427, opened 2025-03-24
follow up compose wiktionary dialog 

Williamrai:
### What does this do?
- fixes the scrollable issue with BottomSheetDialogFragment and composeView 
- Adds a data class to simplify the management of Compose UI state.

**Phabricator:**
https://phabricator.wikimedia.org/T386527

### DIFF
--- a/app/src/main/java/org/wikipedia/wiktionary/WiktionaryViewModel.kt
+++ b/app/src/main/java/org/wikipedia/wiktionary/WiktionaryViewModel.kt
@@ -6,26 +6,26 @@ import androidx.lifecycle.viewModelScope
 import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import org.wikipedia.Constants
 import org.wikipedia.dataclient.ServiceFactory
 import org.wikipedia.dataclient.WikiSite
 import org.wikipedia.dataclient.restbase.RbDefinition
 import org.wikipedia.page.PageTitle
-import org.wikipedia.util.Resource
 import org.wikipedia.util.StringUtil
 import org.wikipedia.util.log.L
 import java.util.Locale
 
 class WiktionaryViewModel(savedStateHandle: SavedStateHandle) : ViewModel() {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/wiktionary/WiktionaryViewModel.kt
```diff
@@ -76,3 +75,9 @@ class WiktionaryViewModel(savedStateHandle: SavedStateHandle) : ViewModel() {
         private const val GLOSSARY_OF_TERMS = ":Glossary"
     }
 }
+
+data class WiktionaryDialogState(
+    val isLoading: Boolean = true,
+    val list: List<RbDefinition.Usage> = emptyList(),
+    val error: Throwable? = null
+)
```
cooltey, line 83: Just curious, instead of creating a new data class for the data, is it a good practice to send the `uiState` to the `WiktionaryDialogContent` so we don't need extra parameters and data class?

Williamrai: We could pass the uiState directly and structure it like this and it will have less parameters:
```
@Composable
fun WiktionaryDialogContent(
    uiState: WiktionaryDialogState,
    onDialogLinkClick: (url: String) -> Unit
) {
    
}
```
However, I believe it’s better for the topMost composable (like WiktionaryDialogScreen) to hold the uiState as a whole, and then pass individual parameters to its child composables (like WiktionaryDialogContent). This way, the parent can provide the necessary values for the child, making it clearer what each child component requires. This approach also decouples the UI from relying strictly on uiState, allowing other state values to be passed as needed. By doing this, we avoid tightly coupling the components together. I learned this structure from google compose samples 
1. https://github.com/android/compose-samples/blob/main/Reply/app/src/main/java/com/example/reply/ui/ReplyHomeViewModel.kt

2. https://github.com/android/compose-samples/blob/main/JetNews/app/src/main/java/com/example/jetnews/ui/home/HomeViewModel.kt . On this one  if you explore how the uiState is passed down to HomeScreenWithList, you’ll notice that the deepest child defines only the parameters it requires.

I also think it depends on the project and the requirement so the approach we choose should be based on what works best for the specific use case.

cooltey: I see. That makes sense to me. Thanks for providing the detail information :)

PR #5432, opened 2025-03-26
Final, ultimate fixes for db migration.

dbrant:
* There is just one more extreme edge case that the db migration logic was failing to catch (causing a tiny number of crashes in Beta): The old HistoryEntry could in fact contain entries that have the same combination of [lang, namespace, apiTitle], dating back to a time when we *didn't* upsert the "latest" entry for that combination. This kind of case would violate the primary-key constraint on the PageImage table. (This update does not increase the migration time at all.)
* When filtering the items in the History screen, they need to be unique by title *and* language, instead of just by title. Currently, if you navigate to different language versions of the same article (with the same title), it appears as a single entry in History.

### DIFF
--- a/app/src/main/java/org/wikipedia/database/AppDatabase.kt
+++ b/app/src/main/java/org/wikipedia/database/AppDatabase.kt
@@ -259,7 +259,7 @@ abstract class AppDatabase : RoomDatabase() {
                 // a thumbnail), insert them and copy the other metadata.
                 database.execSQL("INSERT INTO PageImage (lang, namespace, apiTitle, description, timeSpentSec)" +
                         " SELECT lang, namespace, apiTitle, description, COALESCE(timeSpentSec, 0) as timeSpentSec FROM" +
-                        " (SELECT lang, namespace, apiTitle, description, COALESCE(timeSpentSec, 0) as timeSpentSec" +
+                        " (SELECT lang, namespace, apiTitle, description, MAX(COALESCE(timeSpentSec, 0)) as timeSpentSec" +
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/database/AppDatabase.kt
```diff
@@ -258,13 +258,13 @@ abstract class AppDatabase : RoomDatabase() {
...
-                        " HistoryEntry_old.description, COALESCE(HistoryEntry_old.timeSpentSec, 0)" +
-                        " FROM HistoryEntry_old" +
+                        " SELECT lang, namespace, apiTitle, description, COALESCE(timeSpentSec, 0) as timeSpentSec FROM" +
+                        " (SELECT lang, namespace, apiTitle, description, COALESCE(timeSpentSec, 0) as timeSpentSec" +
+                        "     FROM HistoryEntry_old GROUP BY lang, namespace, apiTitle) AS HistoryUnique" +
                         " WHERE NOT EXISTS (SELECT 1 FROM PageImage" +
-                        "    WHERE PageImage.lang = HistoryEntry_old.lang AND" +
-                        "          PageImage.namespace = HistoryEntry_old.namespace AND" +
-                        "          PageImage.apiTitle = HistoryEntry_old.apiTitle)")
+                        "     WHERE PageImage.lang = HistoryUnique.lang AND" +
```
Williamrai, line 265: Maybe we can add a MAX before **(COALESCE(timeSpentSec, 0))** so that it will always select the maximum time spent value for the unique entry.

dbrant: Sure!

PR #5436, opened 2025-03-26
Fix: Remove namespace from message displayed when watching a Talk page.

terofeev:
### What does this do?
Removes the redundant "Talk:" namespace prefix from the confirmation message displayed when a user adds a Talk page to their watchlist.

**Phabricator:**
https://phabricator.wikimedia.org/T376362

<details>
  <summary>Before and After screenshots</summary>

**Before**

Before

**After**

After

</details>


dbrant:
Thanks, but I'm not sure this is necessary or correct.
The current behavior is consistent with web:

image

When watching a Talk page, it is correct to refer to it by its namespace, "Talk: ..."
Otherwise, the message in the snackbar in your After screenshot no longer makes sense: "Snow White and _its associated page_" is no longer meaningful.

PR #5447, opened 2025-03-31
Create and add pagination indicator to the BottomAppBar - T387941

voyagerfan:
### What does this do?
This adds a pagination indicator with a size gradient. It also removes the default "flash" transition in the navhost

### Why is this needed?
Pagination indicators are part of the YiR screen design

**Phabricator:**
https://phabricator.wikimedia.org/T387941

### DIFF
--- a/app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
+++ b/app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
@@ -20,7 +20,14 @@ class YearInReviewActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
         setContent {
             BaseTheme {
-                val personalizedScreenList = listOf(readCountData, editCountData, readCountData, editCountData, readCountData)
+                /*
+                personalizedScreenList is temporarily populated with screens
+                for testing purposes. This is will adjusted in future iterations
+                 */
+                val personalizedScreenList = listOf(
+                    readCountData, editCountData, readCountData,
+                    editCountData, readCountData
+                )
                 val getStartedList = listOf(getStartedData)
                 val coroutineScope = rememberCoroutineScope()
                 val navController = rememberNavController()
--- a/app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
+++ b/app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
@@ -69,7 +69,7 @@ class YearInReviewActivity : ComponentActivity() {
                             customBottomBar = { pagerState -> MainBottomBar(
                                 onNavigationRightClick = {
                                     coroutineScope.launch {
-                                        pagerState.scrollToPage(pagerState.currentPage + 1)
+                                        pagerState.animateScrollToPage(pagerState.currentPage + 1)
                                     }
                                 },
                                 pagerState = pagerState,
--- a/app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
+++ b/app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
@@ -5,6 +5,8 @@ import android.content.Intent
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
+import androidx.compose.animation.EnterTransition
+import androidx.compose.animation.ExitTransition
 import androidx.compose.runtime.rememberCoroutineScope
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
```diff
@@ -18,13 +20,18 @@ class YearInReviewActivity : ComponentActivity() {
         super.onCreate(savedInstanceState)
         setContent {
             BaseTheme {
-                val personalizedScreenList = listOf(readCountData, editCountData)
+                val personalizedScreenList = listOf(readCountData, editCountData, readCountData, editCountData, readCountData)
```
cooltey, line 23: Could you please add a comment here to address this is a temporary list and it's for testing only?

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -124,24 +130,35 @@ fun MainBottomBar(
...
-            color = WikipediaTheme.colors.borderColor
-        ),
+        modifier = Modifier
+            .drawBehind {
+                val screenWidth = size.width
+                drawLine(
+                    start = Offset(x = 0f, 0f),
+                    end = Offset(x = screenWidth, 0f),
+                    strokeWidth = 3f,
+                    color = topBorderColor,
```
cooltey, line 142: You can put the color here instead of having a one-time use variable.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -294,3 +352,15 @@ fun YearInReviewScreenContent(
         }
     }
 }
+
+fun paginationSizeGradient(indicatorTotal: Int, iteration: Int, pagerState: PagerState): Int {
```
cooltey, line 356: Can we rename the parameter `indicatorTotal` to `totalIndicators`?

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -294,3 +352,15 @@ fun YearInReviewScreenContent(
...
+    var paginationSize = 8
```
cooltey, line 357: Could you please explain the reason when we set it as `8` as the default size? What if a screen cannot fit `8` indicators in the box?

cooltey (changes requested):
Not sure if this PR is ready for review. If it is ready, please add `Ready for review` tag to your PR.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
voyagerfan: Hi Cooltey, Absolutely! I've added a comment and it will be in the next PR.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: HI Cooltey. I attempted to use `WikipediaTheme.colors.inactiveColor` directly in the `color` property of `drawline()` however I get an error indicating "@Composable invocations can only happen from the context of a @Composable function". I believe this is because the `colors` property of WikimediaTheme has a getter that is annotated with `@Composale` which wont work with the `drawLine()` function. 

I attempted a few other solutions to implement a top border with a modifier but they all had some issues or bloated the code a little more:

- Divider composable
- Spacer composable
- Box Composable

I can keep thinking about this one but Im certain open to suggestions.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Absolutely! I will change this and it will be in the next PR.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
cooltey: I see. There's another solution that can use a `HorizontalDivider` with a height in `1.dp`, and wrap it by a `Column` so you will have a divider on top of the `BottomAppBar`. 

```
Column {
        HorizontalDivider (
            modifier = Modifier
                .height(1.dp)
                .fillMaxWidth(),
            color = WikipediaTheme.colors.inactiveColor
        )
        BorromAppBar(
        ...
}
```

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Hi Cooltey, sorry for the confusion. The pagination dot size is set to 8 which is what was listed in the Figma design.  I can change the name to `paginationIndicatorSize` for more clarity. Im using the `indicatorTotal` parameter to set a threshold for when size transitions occur. Currently, the number of indicators comes from how many YiR pages are being used.  You're absolutely right that it will not fit if total number of indicators is too high. For now, 10-15 indicators should fit.

PR #5448, opened 2025-03-31
Follow up: Year-in-review code format optimization

cooltey:
### What does this do?
This PR shows the code formatting adjustments for project consistency.

@voyagerfan Please review and merge if it looks good to you.


voyagerfan (approved):
Hi @cooltey, these are some very nice modifications - thanks for doing this PR. I liked how you eliminated the `pageData` variable and directly passed `contentData[page]` to the `screenContent` composable. Also, I learned something new: that `context` can be used in the `factory property` of the `AndroidView` without it being explicitly declared!

PR #5450, opened 2025-04-01
Implement Coil image loading library

Williamrai:
### What does this do?
This implements a simple design pattern where we can interchange the libraries for loading images. It will create a standard way to load images that works the same no matter what's happening behind the scenes. The goal is to make it easy to switch between different image loading libraries without changing the rest of the app. 

**NOTE:** removing glide breaks one of the espresso test function that identifies a gif image.  I haven't found a solution for that yet. I would create a separate PR for that once i find a solution. 

**Phabricator:**
https://phabricator.wikimedia.org/T389544
https://phabricator.wikimedia.org/T391880


dbrant:
This is an interesting direction, but there are still various elements to be accounted for, which are specific to Glide:

* We have a module called `OkHttpGlideModule` which makes sure that requests to fetch images are passed through our network stack.
* We have `FaceAndColorDetectImageView` and the corresponding `CenterCropWithFaceTransformation` that uses Glide-specific bitmap transformations.

Can all of the above be similarly abstracted away?

Williamrai:
> * CenterCropWithFaceTransformation

I think we can abstract out the FaceAndColorDetectImageView and CenterCropWithFaceTransformation but I don't know if its possible for OkHttpGlideModule since it uses annotation processor which generates files at compile time. If we follow this pattern these files will still be generated but wont be used since we will replacing this with coil but on coil I think we can do this on runtime. So if we create an interface for config this can work.

app/src/main/java/org/wikipedia/views/imageservice/CenterCropWithFaceTransformation.kt
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    private fun isBitmapEligibleForImageProcessing(bitmap: android.graphics.Bitmap): Boolean {
+        val minSize = 64
+        return bitmap.width >= minSize && bitmap.height >= minSize
+    }
+
+    companion object {
+        private const val ID = "org.wikipedia.views.CenterCropWithFace"
```
cooltey, line 112: Can be removed.

app/src/main/java/org/wikipedia/views/imageservice/ImageService.kt
```diff
@@ -0,0 +1,97 @@
...
+        imageUrl: String?,
+        onSuccess: (Bitmap) -> Unit) {
+        implementation.getBitmapForWidget(context, imageUrl, onSuccess)
+    }
+}
+
+interface ImageLoadListener {
+    fun onSuccess(view: ImageView) {}
+    fun onSuccess(palette: Palette, bmpWidth: Int, bmpHeight: Int) {}
+    fun onError(error: Throwable)
```
cooltey, line 61: Can we have a blank `{}` by default so we don't need to include this every time?

cooltey (changes requested):
The pinch-to-zoom function is not working correctly, which `ImageZoomHelper` may need to be updated in this PR.

dbrant (changes requested):
A couple of initial comments!

dbrant:
> The PhotoView actually seems to work if I just do the standard `imageView.load(url)` provided by Coil, but doesn't work if I use our new `ViewUtil.loadImage(...)` on it, so it's definitely in our logic that needs to be fixed.

_Hint_: It can be fixed by removing a single line from the new code.

Williamrai:
> > The PhotoView actually seems to work if I just do the standard `imageView.load(url)` provided by Coil, but doesn't work if I use our new `ViewUtil.loadImage(...)` on it, so it's definitely in our logic that needs to be fixed.
> 
> _Hint_: It can be fixed by removing a single line from the new code.

Thank you for looking into this, I pushed the fix.

app/src/main/java/org/wikipedia/views/imageservice/CenterCropWithFaceTransformation.kt
Williamrai: done.

app/src/main/java/org/wikipedia/views/imageservice/ImageService.kt
Williamrai: done.

app/src/main/java/org/wikipedia/WikipediaApp.kt
```diff
@@ -163,6 +171,8 @@ class WikipediaApp : Application() {
         WikipediaFirebaseMessagingService.updateSubscription()
 
         EventPlatformClient.setUpStreamConfigs()
+
+        ImageService.setImplementation(CoilImageLoader())
```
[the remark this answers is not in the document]
Williamrai: I removed the default implementation, forcing developers to provide this service before using. I think in this way we can know what service we are using and since we also need to set the factory once when we load the app.

app/src/main/java/org/wikipedia/WikipediaApp.kt
```diff
@@ -279,6 +289,21 @@ class WikipediaApp : Application() {
         }
     }
 
+    override fun newImageLoader(context: PlatformContext): ImageLoader {
```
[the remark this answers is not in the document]
Williamrai: I moved this from the `WikipediaApp` to the `CoilImageLoader` class and set it on its `init block`.

app/src/main/java/org/wikipedia/feed/view/CardLargeHeaderView.kt
```diff
@@ -60,8 +59,8 @@ class CardLargeHeaderView : ConstraintLayout {
                 ContextCompat.getColor(context, R.color.gray600))
     }
 
-    private inner class ImageLoadListener : OnImageLoadListener {
-        override fun onImageLoaded(palette: Palette, bmpWidth: Int, bmpHeight: Int) {
+    private inner class ImageLoadListener : org.wikipedia.views.imageservice.ImageLoadListener {
```
cooltey, line 62: Can we rename the `ImageLoadListener` to `CardImageLoadListener` so we don't have to see the full package path for the `ImageLoadListener`?

app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.kt
```diff
@@ -287,6 +273,21 @@ class GalleryItemFragment : Fragment(), MenuProvider, RequestListener<Drawable?>
         return FragmentUtil.getCallback(this, Callback::class.java)
     }
 
+    private fun createImageLoadListener(): ImageLoadListener {
```
cooltey, line 276: Would it be better to convert this into an `inner class` and create this class once, so we can reuse it?

app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
```diff
@@ -146,15 +149,18 @@ class CustomHtmlParser(private val handler: TagHandler) : TagHandler, ContentHan
                             uri = Service.COMMONS_URL + uri.replace("./", "")
                         }
 
-                        ImagePipelineBitmapGetter(view.context, uri) { bitmap ->
-                            if (!drawable.bitmap.isRecycled) {
+                        ImageService.imagePipeLineBitmapGetter(view.context, uri, onSuccess = { bitmap ->
+                            Log.d("orange", "Received bitmap: width=${bitmap.width}, height=${bitmap.height}, " +
```
cooltey, line 153: Please remove the log?

app/src/main/java/org/wikipedia/views/imageservice/ImageService.kt
```diff
@@ -0,0 +1,100 @@
...
+        cropped: Boolean = true,
+        emptyPlaceholder: Boolean = false,
+        listener: ImageLoadListener? = null
+    )
+
+    fun imagePipeLineBitmapGetter(
+        context: Context,
+        imageUrl: String?,
+        imageTransformer: ImageTransformer? = null,
+        onSuccess: (Bitmap) -> Unit)
```
cooltey, line 90: Please make the ending `)` to the next line.

app/src/main/java/org/wikipedia/views/imageservice/ImageService.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    fun getBitmapForWidget(
+        context: Context,
+        imageUrl: String?,
+        onSuccess: (Bitmap) -> Unit)
```
cooltey, line 95: Same

cooltey (changes requested):

app/src/main/java/org/wikipedia/gallery/GalleryItemFragment.kt
Williamrai: done.

app/src/main/java/org/wikipedia/feed/view/CardLargeHeaderView.kt
Williamrai: done.

app/src/main/java/org/wikipedia/richtext/CustomHtmlParser.kt
Williamrai: removed.

app/src/main/java/org/wikipedia/views/imageservice/ImageService.kt
Williamrai: done.

app/src/main/java/org/wikipedia/views/imageservice/ImageService.kt
Williamrai: done.

cooltey (approved):
Looks good to me.

PR #5454, opened 2025-04-03
Pass login UI messages through parser.

dbrant:
When we receive "UI" messages from the Login API (for entering a 2FA code or captcha), these messages are received as wikitext, and might contain links or other syntax, and must therefore be passed through the parser.

This also takes care of a few other tiny tweaks:
* When displaying a snackbar with text that comes from `fromHtml()`, let's `trim()` it because certain messages, after being parsed, end up containing extraneous empty tags at the end (such as `<p>` or `<br />`) that produce unnecessary empty lines that don't look good in a snackbar.
* The 2FA field in the Login screen was missing top padding.
* When revealing the 2FA field, let's hide the soft keyboard, so that the full snackbar could be displayed, and not overlap the 2FA field.

https://phabricator.wikimedia.org/T390919

### DIFF
--- a/app/src/main/java/org/wikipedia/login/LoginClient.kt
+++ b/app/src/main/java/org/wikipedia/login/LoginClient.kt
@@ -38,10 +38,11 @@ class LoginClient {
                     }
                     cb.success(loginResult)
                 } else if (LoginResult.STATUS_UI == loginResult.status) {
+                    val parsedMessage = loginResult.message?.let { ServiceFactory.get(wiki).parseText(it) }?.text
                     when (loginResult) {
-                        is LoginOAuthResult -> cb.twoFactorPrompt(LoginFailedException(loginResult.message), loginToken)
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/wikipedia/login/LoginClient.kt
```diff
@@ -38,10 +38,11 @@ class LoginClient {
                     }
                     cb.success(loginResult)
                 } else if (LoginResult.STATUS_UI == loginResult.status) {
+                    val parsedMessage = loginResult.message?.let { ServiceFactory.get(wiki).parseText(it) }?.text
```
cooltey, line 41: Would it be a good idea that we add a generic error message if the `parseText` is `null`?

dbrant: Sure, we can fall back to the unparsed message in case parsing fails for whatever reason.

PR #5455, opened 2025-04-03
JankStats

Williamrai:
### What does this do?
This implements JankStats. It copies a class from this sample project from [google](https://github.com/android/performance-samples/tree/358d57c7f5922a0ad4b91d816ee2c4a8f4ce8929/JankStatsSample
) for testing purpose. It will log the jank stats when **MainActivity** goes into background. If it's possible you should test it with a lower spec device.

I tested this on a pixel 2 device and this were in the log. 
`JankStats: reason Activity pausedtotalFrames = 335 jankFrames = 11
`

**Phabricator:**
https://phabricator.wikimedia.org/T390624


Williamrai:
Closing this as the spike is complete.

PR #5456, opened 2025-04-03
Convert WikiCardView to use Compose

cooltey:
### What does this do?
Converts the `WikiCardView` to use Compose. No design review will be needed since this is a custom view conversion.

Some screenshots have been attached in the ticket if you desire to see the implementations.

**Phabricator:**
https://phabricator.wikimedia.org/T389141

### DIFF
--- a/app/src/main/java/org/wikipedia/compose/components/WikiCard.kt
+++ b/app/src/main/java/org/wikipedia/compose/components/WikiCard.kt
@@ -2,20 +2,26 @@ package org.wikipedia.compose.components
 
 import android.os.Build
 import androidx.compose.foundation.BorderStroke
+import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.Card
 import androidx.compose.material3.CardColors
 import androidx.compose.material3.CardDefaults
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/compose/components/WikiCard.kt
Williamrai, line 1: It might be helpful to include previews of the different wikiCardView versions, similar to what's shown in the Phabricator task. This would make it easier to visualize how everything will look. Are you planning to let the child content handle the click events?

cooltey: Thanks for providing the solutions for previews. About the clickable event, I have this approach: https://github.com/wikimedia/apps-android-wikipedia/pull/5458/files#diff-35ffa12c20d1cd356cbbc991ac67fe71df08f73835521cf252ff94f3f05fce07R76-R81, but not sure which one is better. Any suggestions are welcome.

Williamrai: this is fine.

PR #5457, opened 2025-04-03
Enable EmailAuth support.

dbrant:
This enables integration with [EmailAuth](https://www.mediawiki.org/wiki/Extension:EmailAuth), where a login attempt requires a second factor that is sent to the user via email.
This basically reuses our existing logic for 2FA ([OATHAuth](https://www.mediawiki.org/wiki/Extension:OATHAuth)), just with slightly different verbiage and logic.

There is now a new Developer preference for "forcing" email authentication, if necessary for testing.

https://phabricator.wikimedia.org/T390820

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/SharedPreferenceCookieManager.kt
+++ b/app/src/main/java/org/wikipedia/dataclient/SharedPreferenceCookieManager.kt
@@ -4,6 +4,7 @@ import okhttp3.Cookie
 import okhttp3.CookieJar
 import okhttp3.HttpUrl
 import okhttp3.HttpUrl.Companion.toHttpUrl
+import org.wikipedia.login.LoginClient
 import org.wikipedia.settings.Prefs
 import org.wikipedia.util.log.L
 
@@ -104,6 +105,10 @@ class SharedPreferenceCookieManager(
                 // from wikipedia.org unconditionally.
                 buildCookieList(cookieList, cookiesForDomainSpec, CENTRALAUTH_PREFIX)
             }
+        }
+        if (LoginClient.enqueueForceEmailAuth && (url.toString().contains("action=clientlogin"))) {
+            cookieList.add(Cookie.Builder().name("forceEmailAuth").value("1").domain(domain).secure().build())
+            LoginClient.enqueueForceEmailAuth = false
         }
         return cookieList
     }
--- a/app/src/main/java/org/wikipedia/dataclient/okhttp/CommonHeaderRequestInterceptor.kt
+++ b/app/src/main/java/org/wikipedia/dataclient/okhttp/CommonHeaderRequestInterceptor.kt
@@ -4,19 +4,24 @@ import okhttp3.Interceptor
 import okhttp3.Response
 import org.wikipedia.WikipediaApp
 import org.wikipedia.dataclient.RestService
+import org.wikipedia.login.LoginClient
 import java.io.IOException
 
 internal class CommonHeaderRequestInterceptor : Interceptor {
     @Throws(IOException::class)
     override fun intercept(chain: Interceptor.Chain): Response {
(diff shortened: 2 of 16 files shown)


app/src/main/java/org/wikipedia/dataclient/SharedPreferenceCookieManager.kt
```diff
@@ -105,6 +106,10 @@ class SharedPreferenceCookieManager(
                 buildCookieList(cookieList, cookiesForDomainSpec, CENTRALAUTH_PREFIX)
             }
         }
+        if (LoginClient.enqueueForceEmailAuth && (url.toString().contains("action=clientlogin"))) {
```
dbrant, line 109: This is some not-very-clean logic for injecting a cookie for "forcing" email authentication. Unfortunately this is basically the only way it can be done.

app/src/main/java/org/wikipedia/dataclient/okhttp/CommonHeaderRequestInterceptor.kt
```diff
@@ -4,20 +4,25 @@ import okhttp3.Interceptor
...
 
 internal class CommonHeaderRequestInterceptor : Interceptor {
     @Throws(IOException::class)
     override fun intercept(chain: Interceptor.Chain): Response {
         val app = WikipediaApp.instance
+        val url = chain.request().url.toString()
         val builder = chain.request().newBuilder()
                 .header("User-Agent", app.userAgent)
                 .header("X-WMF-UUID", app.appInstallID)
-        if (chain.request().url.encodedPath.contains(RestService.PAGE_HTML_ENDPOINT)) {
```
dbrant, line 16: the `url.encodedPath` and `url.host` calls actually involve a lot of string-splitting allocations, so let's just operate on the entire `url.toString()` which is more performant.

app/src/main/java/org/wikipedia/dataclient/okhttp/CommonHeaderRequestInterceptor.kt
```diff
@@ -4,20 +4,25 @@ import okhttp3.Interceptor
...
+        if (url.contains(RestService.PAGE_HTML_ENDPOINT)) {
             builder.header("Accept", RestService.ACCEPT_HEADER_MOBILE_HTML)
-        } else if (chain.request().url.host.contains("maps.wikimedia.org")) {
+        } else if (url.contains("maps.wikimedia.org")) {
             builder.header("Referer", "https://maps.wikimedia.org/")
         }
+        if (LoginClient.enqueueForceEmailAuth && url.contains("action=clientlogin")) {
+            builder.removeHeader("User-Agent")
```
dbrant, line 24: Again, unclean logic for "forcing" email authentication. It looks like, if the server detects the apps' user-agent (`WikipediaApp/...`) then it's more likely to _ignore_ email authentication. To force it to trigger, let's just remove the user-agent entirely.

cooltey (changes requested):
Found two issues:
If you see this final message after multiple times of failing on authentication, the code input field is still enabled and you can still click on submit button. 


The snackbar does not dismiss automatically (and alone with the OK action button), same as the "authentication failed" snackbar. Is this expected?

dbrant:
> Found two issues: If you see this final message after multiple times of failing on authentication, the code input field is still enabled and you can still click on submit button.

Good find, and... done! If there are too many failures, the 2fa field will now reset to a hidden state.

> The snackbar does not dismiss automatically (and alone with the OK action button), same as the "authentication failed" snackbar. Is this expected?

Yes, this is our standard snackbar behavior when showing errors (i.e. not part of this PR). If a message is longer than a certain length, we make the snackbar persist indefinitely so that the user can have a chance to acknowledge it.

cooltey (approved):

PR #5458, opened 2025-04-04
Convert MessageCardView to use Compose

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T389148


app/src/main/java/org/wikipedia/compose/components/WikiCard.kt
```diff
@@ -48,6 +62,112 @@ fun WikiCard(
...
+                        ) {
+                            Text(
+                                text = positiveButtonText,
+                                fontSize = 16.sp
+                            )
+                        }
+                    }
+
+                    if (negativeButtonText != null && isNegativeButtonVisible) {
+                        AppTextButton(
```
Williamrai, line 156: If we have a **negativeButtonText** then do we need a separate boolean **isNegativeButtonVisible**.

cooltey: Thanks, you are correct! Done.

PR #5461, opened 2025-04-04
Follow-up: YiR pagination and code formats update

cooltey:
### What does this do?
Simplified some code and tried to make the coding style match the project. 
@voyagerfan Please review and merge if it looks good to you.


voyagerfan (approved):
Hi Cooltey, the changes look good! I like how you further condensed the `paginationSizeGradient`. I originally refactored my version from a large if/else block using `when`, but returning the `when` right away makes it more compact. Also your consolidation with `animationDuration` works well. I was experimenting with different values, but sharing one variable for each animationState looks to be the best approach since they share the same value.

PR #5473, opened 2025-04-11
Create a viewModel, API calls and DB queries - T390643

voyagerfan:
### What does this do?
Adds DB and API call specific to YearInReview. Adds a viewModel and is integrated with the UI

### Why is this needed?
Adopts a viewModel for separation of concerns. Implements more of the Figma design by fetching personalized data for screen presentation.

**Phabricator:**
https://phabricator.wikimedia.org/T390643


voyagerfan:
Hi @cooltey, please let me know what you think of the viewModel architecture. Thanks!

app/src/main/java/org/wikipedia/dataclient/Service.kt
```diff
@@ -439,6 +439,16 @@ interface Service {
             @Query("uccontinue") uccontinue: String?
     ): MwQueryResponse
 
+    @GET(MW_API_PREFIX + "action=query&list=usercontribs&ucprop=ids|title|timestamp|comment|size|flags|sizediff|tags&meta=userinfo&uiprop=groups|blockinfo|editcount|latestcontrib|rights|registrationdate")
```
cooltey, line 442: Are all of the additional values in `uiprop` really necessary for this feature?

app/src/main/java/org/wikipedia/history/db/HistoryEntryDao.kt
```diff
@@ -21,6 +21,12 @@ interface HistoryEntryDao {
     @Query("SELECT * FROM HistoryEntry WHERE authority = :authority AND lang = :lang AND apiTitle = :apiTitle AND timestamp = :timestamp LIMIT 1")
     suspend fun findEntryBy(authority: String, lang: String, apiTitle: String, timestamp: Long): HistoryEntry?
 
+    @Query("SELECT COUNT(*) FROM HistoryEntry WHERE timestamp BETWEEN :startDate AND :endDate ")
+    suspend fun getHistoryCount(startDate: Long, endDate: Long): Int
+
+    @Query("SELECT DISTINCT REPLACE(apiTitle, \"_\", \" \") FROM HistoryEntry ORDER BY random() LIMIT 3")
+    suspend fun getApiTitles(): List<String>
```
cooltey, line 28: You can use `displayTitle` instead of `apiTitle` in the query since the titles are not used for requesting APIs.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
```diff
@@ -24,10 +27,9 @@ class YearInReviewActivity : ComponentActivity() {
                 personalizedScreenList is temporarily populated with screens
                 for testing purposes. This is will adjusted in future iterations
                  */
-                val personalizedScreenList = listOf(
-                    readCountData, editCountData, readCountData,
-                    editCountData, readCountData
-                )
+                val viewModelFactory = YearInReviewViewModelFactory(application)
+                val yirViewModel = ViewModelProvider(this, viewModelFactory)[YearInReviewViewModel::class]
```
cooltey, line 31: Can we have a similar way of creating the viewModel? Please see `AddLanguagesListActivity` for examples.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
```diff
@@ -57,30 +59,45 @@ class YearInReviewActivity : ComponentActivity() {
...
-                                },
-                                pagerState = pagerState,
-                                totalPages = personalizedScreenList.size) },
-                            screenContent = { innerPadding, contentData ->
-                                YearInReviewScreenContent(
-                                    innerPadding = innerPadding,
-                                    screenData = contentData
+                        val screenState = yirViewModel.uiScreenListState.collectAsState().value
+                        when (screenState) {
+                            is Resource.Loading -> {} // Re-route to dedicated loading screen or adding loading composable
```
cooltey, line 70: For loading, you can have something like this (this can be in another follow up PR).
```
Box(
    modifier = modifier
        .fillMaxSize()
        .padding(paddingValues)
) {
    CircularProgressIndicator(
        modifier = Modifier
            .align(Alignment.BottomEnd)
            .padding(24.dp),
        color = WikipediaTheme.colors.progressiveColor
    )
}
 ```

File: app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt [hunk 5]
```diff
@@ -57,30 +59,45 @@ class YearInReviewActivity : ComponentActivity() {
...
+                                        YearInReviewScreenContent(
+                                            innerPadding = innerPadding,
+                                            screenData = contentData,
+                                        )
+                                    },
                                 )
-                            },
-                        )
+                            }
+                            else -> {} // Re-Route to a dedicated error screen or load Error composable
```
cooltey, line 92: Same as the error view:
```
Box(
    modifier = modifier
        .fillMaxSize()
        .padding(paddingValues)
        // Add bottom padding when keyboard is visible for android 15 and above
        .padding(bottom = imeHeight),
    contentAlignment = Alignment.Center
) {
    WikiErrorView(
        modifier = Modifier
            .fillMaxWidth(),
        caught = uiState.error,
        errorClickEvents = wikiErrorClickEvents
    )
}

```

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -0,0 +1,122 @@
...
+        )
+    }
+
+    fun fetchPersonalizedData() {
+        val personalizedStatistics = YearInReviewStatistics()
+        viewModelScope.launch(handler) {
+            _uiScreenListState.value = Resource.Loading()
+            val readCountJob = async {
+                personalizedStatistics.readCount = AppDatabase.instance.historyEntryDao().getHistoryCount(startTimeInMillis, endTimeInMillis)
+                if (personalizedStatistics.readCount < 3) throw IllegalStateException("0")
```
cooltey, line 49: Could you please explain why this throws `0` if the readCount < 3?

cooltey (changes requested):

app/src/main/java/org/wikipedia/dataclient/Service.kt
```diff
@@ -439,6 +439,16 @@ interface Service {
...
+    suspend fun getUserContribsByTimeFrame(
+        @Query("ucuser") username: String,
+        @Query("uclimit") maxCount: Int,
+        @Query("ucstart") startDate: String,
```
dbrant, line 446: Fun fact: instead of a `String` type for dates to be passed to the API, you can actually declare this as type `Instant` (java.time.Instant), and they will _automatically_ serialize in the correct ISO8601 format ("dateTtimeZ")

app/src/main/java/org/wikipedia/dataclient/Service.kt
```diff
@@ -439,6 +439,16 @@ interface Service {
...
+        @Query("ucend") endDate: String,
+        @Query("ucnamespace") ns: Int?,
```
dbrant, line 448: If you give these nullable parameters a default value of `= null`, you won't need to specify null explicitly wherever it's called.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -0,0 +1,122 @@
...
+@OptIn(ExperimentalTime::class)
+class YearInReviewViewModel(appContext: Application) : AndroidViewModel(appContext) {
+    private val wikiAppContext = appContext
+    private var _uiScreenListState = MutableStateFlow(Resource<List<YearInReviewScreenData>>())
+    val uiScreenListState: StateFlow<Resource<List<YearInReviewScreenData>>> = _uiScreenListState.asStateFlow()
+
+    val currentYear = Calendar.getInstance().get(Calendar.YEAR).toString()
+    val startTime = "$currentYear-01-01T00:00:00Z"
+    val endTime = "$currentYear-12-31T23:59:59Z"
+    val startTimeInMillis = Instant.parse(startTime).toEpochMilliseconds()
```
dbrant, line 33: ...and instead of parsing from a string here, you can now say `startTimeInMillis = startTime.toInstant(ZoneOffset.UTC).toEpochMilli()`

And I'm pretty sure this will remove the dependency on "ExperimentalTime"

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -0,0 +1,122 @@
...
+                )
+                return@async readCountData
+            }
+            val editCountJob = async {
+                val homeSiteCall = async {
+                    ServiceFactory.get(WikipediaApp.instance.wikiSite)
+                        .getUserContribsByTimeFrame(
+                            username = AccountUtil.userName,
+                            maxCount = 500,
+                            startDate = endTime,
```
dbrant, line 70: ...and now that we can pass an `Instant` directly into the API call, you can pass `startTime`/`endTime` here, without any hardcoding of strings.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -0,0 +1,122 @@
...
+                            endDate = startTime,
+                            ns = null,
```
dbrant, line 72: These can be default parameters, and don't need to be passed explicitly.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -0,0 +1,122 @@
...
+                    personalizedStatistics.readCount.toString()
+                )
+                readCountData.bodyText = wikiAppContext.getString(
+                    R.string.year_in_review_read_count_bodytext,
+                    personalizedStatistics.readCount.toString(),
+                    personalizedStatistics.readCountApiTitles[0],
+                    personalizedStatistics.readCountApiTitles[1],
+                    personalizedStatistics.readCountApiTitles[2]
+                )
+                return@async readCountData
```
dbrant, line 62: A cleaner syntax for returning from an `async{}` block is just to state `readCountData` on its own line. (no need to say `return@async`)
Same in the other block(s).

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModelFactory.kt
```diff
@@ -0,0 +1,14 @@
+package org.wikipedia.yearinreview
+
+import android.app.Application
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
+
+class YearInReviewViewModelFactory(private var appContext: Application) : ViewModelProvider.Factory {
```
dbrant, line 7: A factory class for creating this ViewModel should not be necessary.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -0,0 +1,122 @@
...
+import org.wikipedia.dataclient.ServiceFactory
+import org.wikipedia.util.Resource
+import org.wikipedia.util.log.L
+import java.util.Calendar
+import kotlin.time.ExperimentalTime
+import kotlin.time.Instant
+
+@OptIn(ExperimentalTime::class)
+class YearInReviewViewModel(appContext: Application) : AndroidViewModel(appContext) {
+    private val wikiAppContext = appContext
```
dbrant, line 26: This can just be `wikiAppContext = WikipediaApp.instance`, and you no longer need to inherit from AndroidViewModel; it can just be ViewModel.

dbrant (changes requested):

app/src/main/java/org/wikipedia/dataclient/Service.kt
voyagerfan: You're right, this isnt necessary for the goal we're trying to accomplish here. I originally borrowed from another API call but modified it to accept time bounds for data. I went ahead and shortened the call (below) which will be in the next PR.

@GET(MW_API_PREFIX + "action=query&list=usercontribs&meta=userinfo&uiprop=editcount")

app/src/main/java/org/wikipedia/history/db/HistoryEntryDao.kt
voyagerfan: Sounds good Cooltey, thanks! I changed this to use `displayTitle` and then I realized that this column is a string for a span tag. To get around this, I extracted the content from the span tag using a query but we could also do the extraction in the viewModel. Let me know what you think the next PR.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
voyagerfan: Hi Dmitry, thank you! This is super helpful!

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
voyagerfan: Absolutely! I originally did this since I was manually injecting application context, but now that Im using the `WikipediaApp.instance`, I've worked the viewModel instantiation to look like `AddLanguagesListActivity`.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
voyagerfan: Hey Cooltey, this is a great idea. I did not know there was a native `CircularProgressIndicator` composable. I went ahead a created something similar to this but it will need to be refined. I was thinking about how we could handle different loading scenarios but perhaps its best for another PR.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
voyagerfan: For sure! My understanding is that when readCount < 3, the fallback behavior is to display a hardcoded collective slide instead of individual article slides. Since those collective slides are handled by the `handler`, I needed a way to ensure it would be triggered in that case.

We had discussed possibly using the Elvis operator, but after thinking it through, I felt a more explicit if/else block was clearer and better suited here. Happy to talk through it more if you think another approach might work better!

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
voyagerfan: Thanks Dmitry. Yes this did remove the dependency.

app/src/main/java/org/wikipedia/dataclient/Service.kt
voyagerfan: Hey Dmitry, unfortunately I didnt have success using the type `Instant`. We can talk about this in more detail but I was essentially seeing a type mismatch. 

Also, thank you for the note on auto serialization which worked but with one small hiccup. 
`LocalDateTime.of(currentYear, 1, 1, 0, 0, 0)` is serialized to "2025-01-01T00:00" (dropping the seconds) which the API doesnt seem to like. I've changed the start time to LocalDateTime.of(currentYear, 1, 1, 0, 0, 1) for now.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -0,0 +1,122 @@
...
+    private var _uiScreenListState = MutableStateFlow(Resource<List<YearInReviewScreenData>>())
+    val uiScreenListState: StateFlow<Resource<List<YearInReviewScreenData>>> = _uiScreenListState.asStateFlow()
+
+    val currentYear = Calendar.getInstance().get(Calendar.YEAR).toString()
+    val startTime = "$currentYear-01-01T00:00:00Z"
```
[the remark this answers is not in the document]
voyagerfan: Thanks for this! Its much cleaner looking instead of using strings!

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -0,0 +1,122 @@
...
+import kotlin.time.ExperimentalTime
+import kotlin.time.Instant
+
+@OptIn(ExperimentalTime::class)
+class YearInReviewViewModel(appContext: Application) : AndroidViewModel(appContext) {
+    private val wikiAppContext = appContext
+    private var _uiScreenListState = MutableStateFlow(Resource<List<YearInReviewScreenData>>())
+    val uiScreenListState: StateFlow<Resource<List<YearInReviewScreenData>>> = _uiScreenListState.asStateFlow()
+
+    val currentYear = Calendar.getInstance().get(Calendar.YEAR).toString()
```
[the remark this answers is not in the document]
voyagerfan: Understood. I've made the change and it will be in the next PR

app/src/main/java/org/wikipedia/dataclient/Service.kt
voyagerfan: Good catch, this was oversight on my part. I've set the nullable parameter with a default value of null and removed unnecessary parameters from the API calls in the viewModel

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
voyagerfan: Understood. After cleaning up Service.kt (setting default values as null), I've removed this in the viewModel.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
voyagerfan: Thanks for the tip! I assumed that wouldnt work and I'd need a return statement but I've learned something new today! Instances of these have been fixed now.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
voyagerfan: Im currently passing type `LocalDateTime` since I had a type mismatch when using `Instant`, but perhaps I've missed something - we can cover this in a more detail.

app/src/main/java/org/wikipedia/dataclient/Service.kt
voyagerfan: Actually, I was able to get this to work using `Instant`. Using instant also solved the problem of dropping the seconds when serializing so the API calls are working.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
voyagerfan: I was able to use `Instant` instead of `LocalDateTime`

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
voyagerfan: @cooltey I started to add the WikiErrorView but realized that the viewModel does not currently throw an error, it either shows `Resource.Loading` or `Resource.Success`. The fallback behavior is to show non-personalized slides if the api calls fail so im not sure how to incorporate the error view. Im open to suggestions or perhaps we leave this for another PR?

cooltey (commented):
The logic looks good to me now, before merging it to the feature branch, I just want to confirm few things:

1. Regarding the Figma design, it has two versions for logged-out users: 
  - English users: using data from micro-site
  - non-English users: using static data.
  
  Do you have a plan to implement screens for different versions for English vs non-English users in this PR?
  
2. For screens for logged in users, it seems to have more screens than your implementation, do you want to add them in this PR?

voyagerfan:
Hi Cooltey,
1. I believe this is correct.

I didnt plan to implement different versions for English vs non-English in this PR, however I did intend to introduce branching logic in [ticket T392406](https://phabricator.wikimedia.org/T392406) with the breadcrumb. I was thinking once the branching logic is in place, we can add more screens in T392406, but Im open to any other ideas you have.

2. I can definitely add more screens to this PR. My original intent was to address at least 2 screens for personalized and non-personalized views (as part of the [EPIC "must haves"](https://phabricator.wikimedia.org/T387034)) and to make sure the viewModel architecture is sound before adding more to it. Do you think we should we should add more screens now? I'd be happy to get to work on that.

cooltey:
Per the discussion, merge this PR and I will do:
1. Create a follow-up PR.
2. You will need to work on the rest of the screen for different scenarios

PR #5478, opened 2025-04-15
Compose: simplify WikiErrorView and context handling.

dbrant:
This significantly simplifies our Compose WikiErrorView:
* This view does not need its own ViewModel, because it's entirely static and does not make use of any suspending functions, and doesn't need a `flow` for its state.
* There is no need for a separate "parent" view (?).

In a few other components, we are unnecessarily passing around `context` and `scope` variables, where they can just be used in-place using functions like `stringResource()`.


app/src/main/java/org/wikipedia/compose/components/error/ComposeWikiErrorView.kt
```diff
@@ -146,12 +148,109 @@ fun ComposeWikiErrorView(
...
+            isOffline(it) -> {
+                return ComposeErrorType.Offline()
+            }
+            isEmptyException(it) -> {
+                return ComposeErrorType.Empty()
+            }
+            isNotLoggedIn(it) -> {
+                return ComposeErrorType.LoggedOut()
+            }
+            else -> {}
```
cooltey, line 195: Do we need to keep an empty `else` here?

app/src/main/java/org/wikipedia/compose/components/error/ComposeWikiErrorView.kt
```diff
@@ -146,12 +148,109 @@ fun ComposeWikiErrorView(
...
     BaseTheme(
         currentTheme = Theme.DARK
     ) {
-        ComposeWikiErrorParentView(
+        WikiErrorView(
             caught = Exception()
         )
     }
 }
+
```
cooltey, line 160: Would it be easier to find if we move the `@Preview` function to the bottom of the class?

PR #5479, opened 2025-04-16
Follow-up: YiR entry dialog code optimization

cooltey:
### What does this do?
A follow-up PR of #5465 that does the following optimization:
- Renames classes to `EntryDialog` instead of a plain `Dialog`.
- Renames `BottomSheetScaffold` to `YearInReviewEntryDialogScreen`, since there might be another bottomsheet that exists in the YiR package, and also for easier access.
- Renames the string resources with `entry_dialog` in them.
- Optimizes the scrolling experience.
- Optimizes the padding for title, body text, and the image - the current image is cut because of the `centerCrop`.

@voyagerfan Please help review, and we can discuss this in our 1:1 for you to have a better picture of this PR.


voyagerfan (approved):
Hi Cooltey, the naming convention you're using looks good - I'll be mindful that as well as the padding values in the next PR.

PR #5481, opened 2025-04-17
Breadcrumbs in Compose.

dbrant:
Some initial ideas for doing Breadcrumbs in our Compose screens.

* We can make `BreadCrumbLogEvent` expose a completely generic function to send an arbitrary event, taking only the current Context and a String which identifies a component. (the Context is used automatically to get the name of the parent Activity or Fragment, which works for Compose and non-Compose.)
* After that, it's basically a single line of code to send a Breadcrumb event for any interaction we like; we just need to remember to do it.
* Since we're building various components for our screens (e.g. TopAppBar, ErrorView), we can put Breadcrumb events into those components, which will be inherited automatically by any screens that use those components.
* We should keep an eye on developments in the Compose framework that might allow us to do breadcrumbs in the future like we currently do for regular Views. Until that time, this might be the simplest / most balanced path forward.


Williamrai:
My idea was to create an extension to a modifier like this
```
fun Modifier.onClickWithEvent(
    context: Context,
    composableName: String,
    enabled: Boolean = true,
    role: Role? = null,
    onClick: () -> Unit
): Modifier = composed {
    val localIndication = LocalIndication.current
    val interactionSource = if (localIndication is IndicationNodeFactory) {
        null
    } else {
        remember { MutableInteractionSource() }
    }
    Modifier.clickable(
        enabled = enabled,
        role = role,
        indication = localIndication,
        interactionSource = interactionSource,
        onClick = {
            BreadCrumbLogEvent.logClick(context, composableName)
            onClick()
        }
    )
}
```
and use it like this
```
modifier = Modifier
                    .onClickWithEvent(
                        context = context,
                        composableName = "searchBackButton",
                        onClick = {
                            onBackButtonClick()
                        }
                    ),
```
However, to make this work we would need to create custom component for all UI that already has actions like IconButton, Button and others. So i don't think this extension would make things easier.

PR #5482, opened 2025-04-17
Correctly show/hide wikitext toolbar based on IME state.

dbrant:
This updates (and simplifies!) our logic for showing and hiding the Wikitext syntax toolbar on top of the system keyboard.
It will now be based directly on the insets provided by the system IME, which fixes the currently-broken behavior on API 35.


app/src/main/java/org/wikipedia/edit/SyntaxHighlightViewAdapter.kt
```diff
@@ -43,16 +43,12 @@ class SyntaxHighlightViewAdapter(
...
-        }
-
-        editText.onFocusChangeListener = View.OnFocusChangeListener { _, hasFocus ->
-            showOrHideSyntax(hasFocus)
+        ViewCompat.setOnApplyWindowInsetsListener(rootView) { _, insets ->
+            val imeVisible = insets.isVisible(WindowInsetsCompat.Type.ime())
+            hideAllSyntaxModals()
+            wikiTextKeyboardView.isVisible = imeVisible && editText.isFocused
+            rootView.onApplyWindowInsets(insets.toWindowInsets())
+            insets
```
Williamrai, line 51: This code is not executing on APIs below 35, due to which the WikiTextKeyboardView becomes visible right after we open the edit screen.
[img](https://github.com/user-attachments/assets/e3424905-d824-486a-9a77-945d202c5938)

dbrant: whoops; fixed!

PR #5497, opened 2025-04-24
WikiGames: enhanced sharing of game results

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T392583


dbrant:
Definitely a few "unconventional" techniques here, but... works great!

PR #5499, opened 2025-04-25
Fix potential crash(es) due to coroutine without exception handler.

dbrant:
There were a couple of places in our code where we launch a coroutine (which performs network access) without an exception handler. If the network access fails and throws an exception, the exception will be unhandled, leading to a crash.

**Phabricator:**
https://phabricator.wikimedia.org/T392472


app/src/main/java/org/wikipedia/talk/TalkTopicHolder.kt
```diff
@@ -123,7 +125,10 @@ class TalkTopicHolder internal constructor(
     }
 
     private fun showOverflowMenu(anchorView: View) {
-        context.lifecycleScope.launch {
+        context.lifecycleScope.launch(CoroutineExceptionHandler { _, throwable ->
+            L.e(throwable)
+            FeedbackUtil.showError(context, throwable)
```
dbrant, line 130: We should show a snackbar with the error here, but not in the other place, because this spot is directly caused by a user action, whereas the other spot is an indirect query from other flows.

PR #5500, opened 2025-04-25
Fix crash when long-pressing to open from Places Search.

dbrant:
* Go to Places
* Search for something in the Search interface
* Long-press on a result and select "Open"

This is because the `LongPressHandler` doesn't preserve the `location` of the search result, and therefore doesn't pass it back into its callback.
I would argue the proper fix for this behavior is just to open the result in PageActivity for reading, since intuitively if someone long-presses and chooses "Open", the intention is to "open for reading".

**Phabricator:**
https://phabricator.wikimedia.org/T392660


Williamrai:
yes i also think it should go to PageActivity when someone long press the search result.

PR #5501, opened 2025-04-25
Dev playground locally track categories

Williamrai:
### What does this do?
Adds a dev View to test categories functionality. To access go to developer settings and search "category" or scroll down to the bottom. 

[Preview](https://drive.google.com/file/d/1X1hQOBpxmf15VFyFLelbur3GlC3Q2FbS/view?usp=sharing)


cooltey:
This works for me, and the good thing is that this was created using Compose. Please fix the single-line issue in the input fields.

cooltey (changes requested):
All looks good, some minor improvements can be made:

1. When it's in landscape mode, the screen is not scrollable.
2. Can we have default values in the input fields (Year: 2025, Number of Rows: 5...etc)

PR #5507, opened 2025-04-28
Fix edits requirements update

Williamrai:
### What does this do?
- adds the 50+ Edits requirement logic to use Article Descriptions on SuggestedEditsFeedClient
- adds api call on  SuggestedEditsFeedClient
- adds a Preference to save the cardTypes in order to not perform the api call multiple times when scrolling

**Phabricator:**
https://phabricator.wikimedia.org/T391050


app/src/main/java/org/wikipedia/settings/Prefs.kt
```diff
@@ -773,4 +773,8 @@ object Prefs {
             OnThisDayGameNotificationState.valueOf(it)
         } ?: OnThisDayGameNotificationState.NO_INTERACTED
         set(value) = PrefsIoUtil.setString(R.string.preference_key_otd_notification_state, value.name)
+
```
dbrant, line 776: This will technically work, but I think it's a bit overkill to cache this information in a Preference.
I think this may be a rare case when it's acceptable to cache it in a static (`companion`) variable, right in the `SuggestedEditsFeedClient` class itself, making it simpler and self-contained.

(see additional notes in that class) This PR shouldn't be changing more than 2 files at the most.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
```diff
@@ -17,19 +29,39 @@ class SuggestedEditsFeedClient(
...
     override fun request(context: Context, wiki: WikiSite, age: Int, cb: FeedClient.Callback) {
+        clientJob?.cancel()
         this.age = age
         this.cb = cb
 
         if (UserContribStats.isDisabled() || UserContribStats.maybePauseAndGetEndDate() != null) {
             cb.success(emptyList())
             return
         }
+        val cardTypes = JsonUtil.decodeFromString<MutableList<DescriptionEditActivity.Action>>(Prefs.suggestedEditsFeedCardTypes) ?: mutableListOf()
```
dbrant, line 40: This can be a static variable in this class, instead of in Prefs.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
```diff
@@ -17,19 +29,39 @@ class SuggestedEditsFeedClient(
...
+        if (cardTypes.isEmpty()) {
```
dbrant, line 41: This can be `if (age == 0)`

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
```diff
@@ -17,19 +29,39 @@ class SuggestedEditsFeedClient(
...
+            clientJob = coroutineScope.launch(Dispatchers.IO + CoroutineExceptionHandler { _, throwable ->
```
dbrant, line 42: We don't need to explicitly specify `Dispatchers.IO`. Network calls will be placed on the `IO` dispatcher automatically.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
```diff
@@ -17,19 +29,39 @@ class SuggestedEditsFeedClient(
...
+                L.e(throwable)
+                cb.error(throwable)
+            }) {
+                val homeSiteCall = async { ServiceFactory.get(WikipediaApp.instance.wikiSite).getUserContributions(AccountUtil.userName, 50, null, null) }
```
dbrant, line 46: If you have only one network call, you don't need to put it into an `async { }` block and then `await` it. You can just call the function directly, and it will suspend automatically.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
```diff
@@ -17,19 +29,39 @@ class SuggestedEditsFeedClient(
...
-            DescriptionEditActivity.Action.ADD_IMAGE_TAGS
-        ), wiki, age)))
+                val usesLocalDescriptions = DescriptionEditUtil.wikiUsesLocalDescriptions(WikipediaApp.instance.wikiSite.languageCode)
+                val sufficientContributionsForArticleDescription = totalContributions > (if (usesLocalDescriptions) 50 else 3)
+                if (sufficientContributionsForArticleDescription) {
+                    cardTypes.add(DescriptionEditActivity.Action.ADD_DESCRIPTION)
+                }
+                cardTypes.add(DescriptionEditActivity.Action.ADD_CAPTION)
+                cardTypes.add(DescriptionEditActivity.Action.ADD_IMAGE_TAGS)
+                withContext(Dispatchers.Main) {
```
dbrant, line 57: You don't need to explicitly change dispatchers to `Main` here.

app/src/main/java/org/wikipedia/settings/Prefs.kt
Williamrai: done, used static variable.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
Williamrai: done.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
Williamrai: Thank you.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
Williamrai: Got it, thank you.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
Williamrai: done.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
```diff
@@ -25,14 +33,34 @@ class SuggestedEditsFeedClient(
...
-            DescriptionEditActivity.Action.ADD_IMAGE_TAGS
-        ), wiki, age)))
+        if (age == 0) {
+            cardTypes.clear()
+            clientJob = coroutineScope.launch(CoroutineExceptionHandler { _, throwable ->
+                L.e(throwable)
+                cb.error(throwable)
+            }) {
+                val homeSiteCall = ServiceFactory.get(WikipediaApp.instance.wikiSite).getUserContributions(AccountUtil.userName, 50, null, null)
+                val totalContributions = homeSiteCall.query?.userInfo?.editCount ?: 0
```
cooltey, line 43: From the ticket requirement, we should only check it if the primary language is `en`.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
```diff
@@ -25,14 +33,34 @@ class SuggestedEditsFeedClient(
...
+
+                val usesLocalDescriptions = DescriptionEditUtil.wikiUsesLocalDescriptions(WikipediaApp.instance.wikiSite.languageCode)
+                val sufficientContributionsForArticleDescription = totalContributions > (if (usesLocalDescriptions) 50 else 3)
+                if (sufficientContributionsForArticleDescription) {
```
cooltey, line 47: For the `sufficientContributionsForArticleDescription`, I am wondering if we can have this in cache so we don't need to go through API requests and the other logic every time when reach the Suggested edits card.

cooltey (changes requested):

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
Williamrai: resolved through discussion.

app/src/main/java/org/wikipedia/feed/suggestededits/SuggestedEditsFeedClient.kt
Williamrai: resolved through discussion.

cooltey (approved):

dbrant (approved):

PR #5512, opened 2025-05-01
Calendar for WikiGames Archive

Williamrai:
### What does this do?
Adds a custom background for the days in the MaterialDatePicker for WikiGames. 

**Phabricator:**
https://phabricator.wikimedia.org/T392584

### DIFF
--- a/app/src/main/java/org/wikipedia/games/onthisday/DateDecorator.kt
+++ b/app/src/main/java/org/wikipedia/games/onthisday/DateDecorator.kt
@@ -1,0 +1,75 @@
+package org.wikipedia.games.onthisday
+
+import android.content.Context
+import android.content.res.ColorStateList
+import android.os.Parcel
+import android.os.Parcelable
+import androidx.core.content.ContextCompat
+import com.google.android.material.datepicker.DayViewDecorator
+import org.wikipedia.R
+import java.util.Calendar
+
+class DateDecorator(
+    private val startDate: Calendar,
+    private val endDate: Calendar,
+    private val scoreData: Map<String, Int>
+) : DayViewDecorator() {
+
+    private val calendar = Calendar.getInstance()
+
+    private fun isDateInRange(year: Int, month: Int, day: Int): Boolean {
+        synchronized(calendar) {
+            calendar.set(year, month, day, 0, 0, 0)
+            calendar.set(Calendar.MILLISECOND, 0)
+
+            return !calendar.before(startDate) && !calendar.after(endDate)
+        }
+    }
+
+    private fun getDateKey(year: Int, month: Int, day: Int): String {
+        return "$year-$month-$day"
+    }
+
+    override fun getBackgroundColor(
+        context: Context,
+        year: Int,
+        month: Int,
+        day: Int,
+        valid: Boolean,
+        selected: Boolean
+    ): ColorStateList? {
+        if (!isDateInRange(year, month, day)) {
+            return null
+        }
+
+        val dateKey = getDateKey(year, month, day)
+        val score = scoreData[dateKey]
+
+        return when (score) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/games/onthisday/DateDecorator.kt
```diff
@@ -0,0 +1,75 @@
...
+import android.content.res.ColorStateList
+import android.os.Parcel
+import android.os.Parcelable
+import androidx.core.content.ContextCompat
+import com.google.android.material.datepicker.DayViewDecorator
+import org.wikipedia.R
+import java.util.Calendar
+
+class DateDecorator(
+    private val startDate: Calendar,
```
dbrant, line 13: Can these parameters be `Date` instead of `Calendar`?

app/src/main/java/org/wikipedia/games/onthisday/DateDecorator.kt
```diff
@@ -0,0 +1,75 @@
...
+    private fun isDateInRange(year: Int, month: Int, day: Int): Boolean {
+        synchronized(calendar) {
+            calendar.set(year, month, day, 0, 0, 0)
+            calendar.set(Calendar.MILLISECOND, 0)
+
+            return !calendar.before(startDate) && !calendar.after(endDate)
+        }
+    }
+
+    private fun getDateKey(year: Int, month: Int, day: Int): String {
```
dbrant, line 29: Can this function return a `Long` instead of a `String`?

app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameOnboardingFragment.kt
```diff
@@ -82,4 +88,46 @@ class OnThisDayGameOnboardingFragment : Fragment() {
             }
         }
     }
+
+    // helper function to create a standard key for calendar views
+    private fun getScoreDataKey(year: Int, month: Int, day: Int): String {
```
dbrant, line 93: It looks like this is a duplicate of the key function in `DateDecorator`.

dbrant (changes requested):

app/src/main/java/org/wikipedia/games/onthisday/DateDecorator.kt
Williamrai: Yes, that should work. The key could either be a combination of the year, month, and day in long format, or in date millisecond format. Is that what you had in mind?

app/src/main/java/org/wikipedia/games/onthisday/DateDecorator.kt
```diff
@@ -0,0 +1,75 @@
...
+        return "$year-$month-$day"
+    }
```
Williamrai, line 31: Yes, that should work. The key could either be a combination of the year, month, and day in long format, or in date format depending on the input. Is that what you had in mind?

dbrant: Yeah, not a big deal, but if this is intended to be a sort of "hash" function, it would be more efficient to make it a `Long`. It can work any way we like, e.g. `year * 10000 + month * 100 + day`

Williamrai: okay, sure. thank you.

app/src/main/java/org/wikipedia/games/onthisday/DateDecorator.kt
Williamrai: done.

cooltey:
@Williamrai Please resolve the conflicts.

PR #5520, opened 2025-05-06
Follow-up: viewModel for YearInReview

cooltey:
### What does this do?
This PR utilizes the coding structure to match the project by changing:
1. Creates a `viewModel` outside the `onCreate`.
2. Keeps the `YearInReviewScreenData` clean and moves the static data to the `ViewModel` in the `companion object`.
3. Update variables `private` vs `public`.
4. Removes a redundant `async` for the edit counts calls.

@voyagerfan please review and merge this PR if it looks good to you.

### DIFF
--- a/app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
+++ b/app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
@@ -19,6 +19,8 @@ import org.wikipedia.util.Resource
 
 class YearInReviewActivity : BaseActivity() {
 
+    private val viewModel: YearInReviewViewModel by viewModels()
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContent {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
voyagerfan, line 1: Hey Cooltey this looks good. I liked that you made the viewModel private.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScreenData.kt
voyagerfan, line 1: I think keeping this file strictly as a data class keeps it cleaner.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
voyagerfan, line 1: Hi Cooltey, thanks for rearranging this. I will take note of how variables are declared. Also its great that you added the init block for fetching the personalized data. I had this in earlier iterations, removed it while doing some testing, and forgot to put it back.

PR #5521, opened 2025-05-07
Create and integrate a functionality for the YiR share icon entry point - T393224

voyagerfan:
### What does this do?
This adds sharing functionality to the share icon for the Year in Review screens

### Why is this needed?
This address the share flow of the Figma design.

**Phabricator:**
https://phabricator.wikimedia.org/T393224


voyagerfan:
Currently on hold for design questions and access to Figma details.

voyagerfan:
~~Currently trouble shooting screenshot inconsistencies~~

app/build.gradle
```diff
@@ -169,6 +169,7 @@ apply from: '../gradle/src/ktlint.gradle'
 
 dependencies {
 
+
     // To keep the Maven Central dependencies up-to-date
     // use http://gradleplease.appspot.com/ or http://search.maven.org/.
```
Williamrai: could you please fix this format?

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -394,3 +527,39 @@ private fun paginationSizeGradient(totalIndicators: Int, iteration: Int, pagerSt
...
     }
 }
+
+@Composable
+fun CaptureComposableToBitmap(
+    screen: @Composable () -> Unit,
+    onBitmapReady: (Bitmap) -> Unit
+) {
+    AndroidView(
+        factory = { context ->
```
Williamrai: there’s actually another way to capture a screenshot of a Composable without using AndroidView. These two resources go over the approach:1.https://developer.android.com/develop/ui/compose/graphics/draw/modifiers#composable-to-bitmap 
2. https://stackoverflow.com/a/79233218
If you decide to try this method, it might require a few small changes to your current setup and also should probably add a loading ui.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
Williamrai, line 1: Just curious, will there always be a static image resource?

voyagerfan: I believe so. The gif resources I have come with a duplicate static image.

app/build.gradle
voyagerfan: Nice catch! will do!

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Hi @Williamrai Thanks for the links. I actually like this a bit better than the AndroidView. Since this method yields a `ImageBitmap` I had to convert it a `Bitmap` so its compatible for processing with `ShareUtil`. I also added a loading icon just in case its take a while to compose. Please let me know what you think.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -78,6 +101,24 @@ fun YearInReviewScreen(
...
+                bmp = bitmap,
+                imageFileName = "year_in_review",
+                subject = context.getString(R.string.year_in_review_share_subject),
+                text = context.getString(R.string.year_in_review_share_url)
+            )
+            startCapture = false
+        }
+    }
 
     Scaffold(
```
Williamrai: here you can create a `Box` composable and then add the `drawWithContent` modifier along with `graphics layer `
```
.drawWithContent {
                graphicsLayer.record {
                    this@drawWithContent.drawContent()
                }
                drawLayer(graphicsLayer)
            }
```
The content of the box should be the composable function that you created for taking screenshot. So basically you can remove the `graphics layer` code from the `CreateScreenShotBitmap` and place it here and leave the content like your previous function. Inside this content you can create 
`onImageLoaded: () -> Unit` and execute this whenever the asyncImage is loaded you can use `onSuccess = {}` callback of the AsyncImage. 

also consider using LaunchedEffect that runs when you start the capturing process and when the image is completely loaded and inside here you can create the bitmap. 

Let me know if that makes sense.

voyagerfan: I think I understand but I want to make sure Im not missing any context. It sounds like I'd be nesting composables and Im a little concerned about propagating the callback too deeply from within the hierarchy. What I have done thus far is added a callback and used it with `LaunchedEffect` as you mentioned - good idea by the way.

To keep things moving, I scheduled a sync for us on Monday. Definitely open to incorporating the idea, just want make sure I fully understand.

Williamrai: Yes, the changes you made looks good.

voyagerfan:
@cooltey just an FYI. I've removed the `AndroidView` and am creating the screenshot directly with the composable `CreateScreenShotBitmap`.  Also, instead of using a `delay` to ensure the composable is ready for screen capture, I've added a callback to from `SubcomposeAsyncImage`.

cooltey:
Looks good to me!

PR #5529, opened 2025-05-08
Follow-up 1: some light cleanup of Category db table.

dbrant:
This cleans up some of the naming, and simplifies a few things just a bit.


app/src/main/java/org/wikipedia/dataclient/mwapi/MwQueryPage.kt
```diff
@@ -24,7 +24,7 @@ class MwQueryPage {
     @SerialName("pageid") val pageId = 0
     @SerialName("pageprops") val pageProps: PageProps? = null
     @SerialName("entityterms") val entityTerms: EntityTerms? = null
-    @SerialName("categories") val categoriesProps: List<CategoryResponse>? = null
+    @SerialName("categories") val categories: List<CategoryResponse>? = null
```
cooltey, line 27: Do we still need the `@SerialName` if the name equals the variable name?

dbrant: haha that was my intention!

PR #5532, opened 2025-05-09
Discover pre-work: change the icons location in the readinglist header item view

cooltey:
### What does this do?
This PR moves the two icons to the bottom of the descriptions, and updates the icons color from `placeholder_color` to `primary_color`.

**Phabricator:**
https://phabricator.wikimedia.org/T393502


cooltey:
Confirmed with designer, ready for review

PR #5536, opened 2025-05-09
strings for archive game

Williamrai:
### What does this do?
Adds strings for archive games

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1862,4 +1862,11 @@
     <string name="on_this_day_game_survey_q2">Would you be interested in playing different Wikipedia games?</string>
     <string name="on_this_day_game_archive_calendar_title">Choose a game by date to play.</string>
     <string name="on_this_day_game_score_toast_message">You scored %1$d/%2$d on this day.</string>
+    <string name="on_this_day_game_continue_btn_text">Continue today\’s game</string>
+    <string name="on_this_day_game_current_progress_message">You\'re on question %d.</string>
+    <string name="on_this_day_game_archive_btn_text">Play the archive</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1862,4 +1862,11 @@
     <string name="on_this_day_game_survey_q2">Would you be interested in playing different Wikipedia games?</string>
     <string name="on_this_day_game_archive_calendar_title">Choose a game by date to play.</string>
     <string name="on_this_day_game_score_toast_message">You scored %1$d/%2$d on this day.</string>
+    <string name="on_this_day_game_continue_btn_text">Continue today\’s game</string>
```
cooltey, line 1865: Please use `\'` instead of `\’`, that's slightly different 😅

app/src/main/res/values/strings.xml
```diff
@@ -1862,4 +1862,11 @@
...
+    <string name="on_this_day_game_current_progress_message">You\'re on question %d.</string>
+    <string name="on_this_day_game_archive_btn_text">Play the archive</string>
+    <string name="on_this_day_game_review_results_btn_text">Review results</string>
+    <string name="on_this_day_game_score_message">You scored %1$d/%2$d on today\'s game.</string>
```
cooltey, line 1869: Is this a duplicate of `on_this_day_game_score_toast_message`? The only difference is `on today's` vs `on this`, but I do not see this on Figma.

cooltey (changes requested):

app/src/main/res/values/strings.xml
Williamrai: done.

cooltey (approved):

PR #5537, opened 2025-05-09
Integrate DonateDialog class into YiR donate button - T393815

voyagerfan:
### What does this do?
This PR integrates the `DonateDialog` class with the donate button in the year in review `MainBottomBar`

### Why is this needed?
This adds the ability for a user to donate.

**Phabricator:**
https://phabricator.wikimedia.org/T393815

### DIFF
--- a/app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
+++ b/app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
@@ -54,11 +54,14 @@ import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.viewinterop.AndroidView
+import androidx.fragment.app.FragmentActivity
 import androidx.navigation.NavHostController
 import com.bumptech.glide.Glide
 import kotlinx.coroutines.launch
 import org.wikipedia.R
 import org.wikipedia.compose.theme.WikipediaTheme
+import org.wikipedia.donate.DonateDialog
+import org.wikipedia.page.ExclusiveBottomSheetPresenter
 import kotlin.math.absoluteValue
 
 @OptIn(ExperimentalMaterial3Api::class)
@@ -139,6 +142,8 @@ fun MainBottomBar(
     pagerState: PagerState,
     totalPages: Int
 ) {
+    val context = LocalContext.current
+    val donateActivity = context as? FragmentActivity
     Column {
         HorizontalDivider(
             modifier = Modifier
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -156,7 +161,16 @@ fun MainBottomBar(
...
+                                donateActivity?.let {
+                                    ExclusiveBottomSheetPresenter.show(
+                                        it.supportFragmentManager,
+                                        DonateDialog.newInstance(
+                                            campaignId = "enUS_appmenu_yir_android"
+                                        )
+                                    )
+                                }
+                            })
                             .padding(start = 15.dp)
```
Williamrai, line 174: It might be worth reconsidering this approach, as directly referencing activities within a composable could lead to unintended side effects, especially during recomposition. It could also limit reusability and could make it harder to  implement proper unidirectional data flow pattern where events should go up to the parent level rather than being handled directly within the composable.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -139,6 +142,8 @@ fun MainBottomBar(
     pagerState: PagerState,
     totalPages: Int
 ) {
+    val context = LocalContext.current
+    val donateActivity = context as? FragmentActivity
```
cooltey, line 146: In `BaseActivity`, we have a function called `launchDonateDialog`, you can use it directly by passing `BaseActivity` to the method.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -156,7 +161,16 @@ fun MainBottomBar(
                 ) {
                     Row(
                         modifier = Modifier
-                            .clickable(onClick = { /* TODO() */ })
+                            .clickable(onClick = {
+                                donateActivity?.let {
+                                    ExclusiveBottomSheetPresenter.show(
+                                        it.supportFragmentManager,
+                                        DonateDialog.newInstance(
+                                            campaignId = "enUS_appmenu_yir_android"
```
cooltey, line 169: Instead of giving a complete combined string for the campaign ID, we should send `yir` here and let `CampaignCollection.getFormattedCampaignId` to do the rest. 

Please make sure the `yir` is defined as a constant value.

cooltey (changes requested):

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Hi @Williamrai, I hadn't considered recomposition when casting the context as a fragment activity. What you're saying makes total sense, thanks for pointing that out. Originally I was trying to emulate the button click functionality from `launchDonateDialog`.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Hi @cooltey, thanks for the suggestion. I noticed that `BaseActivity` is abstract, so what I ended up doing was passing an instance of `this@YearInReviewActivity` to the `MainBottomBar` composable, and then calling `launchDonateDialog` from there. Let me know if this approach works for you — happy to adjust if there's a better way.

We can also pass launchDonateDialog("yir") straight into the `MainBottomBar` from `MainActivity`. I have a working version of this too if we want to discuss later.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Hi Cooltey, I have made this change. Thanks for catching this one.

PR #5541, opened 2025-05-12
Add basic instrumentation to YiR screens - T392406

voyagerfan:
### What does this do?
This PR adds breadcrumbing and donor event logging

### Why is this needed?
Part of Epic deliverables

**Phabricator:**
https://phabricator.wikimedia.org/T392406


app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -159,7 +162,19 @@ fun MainBottomBar(
                             .clickable(onClick = { /* TODO() */ })
                             .padding(start = 15.dp)
                             .wrapContentWidth()
-                            .align(Alignment.CenterStart),
+                            .align(Alignment.CenterStart)
+                            .clickable(onClick = {
```
cooltey, line 166: I noticed another `clickable()` exists. Is there any reason why you created a new one here?

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -159,7 +162,19 @@ fun MainBottomBar(
...
+                                EventPlatformClient.submit(
+                                    BreadCrumbLogEvent(
+                                        screen_name = "year_in_review",
+                                        action = "donate_click")
+                                )
+                                DonorExperienceEvent.logAction(
+                                    action = "donate_start_click_yir",
+                                    activeInterface = "wiki_yir",
+                                    campaignId = "enUS_appmenu_yir_android"
+                                )
```
cooltey, line 176: Not sure, but have you confirmed with Shay that we have to send both breadcrumb and donor experience events at a single click?

cooltey (changes requested):

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Hi @cooltey, this duplicate was an oversight on my part. I had left a `/* TODO() */` as a placeholder, but unfortunately didn’t follow up with the actual implementation. I've corrected it now—thanks for catching it.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Hey @cooltey, I confirmed with Shay that we should be sending both.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -157,10 +163,26 @@ fun MainBottomBar(
...
+                                    EventPlatformClient.submit(
+                                        BreadCrumbLogEvent(
+                                            screen_name = "year_in_review",
+                                            action = "donate_click")
+                                    )
+                                    DonorExperienceEvent.logAction(
+                                        action = "donate_start_click_yir",
+                                        activeInterface = "wiki_yir",
+                                        campaignId = "yir"
+                                    )
```
cooltey, line 182: Since you send the `onDonateClick()` to this clickable block, how about moving these events to `YearInReviewActivity`?

voyagerfan: Hi @cooltey. This makes total sense - its much better to have all of the click behavior in one location. I've moved both events to YearInReviewActivity

PR #5551, opened 2025-05-13
Further solidify detection of temp account name.

dbrant:
Some more solid logic for correctly detecting a temp account name.
Now with tests!

### DIFF
--- a/app/src/main/java/org/wikipedia/auth/AccountUtil.kt
+++ b/app/src/main/java/org/wikipedia/auth/AccountUtil.kt
@@ -6,6 +6,7 @@ import android.accounts.AccountManager
 import android.app.Activity
 import android.os.Build
 import androidx.core.os.bundleOf
+import androidx.core.text.isDigitsOnly
 import org.wikipedia.R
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/auth/AccountUtil.kt
```diff
@@ -125,7 +126,7 @@ object AccountUtil {
     }
 
     fun isUserNameTemporary(userName: String): Boolean {
-        return userName.startsWith("~")
+        return userName.length > 6 && userName[0] == '~' && userName[5] == '-' && userName.substring(1, 5).isDigitsOnly()
```
cooltey, line 129: Do you have the references for the username policy or naming rules?

dbrant: Sure: https://www.mediawiki.org/wiki/Help:Temporary_accounts#How_do_temporary_accounts_work?

cooltey: Thanks for the reference!

PR #5562, opened 2025-05-15
Fix: make sure to compare the correct value for updating the list in ReadingListsFragment

cooltey:
### What does this do?
When you have a fresh install, the thumbnail on the list does not get updated correctly, and that's because we have always been comparing the same `numPagesOffline` by itself!


dbrant:
🤯

PR #5570, opened 2025-05-16
Create recurring task for Recommended Reading list

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T394102


cooltey:
@dbrant @Williamrai This is ready for review, and we can always revisit the algorithm anytime if we need to update it.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListViewModel.kt
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private val _uiState = MutableStateFlow(Resource<Unit>())
+    val uiState = _uiState.asStateFlow()
+
+    companion object {
+
+        private const val MAX_RETRIES = 10
+
+        suspend fun generateRecommendedReadingList() {
```
dbrant, line 29: Would it make sense to put all of this logic into the `RecommendedReadingListTask` class instead? (Does the ViewModel need to know about any of this?)

app/src/main/java/org/wikipedia/history/db/HistoryEntryDao.kt
```diff
@@ -18,6 +18,9 @@ interface HistoryEntryDao {
     @Query("SELECT * FROM HistoryEntry WHERE authority = :authority AND lang = :lang AND apiTitle = :apiTitle LIMIT 1")
     suspend fun findEntryBy(authority: String, lang: String, apiTitle: String): HistoryEntry?
 
+    @Query("SELECT * FROM HistoryEntry ORDER BY random()  DESC LIMIT :limit")
```
dbrant, line 21: Extremely minor: double space after random(), and also, for maximum consistency, can we make it uppercase RANDOM()?

dbrant (changes requested):

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListViewModel.kt
cooltey: We will need to generate a new list when the user completes the onboarding process.

app/src/main/java/org/wikipedia/history/db/HistoryEntryDao.kt
cooltey: Done!

PR #5572, opened 2025-05-20
Create onboarding message card for Recommended Reading list

cooltey:
### What does this do?
Use the current `onboardingView` and add a `messageLabel` to the `MessageCardView` for the "NEW!" label.

Pending discussion: Discuss the priorities of "Turn on reading list syncing," "Log in / join Wikipedia," and this card.

**Phabricator:**
https://phabricator.wikimedia.org/T394028


app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
```diff
@@ -811,6 +813,21 @@ class ReadingListsFragment : Fragment(), SortReadingListsDialog.Callback, Readin
...
             binding.onboardingView.isVisible = true
+        } else if (!Prefs.isRecommendedReadingListOnboardingShown && !Prefs.isRecommendedReadingListEnabled) {
+            binding.onboardingView.setMessageLabel(getString(R.string.recommended_reading_list_onboarding_card_new))
+            binding.onboardingView.setMessageTitle(getString(R.string.recommended_reading_list_onboarding_card_title))
+            binding.onboardingView.setMessageText(getString(R.string.recommended_reading_list_onboarding_card_message))
+            binding.onboardingView.setPositiveButton(R.string.recommended_reading_list_onboarding_card_positive_button, {
+                // TODO: open Recommended reading list onboarding page
+                Prefs.isRecommendedReadingListOnboardingShown = true
+            }, false)
+            binding.onboardingView.setNegativeButton(R.string.recommended_reading_list_onboarding_card_negative_button, {
```
Williamrai, line 824: I think `applyListenerToContainer` must be set to true, otherwise, the previous listener may not be properly removed. For example, if the sync reading list login message card was added first, its listener could remain active because it's not being overridden by the Discover click listener.

cooltey: Sure, I'll set it to `true`! (and I am going to create another branch since I got typo on this one 😓 )

cooltey:
Rename the branch. #5583

PR #5579, opened 2025-05-20
Add clickable functionality for "Learn More" and Info Icon buttons - T394815

voyagerfan:
### What does this do?
This PR will add the same clickable functionality to both the "Learn More" button for the onboarding page and the Info Icon in the following screen deck slides.

### Why is this needed?
This supports EPIC T387034

**Phabricator:**
https://phabricator.wikimedia.org/T394815


voyagerfan:
Hi @cooltey, this is ready for review when you have a moment.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
```diff
@@ -44,19 +44,23 @@ class YearInReviewActivity : BaseActivity() {
...
-                            customBottomBar = {
+                            customBottomBar = { pagerState, context ->
                                 OnboardingBottomBar(
                                     onGetStartedClick = {
                                         navController.navigate(
                                             route = YearInReviewNavigation.ScreenDeck.name
                                         )
-                                    }
+                                    },
+                                    context = context
```
cooltey, line 54: Instead of passing `context` layer by layer, can we send `this@YearInReviewActivity` here directly?

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
```diff
@@ -44,19 +44,23 @@ class YearInReviewActivity : BaseActivity() {
...
                                 )
                             },
-                            screenContent = { innerPadding, contentData ->
+                            screenContent = { innerPadding, contentData, context ->
                                 YearInReviewScreenContent(
                                     innerPadding = innerPadding,
-                                    screenData = contentData)
+                                    screenData = contentData,
+                                    context = context,
```
cooltey, line 61: Same here.  Can we send `this@YearInReviewActivity` here directly?

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -251,7 +255,12 @@ fun OnboardingBottomBar(
                     modifier = Modifier
                         .width(152.dp)
                         .height(42.dp),
-                    onClick = { /* TODO() */ }
+                    onClick = {
+                        UriUtil.handleExternalLink(
+                            context = context,
+                            uri = "https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/iOS/Personalized_Wikipedia_Year_in_Review/How_your_data_is_used".toUri()
```
cooltey, line 261: It would be better to put the URL in the XML instead of hard-coding it in the view.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -324,13 +334,20 @@ fun YearInReviewScreenContent(
...
-                        painter = painterResource(R.drawable.baseline_info_24),
-                        tint = WikipediaTheme.colors.primaryColor,
-                        contentDescription = stringResource(R.string.year_in_review_information_icon)
-                    )
+                if (isInfoIconVisible) {
+                    IconButton(
+                        onClick = {
+                            UriUtil.handleExternalLink(
+                                context = context,
+                                uri = "https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/iOS/Personalized_Wikipedia_Year_in_Review/How_your_data_is_used".toUri()
```
cooltey, line 342: Same here.

cooltey (changes requested):

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Hey Cooltey, totally agreed. When you mentioned this, I went and took a look at the strings and noticed examples of URL's there. I followed those implementations but the IDE suggested to do:

`uri = context.getString(R.string.year_in_review_media_wiki_url).toUri()`

instead of other implementations I see like:

`uri = Uri.parse(getString(R.string.create_account_ip_block_help_url)))`

Is either OK to use?

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Agreed! same reply from above.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
voyagerfan: Hi Cooltey, I think we can do this but it may come with some tradeoffs in terms of reusability, testing, previews etc. We can discuss more in detail at our next meeting

cooltey: Per our discussion, I think it would be fine if you could have a context parameter in the Compose screen functions, just like I showed during the meeting. Please try to see if that works for you to make sure the preview can work correctly. Thanks!

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
cooltey: Please follow the IDE suggestion, thanks!

voyagerfan: Understood. I've implemented it the IDE suggestion

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
voyagerfan: Thanks for the productive talk today. I made the changes (removed context from the child composables and replaced them with `this@YearInReviewActivity`, then hardcoded `LocalContext.current` into the parent `YearInReviewScreen`

Unfortunately I cant test right now because of the Coil update but as soon as thats resolved and integrated, I will test to make sure the previews work.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
voyagerfan: Hey Cooltey, absolutely! I've implemented this as well.

voyagerfan:
Holding until testing can be done after reworking to use coil

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
```diff
@@ -47,19 +47,23 @@ class YearInReviewActivity : BaseActivity() {
                         YearInReviewScreen(
                             contentData = listOf(YearInReviewViewModel.getStartedData),
                             navController = navController,
-                            customBottomBar = {
+                            customBottomBar = { pagerState ->
```
cooltey, line 50: The `pagerState` can be removed since it is not used here.

app/src/main/res/values/strings.xml
```diff
@@ -1908,6 +1908,7 @@
     <string name="year_in_review_non_english_collective_readcount_bodytext">Wikipedia had more than 63 million articles across over 300 active languages this year. You joined millions in expanding knowledge and exploring diverse topics</string>
     <string name="year_in_review_non_english_collective_editcount_headline">Editors on the iOS app made more than 452,257 edits</string>
     <string name="year_in_review_non_english_collective_editcount_bodytext">Wikipedia\'s community of volunteer editors made more than 452,257 edits on the iOS app so far this year. The heart and soul of Wikipedia is our global community of volunteer contributors, donors, and billions of readers like yourself – all united to share unlimited access to reliable information.</string>
+    <string name="year_in_review_media_wiki_url">https://www.mediawiki.org/wiki/Wikimedia_Apps/Team/iOS/Personalized_Wikipedia_Year_in_Review/How_your_data_is_used</string>
```
cooltey, line 1911: As we discussed before, this string can be put in a `strings_no_translate.xml` since the URL does not require translation.

And, please confirm with PM/designer about the correct URL, because the one you have here is for the iOS app.

cooltey (changes requested):

app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
voyagerfan: Understood and fixed.

app/src/main/res/values/strings.xml
voyagerfan: Sorry for the misunderstanding — I remember us discussing using `strings_no_translate.xml`, but I didn’t see it specifically noted in the earlier comments. Funny enough, it’s right below `strings.xml`, so I’ve gone ahead and made the change. 

Also, I've confirmed with PM that its OK to use the iOS link for now but it we'll to check in with the PM for a new link before release.

PR #5581, opened 2025-05-20
Notification Recommended reading list

Williamrai:
### What does this do?
Adds notification function for recommended reading list

**Phabricator:**
https://phabricator.wikimedia.org/T394101


app/src/main/java/org/wikipedia/settings/RecommendedReadingListNotificationManager.kt
```diff
@@ -0,0 +1,82 @@
...
+            RecommendedReadingListUpdateFrequency.DAILY -> context.getString(R.string.recommended_reading_list_settings_updates_frequency_daily)
+            RecommendedReadingListUpdateFrequency.WEEKLY -> context.getString(R.string.recommended_reading_list_settings_updates_frequency_weekly)
+            RecommendedReadingListUpdateFrequency.MONTHLY -> context.getString(R.string.recommended_reading_list_settings_updates_frequency_monthly)
+        }
+        val title = context.getString(R.string.recommended_reading_list_notification_title, frequency)
+        NotificationPresenter.showNotification(
+            context = context,
+            builder = NotificationPresenter.getDefaultBuilder(context, 1, NOTIFICATION_TYPE_LOCAL),
+            id = 1,
+            title = title,
```
cooltey, line 48: Set the title (`context.getString(R.string.recommended_reading_list_notification_title, frequency)`) to here directly?

app/src/main/res/xml/developer_preferences.xml
```diff
@@ -508,4 +508,12 @@
             android:title="Category" />
     </PreferenceCategory>
 
+    <PreferenceCategory android:title="Recommended Reading List">
```
cooltey, line 511: Can we have this at the top of the `Playground` section so we can always have the `Playground` at the bottom of the list?

cooltey (commented):
Wondering about the naming: Discover vs. Recommended Reading list...maybe we can discuss it again in the engineers sync.

Williamrai:
> Wondering about the naming: Discover vs. Recommended Reading list...maybe we can discuss it again in the engineers sync.

are you referring to this name ACTION_DISCOVER_READING_LIST?

cooltey:
> > Wondering about the naming: Discover vs. Recommended Reading list...maybe we can discuss it again in the engineers sync.
> 
> are you referring to this name ACTION_DISCOVER_READING_LIST?

Not only this but also other functions that include "discover". For consistency, I think we should use `recommended_readin_list", but I personally like to use "discover". cc @dbrant

app/src/main/java/org/wikipedia/settings/dev/DeveloperSettingsPreferenceLoader.kt
```diff
@@ -214,6 +216,17 @@ internal class DeveloperSettingsPreferenceLoader(fragment: PreferenceFragmentCom
             activity.startActivity(Intent(activity, CategoryDeveloperPlayGround::class.java))
             true
         }
+        findPreference(R.string.preference_key_recommended_reading_list_notification_enabled)
+            .setOnPreferenceChangeListener { _, newValue ->
+                val selectedFrequency = newValue as String
+                val source = when (selectedFrequency) {
```
dbrant, line 222: We already have an enum called `RecommendedReadingListUpdateFrequency` that defines these frequencies. Can we use the enum here (its values _and names_!) instead of hardcoding more strings and arrays?

app/src/main/java/org/wikipedia/settings/RecommendedReadingListNotificationManager.kt
```diff
@@ -0,0 +1,81 @@
...
+
+    fun scheduleRecommendedReadingListNotification(context: Context) {
+        val alarmManager = context.getSystemService(Context.ALARM_SERVICE) as AlarmManager
+        val intent = Intent(context, NotificationPollBroadcastReceiver::class.java)
+            .setAction(ACTION_RECOMMENDED_READING_LIST)
+        val triggerUpdateMillis = System.currentTimeMillis() + timeUntilNextUpdate(Prefs.recommendedReadingListUpdateFrequency).toMillis()
+        val nextUpdateIntervalInMillis = when (Prefs.recommendedReadingListUpdateFrequency) {
+            RecommendedReadingListUpdateFrequency.DAILY -> AlarmManager.INTERVAL_DAY
+            RecommendedReadingListUpdateFrequency.WEEKLY -> 7 * AlarmManager.INTERVAL_DAY
+            RecommendedReadingListUpdateFrequency.MONTHLY -> 30 * AlarmManager.INTERVAL_DAY
```
dbrant, line 27: Hardcoding 30 days per month feels inaccurate. Can we use similar logic to what's used in `RecommendedReadingListTask`?

dbrant (changes requested):

app/src/main/java/org/wikipedia/settings/RecommendedReadingListNotificationManager.kt
cooltey: I think that's a good approach.

PR #5582, opened 2025-05-21
Build screen for selecting the source type for Recommended Reading list

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T394026

### DIFF
--- a/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceScreen.kt
+++ b/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceScreen.kt
@@ -103,12 +103,12 @@ fun SourceSelectionScreen(
                 Box(
                     modifier = Modifier
                         .fillMaxWidth()
-                        .padding(paddingValues)
                         .padding(top = 64.dp),
-                    contentAlignment = Alignment.Center
                 ) {
                     LinearProgressIndicator(
+                        modifier = Modifier.fillMaxWidth(),
                         color = WikipediaTheme.colors.progressiveColor,
+                        trackColor = WikipediaTheme.colors.borderColor
                     )
                 }
             }


cooltey:
The PR is ready for review. It would be better if we merge this to the feature branch before sending to design review, since the onboarding card is not always visible.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceScreen.kt
```diff
@@ -0,0 +1,303 @@
...
+            }
+        }
+
+        is Resource.Error -> {
+            WikiErrorView(
+                modifier = Modifier
+                    .fillMaxWidth(),
+                caught = uiState.throwable,
+                errorClickEvents = wikiErrorClickEvents
+            )
```
Williamrai, line 85: I think we should center the wikiErrorView. It's currently positioned at the top.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceScreen.kt
```diff
@@ -0,0 +1,303 @@
...
+    fromSettings: Boolean,
+    selectedSource: RecommendedReadingListSource,
+    isSavedOptionEnabled: Boolean,
+    isHistoryOptionEnabled: Boolean,
+    onCloseClick: () -> Unit = {},
+    onInterestsClick: () -> Unit = {},
+    onSavedClick: () -> Unit = {},
+    onHistoryClick: () -> Unit = {},
+    onNextClick: () -> Unit = {}
+) {
```
Williamrai, line 116: I think we can reduce the number of  click events by using a mutable state here  `var source by remember { mutableStateOf(selectedSource) }` and replacing all the events with a single  `onSourceClick: (RecommendedReadingListSource) -> Unit,`

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListOnboardingActivity.kt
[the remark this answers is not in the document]
cooltey: Per the discussion in the meeting, this is what we usually do in the project, to create a single Activity for handling the same purpose - in this example, for the onboarding process.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceScreen.kt
```diff
@@ -0,0 +1,303 @@
...
+            contentColor = WikipediaTheme.colors.paperColor
+        ),
+        elevation = CardDefaults.cardElevation(defaultElevation = 0.dp),
+        border = BorderStroke(
+            width = 1.dp,
+            color = WikipediaTheme.colors.borderColor
+        )
+    ) {
+        ListItem(
+            modifier = modifier
```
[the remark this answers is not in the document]
cooltey: Done!

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceScreen.kt
cooltey: Done!

cooltey:
This is ready for review again, and has passed the tests locally.

app/src/main/java/org/wikipedia/compose/extensions/Modifier.kt
```diff
@@ -60,6 +63,16 @@ fun Modifier.pulse(
     }
 }
 
+fun Modifier.clickableWithRipple(
+    onClick: () -> Unit
+): Modifier = composed {
```
dbrant, line 68: Are we totally sure we need this?
As I understand it, we would just need to add `indication = ripple()` wherever we want a ripple.
This seems to instantiate quite a bit of additional logic (another `composed{}` block, and another `MutableInteractionSource`), which feels unnecessary.

dbrant (changes requested):

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceScreen.kt
```diff
@@ -0,0 +1,317 @@
...
+                    modifier = Modifier
+                        .fillMaxWidth()
+                ) {
+                    Spacer(
+                        modifier = Modifier
+                            .fillMaxWidth()
+                            .height(1.dp)
+                            .background(WikipediaTheme.colors.borderColor)
+                    )
+                    Icon(
```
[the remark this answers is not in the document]
cooltey: Done!

PR #5588, opened 2025-05-22
Refactor YearInReview package to use Coil AsyncImage composable - T395048

voyagerfan:
### What does this do?
This PR adds replaces AndroidView with AsyncImage composable to display gifs for the Year in Review Feature.

### Why is this needed?
Year in Review Gifs are currently loaded using Glide/Bumptech but these dependencies have been replaced by Coil

**Phabricator:**
https://phabricator.wikimedia.org/T395048


voyagerfan:
Hi @cooltey, this is PR is ready for review when you have a moment. Thanks!

app/src/main/res/values/strings.xml
```diff
@@ -1889,13 +1889,15 @@
     <string name="year_in_review_entry_dialog_bodytext">See insights about what articles we read and edited, and share highlights from our year on Wikipedia.</string>
     <string name="year_in_review_entry_dialog_continue">Continue</string>
     <string name="year_in_review_read_count_headline">You read %s articles this year</string>
-    <string name="year_in_review_read_count_bodytext">This year, you read %s articles, including %s, %s, and %s. Wikipedia had more than 63 million articles available across over 300 languages. You joined millions in expanding knowledge and exploring diverse topics.</string>
+    <string name="year_in_review_read_count_bodytext">This year, you read %1\$s articles, including %2\$s, %3\$s, and %4\$s. Wikipedia had more than 63 million articles available across over 300 languages. You joined millions in expanding knowledge and exploring diverse topics.</string>
```
dbrant, line 1892: * This should be a `plurals` resource, instead of a plain `string`. Many languages actually conjugate the plural noun `articles` based on the associated number. For example, in Russian, if a number ends in 1 (even if it's 1000001), the noun becomes singular!
  * This should probably also be done for the string above this one (`year_in_review_read_count_headline`).
* There should not be a backslash in the parameters (`%1$s`)
* The number of articles should be a `%1$d` parameter instead of `%1$s`, since it's a number.

dbrant (changes requested):

app/src/main/res/values/strings.xml
voyagerfan: Hey @dbrant Thanks for your comments.

- I agree with using `plural` however I believe the default behavior at this time is that personalized strings will always be plural. My understanding from the iOS version is if an API call or DB query returns less than 3 results, then we fallback to a hardcoded page.  If the API/DB returns a count of 3 or more, we display the count and 3 strings. Im happy to make changes if we want to proactively support flexible use cases.
- I've removed the slashes [x] .
- Nice catch on the %d vs %s - I've made the change [x]. I believe this an artifact from an early version of the viewModel where I was using strings for everything.

dbrant: > * I agree with using `plural` however I believe the default behavior at this time is that personalized strings will always be plural. My understanding from the iOS version is if an API call or DB query returns less than 3 results, then we fallback to a hardcoded page.  If the API/DB returns a count of 3 or more, we display the count and 3 strings. Im happy to make changes if we want to proactively support flexible use cases.

Right; I meant to say that even when the number is greater than 3, certain languages conjugate nouns differently for different types of numbers. In Russian, if a number ends in 1 (not just 1, but any number _ending in 1_, like 7654321), the noun becomes singular, not plural. A similar thing happens in numerous other languages.

PR #5590, opened 2025-05-22
Discover message card design

Williamrai:
### What does this do?
- adds recommended reading list discover card 

**Phabricator:**
https://phabricator.wikimedia.org/T394055


app/src/main/java/org/wikipedia/readinglist/DiscoverReadingListView.kt
```diff
@@ -0,0 +1,154 @@
...
+    WikiCard(
+        elevation = 4.dp
+    ) {
+        Row(
+            modifier = modifier,
+            horizontalArrangement = Arrangement.spacedBy(16.dp),
+        ) {
+            Column(
+                modifier = Modifier
+                    .weight(1.1f),
```
voyagerfan, line 54: Not sure but should this be a fixed dimension based on the design?

app/src/main/java/org/wikipedia/readinglist/DiscoverReadingListView.kt
```diff
@@ -0,0 +1,154 @@
...
+                            )
+                        } else {
+                            AsyncImage(
+                                model = imageUrl,
+                                modifier = Modifier
+                                    .size(70.dp),
+                                contentScale = ContentScale.Crop,
+                                contentDescription = null,
+                                placeholder = ColorPainter(WikipediaTheme.colors.borderColor),
+                                error = ColorPainter(WikipediaTheme.colors.borderColor),
```
voyagerfan, line 146: Hi @Williamrai This looks like a fallback which is a great idea. I was wondering if the color was ok to use? I couldnt find an example of a placeholder/error color.

app/src/main/java/org/wikipedia/compose/extensions/Modifier.kt
```diff
@@ -60,6 +63,23 @@ fun Modifier.pulse(
     }
 }
 
+fun Modifier.rippleClickable(
```
voyagerfan, line 66: Hi @Williamrai I like this extension function! It definitely provides consistency for ripple behavior when clicking. One thought I had was the usage of the function. If we plan to use it a lot then its definitely worth keeping but if it sees limited use then perhaps it make may be easier to use `modifier.clickable()` and set the parameters there?  

This isn't a request to change anything right now. Overall, it’s a good idea and a solid addition!

app/src/main/java/org/wikipedia/readinglist/DiscoverReadingListView.kt
Williamrai: This weight is for horizontal distribution. I'm not using a 1 to 1  because it makes the image slightly wider, which doesn’t look visually good.

app/src/main/java/org/wikipedia/readinglist/DiscoverReadingListView.kt
```diff
@@ -0,0 +1,154 @@
...
+                    fontWeight = FontWeight.Normal,
+                    fontSize = 12.sp,
+                    color = WikipediaTheme.colors.primaryColor
+                )
+            }
+
+            LazyVerticalGrid(
+                modifier = Modifier
+                    .clip(RoundedCornerShape(12.dp))
+                    .weight(0.9f),
```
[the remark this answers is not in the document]
Williamrai: yes it is okay to use weights.

app/src/main/java/org/wikipedia/readinglist/DiscoverReadingListView.kt
Williamrai: I'm using the default color we typically use for placeholders, and I'm also applying the same color for errors. You can refer to the `loadImage` function in the `CoilImageServiceLoader` class.

app/src/main/java/org/wikipedia/compose/extensions/Modifier.kt
Williamrai: A similar function was created by Cooltey in one of his PRs, so once the PR is merged, we'll use one of the two implementations.

app/src/main/java/org/wikipedia/readinglist/DiscoverReadingListView.kt
```diff
@@ -0,0 +1,154 @@
...
+import androidx.compose.ui.unit.sp
+import coil3.compose.AsyncImage
+import org.wikipedia.compose.components.WikiCard
+import org.wikipedia.compose.theme.WikipediaTheme
+
+@Composable
+fun DiscoverReadingListView(
+    modifier: Modifier = Modifier,
+    title: String,
+    @DrawableRes subtitleIcon: Int,
```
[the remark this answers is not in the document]
Williamrai: Yes, we can do that, but I believe it's helpful to keep this as a parameter. This way, the icon can be modified directly from the caller without needing to dive into the implementation. It also allows flexibility in the future if we decide to use different icons for different scenarios.

app/src/main/java/org/wikipedia/readinglist/ReadingListsFragment.kt
```diff
@@ -842,6 +862,43 @@ class ReadingListsFragment : Fragment(), SortReadingListsDialog.Callback, Readin
...
+                RecommendedReadingListUpdateFrequency.WEEKLY -> R.string.recommended_reading_list_page_description_weekly
+                RecommendedReadingListUpdateFrequency.MONTHLY -> R.string.recommended_reading_list_page_description_monthly
+            }
+            BaseTheme {
+                RecommendedReadingListDiscoverCardView(
+                    modifier = Modifier
+                        .clickable {
+                           // @TODO: open recommended list screen
+                            isNewListGenerated = false
+                            Prefs.isNewRecommendedReadingListGenerated = false
```
cooltey, line 887: Instead of having a new preference, can we reuse the value below?
```
val recommendedArticles = RecommendedReadingListViewModel.getNewRecommendedArticles()
```

cooltey:
Merge this and welcome to review it in the feature branch later.

PR #5594, opened 2025-05-27
Delete categories in batches

Williamrai:
### What does this do?
Deletes categories in batches and adds this function to the category dev playground

### DIFF
--- a/app/src/main/java/org/wikipedia/categories/db/CategoryDao.kt
+++ b/app/src/main/java/org/wikipedia/categories/db/CategoryDao.kt
@@ -25,4 +25,7 @@ interface CategoryDao {
 
     @Query("DELETE FROM Category WHERE timeStamp < :timeStamp")
     suspend fun deleteOlderThan(timeStamp: Long)
+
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/categories/db/CategoryDao.kt
```diff
@@ -25,4 +25,7 @@ interface CategoryDao {
 
     @Query("DELETE FROM Category WHERE timeStamp < :timeStamp")
     suspend fun deleteOlderThan(timeStamp: Long)
```
cooltey, line 27: If it has no usage, do you still need to keep it?

app/src/main/java/org/wikipedia/settings/dev/playground/CategoryDeveloperPlayGroundViewModel.kt
```diff
@@ -135,7 +137,7 @@ class CategoryDeveloperPlayGroundViewModel : ViewModel() {
             L.e(throwable)
         }) {
             try {
-                AppDatabase.instance.categoryDao().deleteAll()
+                deleteOldDataInBatches(System.currentTimeMillis(), 100)
```
cooltey, line 140: Any reason why you changed to use `deleteOldDataInBatches` instead of `deleteAll`?

PR #5597, opened 2025-05-28
Simplify detection of Dark theme in WikiCard composable.

dbrant:
This removes the necessity to pass around an `isDarkTheme` parameter wherever a `WikiCard` composable is used.

The only purpose of this parameter is whether or not to apply an elevation to the card, and the reason it became a parameter is to make Previews work, since a Preview is not aware of `WikipediaApp.instance`.

But now, we can encode directly into the `WikipediaTheme` composable whether the current theme is Dark or not, and remove the actual dependency on `WikipediaApp.instance`.

### DIFF
--- a/app/src/main/java/org/wikipedia/compose/theme/WikipediaTheme.kt
+++ b/app/src/main/java/org/wikipedia/compose/theme/WikipediaTheme.kt
@@ -38,8 +38,4 @@ object WikipediaTheme {
     val typography: WikipediaTypography
         @Composable
         get() = LocalWikipediaTypography.current
-
-    val isDark: Boolean
-        @Composable
-        get() = LocalWikipediaColor.current.isDarkTheme
 }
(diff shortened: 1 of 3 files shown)


cooltey (approved):
Works for me.

app/src/main/java/org/wikipedia/compose/theme/WikipediaTheme.kt
```diff
@@ -38,4 +38,8 @@ object WikipediaTheme {
     val typography: WikipediaTypography
         @Composable
         get() = LocalWikipediaTypography.current
+
+    val isDark: Boolean
+        @Composable
+        get() = LocalWikipediaColor.current.isDarkTheme
 }
```
Williamrai, line 45: do we need a separate `isDark`, it is accessible through colors.isDarkTheme?

PR #5602, opened 2025-05-30
Follow-up: Discover settings for Recommended reading list

cooltey:
- Move around the classes and rename classes a bit.
- Update parameters naming style.
- Simplify the way of accessing the source fragment / and the upcoming interests fragment by using the `RecommendedReadingListOnboardingActivity`
- Add ripple animation for dialog radio items and update the frequency item.

### DIFF
--- a/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListOnboardingActivity.kt
+++ b/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListOnboardingActivity.kt
@@ -2,17 +2,32 @@ package org.wikipedia.readinglist.recommended
 
 import android.content.Context
 import android.content.Intent
+import androidx.fragment.app.Fragment
 import org.wikipedia.activity.SingleFragmentActivity
 
-class RecommendedReadingListOnboardingActivity : SingleFragmentActivity<RecommendedReadingListSourceFragment>() {
+class RecommendedReadingListOnboardingActivity : SingleFragmentActivity<Fragment>() {
 
-    public override fun createFragment(): RecommendedReadingListSourceFragment {
-        return RecommendedReadingListSourceFragment.newInstance()
+    public override fun createFragment(): Fragment {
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListOnboardingActivity.kt
```diff
@@ -2,17 +2,32 @@ package org.wikipedia.readinglist.recommended
...
+    public override fun createFragment(): Fragment {
+        val fromSetting = intent.getBooleanExtra(EXTRA_FROM_SETTING, false)
+        val startFromSourceSelection = intent.getBooleanExtra(EXTRA_START_FROM_SOURCE_SELECTION, true)
+        return if (startFromSourceSelection) {
+            RecommendedReadingListSourceFragment.newInstance(fromSetting)
+        } else {
+            // TODO: add this for the article interests screen when it is ready
+            Fragment()
+        }
     }
```
Williamrai, line 19: Just curious, will we be opening the interest screen through this activity?

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceFragment.kt
```diff
@@ -27,6 +28,7 @@ class RecommendedReadingListSourceFragment : Fragment() {
                         onCloseClick = {
                             if (viewModel.fromSettings) {
                                 viewModel.saveSourceSelection()
+                                requireActivity().setResult(RESULT_OK)
                             }
                             requireActivity().finish()
```
Williamrai, line 33: When the user goes back, the snackbar always appears. Should we display it only when there's an actual change in value?

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListOnboardingActivity.kt
cooltey: Yes.

PR #5604, opened 2025-05-30
Gallery Image Rotation Bug Fix

Williamrai:
### What does this do?
The odd image behavior when you rotate the screen was due to disabling the activity recreation for `GalleryActivity`, which prevented `ViewPager2` from properly handling configuration changes. The work around for this is posted on this [thread](https://issuetracker.google.com/issues/175796502), briefly setting the ViewPager2 to a previous position and then back to the current one, effectively resets the state and lands on the correct position. 

**Phabricator:**
https://phabricator.wikimedia.org/T392920

### DIFF
--- a/app/src/main/java/org/wikipedia/main/MainFragment.kt
+++ b/app/src/main/java/org/wikipedia/main/MainFragment.kt
@@ -69,6 +69,7 @@ import org.wikipedia.places.PlacesActivity
 import org.wikipedia.random.RandomActivity
 import org.wikipedia.readinglist.ReadingListBehaviorsUtil
 import org.wikipedia.readinglist.ReadingListsFragment
+import org.wikipedia.readinglist.recommended.RecommendedReadingListAbTest
 import org.wikipedia.search.SearchActivity
 import org.wikipedia.search.SearchFragment
 import org.wikipedia.settings.Prefs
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/wikipedia/gallery/GalleryActivity.kt
```diff
@@ -123,7 +124,7 @@ class GalleryActivity : BaseActivity(), LinkPreviewDialog.LoadPageCallback, Gall
         galleryAdapter = GalleryItemAdapter(this@GalleryActivity)
         binding.pager.adapter = galleryAdapter
         binding.pager.registerOnPageChangeCallback(pageChangeListener)
-        binding.pager.offscreenPageLimit = 2
+        binding.pager.offscreenPageLimit = ViewPager2.OFFSCREEN_PAGE_LIMIT_DEFAULT
```
dbrant: I'm not sure I agree with this -- it seems to make the user experience worse. The reason for increasing the `offscreenPageLimit` is to make the swiping of images smoother: the next image is _pre-loaded_, so that the swiping shows you the image already loaded.
With this change, the next image is only loaded when you start swiping, which looks much less fluid.

dbrant (changes requested):

app/src/main/java/org/wikipedia/gallery/GalleryActivity.kt
Williamrai: You're right, this change does make the user experience feel less fluid. I'll look into alternative solutions.

cooltey (approved):
Looks good to me!

app/src/main/java/org/wikipedia/gallery/GalleryActivity.kt
```diff
@@ -101,6 +102,18 @@ class GalleryActivity : BaseActivity(), LinkPreviewDialog.LoadPageCallback, Gall
         }
     }
 
+    override fun onConfigurationChanged(newConfig: Configuration) {
+        super.onConfigurationChanged(newConfig)
+        if (newConfig.orientation == Configuration.ORIENTATION_LANDSCAPE) {
```
dbrant, line 109: Did you mention that there is a ticket in Google's IssueTracker for this issue?
If so, let's make a comment with a link to the issue, so that we can revisit it later when it's fixed.

PR #5607, opened 2025-06-03
[Recommended reading lists] Interests selection screen.

dbrant:
https://phabricator.wikimedia.org/T394027


app/src/main/java/org/wikipedia/feed/becauseyouread/BecauseYouReadClient.kt
Williamrai, line 1: Just curious, why +1?

Williamrai (commented):
There seems to be inconsistencies in the scrolling behavior. On the Google Pixel 2, and likely other smaller devices, the scroll doesn't stop at the search bar. On the Pixel 9 and other larger device, the app bar title doesn’t appear when scrolling up because the scroll stops at the title. 
[Pixel2](https://github.com/user-attachments/assets/af6f5368-6900-4904-a1fd-0d55dd90482a)
[pixel9](https://github.com/user-attachments/assets/17aae021-3e84-413c-84c7-54c653389258)

dbrant:
> There seems to be inconsistencies in the scrolling behavior. On the Google Pixel 2, and likely other smaller devices, the scroll doesn't stop at the search bar. On the Pixel 9 and other larger device, the app bar title doesn’t appear when scrolling up because the scroll stops at the title.

Sorry, I'm not sure what the issue is? In your screenshots, the scrolling stops because you've reached the bottom of the content.
What do we want the behavior to be?

Williamrai:
> > There seems to be inconsistencies in the scrolling behavior. On the Google Pixel 2, and likely other smaller devices, the scroll doesn't stop at the search bar. On the Pixel 9 and other larger device, the app bar title doesn’t appear when scrolling up because the scroll stops at the title.
> 
> Sorry, I'm not sure what the issue is? In your screenshots, the scrolling stops because you've reached the bottom of the content. What do we want the behavior to be?

I thought the scroll behavior was supposed to stop at the search bar, where the search bar stays fixed and the rest of the content scrolls underneath it like [this](https://github.com/user-attachments/assets/fe752903-81e4-456f-a1ae-ea860777bd26).  If its not that then this is good.

dbrant:
> I thought the scroll behavior was supposed to stop at the search bar, where the search bar stays fixed and the rest of the content scrolls underneath it like [this](https://github.com/user-attachments/assets/fe752903-81e4-456f-a1ae-ea860777bd26). If its not that then this is good.

That looks like a complete coincidence where the scrolling stopped exactly at the top of the search bar. I'm pretty sure it's OK for the screen to keep scrolling, if you have more items at the bottom.

Williamrai:
> > I thought the scroll behavior was supposed to stop at the search bar, where the search bar stays fixed and the rest of the content scrolls underneath it like [this](https://github.com/user-attachments/assets/fe752903-81e4-456f-a1ae-ea860777bd26). If its not that then this is good.
> 
> That looks like a complete coincidence where the scrolling stopped exactly at the top of the search bar. I'm pretty sure it's OK for the screen to keep scrolling, if you have more items at the bottom.

yes, it is okay. The PR looks good, the compose code is clean and easy to follow.

app/src/main/java/org/wikipedia/feed/becauseyouread/BecauseYouReadClient.kt
dbrant: This was because I wanted to use the function `HistoryEntryWithImageDao.findEntryForReadMore()`, but this function does `+ 1` internally, which creates hidden complexity where it doesn't belong. So, I just moved the `+ 1` to the place where it's actually relevant.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListInterestsFragment.kt
```diff
@@ -0,0 +1,522 @@
...
+                colors = TopAppBarDefaults.topAppBarColors(
+                    containerColor = WikipediaTheme.colors.paperColor,
+                    scrolledContainerColor = WikipediaTheme.colors.paperColor
+                )
+            )
+        },
+        containerColor = WikipediaTheme.colors.paperColor
+    ) { paddingValues ->
+        when (uiState) {
+            is Resource.Loading -> {
```
cooltey, line 234: I am not seeing the loading indicator when it is loading.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListInterestsFragment.kt
```diff
@@ -0,0 +1,522 @@
...
+                        onRandomizeClick = onRandomizeClick,
+                        onSearchClick = onSearchClick
+                    )
+                }
+            }
+        }
+    }
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
```
cooltey: This can be removed.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListViewModel.kt
```diff
@@ -19,7 +19,7 @@ import org.wikipedia.util.Resource
 
 class RecommendedReadingListViewModel(savedStateHandle: SavedStateHandle) : ViewModel() {
 
-    val fromSettings = savedStateHandle.get<Boolean>(Constants.ARG_BOOLEAN) ?: false
+    val fromSettings = savedStateHandle.get<Boolean>(RecommendedReadingListOnboardingActivity.EXTRA_FROM_SETTINGS) == true
```
cooltey, line 22: Do you want to hide the bottom dice icon and the arrow icon if it comes from the settings? and also keep the layout to the topbar. You may check the `SourceSelectionScreen` for a similar logic for handling.

cooltey (changes requested):
It is good already! Just some minor comments.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListInterestsFragment.kt
dbrant: Hmm, I definitely see a loading indicator, especially when it does a network call. Maybe this was because I didn't have this PR up-to-date with the feature branch?

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListInterestsFragment.kt
dbrant: yes!

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListViewModel.kt
dbrant: I am campaigning to keep the screen identical whether you arrive from Settings or not.

cooltey (approved):

PR #5609, opened 2025-06-03
Design follow up

Williamrai:
### What does this do?
- ui fixes from the design follow up

**Phabricator:**
https://phabricator.wikimedia.org/T394056

### DIFF
--- a/app/src/main/res/values-qq/strings.xml
+++ b/app/src/main/res/values-qq/strings.xml
@@ -1864,8 +1864,8 @@
   <string name="recommended_reading_list_settings_notifications_subtitle_disable">Subtitle explaining that notifications for the recommended reading list are turned off.</string>
   <string name="recommended_reading_list_settings_notifications_dialog_title">Title of the dialog asking the user to confirm turning off notifications for  the recommended reading list.</string>
   <string name="recommended_reading_list_settings_notifications_dialog_message">Message in the dialog explaining what happens when the recommended reading list notifications are turned off.</string>
-  <string name="recommended_reading_list_settings_notifications_dialog_positive_button">Positive action button in the dialog to turn off the recommended reading list notifications.</string>
-  <string name="recommended_reading_list_settings_notifications_dialog_negative_button">Negative action button in the dialog to keep the recommended reading list notifications on.</string>
+  <string name="recommended_reading_list_settings_notifications_dialog_negative_button">Positive action button in the dialog to turn off the recommended reading list notifications.</string>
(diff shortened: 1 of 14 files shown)


app/src/main/res/values-qq/strings.xml
```diff
@@ -1864,8 +1864,8 @@
   <string name="recommended_reading_list_settings_notifications_subtitle_disable">Subtitle explaining that notifications for the recommended reading list are turned off.</string>
   <string name="recommended_reading_list_settings_notifications_dialog_title">Title of the dialog asking the user to confirm turning off notifications for  the recommended reading list.</string>
   <string name="recommended_reading_list_settings_notifications_dialog_message">Message in the dialog explaining what happens when the recommended reading list notifications are turned off.</string>
-  <string name="recommended_reading_list_settings_notifications_dialog_positive_button">Positive action button in the dialog to turn off the recommended reading list notifications.</string>
-  <string name="recommended_reading_list_settings_notifications_dialog_negative_button">Negative action button in the dialog to keep the recommended reading list notifications on.</string>
+  <string name="recommended_reading_list_settings_notifications_dialog_negative_button">Positive action button in the dialog to turn off the recommended reading list notifications.</string>
```
dbrant: The changes to these strings are pretty confusing: The name of the string is `...negative_button`, but the description says `Positive action button...`.

Instead of renaming these strings (in all the other languages, too), let's just reverse the order that they're used in the alert dialog.

app/src/main/res/values/strings.xml
```diff
@@ -1948,8 +1948,8 @@
     <string name="recommended_reading_list_settings_notifications_subtitle_disable">Notifications for Discover are turned off. You won\'t receive a notification when Discover is updated.</string>
     <string name="recommended_reading_list_settings_notifications_dialog_title">Turn off notifications?</string>
     <string name="recommended_reading_list_settings_notifications_dialog_message">If you turn off notifications, you will no longer receive reminders when Discover is updated.</string>
-    <string name="recommended_reading_list_settings_notifications_dialog_positive_button">Turn them off</string>
-    <string name="recommended_reading_list_settings_notifications_dialog_negative_button">No, keep them on</string>
+    <string name="recommended_reading_list_settings_notifications_dialog_negative_button">Turn it off</string>
```
dbrant: For changing from `them` to `it`, Sarah was referring to the strings for turning off Discover, not Notifications. (`recommended_reading_list_settings_turn_off_dialog_positive_button`).

Turning off Notifications should remain `them`, because Notifications is plural.

Williamrai: thank you for pointing this out. done.

PR #5614, opened 2025-06-04
Add additional pages to YiR screen deck - T396044

voyagerfan:
### What does this do?
This PR adds additional non-personalized screens into the screen deck display

### Why is this needed?
To bulk out the UX.

**Phabricator:**
https://phabricator.wikimedia.org/T396044


app/build.gradle
```diff
@@ -169,6 +169,7 @@ apply from: '../gradle/src/ktlint.gradle'
 
 dependencies {
 
+
```
cooltey: Please remove the unused change.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
```diff
@@ -303,7 +345,7 @@ fun YearInReviewScreenContent(
     innerPadding: PaddingValues,
     screenData: YearInReviewScreenData,
     context: Context,
-    isInfoIconVisible: Boolean = true
+    isShareSheetView: Boolean = false
```
cooltey: Can we have a name like `screenCaptureMode` or `screenShotMode`.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -32,7 +32,7 @@ class YearInReviewViewModel() : ViewModel() {
     private val handler = CoroutineExceptionHandler { _, throwable ->
         L.e(throwable)
         _uiScreenListState.value = Resource.Success(
-            data = listOf(nonEnglishCollectiveReadCountData, nonEnglishCollectiveEditCountData)
+            data = listOf(nonEnglishCollectiveReadCountData, nonEnglishCollectiveEditCountData, nonEnglishCollectiveReadCountData, nonEnglishCollectiveEditCountData)
```
cooltey, line 35: Not sure but are these expected that with two duplicated data?

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan: Hi Cooltey, Absolutely! I've made the change to `screenCaptureMode` and it will be in the next push

app/build.gradle
voyagerfan: Good catch. Done!

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
voyagerfan: Hi @cooltey I've ask and confirmed that this is OK for now since its a prototype.

PR #5615, opened 2025-06-04
Expose Year in Review in the developer settings - T394547

voyagerfan:
### What does this do?
This PR adds a conditional to hide the Year In Review icon unless its in dev settings.

### Why is this needed?
Supports Epic T387034

**Phabricator:**
https://phabricator.wikimedia.org/T394547


app/src/main/java/org/wikipedia/settings/dev/DeveloperSettingsPreferenceLoader.kt
```diff
@@ -214,6 +214,9 @@ internal class DeveloperSettingsPreferenceLoader(fragment: PreferenceFragmentCom
             activity.startActivity(Intent(activity, CategoryDeveloperPlayGround::class.java))
             true
         }
+        findPreference(R.string.preference_key_year_in_review_is_enabled).onPreferenceClickListener = Preference.OnPreferenceClickListener {
+            false
+        }
```
cooltey: This can be removed that the on/off status can be handled automatically through the XML.

voyagerfan: You're right, I've removed it. Thanks for this.

PR #5628, opened 2025-06-05
Reading List Nav tab show red dot icon

Williamrai:
### What does this do?
- displays a red dot on the Reading List navigation tab when a new list is available
- adds logic to the New Discover card to keep the red dot display in sync with the navigation tab


cooltey (approved):
It works for me. @dbrant

PR #5630, opened 2025-06-06
Instrumentation for the Recommended reading list

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T393488


cooltey:
@dbrant @Williamrai This is ready for review. I left some comments in the instrumentation deck and will make another follow-up PR for the new changes.

PR #5631, opened 2025-06-06
Show a confirm dialog when adjust source from settings

cooltey:
### What does this do?
If the user entered the source selection screen from the Settings screen, we should show the confirmation dialog after it changes.

https://www.figma.com/design/Z2n0gZOOAbgsIdHgUS5ocQ/Android-%E2%86%92-Rabbit-Holes-%E2%86%92-T378612?node-id=1210-18522&t=5EYUyKoJosPad3bC-4


cooltey:
Self-merge for instrumentation.

PR #5639, opened 2025-06-06
Update instrumentation for Recommended reading list

cooltey:
### What does this do?
Update the instrumentation based on the feedback.

### DIFF
--- a/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceScreen.kt
+++ b/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceScreen.kt
@@ -52,6 +52,9 @@ import org.wikipedia.settings.Prefs
 import org.wikipedia.theme.Theme
 import org.wikipedia.util.Resource
 
+val sourceOptionsForEvent = mutableListOf<RecommendedReadingListSource>()
+var selectedSourceForEvent = Prefs.recommendedReadingListSource
+
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
 fun SourceSelectionScreen(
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceScreen.kt
```diff
@@ -91,7 +94,17 @@ fun SourceSelectionScreen(
                         modifier = Modifier
```
Williamrai, line 108: i think we can do something like this inside the `sourceSelectionScreen`
```
    val optionsShown = remember(uiState) { buildList {
        if (uiState is Resource.Success) {
            add(RecommendedReadingListSource.INTERESTS)
            if (uiState.data.isSavedOptionEnabled) {
                add(RecommendedReadingListSource.READING_LIST)
            }
            if (uiState.data.isHistoryOptionEnabled) {
                add(RecommendedReadingListSource.HISTORY)
            }
        }
    } }
```
and then we can create     
`var selectedSourceForEvent by remember { mutableStateOf(Prefs.recommendedReadingListSource) }` and we can set this in `onSourceClick` of `SourceSelectionContent`, then we can do this 
```
RecommendedReadingListEvent.submit(
                                        action = "submit_click",
                                        activeInterface = activeInterface,
                                        optionsShown = optionsShown.map { it.eventString }.toString(),
                                        selected = selectedSourceForEvent.eventString,
                                        currentSetting = Prefs.recommendedReadingListSource.eventString
                                    )

```

cooltey: Done! Thanks for the example.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceScreen.kt
```diff
@@ -64,6 +68,21 @@ fun SourceSelectionScreen(
...
+                add(RecommendedReadingListSource.READING_LIST)
+            }
+            if (uiState.data.isHistoryOptionEnabled) {
+                add(RecommendedReadingListSource.HISTORY)
+            }
+        }
+    }
+
+    RecommendedReadingListEvent.submit("impression", activeInterface, optionsShown = optionsShown.map { it.eventString }.toString())
+
```
Williamrai: initially the list would be 0 because state would be in loading phase, we can put this inside a LaunchedEffect. 
```
LaunchedEffect(uiState) {
        if (uiState is Resource.Success) {
            RecommendedReadingListEvent.submit("impression", activeInterface, optionsShown = optionsShown.map { it.eventString }.toString())
        }
    }
```

cooltey: Can I move it to `is Resource.Success -> {` block inside the `Scaffold` without having `LaunchedEffect`?

PR #5642, opened 2025-06-06
Follow-up: dev settings for YiR

cooltey:
### What does this do?
Simplified code and cleaned up spaces.


voyagerfan (approved):
@cooltey these changes look good, especially the consolidation of the if-statement. Thanks for reviewing the code!

PR #5646, opened 2025-06-09
Recommended reading list: optimizing the onboarding experence

cooltey:
### What does this do?
- Change onboarding process and avoid showing zero selected articles if select Interests.

### DIFF
--- a/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceFragment.kt
+++ b/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceFragment.kt
@@ -1,6 +1,5 @@
 package org.wikipedia.readinglist.recommended
 
-import android.app.Activity.RESULT_OK
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
@@ -31,10 +30,6 @@ class RecommendedReadingListSourceFragment : Fragment() {
                         uiState = viewModel.uiSourceState.collectAsState().value,
                         fromSettings = viewModel.fromSettings,
                         onCloseClick = {
-                            if (viewModel.fromSettings) {
-                                viewModel.saveSourceSelection()
-                                requireActivity().setResult(RESULT_OK)
-                            }
                             requireActivity().finish()
                         },
                         onSourceClick = {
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSettingsActivity.kt
```diff
@@ -50,15 +55,17 @@ class RecommendedReadingListSettingsActivity : BaseActivity(), BaseActivity.Call
...
                 RecommendedReadingListSettingsScreen(
                     uiState = uiState,
+                    resetUiState = resetUiState,
                     modifier = Modifier
                         .fillMaxSize(),
                     onBackButtonClick = {
                         RecommendedReadingListEvent.submit("back_click", "discover_settings")
-                        onBackPressed()
+                        viewModel.generateRecommendedReadingList()
                     },
```
Williamrai, line 69: While this will generate a new list whenever the discover settings are modified, it will not refresh the discover list screen if we enter the settings from the list itself. Is this the desired behavior, or should we refresh the screen with the updates from the settings screen? I think this is an edge case. 

https://github.com/user-attachments/assets/a19099dc-ba70-4e44-9a90-64ab579ce881

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSettingsScreen.kt
```diff
@@ -47,25 +50,32 @@ import org.wikipedia.R
...
+import org.wikipedia.compose.components.error.WikiErrorClickEvents
+import org.wikipedia.compose.components.error.WikiErrorView
 import org.wikipedia.compose.extensions.noRippleClickable
 import org.wikipedia.compose.theme.BaseTheme
 import org.wikipedia.compose.theme.WikipediaTheme
 import org.wikipedia.theme.Theme
+import org.wikipedia.util.Resource
 
 @Composable
 fun RecommendedReadingListSettingsScreen(
```
Williamrai, line 62: we also need to add a BackHandler to handle the system back button.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSettingsActivity.kt
Williamrai: Solution in this PR https://github.com/wikimedia/apps-android-wikipedia/pull/5650

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSourceFragment.kt
```diff
@@ -25,19 +25,33 @@ class RecommendedReadingListSourceFragment : Fragment() {
...
+                                }
                                 requireActivity().setResult(RESULT_OK)
                             }
                             requireActivity().finish()
                         },
                         onSourceClick = {
+                            RecommendedReadingListEvent.submit("${it.eventString}_click", activeInterface)
                             if (viewModel.fromSettings) {
                                 if (it == Prefs.recommendedReadingListSource) {
                                     viewModel.updateSourceSelection(newSource = it)
```
Williamrai, line 57: if (it == Prefs.recommendedReadingListSource) will only work for Interest, Prefs.recommendedReadingListSource is never set before this.

Williamrai: resolved through discussion.

PR #5654, opened 2025-06-10
YiR: minor layout tweaks

cooltey:
### What does this do?
- Update padding to some layouts.
- Add a preview for the content.

@voyagerfan Please help to review and merge it, thanks.

### DIFF
--- a/app/src/main/java/org/wikipedia/yearinreview/YearInReviewEntryDialogScreen.kt
+++ b/app/src/main/java/org/wikipedia/yearinreview/YearInReviewEntryDialogScreen.kt
@@ -75,7 +75,7 @@ fun YearInReviewEntryDialogScreen() {
             },
             modifier = Modifier
                 .fillMaxWidth()
-                .padding(horizontal = 20.dp),
+                .padding(horizontal = 16.dp),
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/yearinreview/YearInReviewEntryDialogScreen.kt
voyagerfan, line 1: Hi @cooltey, I noticed this preview on line 95 isn’t wrapped in our theme — not part of this PR exactly, but since we’re doing that now, might be worth updating. (Totally something I missed originally too)

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScaffold.kt
voyagerfan, line 1: All of these changes look good! thanks for catching these and adding the preview!

app/src/main/java/org/wikipedia/yearinreview/YearInReviewEntryDialogScreen.kt
cooltey: Done!

voyagerfan (commented):
Looks good!

PR #5665, opened 2025-06-12
Tweak: simplifiy the logic of requesting morelike API to generate list

cooltey:
### What does this do?
In the current logic, we add duplicate titles to the `sourcesWithOffset` to request the `morelike` API for recommendations, which can be simplified by removing the duplicate titles and adjusting the `take()` number from the response. This PR optimizes it and it can reduce the times of requesting the `morelike` API.

### DIFF
--- a/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListHelper.kt
+++ b/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListHelper.kt
@@ -8,6 +8,7 @@ import org.wikipedia.events.NewRecommendedReadingListEvent
 import org.wikipedia.page.PageTitle
 import org.wikipedia.readinglist.database.RecommendedPage
 import org.wikipedia.settings.Prefs
+import kotlin.math.ceil
 
 object RecommendedReadingListHelper {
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListHelper.kt
```diff
@@ -72,39 +73,42 @@ object RecommendedReadingListHelper {
...
+                    wiki = it.wikiSite,
+                    lang = it.wikiSite.languageCode,
+                    namespace = it.namespace(),
+                    apiTitle = it.prefixedText,
+                    displayTitle = it.displayText,
+                    description = it.description,
+                    thumbUrl = it.thumbUrl
                 )
                 // Update the offset in the source list
                 newSourcesWithOffset.add(SourceWithOffset(sourceWithOffset.title, sourceWithOffset.language, offset))
```
Williamrai: I don't see this being used.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListHelper.kt
```diff
@@ -131,10 +135,10 @@ object RecommendedReadingListHelper {
         }
         Prefs.isNewRecommendedReadingListGenerated = true
         FlowEventBus.post(NewRecommendedReadingListEvent())
-        return finalList
+        return finalList.take(Prefs.recommendedReadingListArticlesNumber)
     }
```
Williamrai, line 147: Instead of selecting the first n elements, we could randomly sample n elements.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListHelper.kt
Williamrai, line 1: I think this is a good optimization. We're no longer making multiple API calls. Instead, we are only calling the API for the articles the user has selected or already has. Then, we simply add the `defaultTakeSize` number of articles to the list, which eliminates the need for the multiple calls we were making before. It also makes sense to fetch that number of articles since we're sorting them by relevance index.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListHelper.kt
cooltey: You are correct! and this should be fixed.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListHelper.kt
cooltey: Sure!

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListHelper.kt
```diff
@@ -58,63 +60,71 @@ object RecommendedReadingListHelper {
                     it.title
                 }
             }
-        }
+        }.distinctBy { it.prefixedText }
```
cooltey, line 63: Reduce the duplicate titles to reduce unnecessary loops.

Williamrai (approved):

PR #5667, opened 2025-06-13
Use correct help URL for Short Descriptions in enwiki.

dbrant:
This updates the URL that the user is taken to, when tapping Info links about short descriptions.
In the case of English Wikipedia, the URL will now go directly to [WP:SHORTDESC](https://en.wikipedia.org/wiki/Wikipedia:Short_description), instead of our general Suggested Edits page on MediaWiki.org.

https://phabricator.wikimedia.org/T390105

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
@@ -154,7 +154,11 @@ class SuggestedEditsCardsFragment : Fragment(), MenuProvider, SuggestedEditsItem
                             R.string.suggested_edits_image_tags_help_url)
                     }
                     else -> {
-                        FeedbackUtil.showAndroidAppEditingFAQ(requireContext())
+                        if ((viewModel.action == ADD_DESCRIPTION && viewModel.langFromCode == "en") || (viewModel.action == TRANSLATE_DESCRIPTION && viewModel.langToCode == "en")) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsCardsFragment.kt
```diff
@@ -154,7 +154,11 @@ class SuggestedEditsCardsFragment : Fragment(), MenuProvider, SuggestedEditsItem
                             R.string.suggested_edits_image_tags_help_url)
                     }
                     else -> {
-                        FeedbackUtil.showAndroidAppEditingFAQ(requireContext())
+                        if ((viewModel.action == ADD_DESCRIPTION && viewModel.langFromCode == "en") || (viewModel.action == TRANSLATE_DESCRIPTION && viewModel.langToCode == "en")) {
```
cooltey, line 157: Would it be better if we simplify this to:
```
if ((viewModel.action == ADD_DESCRIPTION || viewModel.action == TRANSLATE_DESCRIPTION) && viewModel.langFromCode == "en")
```

dbrant: I don't believe so: notice that when adding a description, the language checked is `langFromCode`, but when translating a description, the language checked is `langToCode`.

cooltey: Ah! right, I missed that.

PR #5668, opened 2025-06-13
update compose to use Material typography

Williamrai:
### What does this do?
This replaces the current Wikipedia typography with Material Typography. Since the default font weight of most of the styles in Material Typography is set to normal, I had to override it and in doing so, I also needed to apply additional style adjustments.

Mapping of our current styles to Material3 styles with changes applied:

1. h1 (24sp/32sp Bold) -> headlineSmall (24/32) — made it bold
2. h2 (20sp/28sp Bold) -> titleLarge (22/28) — made it bold and adjusted the font size
3. h3 (16sp/24sp Bold) -> titleMedium (16/24) — made it bold
4. h4 (14sp/24sp Bold) -> titleSmall (14/20) — made it bold and adjusted the lineHeight
5. h5 (12sp/18sp Bold) -> We haven’t used this style, so I didn’t include an equivalent Material3 mapping.
6. bodyLarge and bodyMedium are directly equivalent

Once design completes the M3 alignment work we can update these styles accordingly and also convert the `WikipediaColor` to use `MaterialColor`. 

**Phabricator:**
https://phabricator.wikimedia.org/T396876

### DIFF
--- a/app/src/main/java/org/wikipedia/compose/theme/WikipediaTypography.kt
+++ b/app/src/main/java/org/wikipedia/compose/theme/WikipediaTypography.kt
@@ -1,101 +1,34 @@
 package org.wikipedia.compose.theme
 
-import androidx.compose.runtime.Immutable
-import androidx.compose.runtime.staticCompositionLocalOf
+import androidx.compose.material3.Typography
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.sp
 
-@Immutable
-data class WikipediaTypography(
-    val h1: TextStyle = TextStyle(),
-    val h2: TextStyle = TextStyle(),
-    val h3: TextStyle = TextStyle(),
-    val h4: TextStyle = TextStyle(),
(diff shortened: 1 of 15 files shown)


app/src/main/java/org/wikipedia/compose/theme/WikipediaTypography.kt
```diff
@@ -1,101 +1,34 @@
...
-    val small: TextStyle = TextStyle()
-)
-
-val LocalWikipediaTypography = staticCompositionLocalOf {
-    WikipediaTypography()
-}
-
-val Typography = WikipediaTypography(
-    h1 = TextStyle(
+val WikipediaTypography = Typography(
```
dbrant: The ultimate goal is to get rid of `WikipediaTypography` entirely, instead of using it to override the default Material typography. We want to start using the plain, unmodified typography from Material 3 whenever possible, even if it looks slightly different from the current appearance.

If we have text left over that still needs to use the old styles (e.g. titles of old activities), we should override those styles in-place (in each Text composable) instead of a global style.

Williamrai: done.

cooltey (approved):
Looks good to me.

PR #5674, opened 2025-06-16
Red dot when RRL is available

Williamrai:
### What does this do?
- show red dot when rrl is available to user

### DIFF
--- a/app/src/main/java/org/wikipedia/main/MainFragment.kt
+++ b/app/src/main/java/org/wikipedia/main/MainFragment.kt
@@ -152,8 +152,8 @@ class MainFragment : Fragment(), BackPressedHandler, MenuProvider, FeedFragment.
         binding.mainNavTabLayout.descendants.filterIsInstance<TextView>().forEach {
             it.maxLines = 2
         }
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -151,7 +152,8 @@ class MainFragment : Fragment(), BackPressedHandler, MenuProvider, FeedFragment.
         binding.mainNavTabLayout.descendants.filterIsInstance<TextView>().forEach {
             it.maxLines = 2
         }
-        binding.mainNavTabLayout.setOverlayDot(NavTab.READING_LISTS, Prefs.isRecommendedReadingListEnabled && Prefs.isNewRecommendedReadingListGenerated)
+        val showRedDot = (RecommendedReadingListAbTest().isTestGroupUser() && !Prefs.isRecommendedReadingListOnboardingShown) || (Prefs.isRecommendedReadingListEnabled && Prefs.isNewRecommendedReadingListGenerated)
```
cooltey: This should have a better naming: `shouldShowRedDotForRecommendedReadingList`  or a shorten version of name 😓

Williamrai: done.

PR #5675, opened 2025-06-16
Fix: double check the source articles before showing the discover card

cooltey:
### What does this do?
Adds a simple database check before showing the Discover card to avoid an edge case where the users clean up the history/saved articles after they select the source.

### DIFF
--- a/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListHelper.kt
+++ b/app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListHelper.kt
@@ -17,7 +17,7 @@ object RecommendedReadingListHelper {
     private const val MAX_RETRIES = 10
 
     suspend fun readyToGenerateList(): Boolean {
-        if (!Prefs.isRecommendedReadingListEnabled) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListHelper.kt
```diff
@@ -16,14 +16,39 @@ object RecommendedReadingListHelper {
     private const val SUGGESTION_REQUEST_ITEMS = 50
     private const val MAX_RETRIES = 10
 
+    suspend fun readyToGenerateList(): Boolean {
+        if (!Prefs.isRecommendedReadingListEnabled || Prefs.recommendedReadingListArticlesNumber <= 0) {
+            return false
+        }
+        if (Prefs.resetRecommendedReadingList) {
+            // Make sure the sources have enough articles to generate a recommended reading list.
```
Williamrai: I think we can remove this Prefs.resetRecommendedReadingList, it always false when there is no changes in the discover setting becomes false which will not execute the when condition.

cooltey: Done.

PR #5682, opened 2025-06-17
Disallow translating Descriptions into English if having <50 edits.

dbrant:
Although we correctly prevent anyone with <50 edits from adding descriptions, there is a hole in the logic that would allow someone to **translate** descriptions into English, even with <50 edits.
Let's disallow translating descriptions if there are <50 edits, and if English is a secondary language.

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageRecsFragmentViewModel.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsImageRecsFragmentViewModel.kt
@@ -53,18 +53,18 @@ class SuggestedEditsImageRecsFragmentViewModel(savedStateHandle: SavedStateHandl
     fun fetchRecommendation() {
         _uiState.value = Resource.Loading()
         viewModelScope.launch(handler) {
-            var page: MwQueryPage?
+            var page: MwQueryPage? = null
             var tries = 0
             do {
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -380,6 +380,11 @@ class SuggestedEditsTasksFragment : Fragment() {
             !usesLocalDescriptions && viewModel.blockMessageWikidata.isNullOrEmpty()) {
             if (sufficientContributionsForArticleDescription) {
                 displayedTasks.add(addDescriptionsTask)
+
+                // Disable translating descriptions if the user has <50 edits, and they have English as a secondary language.
+                if (viewModel.totalContributions < 50 && WikipediaApp.instance.languageState.appLanguageCodes.contains("en")) {
+                    addDescriptionsTask.secondaryAction = null
+                }
             }
```
Williamrai, line 388: Should this check also be added to `SuggestedEditsFeedClient`?

dbrant: I don't believe so -- the Feed only offers to add descriptions, not translate them. That logic should already be covered correctly by the check for >50 edits.

Williamrai: okay, got it.

PR #5684, opened 2025-06-18
Don't treat redirects (HTTP 3xx) as errors.

dbrant:
There is an interesting quirk in one of our network interceptors that can potentially lead to subtle bugs, and also leads to confusing errors in Logcat.
(For example, when launching Places, some network calls result in a HTTP 304 (Unmodified) response, which is perfectly normal, but this actually manifests as an error printed in Logcat, which naturally raises alarms.)

See inline comments for explanation.

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/okhttp/UnsuccessfulResponseInterceptor.kt
+++ b/app/src/main/java/org/wikipedia/dataclient/okhttp/UnsuccessfulResponseInterceptor.kt
@@ -3,16 +3,18 @@ package org.wikipedia.dataclient.okhttp
 import okhttp3.Interceptor
 import okhttp3.Response
 import okhttp3.internal.closeQuietly
-import java.io.IOException
 
 class UnsuccessfulResponseInterceptor : Interceptor {
-    @Throws(IOException::class)
+    @Throws(HttpStatusException::class)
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/dataclient/okhttp/UnsuccessfulResponseInterceptor.kt
```diff
@@ -3,16 +3,18 @@ package org.wikipedia.dataclient.okhttp
...
 import okhttp3.Response
 import okhttp3.internal.closeQuietly
-import java.io.IOException
 
 class UnsuccessfulResponseInterceptor : Interceptor {
-    @Throws(IOException::class)
+    @Throws(HttpStatusException::class)
     override fun intercept(chain: Interceptor.Chain): Response {
         val rsp = chain.proceed(chain.request())
-        if (rsp.isSuccessful) {
```
dbrant, line 12: The `isSuccessful` field (provided by OkHttp) becomes `true` only for HTTP 2xx responses, and is **not true** for 3xx responses, for some reason. HTTP 3xx includes things like redirects (301 and 302), as well as Unmodified (304), which are perfectly good responses. We can debate whether these responses should be considered "successful" (I think they should be), but they are definitely not _errors_. Therefore, we should no longer throw an exception when a 3xx response is received here.

PR #5690, opened 2025-06-18
automatically converting Codex color tokens to Compose color values

Williamrai:
### What does this do?
adds a script to generate the following
1. ComposeColors class with raw colors
2. WikipediaColor data class consisting the names of the colors that will reference the raw colors
3. LightColor tokens
4. DarkColor tokens

**Phabricator:**
https://phabricator.wikimedia.org/T396880


scripts/generate-codex-colors.py
```diff
@@ -0,0 +1,107 @@
...
+
+# there are three type of color tokens from Codex
+# 1. theme (raw colors like red100, white, accent, gray50 etc. file is under src/themes/wikimedia-ui.json)
+# 2. base (color roles that references the theme colors like placeholder, progressive, disabled, destructive etc. file is under src/application.json)
+# 3. component (color for specific ui components like table, footer etc. file is under src/components.json)
+# 4. mode (dark mode colors)
+
+# URL
+codex_main_tokens_url = "https://cdn.jsdelivr.net/npm/@wikimedia/codex-design-tokens@latest/dist/theme-wikimedia-ui.json"
+codex_dark_mode_tokens_url = "https://cdn.jsdelivr.net/npm/@wikimedia/codex-design-tokens@latest/dist/theme-wikimedia-ui-mode-dark.json"
```
cooltey: Would it be possible to make these constant values in uppercase?

scripts/generate-codex-colors.py
```diff
@@ -0,0 +1,107 @@
...
+    print("TODO after the m3 alignment work")
+
+# execution
+# 1. generate raw colors
+# 2. generate WikipediaColor data class with codex tokens
+# 3. generate light mode colors and append to WikipediaColor file
+# 4. generate dark mode colors and append to WikipediaColor file
+
+wiki_color_file_path = "../app/src/main/java/org/wikipedia/compose/theme/WikipediaColor.kt"
+raw_color_file_path =  "../app/src/main/java/org/wikipedia/compose/theme/ComposeColors.kt"
```
cooltey: Same here.

cooltey (changes requested):
When I run `./gradlew ktlint` to check the style, it showed me the following message:
```
ComposeColors.kt:1:1: File must end with a newline (\n) (standard:final-newline)
ComposeColors.kt:5:23: Trailing space(s) (standard:no-trailing-spaces)
WikipediaColor.kt:1:1: File must end with a newline (\n) (standard:final-newline)
```

Few thoughts:
1. Can we make this script be able to execute directly without having the `python` in the command line to execute it? For example, I can run `bump-version-code` directly.
2. Would it be possible to include some `print` event so that we can make sure the script is running in process?

scripts/generate-codex-colors.py
```diff
@@ -0,0 +1,107 @@
...
+import re
+
+# there are three type of color tokens from Codex
+# 1. theme (raw colors like red100, white, accent, gray50 etc. file is under src/themes/wikimedia-ui.json)
+# 2. base (color roles that references the theme colors like placeholder, progressive, disabled, destructive etc. file is under src/application.json)
+# 3. component (color for specific ui components like table, footer etc. file is under src/components.json)
+# 4. mode (dark mode colors)
+
+# URL
+codex_main_tokens_url = "https://cdn.jsdelivr.net/npm/@wikimedia/codex-design-tokens@latest/dist/theme-wikimedia-ui.json"
```
dbrant: Instead of `latest`, let's specify an explicit pinned version number of Codex (the current latest is `2.1`), to be totally sure which version our tokens are coming from. So, let's define something like:
`CODEX_VERSION = 2.1`
and use it in place of `latest`.

scripts/generate-codex-colors.py
```diff
@@ -0,0 +1,107 @@
...
+    with open(file_path, 'w') as f:
+        f.write("".join(content))
+
+# this generates the class with the codex tokens that will use the raw colors
+# this will generate class similar to our current WikipediaColor
+def generate_wikipedia_color(main_color_data, file_path):
+    base_colors = extract_colors(main_color_data, "base")
+    content = ["package org.wikipedia.compose.theme\n\n",
+               "import androidx.compose.runtime.Immutable\n"
+               "import androidx.compose.ui.graphics.Color\n\n",
```
dbrant: When generating the Kotlin output, let's insert a comment that mentions the version of Codex that this is based on.

scripts/generate-codex-colors.py
Williamrai: done.

scripts/generate-codex-colors.py
Williamrai: done.

scripts/generate-codex-colors.py
Williamrai: done.

scripts/generate-codex-colors.py
Williamrai: done.

Williamrai:
> When I run `./gradlew ktlint` to check the style, it showed me the following message:
> 
> ```
> ComposeColors.kt:1:1: File must end with a newline (\n) (standard:final-newline)
> ComposeColors.kt:5:23: Trailing space(s) (standard:no-trailing-spaces)
> WikipediaColor.kt:1:1: File must end with a newline (\n) (standard:final-newline)
> ```
> 
> Few thoughts:
> 
> 1. Can we make this script be able to execute directly without having the `python` in the command line to execute it? For example, I can run `bump-version-code` directly.
> 2. Would it be possible to include some `print` event so that we can make sure the script is running in process?

1. I added` #!/usr/bin/env `python at the top so we can run the script directly as `scripts/generate-codex-colors.py` without needing to prepend it with python.
2. added

cooltey (commented):
The script works for me (even on Windows!), but I noticed the script will generate the `ComposeColors` in the `compose/theme` package. Is this something expected?

Williamrai:
> The script works for me (even on Windows!), but I noticed the script will generate the `ComposeColors` in the `compose/theme` package. Is this something expected?

Nope that's not expected, i updated the file path. Thank you for pointing this out.

cooltey (approved):
It looks good to me now.

scripts/generate-codex-colors.py
```diff
@@ -0,0 +1,125 @@
...
+                and 'value' in color_info
+                and color_info['value'].startswith("#")):
+            hex_value = color_info['value']
+            results[name] = hex_value
+    return results
+
+# takes in #f8f9fa and converts to 0xFFf8f9fa
+def convert_hex_to_kotlin_color(hex_color):
+    hex_color = hex_color.lstrip("#")
+    return f"0xFF{hex_color}"
```
dbrant, line 70: Some CSS colors use a shorthand 3-digit format (for example blue700: `#36c`) which is not directly compatible with Kotlin. We need to expand these colors into the 6-digit format, so `36c` should actually become `3366cc`.

Williamrai: fixed the 3digit css color edge case and added lightColors and DarkColors.

dbrant:
Fabulous -- We will likely need to make a few more tweaks to this script before actually using these color classes, but this is a great foundation.

PR #5692, opened 2025-06-18
Add support for captcha when logging in.

dbrant:
https://phabricator.wikimedia.org/T396403


cooltey (changes requested):
Found some issues:
1. After entering an incorrect password multiple times, a CAPTCHA appears. However, if I then enter either the wrong CAPTCHA or the wrong password, no error message is displayed.
2. After a couple of retries from 1., I entered the correct password and the correct captcha text, but somehow I was unable to log in. 
3. The link in the message "Can't see the image?..." guides the user to the `Wikipedia:Request an account` page, not sure if it is the correct wiki page for account login.

dbrant:
> 1. After entering an incorrect password multiple times, a CAPTCHA appears. However, if I then enter either the wrong CAPTCHA or the wrong password, no error message is displayed.

Fixed: this will now display an error. However, unfortunately the error when entering an incorrect captcha is still "Incorrect password". This seems to be a limitation of the API, and will need to be addressed in future updates.

> 2. After a couple of retries from 1., I entered the correct password and the correct captcha text, but somehow I was unable to log in.

I was unable to reproduce this. If you see this behavior again, does it show any errors or messages now?

> 3. The link in the message "Can't see the image?..." guides the user to the `Wikipedia:Request an account` page, not sure if it is the correct wiki page for account login.

Agreed - let's remove this message, which will make it consistent with mobile web.

cooltey (approved):
It works for me now! Thanks 👍 

@Williamrai please help to review and merge if it looks good to you too, thanks.

PR #5694, opened 2025-06-18
Upgrade the Cateogry table

cooltey:
### What does this do?
Migrate from the original table structure to a simpler structure.

**Phabricator:**
https://phabricator.wikimedia.org/T396364

### DIFF
--- a/app/src/main/java/org/wikipedia/categories/db/CategoryDao.kt
+++ b/app/src/main/java/org/wikipedia/categories/db/CategoryDao.kt
@@ -13,9 +13,9 @@ interface CategoryDao {
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     suspend fun insertAll(categories: List<Category>): List<Long>
 
-    @Query("SELECT title, lang, COUNT(*) as count FROM Category where timeStamp >= :startTimeStamp AND timeStamp <= :endTimeStamp " +
+    @Query("SELECT title, lang, COUNT(*) as count FROM Category WHERE year BETWEEN :startYear AND :endYear " +
             "GROUP BY title, lang ORDER BY count DESC")
-    suspend fun getCategoriesByTimeRange(startTimeStamp: Long, endTimeStamp: Long): List<CategoryCount>
+    suspend fun getCategoriesByTimeRange(startYear: Int, endYear: Int): List<Category>
 
     @Query("SELECT * FROM Category")
     suspend fun getAllCategories(): List<Category>
@@ -23,6 +23,6 @@ interface CategoryDao {
     @Query("DELETE FROM Category")
     suspend fun deleteAll()
 
-    @Query("DELETE FROM Category WHERE rowid IN (SELECT rowid FROM Category WHERE timeStamp <:timeStamp LIMIT :batchSize)")
-    suspend fun deleteOlderThanInBatch(timeStamp: Long, batchSize: Int): Int
+    @Query("DELETE FROM Category WHERE rowid IN (SELECT rowid FROM Category WHERE year <:year LIMIT :batchSize)")
+    suspend fun deleteOlderThanInBatch(year: Int, batchSize: Int): Int
 }
--- a/app/src/main/java/org/wikipedia/settings/dev/playground/CategoryDeveloperPlayGroundViewModel.kt
+++ b/app/src/main/java/org/wikipedia/settings/dev/playground/CategoryDeveloperPlayGroundViewModel.kt
@@ -12,14 +12,10 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import org.wikipedia.categories.db.Category
-import org.wikipedia.categories.db.CategoryCount
 import org.wikipedia.database.AppDatabase
-import org.wikipedia.util.DateUtil
 import org.wikipedia.util.UiState
 import org.wikipedia.util.log.L
 import java.time.Year
-import java.util.Calendar
-import java.util.Date
 import java.util.UUID
 
 class CategoryDeveloperPlayGroundViewModel : ViewModel() {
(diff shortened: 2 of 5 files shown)


app/src/main/java/org/wikipedia/recurring/CategoriesTableCleanupTask.kt
```diff
@@ -1,35 +1,28 @@
...
-        calendar.add(Calendar.YEAR, -years)
-        return calendar.timeInMillis
-    }
-
-    suspend fun deleteOldDataInBatches(timeStamp: Long, batchSize: Int) {
+    suspend fun deleteOldDataInBatches(year: Int, batchSize: Int) {
         var deletedCount = 0
         do {
-            deletedCount = AppDatabase.instance.categoryDao().deleteOlderThanInBatch(timeStamp, batchSize)
+            deletedCount = AppDatabase.instance.categoryDao().deleteOlderThanInBatch(year, batchSize)
```
dbrant: I don't really think we need the "batch" logic any longer, nor to have a `do { }` loop here. Sqlite is very efficient, and should not have any trouble deleting thousands of entries in one call.

app/src/main/java/org/wikipedia/categories/db/CategoryDao.kt
```diff
@@ -13,16 +13,31 @@ interface CategoryDao {
...
 
     @Query("DELETE FROM Category")
     suspend fun deleteAll()
 
-    @Query("DELETE FROM Category WHERE rowid IN (SELECT rowid FROM Category WHERE timeStamp <:timeStamp LIMIT :batchSize)")
-    suspend fun deleteOlderThanInBatch(timeStamp: Long, batchSize: Int): Int
+    @Query("DELETE FROM Category WHERE rowid IN (SELECT rowid FROM Category WHERE year <:year LIMIT :batchSize)")
+    suspend fun deleteOlderThanInBatch(year: Int, batchSize: Int): Int
+
+    suspend fun upsertAll(list: List<Category>) {
```
dbrant, line 38: Since this function performs multiple db operations, it can be annotated with `@Transaction`, to indicate to Room that this should be executed in a single transaction, which will be more efficient.

dbrant (changes requested):

app/src/main/java/org/wikipedia/settings/dev/playground/CategoryDeveloperPlayGroundViewModel.kt
```diff
@@ -72,14 +64,19 @@ class CategoryDeveloperPlayGroundViewModel : ViewModel() {
...
-                    System.currentTimeMillis()
+                val finalYear = if (year == Year.now().value) {
+                    year - 1
                 } else {
-                    getPreviousYearMillis(currentYear - year)
+                    year
                 }
                 AppDatabase.instance.categoryDao().insert(
-                    Category(title, languageCode, Date(timeStamp))
+                    Category(
```
Williamrai: this will insert and replaced but will not update the count next time same category is inserted

app/src/main/java/org/wikipedia/settings/dev/playground/CategoryDeveloperPlayGroundViewModel.kt
```diff
@@ -137,7 +135,7 @@ class CategoryDeveloperPlayGroundViewModel : ViewModel() {
             L.e(throwable)
         }) {
             try {
-                deleteOldDataInBatches(System.currentTimeMillis(), 100)
+                deleteOldDataInBatches(Year.now().value, 100)
                 loadCategories()
```
Williamrai, line 136: for deleteAll the year should be `Year.now().value + 1 `because the query has  `<:year `

app/src/main/java/org/wikipedia/settings/dev/playground/CategoryDeveloperPlayGroundViewModel.kt
```diff
@@ -113,7 +109,9 @@ class CategoryDeveloperPlayGroundViewModel : ViewModel() {
                         Category(
                             title = "Category:${categories[randomCategoryIndex]}${UUID.randomUUID()}",
                             lang = languages[randomLanguageIndex],
-                            timeStamp = Date(timeStamp)
+                            year = finalYear,
+                            month = 1,
+                            count = 1
```
Williamrai, line 112: same here, insertAll will insert and replace, it will not update the count.

app/src/main/java/org/wikipedia/recurring/CategoriesTableCleanupTask.kt
cooltey: Done.

app/src/main/java/org/wikipedia/categories/db/CategoryDao.kt
cooltey: Done!

app/src/main/java/org/wikipedia/settings/dev/playground/CategoryDeveloperPlayGroundViewModel.kt
cooltey: Fixed!

app/src/main/java/org/wikipedia/settings/dev/playground/CategoryDeveloperPlayGroundViewModel.kt
cooltey: Fixed!

app/src/main/java/org/wikipedia/settings/dev/playground/CategoryDeveloperPlayGroundViewModel.kt
cooltey: Fixed!

app/src/main/java/org/wikipedia/recurring/CategoriesTableCleanupTask.kt
```diff
@@ -1,36 +1,26 @@
...
-        return calendar.timeInMillis
-    }
-
-    suspend fun deleteOldDataInBatches(timeStamp: Long, batchSize: Int) {
-        var deletedCount = 0
-        do {
-            deletedCount = AppDatabase.instance.categoryDao().deleteOlderThanInBatch(timeStamp, batchSize)
-        } while (deletedCount > 0)
+    suspend fun deleteOldDataInBatches(year: Int) {
+        AppDatabase.instance.categoryDao().deleteOlderThanInBatch(year)
```
Williamrai: It might be better to call this code directly within the run function, since the method no longer deletes in batches. You could also consider renaming the SQLite function from `deleteOlderThanInBatch` to `deleteOlderThan` or something similar.

Williamrai (approved):

PR #5695, opened 2025-06-18
Fix: should use hard-coded text for the event in RRL

cooltey:
### What does this do?
A fix for showing the "translated" frequency text in the event, regarding the analyst's feedback.


cooltey:
Self-merge for a cherry-pick release.

PR #5710, opened 2025-06-23
Espresso Fixes and optimize FeedScreen test

Williamrai:
### What does this do?
The goal of this PR is to address Espresso crashes and improve the stability of the FeedScreen test by reducing its flakiness.


app/src/androidTest/java/org/wikipedia/base/utils/GifMatchers.kt
```diff
@@ -14,12 +15,8 @@ object GifMatchers {
...
 
             override fun matchesSafely(imageView: ImageView): Boolean {
-                val drawable = imageView.drawable
-
-                return when (drawable) {
-                    // @TODO: Add a way to identify if image is drwable
-                    else -> false
-                }
+                var drawable = imageView.drawable
+                return drawable is Animatable
```
cooltey: Would it be simpler not to declare a variable for the drawable and return `imageView.drawable is Animatable` directly?

app/src/androidTest/java/org/wikipedia/base/BaseTest.kt
```diff
@@ -80,6 +81,9 @@ abstract class BaseTest<T : AppCompatActivity>(
         activityScenarioRule.scenario.onActivity {
             activity = it
         }
+        val instance = WikipediaApp.instance
+        val appLanguageCodeList = instance.languageState.appLanguageCodes
+        instance.languageState.removeAppLanguageCodes(appLanguageCodeList.filter { it != "en" })
```
cooltey: How about:
```
WikipediaApp.instance.languageState.let {
    it.removeAppLanguageCodes(it.appLanguageCodes.filter { it != "en" })
}
```

app/src/androidTest/java/org/wikipedia/base/utils/GifMatchers.kt
Williamrai: done

app/src/androidTest/java/org/wikipedia/base/BaseTest.kt
Williamrai: done

PR #5717, opened 2025-06-24
Gif not loading and png support

Williamrai:
### What does this do?
- fixes gif not loading and adds png support for face detect 

Transformations were being applied to GIFs, which converted them into static images and prevented them from loading properly. To fix this, we now skip the transformation logic for GIFs.
```
 when {
    (detectFace == true && shouldDetectFace(url)) -> requestBuilder.transformations(FaceDetectTransformation(), DimImageTransformation())
     else -> requestBuilder.transformations(WhiteBackgroundTransformation(), DimImageTransformation()) 
}
```


dbrant:
@Williamrai Is there an example article that has a .png file that has a face?

Williamrai:
> @Williamrai Is there an example article that has a .png file that has a face?

In today's Top read "Shai Gilgeous-Alexander" with the basketball is a png image.

PR #5719, opened 2025-06-24
Improve the bump-version-code Python script compatibility

cooltey:
### What does this do?
Update the Python script to enhance its compatibility across various platforms.

### Why is this needed?
The `sh` module does not support Windows.

### DIFF
--- a/scripts/bump-version-code.py
+++ b/scripts/bump-version-code.py
@@ -10,13 +10,14 @@ Step 1: (run without arguments)
 
 After this run 'git review' and + the commit.
 
-Requires the Python module 'sh' to run.
+Cross-platform compatible - works on Windows, macOS, and Linux.
 """
-import sh
+import subprocess
 import os
 import re
 import sys
 
+VERSION_CODE_BRANCH = 'bumpVersionCode'
 VERSION_CODE_REGEX = r'(?P<key>versionCode) (?P<value>\d+)'
 
 script_dir = sys.path[0]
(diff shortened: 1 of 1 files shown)


scripts/bump-version-code.py
```diff
@@ -10,13 +10,14 @@
 
 After this run 'git review' and + the commit.
```
dbrant: The comment above is no longer relevant, and can be removed.

scripts/bump-version-code.py
```diff
@@ -50,24 +63,41 @@ def transform_file(file_path, *funcs):
...
+    run_git_command('push', '--set-upstream', 'origin', VERSION_CODE_BRANCH)
+
+    # Switch back to main
+    run_git_command('checkout', 'main')
 
 
 if __name__ == '__main__':
     bump('app/build.gradle')
-    print('BUMP NOTICE! Merge the new `bumpVersionCode` into Main.)')
+    print('BUMP NOTICE! Merge the new `bumpVersionCodeTest` into Main.)')
```
dbrant: What's "bumpVersionCodeTest"? Do we want to use `VERSION_CODE_BRANCH` here, too?

cooltey: Done 😓

scripts/bump-version-code.py
cooltey: Removed!

PR #5720, opened 2025-06-25
Correctly align snackbars in Login / CreateAccount activities.

dbrant:
On Android 15, the snackbars in these activities are overlapping under the navigation buttons.
Let's make the base layout in these activities be a `CoordinatorLayout` instead of `FrameLayout`, which fixes the snackbar behavior.

image

### DIFF
--- a/app/src/main/java/org/wikipedia/util/FeedbackUtil.kt
+++ b/app/src/main/java/org/wikipedia/util/FeedbackUtil.kt
@@ -24,8 +24,11 @@ import org.wikipedia.R
 import org.wikipedia.WikipediaApp
 import org.wikipedia.activity.BaseActivity
 import org.wikipedia.analytics.eventplatform.BreadCrumbLogEvent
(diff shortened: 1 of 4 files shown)


Williamrai:
The snackbar doesn’t have any padding at the bottom and looks like it’s stuck to the bottom navigation bar. Just wanted to check, is this expected?
[snackbar](https://github.com/user-attachments/assets/4ed78b07-2219-4fc3-a925-41b21bb9f72d)

cooltey:
> The snackbar doesn’t have any padding at the bottom and looks like it’s stuck to the bottom navigation bar. Just wanted to check, is this expected? [snackbar](https://github.com/user-attachments/assets/4ed78b07-2219-4fc3-a925-41b21bb9f72d)

+1 the snackbar seems too close to the system navigation bar.

dbrant:
@Williamrai @cooltey  This is also (finally) ready.

PR #5723, opened 2025-06-26
Espresso Editing Test

Williamrai:
### What does this do?
Adds editing tests using Espresso.

This test will cover the following:
1. Verifies all text formatting options (bold, italic, headings, etc.)
2. Validates functionality for images, links, and references, lists, undo, redo
3. Confirms that edits are correctly published
4. Checks search functionality on edited text

**Phabricator:**
https://phabricator.wikimedia.org/T378659


cooltey (commented):
Few observations:
1. For now, the login requires entering the Captcha in a certain condition, how do we resolve it 🤔 
2. The test failed when trying to insert a media into the edit field, not sure if you see that on your side.

Williamrai:
> Few observations:
> 
> 1. For now, the login requires entering the Captcha in a certain condition, how do we resolve it 🤔
> 2. The test failed when trying to insert a media into the edit field, not sure if you see that on your side.

1. I believe the CAPTCHA test may not be necessary in this case, as the login credentials are always valid. If testing the CAPTCHA is required, it would be more appropriate to handle that in a separate test. This particular test is intended to focus only on verifying that the edit functionality works as expected. In cases where we would see CAPTCHA, I think we would have to either bypass the CAPTCHA completely or handle the captcha but handling it would require reading the CAPTCHA which i am not sure how we can achieve it. 

2. No i haven't seen that on my side, what device were you using?

PR #5730, opened 2025-06-27
Remove allowMainThreadQueries and update DAO functions

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T397993

### DIFF
--- a/app/src/main/java/org/wikipedia/database/AppDatabase.kt
+++ b/app/src/main/java/org/wikipedia/database/AppDatabase.kt
@@ -354,7 +354,6 @@ abstract class AppDatabase : RoomDatabase() {
                     MIGRATION_23_24, MIGRATION_24_25, MIGRATION_25_26, MIGRATION_26_27,
                     MIGRATION_26_28, MIGRATION_27_28, MIGRATION_28_29, MIGRATION_29_30,
                     MIGRATION_30_31)
-                .allowMainThreadQueries() // TODO: remove after resolving main thread issues in DAO classes
                 .fallbackToDestructiveMigration(false)
                 .build()
         }
--- a/app/src/main/java/org/wikipedia/readinglist/db/ReadingListPageDao.kt
+++ b/app/src/main/java/org/wikipedia/readinglist/db/ReadingListPageDao.kt
@@ -24,31 +24,28 @@ import org.wikipedia.util.StringUtil
 @Dao
 interface ReadingListPageDao {
     @Insert(onConflict = OnConflictStrategy.REPLACE)
-    fun insertReadingListPage(page: ReadingListPage): Long
+    suspend fun insertReadingListPage(page: ReadingListPage): Long
 
     @Update(onConflict = OnConflictStrategy.REPLACE)
-    fun updateReadingListPage(page: ReadingListPage)
+    suspend fun updateReadingListPage(page: ReadingListPage)
 
     @Delete
-    fun deleteReadingListPage(page: ReadingListPage)
-
-    @Query("SELECT * FROM ReadingListPage")
-    fun getAllPages(): List<ReadingListPage>
+    suspend fun deleteReadingListPage(page: ReadingListPage)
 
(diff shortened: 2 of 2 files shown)


cooltey:
This is ready for review. The changes affect all reading list saving behaviors in the app.

app/src/main/java/org/wikipedia/readinglist/ReadingListBehaviorsUtil.kt
Williamrai, line 1: the save offline download animation is not properly. 

https://github.com/user-attachments/assets/5cb3b0ad-7c14-44b7-9f36-e6f8e88d67ab

cooltey: Thanks! Fixed.

app/src/main/java/org/wikipedia/readinglist/ReadingListsExportImportHelper.kt
```diff
@@ -77,30 +80,31 @@ object ReadingListsExportImportHelper : BaseActivity.Callback {
             .setStyle(NotificationCompat.BigTextStyle().bigText(context.getString(R.string.reading_list_notification_text, numOfLists)))
     }
 
-    fun importLists(activity: BaseActivity, jsonString: String) {
+    fun importLists(activity: AppCompatActivity, jsonString: String) {
```
Williamrai, line 83: I'm wondering if it might be cleaner to convert this into a suspend function and launch it using viewLifecycleOwner.lifecycleScope in the fragment, instead of passing **AppCompatActivity** for launching the coroutine, what do you think? and maybe for other functions as well.

app/src/main/java/org/wikipedia/WikipediaApp.kt
```diff
@@ -251,16 +251,20 @@ class WikipediaApp : Application() {
...
-        Prefs.tempAccountWelcomeShown = false
-        Prefs.tempAccountCreateDay = 0L
-        Prefs.tempAccountDialogShown = false
-        SharedPreferenceCookieManager.instance.clearAllCookies()
-        AppDatabase.instance.notificationDao().deleteAll()
-        FlowEventBus.post(LoggedOutEvent())
-        L.d("Logout complete.")
+        MainScope().launch(CoroutineExceptionHandler { _, t ->
+            L.e(t)
+        }) {
```
dbrant: This function must not launch a coroutine. It needs to be run synchronously, so that the logic of `CsrfTokenClient` can work correctly.

app/src/main/java/org/wikipedia/WikipediaApp.kt
```diff
@@ -251,16 +251,20 @@ class WikipediaApp : Application() {
...
+            AccountUtil.removeAccount()
+            Prefs.isPushNotificationTokenSubscribed = false
+            Prefs.pushNotificationTokenOld = ""
+            Prefs.tempAccountWelcomeShown = false
+            Prefs.tempAccountCreateDay = 0L
+            Prefs.tempAccountDialogShown = false
+            SharedPreferenceCookieManager.instance.clearAllCookies()
+            AppDatabase.instance.notificationDao().deleteAll()
```
dbrant, line 262: ^ It looks like this is the only Dao call that is made in this function, and it doesn't really impact the Csrf logic. So, we can just run this single line as a miniature coroutine, and leave the rest of this function as before.

dbrant (changes requested):

app/src/main/java/org/wikipedia/WikipediaApp.kt
cooltey: Sure! Done.

app/src/main/java/org/wikipedia/WikipediaApp.kt
cooltey: Done.

PR #5736, opened 2025-06-30
Update the Python scripts to be compatible across different platforms

cooltey:
### What does this do?
Updated the existing scripts to be compatible across different platforms.
- Removed `.sh` files and replaced with `.py` scripts.
- Tested and worked on both macOS and Windows 11.
- Ready to be used in the current release.


upload-all.py
```diff
@@ -0,0 +1,209 @@
...
+
+def find_scp_command():
+    """Find the SCP command on different platforms"""
+    if platform.system() == 'Windows':
+        # On Windows, try common locations for SCP
+        possible_paths = [
+            'scp',  # If it's in PATH
+            'C:/Windows/System32/OpenSSH/scp.exe',
+            'C:/Program Files/Git/usr/bin/scp.exe',
+            'C:/Program Files (x86)/Git/usr/bin/scp.exe'
```
cooltey: A bit of concern on these specific `.exe` paths, maybe we can remove them from the script.

PR #5740, opened 2025-07-01
On this day game code structure refactor

Williamrai:
### What does this do?
The gameplay logic is moved out of **OnThisDayGameActivity** into its own fragment called **OnThisDayGamePlayFragment**. The onboarding fragment now acts as the menu. It shows the user the current state of the game and handles navigation from there. So instead of updating the state for navigation, the logic now directly navigates the user to the fragment. Also, **OnThisDayGameOnboardingFragment** is now called **OnThisDayGameMenuFragment**, and **OnThisDayFinalFragment** has been renamed to **OnThisDayGameOverFragment**.

`Note: All game logic remains the same. `


app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameActivity.kt
```diff
@@ -38,35 +25,22 @@ import org.wikipedia.activity.BaseActivity
...
 
     private lateinit var binding: ActivityOnThisDayGameBinding
     private val viewModel: OnThisDayGameViewModel by viewModels()
 
-    private val cardAnimatorSetIn = AnimatorSet()
-    private val cardAnimatorSetOut = AnimatorSet()
-    private lateinit var mediaPlayer: MediaPlayer
-    private var selectedCardView: WikiCardView? = null
-
     @SuppressLint("SourceLockedOrientationActivity", "ClickableViewAccessibility")
```
cooltey: It would be better to remove `ClickableViewAccessibility` because it is not required anymore.

app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameMenuFragment.kt
```diff
@@ -25,14 +25,14 @@ import org.wikipedia.util.Resource
 import org.wikipedia.util.ResourceUtil
 import java.time.LocalDate
 
-class OnThisDayGameOnboardingFragment : OnThisDayGameBaseFragment() {
-    private var _binding: FragmentOnThisDayGameOnboardingBinding? = null
+class OnThisDayGameMenuFragment : OnThisDayGameBaseFragment() {
```
cooltey: `Menu` seems a little bit confusing to me. Would it be better to have a name such as `Initial` or `TitleScreen`...etc?

app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameOverFragment.kt
```diff
@@ -55,20 +56,24 @@ import java.time.Duration
 import java.time.LocalDate
 import java.time.LocalDateTime
 import java.time.LocalTime
+import java.time.MonthDay
+import java.time.format.DateTimeFormatter
 import java.util.Locale
 
-class OnThisDayGameFinalFragment : OnThisDayGameBaseFragment(), OnThisDayGameArticleBottomSheet.Callback {
-    private var _binding: FragmentOnThisDayGameFinalBinding? = null
+class OnThisDayGameOverFragment : OnThisDayGameBaseFragment(), OnThisDayGameArticleBottomSheet.Callback {
```
cooltey: Instead of `Over`, would it be better if we name it `OnThisDayGameResultFragment`?

cooltey (commented):
Noticed an issue: when you complete the game and see the final screen, and then go back to the previous screen by using the system back button, you will see the game questions for a bit. Is the issue related to this change?

app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameActivity.kt
Williamrai: done.

app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameMenuFragment.kt
Williamrai: renamed to OnThisDayGameMainMenuFragment

app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameOverFragment.kt
Williamrai: renamed to OnThisDayGameResultFragment.

dbrant (changes requested):
One minor inconsistency:
When starting the game (on the first question), the questions no longer animate in, and just appear immediately. This is different from the current expected behavior.

Williamrai:
> One minor inconsistency: When starting the game (on the first question), the questions no longer animate in, and just appear immediately. This is different from the current expected behavior.

fixed.

PR #5753, opened 2025-07-08
No longer use WebView in LinkPreviewDialog.

dbrant:
Recent updates in PCS have affected the display of link previews in Dark mode.
Rather than reverting the PCS changes, let's no longer use a WebView in link previews, and figure out a more sensible balance of showing html content in link previews going forward.


cooltey:
Just wanted to make a note:
This is valid: https://phabricator.wikimedia.org/T314003

PR #5759, opened 2025-07-10
(Finally, properly) fix viewing saved articles when offline.

dbrant:
This ensures that saved articles will be always viewable when offline, _even if_ using a VPN (which makes the app believe it's still "online").

Here is how this is accomplished:
* For loading an article, we need two things: the `/page/summary` and the `page/mobile-html` contents, both of which are saved offline.
* However, we also make other network calls when loading a page, such as: checking its Watched status, and getting its Categories. And the problem is that we are making the display of the article _dependent on_ the success of these network calls. But in reality, these calls are really kind of "optional". If either of them fails, there's no reason not to proceed with loading the article.

https://phabricator.wikimedia.org/T354119
https://phabricator.wikimedia.org/T383950

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
+++ b/app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
@@ -30,6 +30,7 @@ import org.wikipedia.util.UriUtil
 import org.wikipedia.util.log.L
 import org.wikipedia.views.ObservableWebView
 import retrofit2.Response
+import java.io.IOException
 import java.time.Instant
 import java.time.LocalDate
 import java.time.ZoneId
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
```diff
@@ -149,18 +150,29 @@ class PageFragmentLoadState(private var model: PageViewModel,
...
+                        try {
+                            if (makeWatchRequest) {
+                                ServiceFactory.get(title.wikiSite)
+                                    .getWatchedStatusWithCategories(title.prefixedText)
+                            } else if (WikipediaApp.instance.isOnline && !AccountUtil.isLoggedIn) {
+                                AnonymousNotificationHelper.maybeGetAnonUserInfo(title.wikiSite)
+                            } else {
+                                MwQueryResponse()
+                            }
+                        } catch (_: IOException) {
```
dbrant, line 162: Very importantly, note that I'm only catching `IOException`, which will limit the catching to network-related issues. Other errors, such as coroutine termination or Mediawiki errors, will continue to have the same effect as before.

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
```diff
@@ -149,18 +150,29 @@ class PageFragmentLoadState(private var model: PageViewModel,
...
+                            L.w("Ignoring network error while fetching watched status.")
                             MwQueryResponse()
                         }
                     }
                     val categoriesRequest = async {
-                        if (!makeWatchRequest && WikipediaApp.instance.isOnline) {
-                            ServiceFactory.get(title.wikiSite).getCategoriesProps(title.text)
-                        } else {
+                        try {
+                            if (!makeWatchRequest && WikipediaApp.instance.isOnline) {
```
cooltey, line 169: Not really related to this PR, but having `!makeWatchRequest` along with `WikipediaApp.instance.isOnline` because `isOnline` is defined in `makeWatchRequest`.

PR #5762, opened 2025-07-10
view model to load page in PageFragment

Williamrai:
### What does this do?
- use view model to load page in the PageFragment 
- introduces state handling in the PageFragment
- code refactors

**Phabricator:**
https://phabricator.wikimedia.org/T303945


app/src/main/java/org/wikipedia/page/pageload/PageDataFetcher.kt
```diff
@@ -0,0 +1,87 @@
...
+                val response = ServiceFactory.get(title.wikiSite).getWatchedStatusWithCategories(title.prefixedText)
+                val page = response.query?.firstPage()
+                WatchStatus(
+                    isWatched = page?.watched == true,
+                    hasWatchlistExpiry = page?.hasWatchlistExpiry() == true,
+                    myQueryResponse = response
+                )
+            } else if (WikipediaApp.instance.isOnline && !AccountUtil.isLoggedIn) {
+                val response = AnonymousNotificationHelper.maybeGetAnonUserInfo(title.wikiSite)
+                WatchStatus(false, false, response)
```
cooltey: Please add parameter names here.

app/src/main/java/org/wikipedia/page/pageload/PageDataFetcher.kt
```diff
@@ -0,0 +1,87 @@
...
+            } else {
+                WatchStatus(false, false, MwQueryResponse())
```
cooltey: Please add parameter names here.

app/src/main/java/org/wikipedia/page/pageload/PageLoadViewModel.kt
```diff
@@ -0,0 +1,404 @@
...
+        }) {
+            val cacheControl = if (request.options.isRefresh) "no-cache" else "default"
+            val pageSummaryDeferred = async { pageDataFetcher.fetchPageSummary(request.title, cacheControl) }
+            val watchStatusDeferred = async { pageDataFetcher.fetchWatchStatus(request.title) }
+
+            val pageSummary = pageSummaryDeferred.await()
+            handlePageSummary(pageSummary, request)
+
+            val watchStatus = watchStatusDeferred.await()
+            handleWatchStatusAndFetchCategories(watchStatus, request.title)
```
cooltey: The `fetchWatchStatus` already contains the categories response, so this function should either pass the `categories` from the `fetchWatchStatus` directly, or fetch it only if the `categories` is empty.

app/src/main/java/org/wikipedia/page/pageload/PageLoadViewModel.kt
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    companion object {
+        val Factory: ViewModelProvider.Factory = viewModelFactory {
+            initializer {
+                val app = this[APPLICATION_KEY] as WikipediaApp
+                PageLoadViewModel(app)
+            }
+        }
+    }
```
cooltey: Could you please explain it a bit, and what's the difference if we set a `val app = WikipediaApp.instance` directly in the view model class?

app/src/main/java/org/wikipedia/page/pageload/PageDataFetcher.kt
Williamrai: done.

app/src/main/java/org/wikipedia/page/pageload/PageDataFetcher.kt
Williamrai: done.

app/src/main/java/org/wikipedia/page/pageload/PageLoadViewModel.kt
Williamrai: Functionally, there won’t be a difference since WikipediaApp instance is an object and can be accessed easily everywhere, but ideally, any external dependencies should be injected into the ViewModel even the **dataFetcher** that i have. I will remove this since this can create confusion.

Williamrai:
> Just did a quick scan of this PR and it looks pretty good! 👍
> 
> One issue I found:
> 
> * When loading an article that has a lead image, the footer does not get loaded correctly, and it shows up after refreshing the page.

the read more is loading lazily for me, i can a slight delay though 
[preview](https://github.com/user-attachments/assets/ca02b30a-4d2f-43a4-959e-2b9a0055a8e2)

app/src/main/java/org/wikipedia/page/pageload/PageLoadViewModel.kt
Williamrai: done, the current logic fetches data based on the applied conditions.

Williamrai:
> > Just did a quick scan of this PR and it looks pretty good! 👍
> > One issue I found:
> > 
> > * When loading an article that has a lead image, the footer does not get loaded correctly, and it shows up after refreshing the page.
> 
> the read more is loading lazily for me, i can a slight delay though [preview](https://github.com/user-attachments/assets/ca02b30a-4d2f-43a4-959e-2b9a0055a8e2)

As per discussion, the issue arises rarely and its difficult to reproduce, will monitor this.

PR #5765, opened 2025-07-14
Initialize fastlane.

dbrant:
This initializes [fastlane](https://docs.fastlane.tools/) in our repo, which will help facilitate things like automated releases, syncing of Play Store assets, etc.

(This currently has no impact on our code, build process, or testing. This is just setting things up.)


fastlane/Fastfile
```diff
@@ -0,0 +1,38 @@
...
+platform :android do
+  desc "Runs all the tests"
+  lane :test do
+    gradle(task: "test")
+  end
+
+  desc "Submit a new Beta Build to Crashlytics Beta"
+  lane :beta do
+    gradle(task: "clean assembleRelease")
+    crashlytics
```
cooltey: Do we have crashlytics setup in the account?

dbrant: We don't - these are just example actions that were auto-generated during fastlane initialization. We can remove this one.

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
```diff
@@ -101,6 +101,7 @@ class PageFragmentLoadState(private var model: PageViewModel,
     }
 
     fun backStackEmpty(): Boolean {
+        println("")
```
cooltey: added by accident?

dbrant: dammit!

PR #5766, opened 2025-07-14
Offline test workflow 

Williamrai:
### What does this do?
- adds a workflow for offline espresso test
- espresso code fixes


.github/workflows/android_offline_test.yml
```diff
@@ -0,0 +1,33 @@
+name: Android offline test
```
cooltey: Would it be possible to have more specific naming, for example, `Android saved article offline test`, in case we need to test the `okhttp` cache offline test in the future? Not sure if this makes sense to you.

Williamrai: done.

PR #5769, opened 2025-07-15
Language search empty search extra padding fix

Williamrai:
### What does this do?
- removes the extra padding 
[extra padding portrait](https://github.com/user-attachments/assets/4894aef4-995a-4df4-aa4c-be36fd843211)
[extra padding landscape](https://github.com/user-attachments/assets/9077c005-0def-40dd-bcc9-83ec0ad40c33)

### DIFF
--- a/app/src/main/java/org/wikipedia/language/LangLinksScreen.kt
+++ b/app/src/main/java/org/wikipedia/language/LangLinksScreen.kt
@@ -1,15 +1,12 @@
 package org.wikipedia.language
 
-import android.os.Build
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
-import androidx.compose.foundation.layout.ime
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/language/LangLinksScreen.kt
```diff
@@ -49,12 +45,7 @@ fun ComposeLangLinksScreen(
     ) {
     val context = LocalContext.current
     var searchQuery by remember { mutableStateOf("") }
-    val (imeHeight, isKeyboardVisible) = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
-        // Handle IME (keyboard) insets
-        val windowInsets = WindowInsets.ime
-        val height = with(LocalDensity.current) { windowInsets.getBottom(this).toDp() }
-        Pair(height, height > 0.dp)
```
cooltey, line 56: I am wondering whether we need to add this logic when we have the search bar in a compose layout?

Williamrai: this was added to push the search empty view up when keyboard was visible but now it is pushing it too far in both portrait and landscape modes.

Williamrai: I initially added the imeHeight as padding to ensure the view stayed visible when the keyboard appeared, but it wasn’t displaying correctly, so I’ve removed that logic. Now, in landscape mode with the keyboard visible, the view is partially visible which I think is acceptable. What do you think?

this is on pixel 2
[landscape view](https://github.com/user-attachments/assets/0ede44e4-875e-45b5-9e0f-35c109bb94be)

cooltey: Ah got it, that's totally acceptable. Thanks!

PR #5770, opened 2025-07-15
Follow-up: updategenerate_wiki_languages.py  script to have a different array format 

cooltey:
### What does this do?
- Use `|` to separate the key and values
- Only add the parent language code as a value if the language variants have fewer than two codes.


scripts/generate_wiki_languages.py
Williamrai, line 1: Just confirming, the structure here would be: main_language_code | variants_code, after the separator we would have variants right?

cooltey: Yes. If the main language code itself is considered as a variant, then it will be added after the `|`

PR #5780, opened 2025-07-19
Refactor on this day game to use LocalDate

Isira-Seneviratne:
### What does this do?
Refactors the On This Day game to use `LocalDate` instead of the legacy `Date` and `Calendar` classes.

### Why is this needed?
It is [not recommended](https://docs.oracle.com/javase/tutorial/datetime/iso/legacy.html) to use the older `Date` and `Calendar` classes.

**Phabricator:**


app/src/main/java/org/wikipedia/games/onthisday/OnThisDayGameBaseFragment.kt
```diff
@@ -58,90 +56,66 @@ abstract class OnThisDayGameBaseFragment : Fragment() {
     protected fun prepareAndOpenArchiveCalendar(viewModel: OnThisDayGameViewModel) {
         lifecycleScope.launch {
             val startDateBasedOnLanguage = LANG_CODES_SUPPORTED.associateWith { dateReleasedForLang(it) }
-            val localDate = startDateBasedOnLanguage[viewModel.wikiSite.languageCode]
-            val startDate = Date.from(localDate?.atStartOfDay(ZoneId.systemDefault())?.toInstant())
+            val startDate = startDateBasedOnLanguage[viewModel.wikiSite.languageCode]!!
```
cooltey: Instead of having a `!!` for the startDate, would it be possible to handle this properly?

PR #5783, opened 2025-07-21
[Feature branch] The Donation Reminder

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T399601

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageActivity.kt
+++ b/app/src/main/java/org/wikipedia/page/PageActivity.kt
@@ -105,6 +105,7 @@ class PageActivity : BaseActivity(), PageFragment.Callback, LinkPreviewDialog.Lo
     private val isCabOpen get() = currentActionModes.isNotEmpty()
     private var exclusiveTooltipRunnable: Runnable? = null
     private var isTooltipShowing = false
+
     private val requestEditSectionLauncher = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) {
         if (it.resultCode == EditHandler.RESULT_REFRESH_PAGE) {
             FeedbackUtil.makeSnackbar(this, getString(R.string.edit_saved_successfully))


app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.kt
```diff
@@ -1,28 +1,43 @@
...
 import org.wikipedia.views.LinearLayoutOverWebView
 import org.wikipedia.views.ObservableWebView
+import java.util.Date
 
-class PageHeaderView : LinearLayoutOverWebView, ObservableWebView.OnScrollChangeListener {
+class PageHeaderView(context: Context, attrs: AttributeSet? = null) : LinearLayoutOverWebView(context, attrs), ObservableWebView.OnScrollChangeListener {
     interface Callback {
         fun onImageClicked()
         fun onCallToActionClicked()
+        @Experimental
```
dbrant: What's the rationale for using the `parcelize.Experimental` annotation here? (or anywhere?)

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderViewModel.kt
```diff
@@ -0,0 +1,195 @@
...
+        var maximumAmount = donationConfig?.currencyMaximumDonation?.get(currencyCode) ?: 0f
+        if (maximumAmount == 0f) {
+            val defaultMin = donationConfig?.currencyMinimumDonation?.get(GooglePayComponent.CURRENCY_FALLBACK) ?: 0f
+            if (defaultMin > 0f) {
+                maximumAmount = (donationConfig?.currencyMinimumDonation?.get(currencyCode) ?: 0f) / defaultMin *
+                        (donationConfig?.currencyMaximumDonation?.get(GooglePayComponent.CURRENCY_FALLBACK) ?: 0f)
+            }
+        }
+
+        val presets = DonationReminderHelper.currencyAmountPresets[currentCountryCode] ?: listOf(0f)
```
dbrant: Instead of `0` as a fallback here, let's use `minimumAmount`, to reduce the chance of having a case where `0 €` is a selectable option.

dbrant (changes requested):
The text in the card doesn't seem to make sense:
"...you asked us to donate after reading €1."

image

app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.kt
Williamrai: https://github.com/wikimedia/apps-android-wikipedia/pull/5787#pullrequestreview-3088829196 The idea was to mark any experimental functions like this so we can easily find and review/remove them later.

dbrant: I see... If we really want to mark something as experimental, it's probably enough to leave a comment, instead of a totally unrelated annotation.

cooltey: Done. Change to use `//TODO:` instead of an annotation

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderViewModel.kt
cooltey: Done.

cooltey:
> The text in the card doesn't seem to make sense: "...you asked us to donate after reading €1."

Done. Didn't notice the verbiage has been changed 😓

UjuiUjuMandan:
The fdroid build failed after this.

https://gitlab.com/fdroid/fdroiddata/-/merge_requests/26668

Reference to proprietary GooglePayComponent directly in main variant is unacceptable, because it bypasses isGooglePayAvailable.

@cooltey Please keep it in DonateViewModel.

(How come there is no test build on pull requests for each variant...?)

PR #5784, opened 2025-07-21
Upgrade fundraising configuration to support A/B testing.

dbrant:
This bumps our `CAMPAIGN_VERSION` number to `2`, and updates the structures and logic accordingly.

https://phabricator.wikimedia.org/T398037

### DIFF
--- a/app/src/main/java/org/wikipedia/dataclient/donate/Campaign.kt
+++ b/app/src/main/java/org/wikipedia/dataclient/donate/Campaign.kt
@@ -34,8 +34,8 @@ class Campaign(
     }
 
     fun getAssetsForLang(lang: String): Assets? {
-        val list = assets[lang].orEmpty()
-        if (list.isEmpty()) {
+        val list = assets[lang]
+        if (list.isNullOrEmpty()) {
             return null
         } else if (list.size == 1) {
             return list[0]
@@ -57,7 +57,7 @@ class Campaign(
                 return i
             }
         }
-        return assets.size - 1
+        return assets.lastIndex
     }
 
     @Serializable


app/src/main/java/org/wikipedia/dataclient/donate/Campaign.kt
```diff
@@ -23,12 +25,45 @@ class Campaign(
...
+            return id
+        }
+        return id + "_" + assetsForLang.id
+    }
+
     fun getAssetsForLang(lang: String): Assets? {
-        return assets[lang]
+        val list = assets[lang].orEmpty()
+        if (list.isEmpty()) {
+            return null
```
cooltey: instead of having a `orEmpty()` on line `37`, can we use a `list.isNullOrEmpty` on line `38`?

app/src/main/java/org/wikipedia/dataclient/donate/Campaign.kt
```diff
@@ -23,12 +25,45 @@ class Campaign(
...
+        val random = Random(WikipediaApp.instance.appInstallID.hashCode() + id.hashCode())
+        val f = random.nextFloat()
+        var sum = 0f
+        for (i in assets.indices) {
+            sum += assets[i].weight
+            if (f <= sum) {
+                return i
+            }
+        }
+        return assets.size - 1
```
cooltey: Very minor, how about using `assets.lastIndex`?

app/src/main/java/org/wikipedia/dataclient/donate/Campaign.kt
dbrant: Sure!

app/src/main/java/org/wikipedia/dataclient/donate/Campaign.kt
dbrant: Correct!

PR #5787, opened 2025-07-21
Donation reminder: In-article prompt

cooltey:
Adjust the `DonationReminderConfig` data class for different active states. Please follow the task description to do the tests.

**Phabricator:**
https://phabricator.wikimedia.org/T399637


app/src/main/res/layout/view_donation_reminder_card.xml
```diff
@@ -0,0 +1,81 @@
...
+            app:chipMinHeight="22dp"/>
+
+        <TextView
+            android:id="@+id/messageTitleView"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:layout_marginBottom="4dp"
+            android:fontFamily="sans-serif-medium"
+            android:textSize="16sp"
+            android:textColor="?attr/primary_color"
```
Williamrai, line 38: what do you think of directly using this style here
`style="@style/TextAppearance.Material3.TitleMedium"`

app/src/main/res/layout/view_donation_reminder_card.xml
```diff
@@ -0,0 +1,81 @@
...
+            android:text="@string/donation_reminders_initial_prompt_title"/>
+
+        <TextView
+            android:id="@+id/messageTextView"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:textSize="14sp"
+            android:textColor="?attr/secondary_color"
+            android:text="@string/donation_reminders_initial_prompt_message"
```
Williamrai, line 47: also here,
`style="@style/TextAppearance.Material3.BodyMedium"`

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderCardView.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    val binding = ViewDonationReminderCardBinding.inflate(LayoutInflater.from(context), this, true)
+
+    init {
+        elevation = 0f
+    }
+
+    fun setTitle(title: String) {
+        val titleWithReservedSpace = "$title    %" // HACK: Reserve space for the icon
+        val spannableString = SpannableString(titleWithReservedSpace)
```
Williamrai, line 27: The gap between the text and the heart icon is bigger from the figma design, i think two space is enough here to match the figma.

Screenshot 2025-08-05 at 9 59 22 AM

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderHelper.kt
```diff
@@ -22,36 +25,131 @@ object DonationReminderHelper {
...
+    private fun resetDonationReminder() {
+        Prefs.donationReminderConfig.let { config ->
+            if (config.articleVisit > 0 && config.articleFrequency > 0 &&
+                config.articleVisit % config.articleFrequency == 0 &&
+                !config.initialPromptActive) {
+                // When reaching the article frequency, reset the configuration
+                Prefs.donationReminderConfig = config.copy(
+                    finalPromptActive = true,
+                    finalPromptCount = 0
+                )
```
Williamrai, line 142: Do we need to reset the article visit or continue counting? Users can change their frequency later.

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderHelper.kt
```diff
@@ -22,36 +25,131 @@ object DonationReminderHelper {
...
+    val isEnabled
+        get() = ReleaseUtil.isDevRelease ||
+                (enabledCountries.contains(GeoUtil.geoIPCountry.orEmpty()) &&
+                        enabledLanguages.contains(WikipediaApp.Companion.instance.languageState.appLanguageCode) &&
+                        LocalDate.now() <= LocalDate.of(2025, 12, 1) && !AccountUtil.isLoggedIn)
+
+    val hasActiveReminder
+        get() = maybeShowInitialDonationReminder(false) || maybeShowDonationReminder(false)
+
+    var shouldShowSettingSnackbar = false
```
Williamrai, line 44: do you think we can rename **hasActiveReminder** to something else like **hasShownReminderCard**, even though when we don't have reminder set this becomes true to show the initial prompt? or if we are checking for reminder i think we can use the values of **articleFrequency** and **donateAmount**

Williamrai (commented):
Would it be helpful to add a **@MARK_EXPERIMENT** comment to any new or modified functions (not already in **DonationReminderHelper**) so we can easily identify and clean them up later?

Williamrai:
If a user sees the donation card on the article screen but doesn't interact with it, and later enables donation reminders from the settings, the **initialPromptActive** and **finalPromptActive** flags remain true. This is because **donationReminderDismissed** is only triggered through actions on the **DonationReminderCard**. As a result, the user will never see either the initial or final donation reminder prompts. Maybe we can aslo set them to false if user enables reminder from the settings. 

Initially **Prefs.donationReminderConfig.isEnabled** is always set to false, which causes **hasActiveReminder** to always be false and prevents the initial donation reminder card from appearing.

app/src/main/res/layout/view_donation_reminder_card.xml
cooltey: Done

app/src/main/res/layout/view_donation_reminder_card.xml
cooltey: Done

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderHelper.kt
cooltey: Done.

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderCardView.kt
cooltey: Done.

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderHelper.kt
cooltey: The logic has been simplified. and add `@Experimental` to the `interface functions`.

cooltey:
> If a user sees the donation card on the article screen but doesn't interact with it, and later enables donation reminders from the settings, the **initialPromptActive** and **finalPromptActive** flags remain true. This is because **donationReminderDismissed** is only triggered through actions on the **DonationReminderCard**. As a result, the user will never see either the initial or final donation reminder prompts. Maybe we can aslo set them to false if user enables reminder from the settings.
> 
> Initially **Prefs.donationReminderConfig.isEnabled** is always set to false, which causes **hasActiveReminder** to always be false and prevents the initial donation reminder card from appearing.

Logic has been changed. Please checkout again to see if the issues are still valid.

cooltey:
> > If a user sees the donation card on the article screen but doesn't interact with it, and later enables donation reminders from the settings, the **initialPromptActive** and **finalPromptActive** flags remain true. This is because **donationReminderDismissed** is only triggered through actions on the **DonationReminderCard**. As a result, the user will never see either the initial or final donation reminder prompts. Maybe we can aslo set them to false if user enables reminder from the settings.

Hi @Williamrai, this has been fixed and the PR is ready for review again.

Williamrai (commented):
Everything looks good. Just wanted to confirm:
1. **(isInitialPrompt || Prefs.donationReminderConfig.finalPromptCount == DonationReminderHelper.MAX_REMINDER_PROMPTS)** on **donationReminderCardNegativeClicked**
We only show the snackbar for the initial prompt, and not when the user dismisses the final donation reminder. is that correct?

PR #5793, opened 2025-07-28
Correctly display error(s) in SearchResultsFragment.

dbrant:
The `SearchResultsFragment` has a WikiErrorView for showing any errors that come from the API calls, but the WikiErrorView is never actually used! So, currently API errors will just be ignored, and the screen will be blank.

This fixes the logic to display errors properly.


dbrant:
> When there is no internet and user has the previous search list then the error view appears on top of the list. [Bug](https://github.com/user-attachments/assets/71034eb2-87ad-4631-8a4c-986bb836e5bb)

@Williamrai This is (finally) ready.

PR #5822, opened 2025-08-07
Donation Reminders Instrumentation

Williamrai:
### What does this do?
- adds instrumentation for donation reminders and also adds ime action send for textField 

**Phabricator:**
https://phabricator.wikimedia.org/T399640


app/src/extra/java/org/wikipedia/donate/GooglePayActivity.kt
```diff
@@ -290,7 +291,11 @@ class GooglePayActivity : BaseActivity() {
                                 binding.checkBoxTransactionFee.isChecked,
                                 binding.checkBoxRecurring.isChecked,
                                 if (viewModel.emailOptInRequired) binding.checkBoxAllowEmail.isChecked else true,
-                                intent.getStringExtra(DonateDialog.ARG_CAMPAIGN_ID).orEmpty().ifEmpty { CAMPAIGN_ID_APP_MENU })
+                                intent.getStringExtra(DonateDialog.ARG_CAMPAIGN_ID).orEmpty().ifEmpty {
+                                    if (DonationReminderHelper.isEnabled) {
+                                        DonationReminderHelper.CAMPAIGN_ID
+                                    } else CAMPAIGN_ID_APP_MENU
+                                })
```
cooltey: This will send `appmenu_reminder` even if the users donate from the More menu, which is incorrect.

It would be better if you could assign the `appmenu_reminder` in `setupDirectGooglePayButton` inside the `DonateDialog`, which can track whether the donation was actually from the feature.

cooltey (changes requested):
This looks very good! Good job 👍 Just have a minor change requested.

app/src/extra/java/org/wikipedia/donate/GooglePayActivity.kt
Williamrai: done, also added donateOtherButton inside the setupDirectGooglePayButton since the event can also be called from the more menu.

PR #5824, opened 2025-08-08
Create separate activity for Suggested Edits

cooltey:
### What does this do?
Create a new `Activity` for Suggested Edits.

**Phabricator:**
https://phabricator.wikimedia.org/T401017

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
@@ -1,12 +1,16 @@
 package org.wikipedia.suggestededits
 
 import android.app.Activity
-import android.net.Uri
 import android.os.Bundle
 import android.view.LayoutInflater
+import android.view.Menu
+import android.view.MenuInflater
+import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
 import androidx.activity.result.contract.ActivityResultContracts
+import androidx.core.net.toUri
+import androidx.core.view.MenuProvider
 import androidx.core.view.isVisible
 import androidx.core.widget.NestedScrollView
 import androidx.fragment.app.Fragment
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
Williamrai, line 1: [Bug](https://github.com/user-attachments/assets/21b8d51f-cc4c-4a08-b146-6e802c98e48f)
The tooltip still appears here. I think we also need to add the A/B test check in this function **showOneTimeSequentialUserStatsTooltips**.

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -147,9 +166,12 @@ class SuggestedEditsTasksFragment : Fragment() {
...
                     FlowEventBus.events.collectLatest { event ->
-                        if (event is LoggedOutEvent &&
-                            (requireActivity() as MainActivity).isCurrentFragmentSelected(this@SuggestedEditsTasksFragment)) {
-                            refreshContents()
+                        if (event is LoggedOutEvent) {
+                            if (inActivityAbTestGroup) {
+                                refreshContents()
+                            } else if ((requireActivity() as MainActivity).isCurrentFragmentSelected(this@SuggestedEditsTasksFragment)) {
+                                refreshContents()
+                            }
```
Williamrai: Do we need this logic, both are refreshing the content?

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
cooltey: Yes, done!

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
cooltey: Done!

PR #5833, opened 2025-08-12
Activity tab categories module design

Williamrai:
### What does this do?
- Adds top categories of the month for activity tab

**Phabricator:**
https://phabricator.wikimedia.org/T401019


app/src/main/java/org/wikipedia/activitytab/ActivityTabFragment.kt
```diff
@@ -50,9 +54,16 @@ class ActivityTabFragment : Fragment() {
         }
     }
 
+    override fun onResume() {
+        super.onResume()
+        viewModel.load()
```
dbrant: There should not be a need to `load` anything from the ViewModel upon resuming. (The goal of a ViewModel is to load things independently of the activity/fragment lifecycle. Loading upon resuming will cause unnecessary network and database transactions.)

PR #5843, opened 2025-08-14
Donation reminder UI fixes

Williamrai:
### What does this do?
- ui fixes
- adds snackbar on confirmation, saves confirmation only when donation reminders value changes
- fixes min amount validation check
- makes "About this experiment" always available

### DIFF
--- a/app/src/main/java/org/wikipedia/util/FeedbackUtil.kt
+++ b/app/src/main/java/org/wikipedia/util/FeedbackUtil.kt
@@ -166,22 +166,19 @@ object FeedbackUtil {
         val marginForNavbar = if (windowInsets != null) {
             when (navigationType) {
                 0, 1 -> {
-                    val navBarInsets = windowInsets.getInsets(WindowInsetsCompat.Type.navigationBars())
-                    navBarInsets.bottom
+                    windowInsets.getInsets(WindowInsetsCompat.Type.navigationBars())
                 }
-                2 -> {
-                    val gestureInsets = windowInsets.getInsets(WindowInsetsCompat.Type.systemGestures())
-                    gestureInsets.bottom
+                else -> {
+                    windowInsets.getInsets(WindowInsetsCompat.Type.systemGestures())
                 }
-                else -> 0
             }
-        } else 0
+        } else null
 
         params.setMargins(
             params.leftMargin,
             params.topMargin,
             params.rightMargin,
-            params.bottomMargin + marginForNavbar
+            params.bottomMargin + (marginForNavbar?.bottom ?: 0)
         )
         view.layoutParams = params
         return snackbar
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/wikipedia/util/FeedbackUtil.kt
```diff
@@ -147,6 +151,40 @@ object FeedbackUtil {
         return makeSnackbar(findBestView(activity), text, duration, wikiSite)
     }
 
+    fun makeNavigationAwareSnackbar(activity: Activity, text: CharSequence, wikiSite: WikiSite = WikipediaApp.instance.wikiSite): Snackbar {
+        val rootView = activity.findViewById<View>(android.R.id.content)
+        val snackbar = makeSnackbar(rootView, text, LENGTH_DEFAULT, wikiSite)
+        val view = snackbar.view
+        val params = view.layoutParams as ViewGroup.MarginLayoutParams
+        val navigationType = Settings.Secure.getInt(activity.contentResolver, "navigation_mode", 0)
```
cooltey, line 163: This works for me, and it would be better if you could put the reference about which number belongs to which navigation mode. Thanks!

app/src/main/java/org/wikipedia/util/FeedbackUtil.kt
```diff
@@ -147,6 +151,40 @@ object FeedbackUtil {
...
+        val windowInsets = ViewCompat.getRootWindowInsets(rootView)
+
+        val bottomMargin = if (windowInsets != null) {
```
cooltey: Would it be better if you could rename it to something like `marginForNavbar`?

app/src/main/java/org/wikipedia/util/FeedbackUtil.kt
Williamrai: done.

app/src/main/java/org/wikipedia/util/FeedbackUtil.kt
Williamrai: done.

app/src/main/res/values/strings.xml
```diff
@@ -432,7 +432,7 @@
     <string name="theme_chooser_dialog_image_dimming_switch_label">Image dimming (in Dark theme)</string>
     <string name="preference_title_prefer_offline_content">Prefer offline content</string>
     <string name="preference_summary_prefer_offline_content">Save data usage by loading articles that are available offline rather than always loading the latest version of an article</string>
-    <string name="preference_title_delete_local_donation_history">Delete local donation history</string>
+    <string name="preference_title_delete_local_donation_history">Clear donation history</string>
```
Williamrai, line 435: Would it be okay to replace this with a new string, since it’s only being used for the settings title?

cooltey: It should be fine.

Williamrai: okay. Thank you.

PR #5849, opened 2025-08-16
Restrict width for search widget on Android versions below 12

Isira-Seneviratne:
### What does this do?
This change restricts the size of the search widget on Android versions below 12.

### Why is this needed?
This change avoids widget sizing issues on Android versions below 12.

**Phabricator:**


Williamrai:
Thanks for working on this! I appreciate the effort you've put into this solution. I think there might be a simpler approach we could try first, we can set `android:minResizeWidth` in the widget metadata XML to prevent resizing below this limit. This would support all API levels, require fewer code changes, and avoid adding another dependency. What do you think about trying that approach?

Isira-Seneviratne:
> Thanks for working on this! I appreciate the effort you've put into this solution. I think there might be a simpler approach we could try first, we can set `android:minResizeWidth` in the widget metadata XML to prevent resizing below this limit. This would support all API levels, require fewer code changes, and avoid adding another dependency. What do you think about trying that approach?

Sure, the XML could be overriden for API levels 31 and later.

Williamrai:
> > Thanks for working on this! I appreciate the effort you've put into this solution. I think there might be a simpler approach we could try first, we can set `android:minResizeWidth` in the widget metadata XML to prevent resizing below this limit. This would support all API levels, require fewer code changes, and avoid adding another dependency. What do you think about trying that approach?
> 
> Sure, the XML could be overriden for API levels 31 and later.

Thanks again for taking the time to make these updates. I think we may not need the `xml-v31` override for this fix. Adding `android:minResizeWidth` in the base XML should apply correctly across all API levels, so the extra version-specific file might not be necessary.

Williamrai:
Hi, I don’t see the `android:minResizeWidth="160dp"` fix in this PR anymore. I think it may have been removed accidentally.

PR #5851, opened 2025-08-18
Send event when initiating account vanishing.

dbrant:
We'd like to send an event to analytics when the user initiates the Account Vanishing workflow in Settings.

https://phabricator.wikimedia.org/T402049


app/src/main/java/org/wikipedia/settings/LogoutPreference.kt
```diff
@@ -70,6 +72,7 @@ class LogoutPreference : Preference {
                     .setNegativeButton(android.R.string.cancel, null)
                     .setPositiveButton(R.string.account_vanish_request_title) { _, _ ->
                         it.finish()
+                        CreateAccountEvent(LoginActivity.SOURCE_SETTINGS).logVanish()
```
cooltey, line 75: Do you have the answer for `settings` vs. `setting` from Shay?

PR #5854, opened 2025-08-19
For updating DAO functions: improve efficiency of TalkPageSeen

dbrant:
(to be merged into the DAO branch)

### DIFF
--- a/app/src/main/java/org/wikipedia/talk/TalkTopicsViewModel.kt
+++ b/app/src/main/java/org/wikipedia/talk/TalkTopicsViewModel.kt
@@ -105,12 +105,13 @@ class TalkTopicsViewModel(var pageTitle: PageTitle) : ViewModel() {
             val discussionToolsInfoResponse = ServiceFactory.get(pageTitle.wikiSite).getTalkPageTopics(pageTitle.prefixedText,
                     OfflineCacheInterceptor.SAVE_HEADER_SAVE, pageTitle.wikiSite.languageCode, UriUtil.encodeURL(pageTitle.prefixedText))
 
-            seenThreadItemsSha.clear()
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/talk/TalkTopicsViewModel.kt
```diff
@@ -105,13 +105,14 @@ class TalkTopicsViewModel(var pageTitle: PageTitle) : ViewModel() {
             val discussionToolsInfoResponse = ServiceFactory.get(pageTitle.wikiSite).getTalkPageTopics(pageTitle.prefixedText,
                     OfflineCacheInterceptor.SAVE_HEADER_SAVE, pageTitle.wikiSite.languageCode, UriUtil.encodeURL(pageTitle.prefixedText))
 
-            seenThreadItemsSha.clear()
-            seenThreadItemsSha.addAll(talkPageDao.getAll().map { it.sha })
```
dbrant, line 109: It's generally not a good idea to "get all" entries from a db table, and then perform operations on them. If you think about it, this is doing work that sqlite should be doing for us!
Whenever possible, we should let sqlite perform the operation we want, and return just the items we need.

app/src/main/java/org/wikipedia/talk/db/TalkPageSeenDao.kt
```diff
@@ -16,6 +16,9 @@ interface TalkPageSeenDao {
     @Query("SELECT * FROM TalkPageSeen")
     suspend fun getAll(): List<TalkPageSeen>
 
+    @Query("SELECT * FROM TalkPageSeen WHERE sha IN (:shaList)")
```
dbrant, line 19: Room provides a syntax for passing a _list_ of items to compare against!

PR #5857, opened 2025-08-20
Follow-up to DAO updates.

dbrant:
* The `ReadingListBehaviorsUtil` class contains some really tricky logic, which requires multiple db queries to be completed, and _also_ relies on an Activity's lifecycleScope (because it wants to show Snackbars and Dialogs, and make callbacks to the Activity). Therefore, even though it's "correct" to use MainScope here (or some other scope not tied to the Activity), _we should also_ explicitly check the calling Activity's lifecycle before making any callbacks or showing any Snackbars or Dialogs. Otherwise crashes may occur.


app/src/main/java/org/wikipedia/readinglist/ReadingListPreviewSaveDialogView.kt
```diff
@@ -52,7 +52,7 @@ class ReadingListPreviewSaveDialogView(context: Context, attrs: AttributeSet? =
         coroutineScope().launch(CoroutineExceptionHandler {
             _, throwable -> L.w(throwable)
         }) {
-            currentReadingLists = AppDatabase.instance.readingListDao().getAllLists()
+            currentReadingLists = AppDatabase.instance.readingListDao().getListsWithoutContents()
```
dbrant, line 55: Another optimization:
This dialog needs a list of the user's Reading Lists, for the purpose of checking whether a list with a specific name is already taken (It does not need the full contents of all the lists.)
Therefore, we can use this slightly different call, which is much more efficient.

PR #5858, opened 2025-08-20
Follow-up to navigationAwareSnackbar.

dbrant:
I don't actually see the purpose of querying the system variable `navigation_mode` to determine which insets to check. (i.e. it should always be `navigationBars()` insets)
(Correct me if I'm wrong, or if this doesn't work @Williamrai )


Williamrai:
> I don't actually see the purpose of querying the system variable `navigation_mode` to determine which insets to check. (i.e. it should always be `navigationBars()` insets) (Correct me if I'm wrong, or if this doesn't work @Williamrai )

Agreed, this is much cleaner.

PR #5863, opened 2025-08-22
Activity tab timeline

Williamrai:
**What does this do?**
- adds users activity timeline list in the activity screen
- uses a cursor based approach to load data for different sources in the load function of paging source, 

**Phabricator:**
https://phabricator.wikimedia.org/T399767


app/src/main/java/org/wikipedia/activitytab/timeline/ActivityTabTimeline.kt
```diff
@@ -0,0 +1,157 @@
...
+                ImageService.getRequest(LocalContext.current, url = timelineItem.thumbnailUrl)
+            AsyncImage(
+                model = request,
+                placeholder = BrushPainter(SolidColor(WikipediaTheme.colors.borderColor)),
+                error = BrushPainter(SolidColor(WikipediaTheme.colors.borderColor)),
+                contentScale = ContentScale.Crop,
+                contentDescription = null,
+                modifier = Modifier
+                    .size(56.dp)
+                    .clip(RoundedCornerShape(16.dp))
```
cooltey: This should be `8.dp` instead of `16.dp`.

app/src/main/java/org/wikipedia/activitytab/timeline/ActivityTabTimeline.kt
```diff
@@ -0,0 +1,157 @@
...
+                contentDescription = null
+            )
+        }
+        Column(
+            modifier = Modifier
+                .weight(1f)
+        ) {
+            Text(
+                text = timelineItem.apiTitle,
+                style = MaterialTheme.typography.titleMedium,
```
cooltey: The text style is different from the design.

app/src/main/java/org/wikipedia/activitytab/timeline/ActivityTabTimeline.kt
```diff
@@ -0,0 +1,157 @@
...
+                tint = WikipediaTheme.colors.primaryColor,
+                contentDescription = null
+            )
+        }
+        Column(
+            modifier = Modifier
+                .weight(1f)
+        ) {
+            Text(
+                text = timelineItem.apiTitle,
```
cooltey: Should use `displayTitle` instead of the `apiTitle`. And we should use `HtmlText` instead of `Text` since the displayTitle might contain HTML tags.

app/src/main/java/org/wikipedia/history/db/HistoryEntryWithImageDao.kt
```diff
@@ -18,6 +18,17 @@ import java.util.TimeZone
 @Dao
 interface HistoryEntryWithImageDao {
 
+    @Query("SELECT HistoryEntry.*, PageImage.imageName, PageImage.description, PageImage.geoLat, PageImage.geoLon, PageImage.timeSpentSec FROM HistoryEntry LEFT OUTER JOIN PageImage ON (HistoryEntry.namespace = PageImage.namespace AND HistoryEntry.apiTitle = PageImage.apiTitle AND HistoryEntry.lang = PageImage.lang) INNER JOIN(SELECT lang, apiTitle, MAX(timestamp) as max_timestamp FROM HistoryEntry GROUP BY lang, apiTitle) LatestEntries ON HistoryEntry.apiTitle = LatestEntries.apiTitle AND HistoryEntry.timestamp = LatestEntries.max_timestamp WHERE HistoryEntry.timestamp BETWEEN :startTime AND :endTime ORDER BY timestamp DESC")
+    @RewriteQueriesToDropUnusedColumns
+    suspend fun getHistoryEntriesInRange(startTime: Date, endTime: Date): List<HistoryEntryWithImage>
+
+    @Query("""
```
cooltey: Can we keep single `"` instead of three `"""`?

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
```diff
@@ -0,0 +1,183 @@
...
+        return items to nextCursor
+    }
+}
+
+class ReadingListSource(
+    val dao: ReadingListPageDao
+) : TimelineSource {
+    override suspend fun fetch(pageSize: Int, cursor: Cursor?): Pair<List<TimelineItem>, Cursor?> {
+        val offset = (cursor as? Cursor.ReadingListCursor)?.offset ?: 0
+        val items = dao.getPagesWithLimitOffset(pageSize, offset).map { TimelineItem(
```
cooltey: Can `TimelineItem` move to the next line for a proper code format?

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
```diff
@@ -0,0 +1,183 @@
...
+    }
+}
+
+class HistoryEntrySource(
+    private val dao: HistoryEntryWithImageDao
+) : TimelineSource {
+
+    override suspend fun fetch(pageSize: Int, cursor: Cursor?): Pair<List<TimelineItem>, Cursor?> {
+        val offset = (cursor as? Cursor.HistoryEntryCursor)?.offset ?: 0
+        val items = dao.getHistoryEntriesWithOffset(pageSize, offset).map { TimelineItem(
```
cooltey: Can `TimelineItem` move to the next line for a proper code format?

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
```diff
@@ -0,0 +1,183 @@
...
+import java.time.format.TextStyle
+import java.time.temporal.ChronoUnit
+import java.util.Date
+import java.util.Locale
+
+interface TimelineSource {
+    suspend fun fetch(pageSize: Int, cursor: Cursor?): Pair<List<TimelineItem>, Cursor?>
+}
+
+fun formatDate(date: LocalDate): String {
```
cooltey: We have some useful date functions in `DateUtil` class, maybe you can find something useful there instead of creating one.

cooltey (changes requested):

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
Williamrai: done.

app/src/main/java/org/wikipedia/history/db/HistoryEntryWithImageDao.kt
Williamrai: i removed this function as this is not used. i added this for date approach.

app/src/main/java/org/wikipedia/activitytab/timeline/ActivityTabTimeline.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/timeline/ActivityTabTimeline.kt
Williamrai: Since we were using the material theme, i used the closest one to h3-article.

app/src/main/java/org/wikipedia/activitytab/timeline/ActivityTabTimeline.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/timeline/ActivityTabTimeline.kt
cooltey: The fontFamily should be `FontFamily.Serif`

Williamrai: done.

cooltey (changes requested):
Some final comments:
1. The item height seems too high and the header title should have bottom padding for a better display (similar to the #5866)



2. The title does not align to the center vertically if it does not have a description.



3.  Some items have a larger padding around them, is this expected?


4.  When I clicked on the March 3, 2022 talk item and it turned out was actually an edit from March 2, 2022. I think this might be something related to the timezone you have applied to the logic.

cooltey:
And one minor comment: can we name this `TimelineModule` to keep consistent across different modules? and not sure if it is fine to not have the subpackage for the `timeline` module.

Williamrai:
> The item height seems too high and the header title should have bottom padding for a better display (similar to the https://github.com/wikimedia/apps-android-wikipedia/pull/5866)

Fixed by changing vertical padding to 8.dp for each row and adding a bottom space of 8.dp height to ensure the spacing between each rows. Added padding 8.dp to the bottom of the date separator. 

> The title does not align to the center vertically if it does not have a description.

Fixed.

> Some items have a larger padding around them, is this expected?

Based on figma when we add button, it does have bigger padding but i have reduced the padding from 32.dp to 22.dp closer to figma. Fixed. 

> When I clicked on the March 3, 2022 talk item and it turned out was actually an edit from March 2, 2022. I think this might be something related to the timezone you have applied to the logic.

 The issue was due to local time zone to utc conversion. Fixed.

> And one minor comment: can we name this TimelineModule to keep consistent across different modules? and not sure if it is fine to not have the subpackage for the timeline module.

I have renamed the package, i think we can keep the subpackage since it contains more than one class related to timeline.

app/src/main/java/org/wikipedia/activitytab/ActivityTabFragment.kt
```diff
@@ -450,7 +467,88 @@ class ActivityTabFragment : Fragment() {
...
+                                        timelineItem = displayItem.item,
+                                        onItemClick = { item ->
+                                            when (item.activitySource) {
+                                                ActivitySource.EDIT -> {
+                                                    startActivity(
+                                                        ArticleEditDetailsActivity.newIntent(
+                                                            requireContext(),
+                                                            PageTitle(
+                                                                item.apiTitle,
+                                                                viewModel.wikiSite
```
cooltey: Please build the `PageTitle` properly with the description, displayTitle, namespace and so on.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
```diff
@@ -0,0 +1,198 @@
...
+    val wiki = timelineItem.wiki
+    wiki?.languageCode = timelineItem.lang
+    return PageTitle(
+        timelineItem.apiTitle,
+        wiki!!,
+        timelineItem.thumbnailUrl,
+        timelineItem.description,
+        timelineItem.displayTitle
+    )
+}
```
cooltey: For these extensions, I am not sure it is a good idea to include them inside this "repository" class. Would it be possible to include the `toHistoryEntry` and `toPageTitle` inside the `TimelineItem` class? and move the date-related functions to maybe a date extension?

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
```diff
@@ -0,0 +1,198 @@
...
+                }
+            )
+        }
+        val nextCursor =
+            if (items.size < pageSize) null else Cursor.HistoryEntryCursor(offset + items.size)
+        return items to nextCursor
+    }
+}
+
+class ApiTimelineSource(
```
cooltey: Not sure, but would it be better if we name it `UserContribPagingSource`?

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
```diff
@@ -0,0 +1,198 @@
...
+            DateUtil.getMonthOnlyDateString(localDate)
+        }
+
+        else -> {
+            DateUtil.getShortDateString(localDate)
+        }
+    }
+}
+
+class HistoryEntrySource(
```
cooltey: The naming seems a bit unclear. Would it be possible to name it `HistoryEntryPagingSource`?

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
```diff
@@ -0,0 +1,198 @@
...
+
+        localDate.year == now.year -> {
+            DateUtil.getMonthOnlyDateString(localDate)
+        }
+
+        else -> {
+            DateUtil.getShortDateString(localDate)
+        }
+    }
+}
```
cooltey: This whole function can be moved to the `DateUtils` class.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
```diff
@@ -0,0 +1,198 @@
...
+                source = -1
+            )
+        } ?: emptyList()
+
+        val nextCursor = response.continuation?.ucContinuation?.let { Cursor.ApiCursor(it) }
+        return items to nextCursor
+    }
+}
+
+class ReadingListSource(
```
cooltey: `ReadingListPagingSource`

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
```diff
@@ -0,0 +1,198 @@
+package org.wikipedia.activitytab.timeline
```
cooltey: There's a typo in the class name. It should be `TimelineRepository` instead.

app/src/main/java/org/wikipedia/history/db/HistoryEntryWithImageDao.kt
```diff
@@ -17,6 +17,11 @@ import java.util.TimeZone
 
 @Dao
 interface HistoryEntryWithImageDao {
+    @Query("SELECT HistoryEntry.*, PageImage.imageName, PageImage.description,PageImage.geoLat, PageImage.geoLon,PageImage.timeSpentSec FROM HistoryEntry LEFT OUTER JOIN PageImage ON (HistoryEntry.namespace = PageImage.namespace AND HistoryEntry.apiTitle = PageImage.apiTitle AND HistoryEntry.lang = PageImage.lang)INNER JOIN (SELECT lang, apiTitle, MAX(timestamp) as max_timestamp FROM HistoryEntry GROUP BY lang, apiTitle) LatestEntries ON HistoryEntry.apiTitle = LatestEntries.apiTitle AND HistoryEntry.timestamp = LatestEntries.max_timestamp ORDER BY timestamp DESC LIMIT :limit OFFSET :offset")
```
cooltey: Would it be possible to make sure the texts inside the query have proper spaces between different column names or syntax?

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
```diff
@@ -0,0 +1,199 @@
...
+import org.wikipedia.R
+import org.wikipedia.compose.components.HtmlText
+import org.wikipedia.compose.theme.BaseTheme
+import org.wikipedia.compose.theme.WikipediaTheme
+import org.wikipedia.theme.Theme
+import org.wikipedia.util.DateUtil
+import org.wikipedia.views.imageservice.ImageService
+import java.util.Date
+
+// @TODO: MARK_ACTIVITY_TAB retrieve description and thumbnail for contributions through API
```
cooltey, line 46: Is this still valid?

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
```diff
@@ -0,0 +1,198 @@
...
+    var authority: String = "",
+    var lang: String = "",
+    var apiTitle: String = "",
+    var displayTitle: String = "",
+    var namespace: String = "",
+    val wiki: WikiSite? = null
+)
+
+sealed class Cursor {
+    data class ApiCursor(val token: String?) : Cursor()
```
cooltey: Same, maybe consider `UserContribCursor()`.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelinePagingSource.kt
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    override suspend fun load(params: LoadParams<TimelinePageKey>): LoadResult<TimelinePageKey, TimelineItem> {
+        return try {
+            val key = params.key ?: TimelinePageKey()
+            val allItems = mutableListOf<TimelineItem>()
+            val newCursors = mutableMapOf<String, Cursor>()
+
+            sources.forEach { source ->
+                val sourceName = source.javaClass.simpleName
```
cooltey: Instead of using `javaClass.simpleName`, would it be possible to add `val id: String` or `val key: String` to the `TimelineSource`?

cooltey (changes requested):

app/src/main/java/org/wikipedia/activitytab/timeline/TimelinePagingSource.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
Williamrai: yes, waiting on dmitry.

app/src/main/java/org/wikipedia/history/db/HistoryEntryWithImageDao.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
Williamrai: done

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
Williamrai: I renamed it. I named this API because I assumed that later we might need to make multiple API calls here to load the page summary, in case a single call isn’t sufficient.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/ActivityTabFragment.kt
Williamrai: I think this is still a TODO since i can only get title name from the userContrib api, still waiting on dmitry input on how to efficiently get the other properties.

Williamrai: adding logic to retrieve this value after discussion.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
Williamrai: adding logics to retrieve this value after discussion.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRepository.kt
```diff
@@ -0,0 +1,200 @@
...
+        val items = dao.getPagesWithLimitOffset(pageSize, offset).map {
+            TimelineItem(
+                id = it.mtime + it.atime,
+                pageId = 0,
+                apiTitle = it.apiTitle,
+                displayTitle = it.displayTitle,
+                description = it.description,
+                thumbnailUrl = it.thumbUrl,
+                timestamp = Date(it.mtime),
+                wiki = it.wiki,
```
cooltey: This will introduce issue if you always assign the primary language wiki here.

You should assign from the reading list `lang` here by creating it like the following:
```
wiki = WikiSite.forLanguageCode(it.lang)
```

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRepository.kt
```diff
@@ -0,0 +1,200 @@
...
+        )
+        entry.title.thumbUrl = thumbnailUrl
+        entry.title.description = description
+
+        return entry
+    }
+
+    fun toPageTitle(): PageTitle {
+        val wiki = wiki
+        wiki?.languageCode = lang
```
cooltey: If you follow the comment above, this won't be necessary.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRepository.kt
cooltey: or maybe you can assign the `lang` with the reading list `lang` and create the `WikiSite` inside `toPageTitle()`.

cooltey (changes requested):
Some final comments and @dbrant please review and merge if it looks good to you.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRepository.kt
Williamrai: i did wiki = WikiSite.forLanguageCode(it.lang) and removed the code from line 180 and 181 inside the toPageTitle() function.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRepository.kt
Williamrai: i followed the above code fix and removed this line.

app/src/main/java/org/wikipedia/activitytab/ActivityTabFragment.kt
Williamrai: added logic.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineRespository.kt
Williamrai: added logic.

cooltey (approved):
LGTM

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
Williamrai: added.

cooltey:
Just noticed this: If you have empty data in your timeline, it should show an empty card. [Figma](https://www.figma.com/design/MXUBnIJvHfMBHhPZta7yRm/Android---%3E-Activity-Tab?node-id=47-5010&m=dev).

Williamrai:
> @Williamrai Looks good, except dark mode:
> 
> image

Fixed.

Williamrai:
> Just noticed this: If you have empty data in your timeline, it should show an empty card. [Figma](https://www.figma.com/design/MXUBnIJvHfMBHhPZta7yRm/Android---%3E-Activity-Tab?node-id=47-5010&m=dev).

Added.

PR #5883, opened 2025-08-29
Activity tab onboarding

Williamrai:
### What does this do?
- adds onboarding activity and views

**Phabricator:**
https://phabricator.wikimedia.org/T401612

**Note: Waiting on resources from design**


app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -123,6 +125,12 @@ class MainFragment : Fragment(), BackPressedHandler, MenuProvider, FeedFragment.
         }
     }
 
+    private val onboardingLauncher = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) { result ->
```
dbrant: Naming: can we name this something much more specific than `onboardingLauncher`? Maybe `activityTabOnboardingLauncher`?

dbrant (changes requested):
A tiny note inline, and also:

On my device, the whole layout overflows and requires scrolling to reach the Continue button. Can we keep the Continue / Learn more buttons floating at the bottom, above the rest of the content, and allow the content to scroll behind them? (i.e. put those buttons in a container with 16dp padding, and keep it pinned to the bottom.)

image

Williamrai:
> A tiny note inline, and also:
> 
> On my device, the whole layout overflows and requires scrolling to reach the Continue button. Can we keep the Continue / Learn more buttons floating at the bottom, above the rest of the content, and allow the content to scroll behind them? (i.e. put those buttons in a container with 16dp padding, and keep it pinned to the bottom.)
> 
> image

Fixed.

app/src/main/java/org/wikipedia/main/MainFragment.kt
Williamrai: done, thank you.

PR #5900, opened 2025-09-04
Include user-agent in language-update scripts, and set up Test2wiki.

dbrant:
A couple of things:
* Since the Foundation's user-agent policy is now enforced, we must send a `User-Agent` header any time we make an API call from our python scripts, otherwise they will fail.
* We should include [test2wiki](https://test2.wikipedia.org/) as part of our languages, in addition to testwiki, since this is also an important testing ground for certain features.
* While we're here, let's update our language list to the latest.


scripts/generate_wiki_languages.py
```diff
@@ -7,6 +7,9 @@
 import json
 import requests
 
+headers = {
+    'User-Agent': 'WikipediaApp/2.7 (android-support@wikimedia.org) language-list-script/1.0'
+}
```
cooltey: Minor thought: would it be possible to have this `headers` as a global variable so it can be used in multiple `.py` files?

dbrant: done!

PR #5904, opened 2025-09-04
Activity tab shimmer loading

Williamrai:
### What does this do?
- adds loading shimmer effect for loading

### DIFF
--- a/app/src/main/java/org/wikipedia/activitytab/ActivityTabFragment.kt
+++ b/app/src/main/java/org/wikipedia/activitytab/ActivityTabFragment.kt
@@ -185,7 +185,7 @@ class ActivityTabFragment : Fragment() {
             modifier = Modifier
                 .fillMaxSize()
                 .background(WikipediaTheme.colors.paperColor),
-            containerColor = WikipediaTheme.colors.paperColor,
+            containerColor = WikipediaTheme.colors.paperColor
         ) { paddingValues ->
             var isRefreshing by remember { mutableStateOf(false) }
             val state = rememberPullToRefreshState()
--- a/app/src/main/java/org/wikipedia/activitytab/ActivityTabFragment.kt
+++ b/app/src/main/java/org/wikipedia/activitytab/ActivityTabFragment.kt
@@ -20,6 +20,7 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material3.Button
 import androidx.compose.material3.ButtonDefaults
@@ -40,6 +41,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.ComposeView
@@ -70,6 +72,7 @@ import org.wikipedia.activity.BaseActivity
 import org.wikipedia.activity.FragmentUtil.getCallback
 import org.wikipedia.activitytab.timeline.ActivitySource
 import org.wikipedia.activitytab.timeline.TimelineDateSeparator
+import org.wikipedia.activitytab.timeline.TimelineItem
 import org.wikipedia.activitytab.timeline.TimelineModule
 import org.wikipedia.activitytab.timeline.TimelineModuleEmptyView
 import org.wikipedia.auth.AccountUtil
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/wikipedia/compose/extensions/Modifier.kt
```diff
@@ -103,3 +118,32 @@ private fun PreviewPulse() {
...
     }
 }
+
+fun Modifier.shimmerEffect(
+    shimmerColors: List<Color>? = null,
+    durationMs: Int = 1200
+): Modifier = composed {
+    val colors = shimmerColors ?: WikipediaTheme.colors.getShimmerColors()
+    var size by remember { mutableStateOf(IntSize.Zero) }
+    val transition = rememberInfiniteTransition(label = "shimmer")
```
cooltey: The `label` may not be necessary in this case since it has a pre-defined label in the function.

app/src/main/java/org/wikipedia/compose/extensions/Modifier.kt
```diff
@@ -103,3 +118,32 @@ private fun PreviewPulse() {
...
+
+    val startOffsetX by transition.animateFloat(
+        initialValue = -2 * size.width.toFloat(),
+        targetValue = 2 * size.width.toFloat(),
+        animationSpec = infiniteRepeatable(
+            animation = tween(durationMs, easing = LinearEasing),
+            repeatMode = RepeatMode.Restart
+        ),
+        label = "shimmer"
```
cooltey: Same.

app/src/main/java/org/wikipedia/compose/theme/WikipediaColor.kt
```diff
@@ -118,3 +119,20 @@ val SepiaColors = WikipediaColor(
     additionColor = ComposeColors.Blue300_15,
     overlayColor = ComposeColors.Black_30
 )
+
+@Composable
+fun WikipediaColor.getShimmerColors(): List<Color> {
```
cooltey: Can we name it `shimmerColors` instead?

app/src/main/java/org/wikipedia/compose/extensions/Modifier.kt
```diff
@@ -103,3 +118,32 @@ private fun PreviewPulse() {
...
+): Modifier = composed {
+    val colors = shimmerColors ?: WikipediaTheme.colors.getShimmerColors()
+    var size by remember { mutableStateOf(IntSize.Zero) }
+    val transition = rememberInfiniteTransition(label = "shimmer")
+
+    val startOffsetX by transition.animateFloat(
+        initialValue = -2 * size.width.toFloat(),
+        targetValue = 2 * size.width.toFloat(),
+        animationSpec = infiniteRepeatable(
+            animation = tween(durationMs, easing = LinearEasing),
```
cooltey: How about making the `easing` parameter so it can be more customizable?

app/src/main/java/org/wikipedia/activitytab/ActivityTabFragment.kt
```diff
@@ -770,7 +744,66 @@ class ActivityTabFragment : Fragment() {
...
+fun ActivityTabShimmerView() {
+    Box(
+        modifier = Modifier
+            .padding(16.dp)
+            .clip(RoundedCornerShape(size = 12.dp))
+            .fillMaxWidth()
+            .shimmerEffect()
+            .size(120.dp)
+    )
+}
```
cooltey, line 809: Maybe we can make this into a component called `LoadingShimmerView` with some parameters such as rounded corner and size...etc.

cooltey (changes requested):

app/src/main/java/org/wikipedia/activitytab/ActivityTabFragment.kt
Williamrai: I implemented this only for the activity tab since the shimmer modifier can be applied to any composable, and I felt a common shimmer component wasn’t necessary. Would you like me to create a shimmer component instead?

app/src/main/java/org/wikipedia/compose/extensions/Modifier.kt
Williamrai: done.

app/src/main/java/org/wikipedia/compose/theme/WikipediaColor.kt
Williamrai: done.

app/src/main/java/org/wikipedia/compose/extensions/Modifier.kt
Williamrai: done.

app/src/main/java/org/wikipedia/compose/extensions/Modifier.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/ActivityTabFragment.kt
cooltey: Ah that also makes sense to me. 😅

PR #5907, opened 2025-09-05
Activity tab timeline design fixes

Williamrai:
### What does this do?
- design fixes, missing description/thumbnail bug fix

### DIFF
--- a/app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
+++ b/app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
@@ -3,6 +3,7 @@ package org.wikipedia.activitytab.timeline
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
```diff
@@ -43,24 +44,43 @@ import org.wikipedia.util.DateUtil
...
+) {
+    Row(
+        modifier = modifier,
         horizontalArrangement = Arrangement.spacedBy(16.dp),
         verticalAlignment = Alignment.CenterVertically
     ) {
         val icon = when (timelineItem.activitySource) {
             ActivitySource.EDIT -> R.drawable.ic_mode_edit_white_24dp
-            ActivitySource.SEARCH -> R.drawable.search_bold
+            ActivitySource.SEARCH -> R.drawable.outline_search_24
```
cooltey, line 83: I believe you can use the existing `ic_search_white` instead of importing a new one from IDE.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
```diff
@@ -138,6 +136,43 @@ fun TimelineModule(
     }
 }
 
+@Composable
+fun TimelineBottomView(
```
cooltey: The function name seems confusing. Would it be possible to name it `TimelineButtonView` since it only contains a button?

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
```diff
@@ -43,24 +44,43 @@ import org.wikipedia.util.DateUtil
...
+        if (timelineItem.activitySource == ActivitySource.EDIT) {
+            TimelineBottomView(
+                onViewChangesBtnClick = { onItemClick(timelineItem) }
+            )
+        }
+    }
+}
+
+@Composable
+fun TimelineHeader(
```
cooltey: The function name seems confusing. Would it be possible to name it `TimelineContent` or something related?

cooltey (changes requested):

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
Williamrai: sarah wanted this outline_search_24 icon instead.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activitytab/timeline/TimelineModule.kt
Williamrai: done.

PR #5920, opened 2025-09-11
Hygiene: Year-in-review code clean up

cooltey:
### What does this do?
This is an initial PR for code review of the Year in Review feature. This PR cleans up some comments and optimizes the parameters.

### DIFF
--- a/app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
+++ b/app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
@@ -37,10 +37,6 @@ class YearInReviewActivity : BaseActivity() {
         super.onCreate(savedInstanceState)
         setContent {
             BaseTheme {
-                /*
-                personalizedScreenList is temporarily populated with screens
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/yearinreview/YearInReviewActivity.kt
```diff
@@ -170,16 +164,16 @@ class YearInReviewActivity : BaseActivity() {
         }
     }
 
+    fun finish(scope: CoroutineScope, delayTimeMillis: Long = 200) {
+        scope.launch {
+            delay(delayTimeMillis)
```
dbrant, line 169: Can I ask what was the reasoning for delaying 200ms before finishing?

cooltey: I recall this is something related to the survey dialog. Maybe we can optimize the flow a bit in other PR.

cooltey:
Close this because we have #5931

PR #5925, opened 2025-09-11
Show Reading list once user start interacting with the app

Williamrai:
### What does this do?
- Using the login timestamp as the minimum time when querying the reading list table caused all synced items to appear, since the sync completed almost instantly and populated the table with timestamps later than the login timestamp.
- so instead added a preference flag that would prevent loading saved pages immediately after a fresh install and login.

### DIFF
--- a/app/src/main/java/org/wikipedia/activitytab/ActivityTabViewModel.kt
+++ b/app/src/main/java/org/wikipedia/activitytab/ActivityTabViewModel.kt
@@ -251,13 +251,10 @@ class ActivityTabViewModel() : ViewModel() {
 
     private fun createTimelineSources(): List<TimelineSource> {
         val sources = mutableListOf<TimelineSource>()
-
         sources.add(HistoryEntryPagingSource(AppDatabase.instance.historyEntryWithImageDao()))
-
-        if (Prefs.hasUserInteractedWithTheAppAfterLogIn) {
+        if (Prefs.hasUserInteractedWithPagesAfterLogIn) {
             sources.add(ReadingListPagingSource(AppDatabase.instance.readingListPageDao()))
         }
-
         val userContribPagingSource = UserContribPagingSource(wikiSiteForTimeline, AccountUtil.userName, AppDatabase.instance.historyEntryWithImageDao())
         sources.add(userContribPagingSource)
         return sources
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/settings/Prefs.kt
```diff
@@ -858,4 +858,8 @@ object Prefs {
     var isActivityTabOnboardingShown
         get() = PrefsIoUtil.getBoolean(R.string.preference_key_activity_tab_onboarding_shown, false)
         set(value) = PrefsIoUtil.setBoolean(R.string.preference_key_activity_tab_onboarding_shown, value)
+
+    var hasUserInteractedWithPagesAfterLogIn
```
dbrant, line 862: Naming: the name of this preference mentions "login", but I don't see any logic that actually relates this preference to the act of logging in or out.
If I log out, this preference doesn't get reset to `false`. (Is it supposed to?)

It also mentions "interacted with pages", but this preference seems to get set when the user adds a page to a reading list, and when PageFragment is opened (?). Is that the only way that the user can interact with pages?

app/src/main/java/org/wikipedia/page/PageFragment.kt
```diff
@@ -259,6 +260,7 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
 
         // adding this here, so that this call would always be before any donation reminder config updates
         DonationReminderHelper.maybeShowSurveyDialog(requireActivity())
+        ActivityTabViewModel.markUserInteractedAfterLogin()
```
dbrant, line 263: The `onCreateView` function usually is not an indication that the user has interacted with it. This will be very confusing to someone who looks at this code after us.

app/src/main/java/org/wikipedia/activitytab/ActivityTabViewModel.kt
```diff
@@ -313,6 +317,12 @@ class ActivityTabViewModel() : ViewModel() {
 
     companion object {
         const val CAMPAIGN_ID = "appmenu_activity"
+
+        fun markUserInteractedAfterLogin() {
```
dbrant, line 321: Naming: this function now mentions "interacted" instead of "interacted with pages", which obscures the intention of the preference.

I'm not sure someone looking at this code would have any idea what this function is for, or why it's part of ActivityTabViewModel.

Williamrai:
Closing this PR as Dmitry proposed an alternative solution.

PR #5926, opened 2025-09-11
Activity tab ui jump possible solution

Williamrai:
### What does this do?
- adds an event bus to update activity tab data without triggering a full reload on onResume.


Williamrai:
closing since this was an experiment trying out using event bus for the activity tab.

PR #5927, opened 2025-09-12
Activity Tab: show saved articles only if added manually, not from sync.

dbrant:
Our `ReadingListPage` object has a field called `atime` (Added time) that is _basically unused_ (for most of our sorting logic, we use `mtime`, modified-time). Let's use this field to indicate whether this page was added manually, or via remote sync. If it was added manually, then the `atime` will be the timestamp when it was added. But if it was added by remote sync, the `atime` will be zero.

This way, when we collect `ReadingListPage` items to show in the Activity Tab, we will only gather pages that have a nonzero `atime` which will automatically filter out pages added by remote sync.

https://phabricator.wikimedia.org/T399767

### DIFF
--- a/app/src/main/java/org/wikipedia/readinglist/ReadingListFragmentViewModel.kt
+++ b/app/src/main/java/org/wikipedia/readinglist/ReadingListFragmentViewModel.kt
@@ -120,6 +120,7 @@ class ReadingListFragmentViewModel : ViewModel() {
                             remoteId = 0
                         ).apply {
                             mtime = System.currentTimeMillis()
+                            atime = mtime
                         }
                     }
                     val recommendedList = ReadingList(


app/src/main/java/org/wikipedia/readinglist/ReadingListFragmentViewModel.kt
```diff
@@ -120,7 +120,6 @@ class ReadingListFragmentViewModel : ViewModel() {
                             remoteId = 0
                         ).apply {
                             mtime = System.currentTimeMillis()
-                            atime = mtime
                         }
                     }
```
Williamrai: Would it make sense to include this in the user activity, since bookmarking from the discover page is similar to adding it manually?

Williamrai:
This looks great! I just have a small question. Thank you for providing this solution.

app/src/main/java/org/wikipedia/readinglist/ReadingListFragmentViewModel.kt
dbrant: Yeah OK you're right, let's keep this.

PR #5928, opened 2025-09-12
Activity tab timeline scroll jump fix 

Williamrai:
### What does this do?
- Created a variable **shouldRefreshTimelineSilently**. This is set when a user clicks on a timeline item and navigates to another activity.
- When this variable is set, the loading UI is not shown while coming back, which prevents the timeline from jumping around. The loading UI will only appear if the user returns to the activity tab from somewhere else, not when navigating away from timeline items.
- Another cause of the screen jump was the shimmer animation. Its height was smaller than the actual content, so it kept shrinking and growing when loading. Added a size parameter so the shimmer height matches the content more closely, making the transition smoother.


cooltey (commented):
Found a potential issue, but not sure if it is expected:

If I read a couple of articles today, and scroll down a little bit to keep the view having the "Today"'s timeline, and then I click on the top-most timeline item and go back, the view will not remain at the same position as expected. I can share a recording if you'd like to see it.

Williamrai:
> Found a potential issue, but not sure if it is expected:
> 
> If I read a couple of articles today, and scroll down a little bit to keep the view having the "Today"'s timeline, and then I click on the top-most timeline item and go back, the view will not remain at the same position as expected. I can share a recording if you'd like to see it.

yes, could you share the recording? The fix is more of a workaround than a perfect solution. I adjusted the size of the loading shimmer to closely match the final content size, and also stopped the loading UI for the timeline when the user returns.

cooltey (approved):
Cannot find any other solutions at this moment, maybe we should think about not having the `viewModel.loadAll()` in `onResume()`. Approved for now and will let @dbrant review this too.

cooltey (approved):

PR #5931, opened 2025-09-15
Year in Review Architecture improvements

Williamrai:
### What does this do?
- Improves the Year in Review architecture by introducing a common scaffold composable that cleanly separates the onboarding screen from the YIR deck screen.
- Refactors the logic for displaying survey dialogs.

### DIFF
--- a/app/src/main/java/org/wikipedia/yearinreview/YearInReviewScreenDeck.kt
+++ b/app/src/main/java/org/wikipedia/yearinreview/YearInReviewScreenDeck.kt
@@ -1,0 +1,466 @@
+package org.wikipedia.yearinreview
+
+import android.graphics.Bitmap
+import android.graphics.BlurMaskFilter
+import androidx.compose.animation.animateColorAsState
+import androidx.compose.animation.core.animateDpAsState
+import androidx.compose.animation.core.tween
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.IntrinsicSize
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Row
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/wikipedia/yearinreview/YearInReviewScreenDeck.kt
```diff
@@ -548,38 +506,24 @@ private fun paginationSizeGradient(totalIndicators: Int, iteration: Int, pagerSt
 
 @Preview
 @Composable
-fun PreviewContent() {
+fun PreviewScreenShot() {
+    val context = LocalContext.current
```
cooltey: This can be removed.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -104,27 +112,28 @@ class YearInReviewViewModel() : ViewModel() {
...
-                editCountData.headLineText = WikipediaApp.instance.resources.getQuantityString(
-                    R.plurals.year_in_review_edit_count_headline,
-                    personalizedStatistics.editCount,
-                    personalizedStatistics.editCount
-                )
-                editCountData.bodyText = WikipediaApp.instance.resources.getQuantityString(
-                    R.plurals.year_in_review_edit_count_bodytext,
-                    personalizedStatistics.editCount,
-                    personalizedStatistics.editCount
+            val editCount = homeSiteResponse.query?.userInfo!!.editCount + wikidataResponse.query?.userInfo!!.editCount + commonsResponse.query?.userInfo!!.editCount
```
cooltey: Would it be possible to keep the existing way of adding up edit counts because they look clearer?

app/src/main/java/org/wikipedia/yearinreview/YearInReviewScreenDeck.kt
Williamrai: done.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
Williamrai: done.

PR #5932, opened 2025-09-15
Updates min sdk to 23

Williamrai:
### What does this do?
- updates min sdk to 23
- Removes code related to supporting API levels below 23

### DIFF
--- a/app/src/main/java/org/wikipedia/search/SearchFragment.kt
+++ b/app/src/main/java/org/wikipedia/search/SearchFragment.kt
@@ -4,7 +4,6 @@ import android.app.Activity.RESULT_OK
 import android.content.Intent
 import android.graphics.Color
 import android.location.Location
-import android.os.Build
 import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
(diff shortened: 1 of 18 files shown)


app/src/main/java/org/wikipedia/search/SearchFragment.kt
```diff
@@ -147,7 +146,7 @@ class SearchFragment : Fragment(), SearchResultsFragment.Callback, RecentSearche
             requireArguments().putString(ARG_QUERY, intent.getStringExtra(Intent.EXTRA_TEXT))
             requireArguments().putSerializable(Constants.INTENT_EXTRA_INVOKE_SOURCE, InvokeSource.INTENT_SHARE)
         } else if (Intent.ACTION_PROCESS_TEXT == intent.action && Constants.PLAIN_TEXT_MIME_TYPE ==
-                intent.type && Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                intent.type) {
```
cooltey: Maybe we can reduce the line and move this to line `148`.

app/src/main/java/org/wikipedia/util/ShareUtil.kt
```diff
@@ -176,11 +173,7 @@ object ShareUtil {
                     .setComponent(ComponentName(activityInfo.packageName, activityInfo.name))
             targetIntents.add(LabeledIntent(targetIntent, activityInfo.packageName, info.labelRes, info.icon))
         }
-        val chooserIntent = (if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
-            Intent.createChooser(Intent(), chooserTitle)
-        } else {
-            Intent.createChooser(targetIntents.removeAt(0), chooserTitle)
-        }) ?: return null
+        val chooserIntent = (Intent.createChooser(Intent(), chooserTitle)) ?: return null
```
cooltey: The outer `()` is not required. Would it be possible to remove it?

app/src/main/java/org/wikipedia/search/SearchFragment.kt
Williamrai: done.

app/src/main/java/org/wikipedia/util/ShareUtil.kt
Williamrai: done.

PR #5933, opened 2025-09-16
Prevent possible crash when taking screenshots for sharing.

dbrant:
This should fix a very obscure crash that happens on certain devices that use hardware bitmaps for drawing some `ImageView` components.

When we build a screen that we want to be able to screenshot for sharing purposes, all bitmaps shown on the screen must be software bitmaps, so that their contents can be drawn onto the screenshot bitmap.

This PR adds an extension function that walks through a `ViewGroup` and ensures that any `ImageView`s are using software `BitmapDrawable`s.

### DIFF
--- a/app/src/main/java/org/wikipedia/activitytab/ActivityTabCustomizationActivity.kt
+++ b/app/src/main/java/org/wikipedia/activitytab/ActivityTabCustomizationActivity.kt
@@ -26,11 +26,13 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import org.wikipedia.R
+import org.wikipedia.WikipediaApp
 import org.wikipedia.activity.BaseActivity
 import org.wikipedia.analytics.eventplatform.ActivityTabEvent
 import org.wikipedia.compose.components.WikiTopAppBar
 import org.wikipedia.compose.theme.BaseTheme
 import org.wikipedia.compose.theme.WikipediaTheme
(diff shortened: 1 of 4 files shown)


Williamrai (approved):

app/src/main/java/org/wikipedia/extensions/View.kt
```diff
@@ -21,3 +27,20 @@ fun View.setTextDirectionByLang(lang: String) {
 fun View.setLayoutDirectionByLang(lang: String) {
     layoutDirection = TextUtils.getLayoutDirectionFromLocale(Locale(lang))
 }
+
+fun View.ensureSoftwareBitmaps() {
+    if (this is ViewGroup) {
```
cooltey: Instead of using a `if` statement, can we use `when` in this case?

app/src/main/java/org/wikipedia/extensions/View.kt
```diff
@@ -21,3 +27,20 @@ fun View.setTextDirectionByLang(lang: String) {
...
+        for (i in 0 until childCount) {
+            getChildAt(i).ensureSoftwareBitmaps()
+        }
```
cooltey: How about using this:
```
children.forEach {
    it.ensureSoftwareBitmaps()
}
```

app/src/main/java/org/wikipedia/extensions/View.kt
dbrant: sure!

app/src/main/java/org/wikipedia/extensions/View.kt
dbrant: sure!

PR #5934, opened 2025-09-16
Prevent possible ANRs from serializing stream configs.

dbrant:
We get some occasional ANRs that point to the serialization of StreamConfigs. This makes sense, because the list of StreamConfigs is a **very** heavy object, and even though `kotlinx.serialization` is quite efficient, serializing an object of this size can still be very CPU-intensive, and shouldn't be done on the main thread.

Let's make sure that serialization of StreamConfigs is done in a separate thread. This should fix those ANRs, and perhaps even improve general startup performance for other devices.

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
+++ b/app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
@@ -8,6 +8,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import org.wikipedia.BuildConfig
 import org.wikipedia.WikipediaApp
 import org.wikipedia.dataclient.ServiceFactory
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
```diff
@@ -111,21 +112,23 @@ object EventPlatformClient {
...
-        STREAM_CONFIGS.putAll(Prefs.streamConfigs)
         MainScope().launch(CoroutineExceptionHandler { _, t ->
             L.e(t)
         }) {
+            STREAM_CONFIGS.clear()
+            // Ensure that serialization of configs is done off the main thread
+            withContext(Dispatchers.Default) {
+                STREAM_CONFIGS.putAll(Prefs.streamConfigs)
+            }
             refreshStreamConfigs()
```
Williamrai, line 132: do we need to load from Prefs here since we are immediately refreshing this from the API?

dbrant: Yes -- What if we're offline and the API call fails?

Williamrai: yes, that makes sense.

PR #5936, opened 2025-09-17
Activity Tab: don't show Game module if unavailable in current language.

dbrant:
(Uses very similar logic as showing/hiding the Donation module. The logic is getting a bit complex; I wonder if we could simplify it in a future PR.)

https://phabricator.wikimedia.org/T404764

### DIFF
--- a/app/src/main/java/org/wikipedia/language/AppLanguageState.kt
+++ b/app/src/main/java/org/wikipedia/language/AppLanguageState.kt
@@ -130,6 +130,7 @@ class AppLanguageState(context: Context) {
         if (_appLanguageCodes.size > 1) {
             _appLanguageCodes.removeAll(codes)
             Prefs.appLanguageCodeList = _appLanguageCodes
+            WikipediaApp.instance.resetWikiSite()
         }
     }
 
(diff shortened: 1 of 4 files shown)


cooltey (changes requested):
Found an issue related to changing primary languages vs game module visibility:

1. First, have two languages in the app: English, Chinese (Taiwan)
2. Go to the Activity tab to see if the game module is available.
3. Go to the Customize (settings) to see if the game module option is available or not.
4. Go to Settings to add Japan and make it the primary language.
5. Repeat 2 and 3
6. Go to Settings to remove Japan and do not change the language order.
7. Repease 2 and 3, and you will find the game module is not available, and the Customize screen is also not showing the game module option.

dbrant:
> Found an issue related to changing primary languages vs game module visibility:

@cooltey  😮 this actually exposes a bug in the Language adding/removing logic. When you remove the "primary" language, the top-level `WikipediaApp.instance.wikiSite` doesn't get updated (!), and remains the stale language that was removed!

This is technically not part of this PR, nevertheless I put in an easy fix.

app/src/main/java/org/wikipedia/language/AppLanguageState.kt
```diff
@@ -130,6 +130,7 @@ class AppLanguageState(context: Context) {
         if (_appLanguageCodes.size > 1) {
             _appLanguageCodes.removeAll(codes)
             Prefs.appLanguageCodeList = _appLanguageCodes
+            WikipediaApp.instance.resetWikiSite()
```
dbrant, line 133: This explicit call was missing the whole time, and was causing a subtle bug.
The `resetWikiSite()` function is inexpensive, and low-risk enough to include in this PR.

PR #5938, opened 2025-09-17
Activity Tab: tweak verbiage in Edits screen.

dbrant:
https://phabricator.wikimedia.org/T404766

### DIFF
--- a/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
+++ b/app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
@@ -123,6 +123,8 @@ class SuggestedEditsTasksFragment : Fragment(), MenuProvider {
         if (inActivityAbTestGroup) {
             notificationButtonView = NotificationButtonView(requireContext())
             requireActivity().addMenuProvider(this, viewLifecycleOwner, Lifecycle.State.RESUMED)
+            binding.layoutTasksContainer.whatIsTitleText.text = getString(R.string.edits_screen_what_is_title)
+            binding.layoutTasksContainer.whatIsBodyText.text = getString(R.string.edits_screen_what_is_body)
         }
 
         binding.layoutContributionsContainer.contributionsContainer.setOnClickListener {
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
```diff
@@ -123,20 +123,29 @@ class SuggestedEditsTasksFragment : Fragment(), MenuProvider {
...
+            }
         }
         tasksContainer.learnMoreButton.setOnClickListener {
-            if (ActivityTabABTest().isInTestGroup()) {
+            if (inActivityAbTestGroup) {
                 ActivityTabEvent.submit(activeInterface = "edit_home", action = "learn_more_click", editCount = viewModel.totalContributions)
+                UriUtil.visitInExternalBrowser(requireContext(), getString(R.string.edit_screen_learn_more_url).toUri())
+            } else {
+                FeedbackUtil.showAndroidAppEditingFAQ(requireContext())
             }
```
Williamrai: Since both click listeners uses same logic, maybe we could move this into a shared function?

app/src/main/res/layout/suggested_edits_tasks_container.xml
```diff
@@ -53,6 +54,7 @@
                 android:textStyle="bold" />
 
             <TextView
+                android:id="@+id/whatIsBodyText"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
                 android:layout_marginTop="8dp"
```
Williamrai, line 60: Longer text is showing extra spacing on the right side, making it look uneven which is the default TextView behavior. I found that setting the **hyphenationFrequency** parameter in **TextView** fixes this but this is only available after api 23. Since we are updating the minSdk to 23 maybe we could update this later, maybe leave a TODO here, what do you think?

[without hyphenation](https://github.com/user-attachments/assets/fe07a301-f3ba-4a77-9d20-6b427774852a)
[with hyphenation](https://github.com/user-attachments/assets/6b66e594-cfe9-47ad-9696-6488fbb3fff6)

app/src/main/java/org/wikipedia/suggestededits/SuggestedEditsTasksFragment.kt
dbrant: sure!

app/src/main/res/layout/suggested_edits_tasks_container.xml
dbrant: This is a good idea, but let's do it after this release. In fact, instead of a TODO, would you mind creating a task for this? There are probably numerous other TextViews where this would be beneficial, and also we would need to decide which hyphenation value to use (normal / normalFast / etc?), and we could even think about putting `hyphenationFrequency` into a `style` that is used by all TextViews.  (And how does hyphenation work in Compose?)

Williamrai: sure.

PR #5946, opened 2025-09-19
YiR: Pre-design scaffolding

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T404900

### DIFF
--- a/app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
+++ b/app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
@@ -7,12 +7,9 @@ import kotlinx.coroutines.async
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.launch
-import org.wikipedia.Constants
 import org.wikipedia.R
 import org.wikipedia.WikipediaApp
-import org.wikipedia.auth.AccountUtil
 import org.wikipedia.database.AppDatabase
-import org.wikipedia.dataclient.ServiceFactory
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -79,80 +118,388 @@ class YearInReviewViewModel() : ViewModel() {
...
-                _uiScreenListState.value = UiState.Success(
-                    data = listOf(readCountJob.await(), nonEnglishCollectiveEditCountData, nonEnglishCollectiveReadCountData, nonEnglishCollectiveEditCountData)
-                )
+                nonEnglishCollectiveReadCountData
             }
         }
     }
 
+    suspend fun nonLoggedInEnglishGeneralSlides(): List<YearInReviewScreenData.StandardScreen> {
+        return withContext(Dispatchers.IO) {
```
dbrant: So wait, can we clarify why all of these functions need to be `suspend`? (and why they need to be inside the `IO` dispatcher?) (and why they need to live in the ViewModel?)

cooltey: That's my initial idea of where we may put API/database requests in it, but I can make them as regular functions and in a separate class.

dbrant: Will there really be that many API/db requests? By the time the entire YiR is ready to be composed, all the data should already be available in the ViewModel.

cooltey: You are right and updated.

PR #5972, opened 2025-10-02
YiR: Create Articles Saved slide

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T405308


app/src/main/java/org/wikipedia/readinglist/db/ReadingListPageDao.kt
```diff
@@ -98,8 +98,8 @@ interface ReadingListPageDao {
     @Query("SELECT * FROM ReadingListPage WHERE atime > 0 ORDER BY atime DESC LIMIT :limit OFFSET :offset")
     suspend fun getPagesByLocallySavedTime(limit: Int, offset: Int): List<ReadingListPage>
 
-    @Query("SELECT DISTINCT displayTitle FROM ReadingListPage ORDER BY atime DESC LIMIT :limit")
-    suspend fun getLatestArticleTitles(limit: Int): List<String>
+    @Query("SELECT DISTINCT displayTitle FROM ReadingListPage ORDER BY atime DESC")
+    suspend fun getAllDistinctArticleTitles(): List<String>
```
dbrant: It looks like this will now get _all_ reading list pages, and will store them in the `YearInReviewModel` class, which will be serialized (!) to Prefs.
We know that some users have thousands of saved pages, so this will become extremely inefficient for them.

It seems the only purpose of this list is to get the `size()` and a few random pages. Can we change this logic to get just those things, instead of everything?

cooltey: That makes sense to me, and done!

PR #5977, opened 2025-10-02
Reduce overriding of Material Dialog styles.

dbrant:
This reduces the unnecessary overriding of styles that we do for Material Dialogs, and will make these dialogs look more like the standard Material 3 dialog.

Before / after:

image image


app/src/main/res/values/styles.xml
```diff
@@ -146,11 +146,10 @@
     </style>
 
     <style name="H2.DialogTitle" parent="MaterialAlertDialog.Material3.Title.Text">
-        <item name="android:fontFamily">sans-serif</item>
-        <item name="android:textStyle">bold</item>
-        <item name="android:textSize">20sp</item>
-        <item name="lineHeight">28sp</item>
         <item name="android:textColor">?attr/primary_color</item>
+        <item name="android:singleLine">false</item>
+        <item name="android:maxLines">5</item>
```
dbrant, line 151: These new bits are actually necessary, because the default MaterialAlertDialog makes the title be `singleLine`, and even if `maxLines` is less than ~5, it will dynamically adjust the size of the text, which we may want in the future, but not now.

PR #5978, opened 2025-10-03
YiR: Add Setting to enable/disable.

dbrant:
https://phabricator.wikimedia.org/T405493


cooltey (commented):
Looks good to me. Do we want to remove the toggle of enabling/disabling YiR in the developer settings too?

dbrant:
> Looks good to me. Do we want to remove the toggle of enabling/disabling YiR in the developer settings too?

Sure!

PR #5982, opened 2025-10-06
YiR: Create Reading Patterns slide

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T405426

### DIFF
--- a/app/src/main/res/values-qq/strings.xml
+++ b/app/src/main/res/values-qq/strings.xml
@@ -1900,8 +1900,8 @@
   <string name="year_in_review_slide_top_categories_body">Body text of the top categories slide for the Year in Review. The %1$d will be replaced by the current year and %2$s will be replaced by the list of top categories.</string>
   <string name="year_in_review_slide_popular_english_articles_headline">Title of the most popular English Wikipedia articles for the Year in Review.</string>
   <plurals name="year_in_review_slide_popular_english_articles_body">
-    <item quantity="one">Title of the most popular English Wikipedia articles for the Year in Review. The %1$d will be replaced by the number of articles and %2$s will be replaced by the list of the most popular English articles.</item>
-    <item quantity="other">Title of the most popular English Wikipedia articles for the Year in Review. The %1$d will be replaced by the number of articles and %2$s will be replaced by the list of the most popular English articles.</item>
+    <item quantity="one">Body text of the most popular English Wikipedia articles for the Year in Review. The %1$d will be replaced by the number of articles and %2$s will be replaced by the list of the most popular English articles.</item>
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/wikipedia/yearinreview/YearInReviewSlides.kt
```diff
@@ -81,13 +84,27 @@ class YearInReviewSlides(
...
-    private fun readingPatternsScreen(vararg params: Int): YearInReviewScreenData.StandardScreen {
-        // TODO: yir106 + yir107
-        return YearInReviewScreenData.StandardScreen(
-            animatedImageResource = R.drawable.year_in_review_puzzle_pieces,
-            staticImageResource = R.drawable.year_in_review_puzzle_pieces,
-            headlineText = "You have clear reading patterns",
-            bodyText = "TBD"
+    private fun readingPatternsScreen(): YearInReviewScreenData.StandardScreen {
+        // TODO: check if the time needs to follow the locale
+        val favoriteTimeText = when (yearInReviewModel.favoriteTimeToRead) {
```
cooltey, line 92: Pending. Please share other ideas on how to set up the hour based on the timezone properly.

app/src/main/java/org/wikipedia/yearinreview/YearInReviewViewModel.kt
```diff
@@ -172,6 +172,29 @@ class YearInReviewViewModel() : ViewModel() {
...
+
+                val mostReadingMonth = async {
+                    AppDatabase.instance.historyEntryDao()
+                        .getMostReadingMonthSince(startTimeInMillis, endTimeInMillis)
+                }
+
+                val favoriteTimeToReadHour = favoriteTimeToRead.await() ?: 0
+
+                val favoriteDayToReadIndex = favoriteDayToRead.await()?.let {
+                    it % 7
```
dbrant: just curious: will this ever be greater than 7?  i.e. why do we need `% 7` here, but not `% 24` for hours or `% 12` for month?

cooltey: It will return `0` to `6`, which indicates `Sunday` to `Monday`

I have replaced it with an `if` statement because `DayOfWeek` is `Monday` to `Sunday`, which counts from `1` to `7`

dbrant: I see - we're converting between two different conventions (sqlite does it differently from java.time (because why would anything be consistent?)) 👍

app/src/main/java/org/wikipedia/yearinreview/YearInReviewSlides.kt
dbrant: will do in a followup.

PR #6149, opened 2025-11-25
Bug fix: Possible solution for accurate title text on Donation Reminder 

Williamrai:
### What does this do?
- adds a new variable **continuousArticleVisit** for tracking continuous article visit per milestone 
- extract logic from **shouldShowReminderNow()** to a new function **recordReminderShown()** which will be called after UI has updated 
- update logic to show accurate titleText


Williamrai:
Closing this since we haven’t seen any similar issues occur.

PR #6244, opened 2026-01-27
Combine date age methods

Isira-Seneviratne:
### What does this do?
* Combines the age-based date retrieval methods in DateUtil into a couple of common methods which accept a ZoneId.
* Removes the UtcDate class.

### Why is this needed?
* Combining the methods allows for reduced code duplication.
* The UtcDate class can be replaced with LocalDate.

**Phabricator:**


dbrant:
Cleaning up old/no-longer-applicable PRs.

PR #6348, opened 2026-02-26
Fix: notification links with ampersands in page titles work correctly

hpr:
### What does this do?
Fixes notification links where page titles have ampersands. These shouldn't be decoded because it breaks parsing by the notification link handler.

Example of bug in Wikipedia Beta:
EmptyListError

Example of fixed behavior:
T418432

### Why is this needed?
To prevent "Empty list doesn't contain element at index 0." bugs on notifications

**Phabricator:**
https://phabricator.wikimedia.org/T418432


cooltey (changes requested):
Thank you for submitting the PR to fix it! Removing the `decodeURL()` here may solve the issue in English articles that contain `&` in the title, but for the article in other wikis, the URL will be encoded, and if we remove this here, the links will become unavailable and cause more issues.

hpr:
@cooltey, thanks, I tried another approach by storing the encoded URL separately. Compiled this and observed that the fix still works with the above article, if there's another test case I should try let me know.

app/src/main/java/org/wikipedia/notifications/db/Notification.kt
```diff
@@ -84,12 +84,18 @@ class Notification(var id: Long = 0,
...
         val tooltip: String = ""
         // The icon could be a string or `false`.
         private val icon: JsonElement? = null
 
         init {
+            encodedUrl = url
+            if (encodedUrl.startsWith("//")) {
+                encodedUrl = WikiSite.DEFAULT_SCHEME + "://" + encodedUrl.substring(2)
+            }
             url = UriUtil.decodeURL(url)
```
dbrant: Thanks; This is definitely the right idea -- the URL is getting decoded and/or encoded at inappropriate times.

However, I'm actually not sure the `url` field even needs to be decoded at all, in the first place.
In fact, I think this entire issue could be resolved by removing this `UriUtil.decodeURL(url)` call on line 99, with no further need for a separate `encodedUrl` field.
Unfortunately there are no comments on why this was done originally, but AFACT no other functionality will be impacted by removing this.

dbrant:
Thanks! Let's standby on this -- I actually think your original fix is the right idea; we should remove instances of decoding/encoding URLs in places where it's not clear why it's done. And we should definitely not keep **both** encoded and decoded url fields in a model class like `Notification`, again without a super obvious reason.

I have a feeling that the proper fix involves just a little more digging into where all these URLs eventually end up, and performing the URL decoding in that _single_ place, instead of multiple places that shouldn't need to be "aware" of url encoding.

mohammadmanzu1-creator (commented):
Moh

PR #6367, opened 2026-03-04
[Feature branch] Reading challenge widget

Williamrai:
### What does this do?
- adds reading challenge widget 
- this branch also has Developer setting called **"Reading Challenge Widget Configurations"** for updating widget states easily
 
**Phabricator:**
https://phabricator.wikimedia.org/T418634


dbrant (changes requested):
I'm a bit confused why there are two separate widgets ("small" and "large" variants)... Aren't widgets supposed to be dynamically resizable by the user? Couldn't we automatically serve the appropriate "small" vs "large" size based on the requested dimensions of the widget?

Williamrai:
> I'm a bit confused why there are two separate widgets ("small" and "large" variants)... Aren't widgets supposed to be dynamically resizable by the user? Couldn't we automatically serve the appropriate "small" vs "large" size based on the requested dimensions of the widget?

> Initially I confirmed with the PM and the designer about resizing widgets, and they said we only need to support two variants: small and large, no resizing. Maybe my question about resizing was misunderstood, but based on that I went ahead with two separate widgets. Since we’ve now clarified, we can also implement resizing.

app/src/main/java/org/wikipedia/widgets/readingchallenge/largewidget/ReadingChallengeLargeWidget.kt
cooltey, line 1: Would it be possible to name this `ReadingChallengeLargeWidgetContent` and put the layout functions in this file? Maybe we can just move this class out of its sub-folder.

app/src/main/java/org/wikipedia/widgets/readingchallenge/smallwidget/ReadingChallengeSmallWidget.kt
cooltey, line 1: Same above. Would it be possible to name this `ReadingChallengeSmallWidgetContent` and put the layout functions in this file? Maybe we can just move this class out of its sub-folder.

app/src/main/java/org/wikipedia/widgets/readingchallenge/largewidget/ReadingChallengeLargeWidget.kt
Williamrai: done. thank you.

app/src/main/java/org/wikipedia/widgets/readingchallenge/smallwidget/ReadingChallengeSmallWidget.kt
Williamrai: done. thank you.

cooltey (approved):

dbrant (approved):

PR #6440, opened 2026-03-24
Handle notification permission in BaseActivity

rajvirsingh2:
### What does this do?
Completes the TODO in BaseActivity to show user-facing messages when notification permission is denied. Adds openAppNotificationSettings() to DeviceUtil to support navigating the user directly to the app's notification settings.

1. Shows a snackbar with a rationale message and a Settings action when the user soft denies.
2. Shows a snackbar directing the user to app settings when they select "Don't ask again".

### Why is this needed?
Previously, when a user denied the notification permission, nothing happened and the denial was silently ignored. This change improves the user experience by giving clear, actionable feedback in both denial scenarios.

**Phabricator:**

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/ForYouCard.kt
+++ b/app/src/main/java/org/wikipedia/feed/ForYouCard.kt
@@ -72,6 +72,7 @@ fun ForYouCardContent(
     card: ForYouCard? = null,
     footerIcon: Painter? = null,
     footerText: String? = null,
+    footerContent: @Composable () -> Unit = {},
     onPageClick: (PageTitle) -> Unit = {},
     onShareClick: (PageTitle) -> Unit = {},
     onSaveClick: (PageTitle) -> Unit = {},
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -78,8 +80,25 @@ abstract class BaseActivity : AppCompatActivity(), ConnectionStateMonitor.Callba
         }
     }
 
+    @RequiresApi(Build.VERSION_CODES.O)
```
cooltey: Instead of having an annotation here for a specific SDK version, could you please have the logic in the `if(!granted){` block like this:
```
if ( if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
...
} else {
...
```

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -193,6 +212,7 @@ abstract class BaseActivity : AppCompatActivity(), ConnectionStateMonitor.Callba
         BreadCrumbLogEvent.logScreenShown(this)
     }
 
+    @RequiresApi(Build.VERSION_CODES.O)
```
cooltey: Same here.

cooltey (changes requested):

PR #6446, opened 2026-03-25
[Feature branch] Explore Feed Redesign

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T407990


app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
```diff
@@ -0,0 +1,568 @@
...
+            val modules = forYouCollectionSaved.modulesPerLanguage[wikiSite.value.languageCode].orEmpty()
+            val newModules = mutableListOf<ForYouModule>()
+            modules.forEach { module ->
+                val filteredCards = module.cards.filterNot { hiddenCards.contains(it.hideKey) }
+                if (filteredCards.isNotEmpty()) {
+                    newModules.add(module.withCards(filteredCards))
+                }
+            }
+            return newModules
+        }
```
cooltey, line 463: This will prevent the feed from showing `Because you read` and `Continue reading` that happened on the same day.

PR #6486, opened 2026-04-07
Games hub use correct action name

Williamrai:
### What does this do?
- replace play_click to archive_click


app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -454,7 +454,7 @@ class MainFragment : Fragment(), BackPressedHandler, MenuProvider, FeedFragment.
     }
 
     override fun onWikiGamesCardFooterClicked() {
-        WikiGamesEvent.submit(action = "more_click", "games_feed")
+        WikiGamesEvent.submit(action = "more_click", "game_feed")
```
cooltey, line 457: "update game_feed to games_feed"?

Williamrai: the correct name is "**game_feed**", I forgot to update the description.

PR #6516, opened 2026-04-21
[Instrumentation] Reading Challenge Widget

Williamrai:
### What does this do?
- adds instrumentation for reading widget challenge

**Phabricator:**
https://phabricator.wikimedia.org/T419306


app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
```diff
@@ -135,6 +137,7 @@ object EventPlatformClient {
                 TestKitchenAdapter.client.updateSourceConfig(STREAM_CONFIGS)
             }
             refreshStreamConfigs()
+            streamConfigsDeferred.complete(Unit)
         }
```
Williamrai, line 138: I added this here because there was a race condition between setting this up and submitting event in ReadingChallengeOnboardingActivity which was always logging this `No stream config exists for this stream, the event is ignored and dropped.` and events were being ignored. And in onboarding activity it would wait before sending event.

dbrant: ?? If this fixes such a fundamental bug in EventPlatformClient, then it should be a separate PR, with a separate explanation. Let me dig into the analytics/TestKitchenClient code a bit and make a fully robust fix, so that this PR won't need its own fixes.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeOnboardingActivity.kt
```diff
@@ -75,6 +76,12 @@ class ReadingChallengeOnboardingActivity : BaseActivity() {
         super.onCreate(savedInstanceState)
         DeviceUtil.setEdgeToEdge(this)
         Prefs.readingChallengeOnboardingShown = true
+
+        lifecycleScope.launch {
+            EventPlatformClient.streamConfigsDeferred.await()
```
dbrant: No particular Activity should have any knowledge of the internals of `EventPlatformClient`.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeWidgetRepository.kt
```diff
@@ -65,6 +55,7 @@ class ReadingChallengeWidgetRepository(private val context: Context) {
     }
 
     fun resolveState(userData: ReadingChallengeUserData): ReadingChallengeState {
+        println("orange resolveState called")
```
dbrant: Nope

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
```diff
@@ -0,0 +1,54 @@
+package org.wikipedia.widgets.readingchallenge
+
+import org.wikimedia.testkitchen.instrument.InstrumentImpl
+import org.wikipedia.analytics.testkitchen.TestKitchenAdapter
+import org.wikipedia.settings.Prefs
+
+object ReadingChallengeAnalyticsHelper {
+    val instrument: InstrumentImpl by lazy {
+        TestKitchenAdapter.client.getInstrument("apps-widgetchallenge")
+            .startFunnel("widget_challenge")
```
dbrant: Let's not keep a long-lived instance of an Instrument for the lifetime of the whole app. Among other things, this is misusing the idea of a "funnel". This funnel will "start" when this object is initialized, and then every event sent through this instrument will have `time_spent_ms` values that are nonsensical. A funnel should start when a UI element is shown (an activity or dialog).

This is why it doesn't make sense for `heartbeat`-type events to have a funnel (see my comments in the instrumentation deck).

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    fun sendAnalytics(state: ReadingChallengeState) {
```
dbrant: Can this have a more precise name, such as `sendHeartbeatEvent()`?

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
```diff
@@ -0,0 +1,54 @@
...
+            is ReadingChallengeState.StreakOngoingReadToday -> {
+                instrument.submitInteraction(
+                    action = "heartbeat",
+                    actionSource = "widget_challenge",
+                    actionContext = mapOf("streak_count" to Prefs.readingChallengeStreak)
+                )
+            }
+            ReadingChallengeState.NotEnrolled,
+            ReadingChallengeState.NotLiveYet,
+            ReadingChallengeState.EnrolledNotStarted -> {
```
dbrant: This function feels kind of complicated, but yet also doesn't seem to send enough information.

For the above three states (NotEnrolled, NotLiveYet, EnrolledNotStarted), we seem to be sending the exact same event? Don't we want to differentiate between those states?

Couldn't we just always send a `heartbeat` event, with a `streak_count` in all cases (even if the streak_count is 0, because why not?), and a different `elementId` for each possible ReadingChallengeState? Doesn't that cover all possible states, in a more simple way?

dbrant (changes requested):

app/src/main/java/org/wikipedia/analytics/eventplatform/EventPlatformClient.kt
Williamrai: removed. this was resolved in https://github.com/wikimedia/apps-android-wikipedia/pull/6528

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeOnboardingActivity.kt
Williamrai: removed.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeWidgetRepository.kt
Williamrai: removed.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
Williamrai: done.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
Williamrai: I have confirmed this with Shay and she said its overkill to track all the names of the widgets and only wants to track the ongoing and conclusion state with streak. She also said for these three state streak count can be omitted as it's 0.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
```diff
@@ -0,0 +1,54 @@
...
+            }
+            else -> {}
+        }
+    }
+
+    fun logAppOpenFromWidget() {
+        TestKitchenAdapter.client.getInstrument("apps-open")
+            .submitInteraction(
+                action = "app_open",
+                actionSource = "widget"
```
dbrant: When sending the `app_open` event, don't we want to specify _which_ widget it's coming from? I'm guessing we'll want to send `app_open` events when opening from other widgets in the future.
At the very least, let's add something like `actionSubtype = "reading_challenge"`

Williamrai: added `actionSubtype = "reading_challenge"`

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
Williamrai: removed the instrument property from the object and instead created a function. Onboarding activity, install dialog and reward dialog was calling instrument from this object. Now they instantiate their own instrument which would start funnel when the user actually sees them.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
```diff
@@ -0,0 +1,53 @@
...
+            else -> {}
+        }
+    }
+
+    fun logAppOpenFromWidget() {
+        TestKitchenAdapter.client.getInstrument("apps-open")
+            .submitInteraction(
+                action = "app_open",
+                actionSource = "widget",
+                actionSubtype = "reading_challenge"
```
cooltey, line 26: I cannot find this `reading_challenge` for the subtype. Can you link the reference here?

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeInstallWidgetDialog.kt
```diff
@@ -55,9 +56,12 @@ import org.wikipedia.theme.Theme
 
 class ReadingChallengeInstallWidgetDialog : ExtendedBottomSheetDialogFragment(startExpanded = true) {
 
+    private val instrument = TestKitchenAdapter.client.getInstrument("apps-widgetchallenge")
+        .startFunnel("widget_challenge")
```
cooltey, line 61: I believe we can set a `.setDefaultActionSource("widget_challenge_install")` here

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeOnboardingActivity.kt
```diff
@@ -75,6 +76,12 @@ class ReadingChallengeOnboardingActivity : BaseActivity() {
         super.onCreate(savedInstanceState)
         DeviceUtil.setEdgeToEdge(this)
         Prefs.readingChallengeOnboardingShown = true
+
+        _instrument = TestKitchenAdapter.client.getInstrument("apps-widgetchallenge")
+            .startFunnel("widget_challenge")
```
cooltey, line 69: We can set a default source here for `widget_challenge_announce`

cooltey (changes requested):

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
Williamrai: https://github.com/wikimedia/apps-android-wikipedia/pull/6516#discussion_r3138827109 this was suggested by dmitry.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeOnboardingActivity.kt
Williamrai: done.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeInstallWidgetDialog.kt
Williamrai: done.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
cooltey: Got it!

cooltey (approved):
LGTM. Waiting for @dbrant to merge.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
Williamrai: removed **funnel_name** from heartbeat event.

app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
Williamrai: adds **stateName** as **elementId** for tracking different widget state.

PR #6526, opened 2026-04-23
[Reading Challenge Widget] update string parameter

Williamrai:
### What does this do?
- adds parameter to string **reading_challenge_onboarding_read_description** so that translator does not have to update the date as well
- removes already translated string which contained dates as well


dbrant:
I'm not sure "removing" the translated string from the other languages will work this way. I may be wrong, but I think removing a string only works if it's removed from _all_ languages at once.

When the next TranslateWiki sync happens, the existing translations will be pulled in again, and will all fail the TranslationTests because they don't contain the correct parameter.

I think the safer approach is just to create a completely new string, and let it be re-translated. (and remove this string at a later time)

PR #6527, opened 2026-04-23
NewExploreFeed: Overflow menus

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T418877


app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
```diff
@@ -17,17 +17,18 @@ import org.wikipedia.feed.personalization.homepreference.HomePreferenceType
...
-    val featuredArticle: PageSummary? = null,
-    val news: List<NewsItem> = emptyList(),
-    val topRead: TopRead? = null,
-    val featuredImage: FeaturedImage? = null,
-    val onThisDay: List<OnThisDay.Event> = emptyList()
+    val featuredArticle: Pair<String, PageSummary?>,
+    val news: Pair<String, List<NewsItem>?>,
+    val topRead: Pair<String, TopRead?>,
+    val featuredImage: Pair<String, FeaturedImage?>,
+    val onThisDay: Pair<String, List<OnThisDay.Event>?>
```
dbrant: ^ This is feeling a bit too complicated; It's not obvious what the `String` portion is intended to do, and it's seeming like this is trying to pass the viewing/hiding logic to the View (composable) instead of keeping it in the ViewModel.

The rest of the logic also seems to confuse the idea of "hiding a module" and "hiding this card".
"Hide this card" is supposed to hide just the _single_ card, and "Hide this module" should hide _all cards_ of this particular type.

For now, could we remove this new logic for hiding modules, and just make a stub "Hide this card" and "Hide this module" menu items with a TODO?

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
cooltey: The idea was to add "key" string for each module, and use that key plus the age so we can filter them out in the `loadCommunityContent` function in the `viewModel`.

And yes, I can revert the last commit.

PR #6528, opened 2026-04-24
TestKitchen: handle case of streamConfigs being empty, not null.

dbrant:
This fixes a potential race condition when Stream Configurations are first initialized when instantiating the TestKitchenClient. When they are first initialized the list of StreamConfigs on the EventPlatformClient side is **empty** (but not null). After a short period, the configs are populated either from Prefs or from the network.

On the Test Kitchen side, we need to improve the check whether the stream configs are null **or empty**, otherwise events submitted during that brief window might be discarded.

### DIFF
--- a/analytics/testkitchen/src/main/java/org/wikimedia/testkitchen/config/SourceConfig.kt
+++ b/analytics/testkitchen/src/main/java/org/wikimedia/testkitchen/config/SourceConfig.kt
@@ -1,7 +1,3 @@
 package org.wikimedia.testkitchen.config
 
-class SourceConfig(val streamConfigs: Map<String, StreamConfig>) {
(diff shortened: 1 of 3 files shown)


analytics/testkitchen/src/main/java/org/wikimedia/testkitchen/config/SourceConfig.kt
```diff
@@ -1,7 +1,3 @@
 package org.wikimedia.testkitchen.config
 
-class SourceConfig(val streamConfigs: Map<String, StreamConfig>) {
-    fun getStreamConfigByName(streamName: String): StreamConfig? {
```
dbrant, line 4: (this function is entirely unnecessary -- callers can just use the `streamConfigs[streamName]` syntax themselves.)

PR #6531, opened 2026-04-26
Using higher contrast text background color when in black or dark mode.

uwemartin-lgtm:
### What does this do?
When visualising comparison, the added or removed text has a green/red background color.
The original code uses a relatively low alpha setting in the color coding which is suitable for visualisation in light mode with white background.
For dark or black mode, the modified code will increase the alpha setting in the color coding.

See below results: black and dark mode have now higher contrast while light mode color setting is not touched.
Screenshot_1777212082_dark
Screenshot_1777212225_black
Screenshot_1777212354_light
  

### Why is this needed?
User reported low visibility with current color setting in black or dark mode.

**Phabricator:**
https://phabricator.wikimedia.org/T398618


cooltey:
Thank you for submitting the PR, and our designer has approved the change!

PR #6534, opened 2026-04-27
[Reading Challenge] Fix: Reward dialog not opening

Williamrai:
### What does this do?
- Fixes reading challenge reward dialog not showing from widget when app is in background

**Phabricator:**
https://phabricator.wikimedia.org/T418641


cooltey:
The fix looks good to me, but I noticed these in the [task description](https://phabricator.wikimedia.org/T418641):
> Display prize information on top of Activity Tab

and 
> User must be logged-in to view prize information

PR #6535, opened 2026-04-27
TestKitchen: Flush events periodically, not just when closing activity.

dbrant:
At the moment, Test Kitchen events are sent (flushed) only when pausing an Activity (`onPause`).
If we start sending events from inside Widget logic (independent of activities), they might get lost if the system terminates the app process before any activities are created.

Let's make an explicit timer Runnable (just like we currently have in EventPlatformClient) that flushes events every 30 seconds. Having this kind of timer is not _the best_ pattern in the Android ecosystem, but it's extremely lightweight, and will work for our purposes.

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/testkitchen/TestKitchenAdapter.kt
+++ b/app/src/main/java/org/wikipedia/analytics/testkitchen/TestKitchenAdapter.kt
@@ -25,7 +25,13 @@ import org.wikipedia.util.ReleaseUtil
 import org.wikipedia.util.log.L
 
 object TestKitchenAdapter : ClientDataCallback, EventSender {
+    const val FLUSH_DELAY_MILLIS = 30000L
     val logger = LogAdapterImpl()
+
+    val flushEventsRunnable: Runnable = {
+        client.flushEventQueue()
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/analytics/testkitchen/TestKitchenAdapter.kt
```diff
@@ -25,15 +25,25 @@ import org.wikipedia.util.ReleaseUtil
...
+
     val client = TestKitchenClient(
         eventSender = this,
         sourceConfigInit = null,
         clientDataCallback = this,
         logger = logger
     )
 
+    init {
+        flushEventsRunnable.run()
```
cooltey, line 44: Will this keep flushing events when the app is in the background? I tried to add a log to the `flushEventsRunnable` and I can see events being sent out continuously.

dbrant: Yes, that's the idea. As long as there are events to be flushed, it will try to flush them.

cooltey: Got it. Thanks.

PR #6538, opened 2026-04-28
Allow pinch to zoom when reading articles

folecr:
* Enable zooming in WebView configuration
 * Replace the meta viewport content tag with what is served to mobile web
 * ...  by https://github.com/wikimedia/mediawiki/blob/master/includes/Skin/Skin.php#L386
 * ... `content="width=device-width, initial-scale=1.0, user-scalable=yes, minimum-scale=0.25, maximum-scale=5.0"`

Bug : T424079

### What does this do?

Allow pinch to zoom in the article webview in Wikipedia Android

I frequently use pinch to zoom on the web (both desktop and mobile) to zoom into article text or images without relayout of the entire article webpage but the mobile app does not allow usage of this gesture. This PR allows using pinch-to-zoom in the mobile app exclusively in the article webview.

Note : corresponding iOS PR https://github.com/wikimedia/wikipedia-ios/pull/5821
Note : On Android the webview cannot be configured to ignore viewport scale limits so this change requires modifying HTML content with `user-scalable=yes` before loading in webview.
A broader, perhaps more correct fix, is to set user-scalable=yes in the content being served. It looks like user-scalable in the HTML content being served is being set by https://github.com/wikimedia/mediawiki-services-mobileapps/blob/master/lib/transformations/pcs/head.js#L71 introduced in change [b3bed0a05861401f83d84e93d436d08bb77da1c4](https://github.com/wikimedia/mediawiki-services-mobileapps/commit/b3bed0a05861401f83d84e93d436d08bb77da1c4). If there is interest in doing so, I could potentially make a change in that repo if there is consensus that it is the right spot, instead of this repo.

### Why is this needed?

When browsing wikipedia on mobile web and desktop web browsers, users may zoom into content with the pinch to zoom gesture. However this is not currently possible in app. This PR enables the same functionality in the wikipedia mobile app, for articles.

Note W3C recommendation https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/viewport#user-scalable

**Phabricator:**
https://phabricator.wikimedia.org/T424079

### DIFF
--- a/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
+++ b/app/src/main/java/org/wikipedia/bridge/JavaScriptActionHandler.kt
@@ -204,6 +204,12 @@ object JavaScriptActionHandler {
                 "})();"
     }
 
+    fun enableUserScaling() =
+        "(function(){" +
+        "var m=document.querySelector('meta[name=\"viewport\"]');" +
+        "if(m){m.setAttribute('content','width=device-width, initial-scale=1.0, user-scalable=yes, minimum-scale=0.25, maximum-scale=5.0');}" +
+        "})();"
+
     @Serializable
     class ImageHitInfo(val left: Float = 0f, val top: Float = 0f, val width: Float = 0f, val height: Float = 0f,
                        val src: String = "")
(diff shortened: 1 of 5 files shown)


folecr:
Gate change behind a developer settings flag as in https://github.com/wikimedia/wikipedia-ios/pull/5821

PR #6542, opened 2026-04-28
NewExploreFeed: Tabs icon

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T424351


app/src/main/java/org/wikipedia/compose/components/TabsBox.kt
Williamrai, line 1: Just curious will the count ever exceed 99?

app/src/main/java/org/wikipedia/compose/components/TabsBox.kt
```diff
@@ -0,0 +1,89 @@
...
+        modifier = modifier
+            .background(color = backgroundColor, shape = shape)
+            .border(width = 1.5.dp, color = borderColor, shape = shape),
+        contentAlignment = Alignment.Center
+    ) {
+        Text(
+            modifier = Modifier.padding(start = 3.dp, end = 4.dp),
+            text = count.toString(),
+            autoSize = TextAutoSize.StepBased(minFontSize = 3.sp, maxFontSize = 10.sp, stepSize = 1.sp),
+            fontFamily = FontFamily.SansSerif,
```
Williamrai, line 46: For users with larger font scale settings, a minimum size of 3.sp may render the text too small to read comfortably and could impact usability. It might be better to just set regular fontSize to 10.sp or adjust the minSize to ensure better readability. What do you think?
`autoSize = TextAutoSize.StepBased(minFontSize = 3.sp, maxFontSize = 10.sp, stepSize = 1.sp)`

Williamrai, line 46: For users with larger font scale settings, a minimum size of 3.sp may be too small to read comfortably and could affect usability. It might be better to use a fixed fontSize like 10.sp or increase the minSize to improve readability.
Also, since this calculation runs on every resume of the home fragment, it could have some performance overhead. What do you think?
`autoSize = TextAutoSize.StepBased(minFontSize = 3.sp, maxFontSize = 10.sp, stepSize = 1.sp)`

app/src/main/java/org/wikipedia/compose/components/TabsBox.kt
cooltey: It is the existing design that we only allow at most `99` tabs in the app.

app/src/main/java/org/wikipedia/compose/components/TabsBox.kt
cooltey: I updated the `minFontSize` to `5.sp`, I think it's better to have a smaller font size instead of not showing anything in the layout if the user has a larger device font size.

app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
```diff
@@ -68,6 +69,8 @@ class HomeViewModel : ViewModel() {
     // Batch counter for "For you" recommendations.
     private var forYouBatchIndex = 0
 
+    val tabsState = MutableStateFlow(WikipediaApp.instance.tabCount to false)
+
     private val communityHandler = CoroutineExceptionHandler { _, throwable ->
```
Williamrai, line 111: tabsState.first and tabsState.second at the call sites don't communicate what they mean. What do you think of using a small data class instead? 
e.g. 
`data class TabsState(val tabCount: Int, val showPulse: Boolean) `

This would make the call sites self-documenting and easier to extend if we add more tab-related state later.

app/src/main/java/org/wikipedia/compose/components/TabsBox.kt
Williamrai: Got it. thank you.

app/src/main/java/org/wikipedia/compose/components/TabsBox.kt
Williamrai: okay thank you.

app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
cooltey: Sure, that makes sense to me. Done!

PR #6545, opened 2026-04-29
[ReadingChallenge] Fix: tap on join challenge should always show the prompt

cooltey:
### What does this do?
When the user clicks on "Join Challenge", the app opens the `MainActivity` but does not clear the existing activity. This PR adds logic to check the extra to see if the action is from the widget in order to show the prompt.

Also, adds a `.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP))` to the `requestReadingChallengeActivity` so we won't show two prompts when the user pauses on the `PageActivity` and clicks on the "Join Challenge".

This PR also makes the views in the `LazyColumn` clickable to avoid missing click behaviors in the container.

#### Steps to reproduce
1. Modify start date to `04-11`.
2. Turn `false` to `readingChallengeOnboardingShown`
3. Go to the `PageActivity` and click `X` on the prompt screen.
4. Go back to the device home screen, and click on "Join Challenge"

**Phabricator:**
https://phabricator.wikimedia.org/T418624#11873358


cooltey:
Note: PM has approved the behavior that when the user is in the `PageActivity`, join the challenge from the widget will bounce back to the feed.

Williamrai (commented):
It looks good. Unrelated to the bug fix, right now tapping the container opens the home screen instead of initiating the join challenge dialog. Would it be possible to have the entire container trigger **JoinChallengeAction** for both the small and large widgets?

cooltey:
> It looks good. Unrelated to the bug fix, right now tapping the container opens the home screen instead of initiating the join challenge dialog. Would it be possible to have the entire container trigger **JoinChallengeAction** for both the small and large widgets?

Yes. Done.

PR #6547, opened 2026-04-30
[ReadingChallenge] Fix: make content scrollable if the text size is too large

cooltey:
### What does this do?
We can use `LazyColumn` in the widget, so when the text is too large, the text area becomes scrollable.

I believe this is not ideal, but an alternative fix for different font size configurations on devices. 

Pending on PM & design approval

https://github.com/user-attachments/assets/0ae1841f-4f39-4271-bfb5-02fff10e3b51

https://github.com/user-attachments/assets/87802a3a-74cd-49b5-9a69-9aded6ef00a7

**Phabricator:**
https://phabricator.wikimedia.org/T421502#11873008


cooltey:
Please note this change does not apply to the `StreakOngoingLargeWidget`, which requires additional architecture changes.

PR #6553, opened 2026-05-01
Notification bell

Williamrai:
### What does this do?
- creates jetpack version of NotificationButtonView
- adds a self animating **NotificationBell** 
- adds ripple indication so that touch feedback is similar to XML **selectableItemBackgroundBorderless** 

**Phabricator:**
https://phabricator.wikimedia.org/T421159

**Demo**
[Screen_recording_20260504_100422.webm](https://github.com/user-attachments/assets/700facb2-ef91-4b96-ae4a-062e197e7566)


app/src/main/java/org/wikipedia/compose/components/NotificationBell.kt
```diff
@@ -0,0 +1,126 @@
...
+        AnimatedVisibility(
+            visible = unreadCount > 0,
+            enter = scaleIn() + fadeIn(),
+            exit = scaleOut() + fadeOut(),
+            modifier = Modifier
+                .align(Alignment.TopEnd)
+                .offset(x = (-6).dp, y = 6.dp)
+        ) {
+            CircleWithNumber(number = unreadCount)
+        }
```
cooltey, line 89: Just curious, is it possible to use `Modifier.pulse()` here?

Williamrai: Yes, pulse can be used here, but it would require moving similar logic from inside the function to the caller.

The original intent of keeping the custom animation inside the function was to simplify things for the caller so they only need to pass the unread count. **NotificationBell** then handles animating only when there’s a difference between the previous and current count For example, when the user is on the Games or Home screen and the notification broadcast receiver sends an updated value.

With the current architecture, **pulse and notification bell** would both still need manual updates in **onResume**. Even though the notification broadcast receiver emits an event whenever a new count is received, we’re not following a Compose-style reactive pattern .So i think for future we might need to think about a refactor where we can actually fully use compose reactive way.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -450,6 +455,40 @@ fun HomeScreen(
     }
 }
 
+@Composable
+fun HomeToolbar(
```
Williamrai, line 514: I copied this from the tabs PR.

app/src/main/java/org/wikipedia/main/MainActivity.kt
```diff
@@ -233,7 +233,7 @@ class MainActivity : SingleFragmentActivity<MainFragment>(), MainFragment.Callba
     }
 
     override fun onUnreadNotification() {
-        fragment.updateNotificationDot(true)
+        (fragment.currentFragment as? HomeFragment)?.onUnreadNotification()
```
cooltey: This change will break the existing behavior of updating the notification icon in reading list and other fragments under `MainFragment`.

app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -600,6 +601,7 @@ class MainFragment : Fragment(), BackPressedHandler, MenuProvider, FeedFragment.
 
     private fun refreshContents() {
         when (val fragment = currentFragment) {
+            is HomeFragment -> fragment.onUnreadNotification()
```
cooltey: We don't need this if you handle the notification events correctly.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -316,7 +330,8 @@ fun HomeScreen(
     onTabClick: () -> Unit = {},
     onUpdateTabCount: () -> Unit = {},
     onCustomizeInterestsClick: () -> Unit = {},
-    onCardImpression: (card: Card) -> Unit = { _ -> }
+    onCardImpression: (card: Card) -> Unit = { _ -> },
+    onNotificationBellClick: () -> Unit = {}
```
cooltey: This can be simplified to `onNotificationClick`.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -260,6 +268,7 @@ class HomeFragment : Fragment() {
     override fun onResume() {
         super.onResume()
         viewModel.updateTabCount()
+        onUnreadNotification()
```
cooltey: If you catch the notification event correctly, you won't need to add it to `onResume`

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -141,6 +144,7 @@ class HomeFragment : Fragment() {
                 val selectedTab by viewModel.selectedTab.collectAsState()
                 val wikiSite by viewModel.wikiSite.collectAsState()
                 val tabsState by viewModel.tabsState.collectAsState()
+                val notificationBellState by viewModel.unreadCount.collectAsState()
```
cooltey: `notificationState` is fine.

cooltey (changes requested):

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: There is a **race condition** between the **notification** sending the event vs the **MainActivity** collecting event. When notification is sending the event the MainActivity is not resumed yet and thats why it misses the event so the override fun onUnreadNotification() never gets called, that's why i added this in **onResume** to update the notification in case we missed that. FlowEventBus has 0 as replay so it cannot resend previous value.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: done.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: done.

app/src/main/java/org/wikipedia/main/MainActivity.kt
Williamrai: added back.

app/src/main/java/org/wikipedia/main/MainFragment.kt
Williamrai: I added this because **refreshContents**  is called in **LoggedOutEvent** and is collected in onCreate of **MainFragment**.  This was needed to handle a timing issue between **logoutEvent** and **AccounUtil updating the login state** . When user logs out the notification still stays (which also exist in main) and once user opens the app again it will go away.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
cooltey: sounds good.

cooltey (approved):
It looks good to me now. @dbrant please review and merge if it looks good to you too.

PR #6554, opened 2026-05-01
[Reading Challenge Widget] Fix mascot disappearing

Williamrai:
### What does this do?
- fixes baby globe being pushed out issue

### Why is this needed?
- The `LocalSize.current` was returning incorrect dimensions on OxygenOS, and it may behave similarly on other operating systems. This PR addresses the issue by assigning a fixed width to the right column based on its maximum required size which is the size of the mascot plus extra spacing, while the left column uses a default weight. This ensures the mascot is always displayed properly.

**Phabricator:**
https://phabricator.wikimedia.org/T421502


app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeLargeWidgetContent.kt
cooltey: I guess this is no longer valid.

Williamrai: removed.

PR #6566, opened 2026-05-06
Remove Notification View from Explore Feed when user logs out

Williamrai:
### What does this do?
There is a timing issue between user logging out and **AccountUtil** updating its state. When user logs out, they are navigated to **MainFragment** where **onPrepareMenu** is remove. At this point `AccountUtil.isLoggedIn` still returns true and the collector which receives the **LoggedOutEvent** does not have the code to remove the notification. This PR ensures the notification button is properly removed when the user logs out.

### DIFF
--- a/app/src/main/java/org/wikipedia/main/MainFragment.kt
+++ b/app/src/main/java/org/wikipedia/main/MainFragment.kt
@@ -149,7 +149,7 @@ class MainFragment : Fragment(), BackPressedHandler, MenuProvider, FeedFragment.
                     when (event) {
                         is LoggedOutEvent,
                         is LoggedOutInBackgroundEvent -> {
-                            notificationButtonView.isVisible = false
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/main/MainFragment.kt
```diff
@@ -149,6 +149,7 @@ class MainFragment : Fragment(), BackPressedHandler, MenuProvider, FeedFragment.
                     when (event) {
                         is LoggedOutEvent,
                         is LoggedOutInBackgroundEvent -> {
+                            notificationButtonView.isVisible = false
```
cooltey: Should we run `requireActivity().invalidateOptionsMenu()` and let the logic decide the visibility there?

Williamrai: yes we can do that.

PR #6575, opened 2026-05-08
Make ReadingLists correctly use remote Create and Modify time.

dbrant:
Well, this is interesting:
It looks like Reading List entries (on the server side) have a `created` and `updated` time, which are persisted in the database. And _we never integrated with these dates_ in our client-side code 🫠

So... let's.

https://phabricator.wikimedia.org/T424161

### DIFF
--- a/app/src/main/java/org/wikipedia/readinglist/database/ReadingListPage.kt
+++ b/app/src/main/java/org/wikipedia/readinglist/database/ReadingListPage.kt
@@ -31,15 +31,12 @@ data class ReadingListPage(
     var remoteId: Long = 0
 ) : Serializable {
 
-    constructor(title: PageTitle, fromRemoteSync: Boolean = false) :
+    constructor(title: PageTitle) :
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/wikipedia/readinglist/database/ReadingListPage.kt
```diff
@@ -31,15 +31,12 @@ data class ReadingListPage(
     var remoteId: Long = 0
 ) : Serializable {
 
-    constructor(title: PageTitle, fromRemoteSync: Boolean = false) :
+    constructor(title: PageTitle) :
             this(title.wikiSite, title.namespace(), title.displayText, title.prefixedText,
                 title.description, title.thumbUrl, lang = title.wikiSite.languageCode) {
         val now = System.currentTimeMillis()
         mtime = now
-        // The `atime` (added time) field is used for sorting pages added *locally* only.
```
dbrant, line 39: It no longer matters whether pages were added locally or remotely, since the dates are now known in both cases.

PR #6576, opened 2026-05-09
Instrumentation for app_open events

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T423720


app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -112,13 +114,42 @@ abstract class BaseActivity : AppCompatActivity(), ConnectionStateMonitor.Callba
...
         if (AppShortcuts.ACTION_APP_SHORTCUT == intent.action) {
-            intent.putExtra(Constants.INTENT_EXTRA_INVOKE_SOURCE, Constants.InvokeSource.APP_SHORTCUTS)
+            intent.putExtra(Constants.INTENT_EXTRA_INVOKE_SOURCE, InvokeSource.APP_SHORTCUTS)
             val shortcutId = intent.getStringExtra(AppShortcuts.APP_SHORTCUT_ID)
             if (!shortcutId.isNullOrEmpty()) {
                 ShortcutManagerCompat.reportShortcutUsed(applicationContext, shortcutId)
             }
         }
 
+        _instrument = TestKitchenAdapter.client.getInstrument("apps-open")
```
dbrant: This feels a bit dangerous -- The `_instrument` field is intended for _subclasses_ of `BaseActivity` to use for instrumenting each Activity's individual needs. (this is specified in the comment above `_instrument`, maybe this should be a javadoc comment)

After you instantiate `_instrument` here, it can be overwritten anytime by a superclass.

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -112,13 +114,42 @@ abstract class BaseActivity : AppCompatActivity(), ConnectionStateMonitor.Callba
...
+            .setDefaultAction("app_open")
+
+        val invokeSource = intent.getSerializableExtra(Constants.INTENT_EXTRA_INVOKE_SOURCE) as InvokeSource?
+        invokeSource?.let {
+            when (it) {
+                InvokeSource.WIDGET -> {
+                    val widgetType = intent.getStringExtra(Constants.INTENT_WIDGET_TYPE)
+                    instrument?.submitInteraction(actionSource = "widget", actionSubtype = widgetType)
```
dbrant: These can be changed to inline instances of Instrument: `TestKitchenAdapter.client.getInstrument("apps-open")...`

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -261,6 +292,12 @@ abstract class BaseActivity : AppCompatActivity(), ConnectionStateMonitor.Callba
 
     override fun onGoOnline() {}
 
+    override fun onNewIntent(intent: Intent) {
+        super.onNewIntent(intent)
+        val invokeSource = intent.getSerializableExtra(Constants.INTENT_EXTRA_INVOKE_SOURCE) as InvokeSource?
+        instrument?.submitInteraction(actionSource = "background", actionSubtype = invokeSource?.value)
```
dbrant: This one is kind of confusing: When exactly does this get fired? For what kinds of situations? If it gets fired whenever _any activity_ receives a new Intent, then why is it called "background"?

Reading the task, it says the following for `background`:
> If the user switches back to Chrome, putting our app in the background (but very much still running), then later taps another Wikipedia link, bringing our app back to the foreground that is another app_open.

...But isn't this already handled with the `external_link` event? What is this adding?

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -112,13 +114,42 @@ abstract class BaseActivity : AppCompatActivity(), ConnectionStateMonitor.Callba
...
+                else -> { }
+            }
+        }
+
+        if (intent.action == Intent.ACTION_MAIN && intent.categories?.contains(Intent.CATEGORY_LAUNCHER) == true) {
+            instrument?.submitInteraction(actionSource = "app_icon")
+        }
+
+        if (intent.action == Intent.ACTION_VIEW) {
+            instrument?.submitInteraction(actionSource = "external_link")
```
dbrant: Should this go into `PageActivity.handleIntent()`?

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -112,13 +114,42 @@ abstract class BaseActivity : AppCompatActivity(), ConnectionStateMonitor.Callba
...
+                InvokeSource.APP_SHORTCUTS -> {
+                    val shortcutId = intent.getStringExtra(AppShortcuts.APP_SHORTCUT_ID)
+                    instrument?.submitInteraction(actionSource = "shortcut", actionSubtype = shortcutId)
+                }
+                else -> { }
+            }
+        }
+
+        if (intent.action == Intent.ACTION_MAIN && intent.categories?.contains(Intent.CATEGORY_LAUNCHER) == true) {
+            instrument?.submitInteraction(actionSource = "app_icon")
```
dbrant: Should this go into `MainActivity.handleIntent()`?

dbrant (changes requested):

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
cooltey: It makes sense to me, fixed!

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
cooltey: Done.

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
cooltey: Done.

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
cooltey: Done. I originally put this in `MainActivity` but it has the same result when I put this in `BaseActivity`. I guess putting this in `MainActivity` is clearer.

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
cooltey: Does the `background` event only apply to `external_link` but not to the other actions, such as `widget` or `shortcut`? If that were the case, we could remove `background`.

PR #6577, opened 2026-05-09
Configure `android:windowSplashScreenBackground` to avoid white flash when the app opens in dark mode

stefansundin:
### What does this do?

If you have your phone in dark mode and then open the app you will get a white flash. This change fixes this by configuring `android:windowSplashScreenBackground`.

### Why is this needed?

Screen recording of what it did before:

[before.webm](https://github.com/user-attachments/assets/624a7f25-8fcd-4363-bf5c-184b10a7b299)

After this change it does not flash:

[after.webm](https://github.com/user-attachments/assets/bf22bd88-061a-4803-9d57-39851263b1b1)

### DIFF
--- a/app/src/main/res/values-night/colors.xml
+++ b/app/src/main/res/values-night/colors.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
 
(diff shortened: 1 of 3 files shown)


cooltey:
Thank you for submitting this PR!

PR #6587, opened 2026-05-13
Additional strings for the Home feed

Williamrai:
### What does this do?
- adds community and for you settings strings

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -2509,6 +2509,19 @@
     <string name="home_feed_settings_whats_driving_title">What\'s driving your feed</string>
     <string name="home_feed_settings_modules_title">Modules</string>
     <string name="home_feed_settings_community_subtitle">Turn on or off Community feed sections</string>
-    <string name="home_feed_settings_for_you_subtitle">Turn on or off For You feed sections</string>
+    <string name="home_feed_settings_for_you_subtitle">Turn on or off For you feed sections</string>
     <string name="home_feed_settings_whats_driving_subtitle">Topics and articles shaping your content</string>
+    <string name="home_feed_settings_community_modules_description">Turning off a module hides it from your Community feed. You can re-enable it here at any time. Some modules may not be available in your language.</string>
+    <string name="home_feed_settings_community_module_featured_article_subtitle">Featured articles are some of the best articles on Wikipedia, selected daily by editors</string>
+    <string name="home_feed_settings_community_module_top_red_subtitle">What is trending today on Wikipedia</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -2509,6 +2509,19 @@
     <string name="home_feed_settings_whats_driving_title">What\'s driving your feed</string>
     <string name="home_feed_settings_modules_title">Modules</string>
     <string name="home_feed_settings_community_subtitle">Turn on or off Community feed sections</string>
-    <string name="home_feed_settings_for_you_subtitle">Turn on or off For You feed sections</string>
+    <string name="home_feed_settings_for_you_subtitle">Turn on or off For you feed sections</string>
     <string name="home_feed_settings_whats_driving_subtitle">Topics and articles shaping your content</string>
+    <string name="home_feed_settings_community_modules_description">Turning off a module hides it from your Community feed. You can re-enable it here at any time. Some modules may not be available in your language.</string>
+    <string name="home_feed_settings_community_module_featured_article_subtitle">Featured articles are some of the best articles on Wikipedia, selected daily by editors</string>
```
dbrant: All of the "title" and "subtitle" strings already exist, and don't need to be duplicated.

app/src/main/res/values/strings.xml
```diff
@@ -2509,6 +2509,19 @@
     <string name="home_feed_settings_whats_driving_title">What\'s driving your feed</string>
     <string name="home_feed_settings_modules_title">Modules</string>
     <string name="home_feed_settings_community_subtitle">Turn on or off Community feed sections</string>
-    <string name="home_feed_settings_for_you_subtitle">Turn on or off For You feed sections</string>
+    <string name="home_feed_settings_for_you_subtitle">Turn on or off For you feed sections</string>
```
dbrant: Both words "For You" should be capitalized. Otherwise this reads really weirdly as a sentence.

dbrant (changes requested):

app/src/main/res/values/strings.xml
Williamrai: originally it was "For You", they asked to change it to "For you". https://phabricator.wikimedia.org/T419761#:~:text=Thanks,-%2C%20William!%20I%20have

dbrant: This makes it very weird grammatically. I left a comment on the task to discuss.

app/src/main/res/values/strings.xml
Williamrai: removed duplicated strings.

app/src/main/res/values/strings.xml
Williamrai: Thank you. updated For you with quotes and replaced sections to modules.

PR #6592, opened 2026-05-13
Fix: do not load from cache if the last entry from database is newer than the cache

cooltey:
### What does this do?
If the app has the cache list for the "For you" feed, it will not show "Continue reading" and "Because you read" modules after the feed is loaded until the next day.

This PR does an additional check for the last entry from the database. We can have another PR to handle the cache mechanism for each module instead of the whole list.


dbrant:
> until the next day

Isn't this expected behavior?

cooltey:
Resolved!

PR #6598, opened 2026-05-15
[Explore Feed] Announcements should not interrupt Home feed onboarding

Williamrai:
### What does this do?
- centralizes announcement logic in a new `AppAnnouncements` object and gates it on onboarding completion
- makes it easier to manage and change as we add more announcements

### Why is this needed?
- We had `maybeShowYearInReview` and `maybeShowReadingChallengePrompt` in `BaseActivity`, both triggered every time `BaseActivity` was created. The conditional checks were getting long and duplicated  any rule change had to be made in two places, and it was easy to forget one.
- Announcements could also appear back-to-back. An announcement opens its own activity, which extends `BaseActivity`, which re-triggers the announcement check and could show another one immediately after. The block list in `AppAnnouncements` prevents this.

**Phabricator:**
https://phabricator.wikimedia.org/T426267


cooltey (commented):
This looks fine, but do we really need a separate class to handle the conditions, if the reading challenge will be removed soon? 🤔

Williamrai:
> This looks fine, but do we really need a separate class to handle the conditions, if the reading challenge will be removed soon? 🤔

I created a dedicated class because I think we may add more popups like this in the future, and having a single dedicated class would make them easier to manage.

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -301,20 +301,20 @@ abstract class BaseActivity : AppCompatActivity(), ConnectionStateMonitor.Callba
...
     }
 
     protected fun maybeShowReadingChallengePrompt() {
         if (ReadingChallengeWidgetRepository.shouldShowOnboardingDialog() &&
-            this !is ReadingChallengeOnboardingActivity && this !is PersonalizationActivity && this !is InitialOnboardingActivity && this !is ExploreFeedBuildingActivity) {
-            requestReadingChallengeActivity.launch(ReadingChallengeOnboardingActivity
-                .newIntent(this)
-                .setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP))
+            this !is ReadingChallengeOnboardingActivity) {
+            showReadingChallenge()
```
cooltey: This function seems a bit confusing to me, because it will only be used in the `MainActivity`.

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
```diff
@@ -301,20 +301,20 @@ abstract class BaseActivity : AppCompatActivity(), ConnectionStateMonitor.Callba
...
-            yearInReviewLauncher.launch((YearInReviewOnboardingActivity.newIntent(this)))
-        }
+    fun showYearInReview() {
+        yearInReviewLauncher.launch((YearInReviewOnboardingActivity.newIntent(this)))
+    }
+
+    fun showReadingChallenge() {
+        requestReadingChallengeActivity.launch(ReadingChallengeOnboardingActivity
+            .newIntent(this)
+            .setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP))
```
cooltey: Maybe we can move the logic into `maybeShowAnnouncement` directly.

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
Williamrai: done.

app/src/main/java/org/wikipedia/activity/BaseActivity.kt
Williamrai: We can move **showYearInReview**, but **showReadingChallenge** is used both here and in MainActivity. In MainActivity, I wanted to explicitly indicate that we are showing the Reading Challenge announcement.

cooltey (approved):

PR #6599, opened 2026-05-15
Home Feed: make Bookmarks properly reactive.

dbrant:
This introduces a proper™ way of handling reading list bookmark states in the Feed, using a StateFlow that is directly integrated into the Room query.

With not too much more code, the bookmark state becomes fully reactive and Compose-worthy.


dbrant:
closing in favor of updated PR.

PR #6601, opened 2026-05-15
[Explore Feed] Empty State "For You" screen

Williamrai:
### What does this do?
- Handles For You empty states when user has not selected any interests and has no reading history (new user)
- Skips home preference page when launching personalization from empty state CTA 
- Adds logic to open the swipe-to-explore dialog only when there is content available on the "For you" screen.

**Phabricator:**
https://phabricator.wikimedia.org/T426456


cooltey (changes requested):
In some cases, if you load the feed without selecting interests first, and then disable the module, and then select interests, and finally enable the module, the feed will show the same "add interests" empty state, which seems incorrect to me.

And then, do we want to show the same message if I have no history, but I enable "Because you read" and "Continue reading", is this expected?

app/src/main/java/org/wikipedia/feed/personalization/PersonalizationActivity.kt
```diff
@@ -42,8 +42,7 @@ class PersonalizationActivity : BaseActivity() {
             )
         } else {
             listOf(
-                PersonalizationPage.INTERESTS,
-                PersonalizationPage.HOME_PREFERENCE
+                PersonalizationPage.INTERESTS
```
cooltey: This change seems a bit off to me. The variable is `EXTRA_SHOW_INTRO_PAGE` but what you do here is only showing the `interests` screen without `home_preference`.

It would be better if we could send something clearer for different conditions.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -147,8 +152,8 @@ class HomeFragment : Fragment() {
 
     private val customizeInterestsLauncher = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) {
         if (it.resultCode == RESULT_OK) {
-            selectTab(HomeTab.FOR_YOU)
             Prefs.homeForYouModulesToday = ""
+            viewModel.selectTabWithoutLoadingContent(HomeTab.FOR_YOU)
```
cooltey: Instead of creating a new function for only skipping loading the content, could we add a parameter in the existing function for this purpose?

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -179,7 +185,7 @@ class HomeFragment : Fragment() {
                         languageState = WikipediaApp.instance.languageState,
                         selectedTab = selectedTab,
                         communityContentState = viewModel.communityState.collectAsState().value,
-                        forYouContentState = viewModel.forYouState.collectAsState().value,
+                        forYouContentState = forYouContentState,
```
cooltey, line 189: That's the reason for this change? If this relates to the code formatting, would you please also do this for `communityContentState`?

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -445,7 +455,9 @@ fun HomeScreen(
     onCardImpression: (card: Card, index: Int) -> Unit = { _, _ -> },
     onCardFooterClick: (card: Card) -> Unit = {},
     onNotificationClick: () -> Unit = {},
-    onManageModulesClick: () -> Unit = {}
+    onManageModulesClick: () -> Unit = {},
+    navigateToCommunityTab: () -> Unit = {},
```
cooltey: I don't think we need a new parameter for this action. We could reuse `onSelectTab()`

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -543,7 +555,9 @@ fun HomeScreen(
                         onPageOverflowDismiss = onPageOverflowDismiss,
                         onCustomizeInterestsClick = onCustomizeInterestsClick,
                         onCardImpression = onCardImpression,
-                        onManageModulesClick = onManageModulesClick
+                        onManageModulesClick = onManageModulesClick,
+                        navigateToCommunityTab = navigateToCommunityTab,
+                        onCustomizeInterestsFromEmptyStateClick = onCustomizeInterestsFromEmptyStateClick
```
cooltey: Can we reuse the existing `onCustomizeInterestsClick` instead of creating a new one?

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -752,17 +766,18 @@ fun CommunityContentTab(
         state.isInitialLoading -> {
             LoadingIndicator(modifier = modifier.fillMaxHeight())
         }
-        state.isEmptyState -> {
+        state.areAllModulesHidden -> {
```
cooltey: There's a conversation in Slack talking about the actual empty content vs hidden modules. Would you be able to handle those conditions here? It's similar to `ForYouEmptyState.NO_DATA`

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -752,17 +766,18 @@ fun CommunityContentTab(
...
                     .fillMaxSize()
                     .background(WikipediaTheme.colors.paperColor)
                     .padding(16.dp)
                     .verticalScroll(rememberScrollState()),
                 title = context.getString(wikiSite.languageCode, R.string.home_feed_screen_empty_state_label),
-                description = context.getString(wikiSite.languageCode, R.string.home_feed_community_screen_empty_state_description),
-                onManageModulesClick = onManageModulesClick
+                description = context.getString(wikiSite.languageCode, R.string.home_feed_community_screen_all_modules_disabled_description),
+                buttonText = context.getString(wikiSite.languageCode, R.string.home_feed_screen_all_modules_disabled_btn_label),
+                onCtaClick = onManageModulesClick
```
cooltey: Instead of calling it `CtaClick`, could we name it `onCallToActionClick` or `onActionButtonClick`?

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -1359,11 +1391,12 @@ fun LanguageDropDownMenu(
 }
 
 @Composable
-fun HomeScreenEmptyState(
+fun AllModulesHiddenView(
```
cooltey: Based on the conversation on Slack about the empty state for `Community`, it's more likely this should be named more generally.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -1393,12 +1426,103 @@ fun HomeScreenEmptyState(
...
             Text(
-                text = stringResource(R.string.home_feed_screen_empty_state_btn_label)
+                text = buttonText
+            )
+        }
+    }
+}
+
+@Composable
+fun ForYouNoDataView(
```
cooltey: Would it be possible to name it `FeedContentEmptyView` or something more general?

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -1393,12 +1426,103 @@ fun HomeScreenEmptyState(
...
+            tint = WikipediaTheme.colors.primaryColor,
+            contentDescription = null
+        )
+        HtmlText(
+            text = text,
+            linkStyle = TextLinkStyles(
+                style = SpanStyle(
+                    fontSize = 14.sp,
+                    color = WikipediaTheme.colors.primaryColor,
+                    textDecoration = TextDecoration.Underline)
```
cooltey: The ending `)` should be in a new line to match the code format.

app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
```diff
@@ -139,13 +145,24 @@ class HomeViewModel : ViewModel() {
...
-        state.copy(cards = visibleModules, isEmptyState = CommunityModuleType.entries.all { hiddenModules.contains(it.name) })
+        state.copy(cards = visibleModules, areAllModulesHidden = CommunityModuleType.entries.all { hiddenModules.contains(it.name) })
     }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(MAX_STOP_TIMEOUT_MILLIS), CommunityContentState())
 
     private val _forYouState = MutableStateFlow(ForYouContentState())
     val forYouState = combine(_forYouState, SettingsRepository.hiddenModules) { state, hiddenModules ->
         val visibleModules = state.modules.filterNot { hiddenModules.contains(it.moduleKey()) }
-        state.copy(modules = visibleModules, isEmptyState = ForYouModuleType.entries.all { hiddenModules.contains(it.name) })
+        val allHidden = ForYouModuleType.entries.all { hiddenModules.contains(it.name) }
+        val interestHidden = hiddenModules.contains(ForYouModuleType.BASED_ON_INTEREST.name)
```
cooltey: These variable names are confusing. Would it be better if we name them more precisely, such as `areAllModulesHidden`?

cooltey (changes requested):

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: The reason for this change was not related to code formatting, but to use it in a conditional check to determine whether the swipe-to-explore dialog should be shown. Updated for **communityContentState** as well.

app/src/main/java/org/wikipedia/feed/personalization/PersonalizationActivity.kt
Williamrai: replaced **EXTRA_SHOW_INTRO_PAGE** with **EXTRA_SHOW_INTERESTS_ONLY**, since we are either showing all three or just one interest.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: I added **selectTabWithoutLoadingContent** because the existing function was both selecting the tab and reloading its content, which caused duplicate loading when the user returned from the interest selection screen.

Now I split the behavior into two separate functions: tab selection is now handled independently, and the reload logic has been moved to **reloadCurrentTab()**. I think later we can refactor this to observe the `selectedTab`.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: I can reuse the **onCustomizeInterestsClick** but we don't have an instrumentation for when empty feed is shown. For now i added a nullable to card parameter and skip the instrumentation until we have an instrumentation.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: done.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: done.

app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
Williamrai: done.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: Could you point out to that conversation? is it this [one](https://wikimedia.slack.com/archives/C7P3N5E6P/p1779136911451239)?

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: Could you point out to that conversation? is it this [one](https://wikimedia.slack.com/archives/C7P3N5E6P/p1779136911451239)?

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: This view is only applicable to For you Feed so would it be okay to name this **ForYouFeedEmptView**?

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: removed.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: Done. converted **forYouEmptyState** to a generic **FeedEmptyState** and using this in **CommunityContentState**

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
Williamrai: renamed to **FeedEmptyStateView**

cooltey (changes requested):
Found an issue that if you have multiple app languages, the "add interests" does not work properly for the non-primary languages. 

The same thing happens when you change the language order in the app languages list.

https://github.com/user-attachments/assets/3b47e267-0a60-4173-be1f-e669dc6d4169

cooltey:
> Found an issue that if you have multiple app languages, the "add interests" does not work properly for the non-primary languages.
> 
> The same thing happens when you change the language order in the app languages list.
> Screen_Recording_20260519_134336_Wikipedia.Dev.mp4

The issue is still here after merging #6610.

PR #6619, opened 2026-05-21
[Explore Feed] Fix localized topic and primary language display

Williamrai:
### What does this do?
- Shows the correct topic string based on the selected language
- displays the correct primary language in the Home tab bar

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
+++ b/app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
@@ -249,8 +249,13 @@ class HomeViewModel : ViewModel() {
     }
 
     fun updateSelectedLanguageIfNeeded() {
-        if (Prefs.homeLanguageCode != wikiSite.value.languageCode) {
-            _wikiSite.value = WikiSite.forLanguageCode(Prefs.homeLanguageCode)
+        if (!WikipediaApp.instance.languageState.appLanguageCodes.contains(wikiSite.value.languageCode)) {
+            updateLanguage(WikipediaApp.instance.languageState.appLanguageCode)
+        }
+
+        val newPrimaryLanguage = WikipediaApp.instance.languageState.appLanguageCode
+        if (newPrimaryLanguage != wikiSite.value.languageCode) {
+            _wikiSite.value = WikiSite.forLanguageCode(newPrimaryLanguage)
         }
     }
 
--- a/app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
+++ b/app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
@@ -249,8 +249,8 @@ class HomeViewModel : ViewModel() {
     }
 
     fun updateSelectedLanguageIfNeeded() {
-        if (!WikipediaApp.instance.languageState.appLanguageCodes.contains(wikiSite.value.languageCode)) {
-            updateLanguage(WikipediaApp.instance.languageState.appLanguageCode)
+        if (Prefs.homeLanguageCode != wikiSite.value.languageCode) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/wikipedia/feed/personalization/interest/InterestSelectionScreen.kt
```diff
@@ -64,6 +66,7 @@ fun InterestOnboardingScreen(
     articlesState: ArticlesState,
     topicsList: List<OnboardingTopic>,
     totalSelectedCount: Int,
+    languageCode: String = "en",
```
dbrant: Let's not hardcode default parameters (`= "en"`) unnecessarily. If we definitely want the caller of this function to pass a language code (which we do), then it shouldn't be optional.

app/src/main/java/org/wikipedia/feed/personalization/interest/InterestSelectionScreen.kt
```diff
@@ -203,16 +207,20 @@ fun InterestOnboardingScreen(
 fun TopicFilterChipRow(
     topics: List<OnboardingTopic>,
     modifier: Modifier = Modifier,
+    languageCode: String = "en",
```
dbrant: Same, etc.

app/src/main/java/org/wikipedia/feed/personalization/interest/InterestSelectionState.kt
```diff
@@ -11,7 +11,8 @@ data class OnboardingTopic(
 data class InterestUiState(
     val topicsList: List<OnboardingTopic> = emptyList(),
     val articlesState: ArticlesState = ArticlesState.Loading,
-    val totalSelectedCount: Int = 0
+    val totalSelectedCount: Int = 0,
+    val languageCode: String = "en"
```
dbrant: Same

app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
```diff
@@ -249,8 +249,8 @@ class HomeViewModel : ViewModel() {
     }
 
     fun updateSelectedLanguageIfNeeded() {
-        if (!WikipediaApp.instance.languageState.appLanguageCodes.contains(wikiSite.value.languageCode)) {
-            updateLanguage(WikipediaApp.instance.languageState.appLanguageCode)
+        if (Prefs.homeLanguageCode != wikiSite.value.languageCode) {
+            _wikiSite.value = WikiSite.forLanguageCode(Prefs.homeLanguageCode)
```
dbrant: This breaks the original purpose that this function was solving:
* Switch your feed to a different language
* Go to manage your languages and remove that language.
* Go back to the feed, and it should refresh automatically.

Is this change actually necessary here?

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/personalization/interest/InterestSelectionScreen.kt
Williamrai: removed.

app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
Williamrai: I added this change to resolve this issue. During onboarding if user changes their primary language and navigate to home fragment the home tab bar will still show english and if they change their primary language then they will still see their previous primary language. I’ll revert this change and add the logic into the original implementation.

[home tab bar.webm](https://github.com/user-attachments/assets/84b33e9a-da42-425e-b551-1ee9948bd82b)

PR #6627, opened 2026-05-22
Feed: Put back Home Preference selection when customizing interests.

dbrant:
At some point, the `HOME_PREFERENCE` page was removed from the personalization activity (when customizing interests after already onboarded).
But this means that the user will never be able to change their default preference of Community vs. For You. We need to add back this selection.

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/personalization/PersonalizationActivity.kt
+++ b/app/src/main/java/org/wikipedia/feed/personalization/PersonalizationActivity.kt
@@ -37,7 +37,8 @@ class PersonalizationActivity : BaseActivity() {
 
         val pages = if (showInterestsOnly) {
             listOf(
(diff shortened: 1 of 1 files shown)


Williamrai:
There was a discussion on this in [slack](https://wikimedia.slack.com/archives/C7P3N5E6P/p1779232969469449) and there is a [task](https://phabricator.wikimedia.org/T426876) for user to select the default preference.

dbrant:
> There was a discussion on this in [slack](https://wikimedia.slack.com/archives/C7P3N5E6P/p1779232969469449) and there is a [task](https://phabricator.wikimedia.org/T426876) for user to select the default preference.

Well I'll be a monkey's uncle.
This breaks my intuition to such a degree that I thought it was a bug. But OK!

PR #6628, opened 2026-05-22
[Explore Feed] Reactive load when tab and language changes

Williamrai:
### What does this do?
- reloads the home feed reactively when the language or selected tab changes, instead of manual calls.

### Why is this needed?
Current code has  `_selectedTab` and `_wikiSIte` as state but we load them manually. Whenever either changed, some caller had to remember to also call the right load/refresh function. This logic is spread across: 
- HomeViewModel.selectTab
- HomeViewModel.reloadCurrentTab
- HomeViewModel.updateLanguage
- HomeFragment.selectTab
So this PR makes loading a reaction to state not manual call.

Note:`HomeFragment.customizeInterestsLauncher` still need an explicit call since interest changed isn't observed as flow.


app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -156,7 +156,6 @@ class HomeFragment : Fragment() {
     private val customizeInterestsLauncher = registerForActivityResult(ActivityResultContracts.StartActivityForResult()) {
         if (it.resultCode == RESULT_INTERESTS_UPDATED) {
             Prefs.homeForYouModulesToday = ""
```
Williamrai, line 68: This would still be manual call because right now we are not observing interest change flow.

cooltey (approved):
It looks good to me!

PR #6629, opened 2026-05-23
Fix: use LaunchedEffect to resolve the overflow animation issue on HomeFeed

cooltey:
### What does this do?
When testing the HomeFeed on Pixel devices or on the emulators, the dismiss animation on the top read's overflow menu does not look really smooth and acts weirdly.

The cause of the issue is probably the `forEach` loop inside the `DropdownMenu`, and we can use `LaunchEffect` to take over the dismiss timing to avoid the issue.

The other solution is using a static overflow menu and controlling the visibility inside the component, which can also be done later.


Williamrai:
This is working, but I also think we should move to your proposed solution in Phase 2.

PR #6630, opened 2026-05-23
Fix: when the tab icon or notification icon appears, the toolbar moves

cooltey:
### What does this do?
When there are no tab/notification icons on the toolbar, and you open a new article in a new tab, then the toolbar will move a bit to push the content down.


Williamrai:
In Goggle Pixel 9 physical device it pushes it down but on emulator its working. 
https://github.com/user-attachments/assets/6f710a33-aa71-4144-b2e5-aed975f760ed

cooltey:
> In Goggle Pixel 9 physical device it pushes it down but on emulator its working. https://github.com/user-attachments/assets/6f710a33-aa71-4144-b2e5-aed975f760ed

Thanks! I could not reproduce it on my device. Can you please help to test it again with my updated approach?

PR #6634, opened 2026-05-26
Home feed: properly use RTL or LTR layout based on language.

dbrant:
When the user sets the Feed language to an RTL language, the inner contents of the feed should then follow an RTL layout.

sorry, it looks like a huge change, but it's really just surrounding the whole block with `CompositionLocalProvider { ... }`


cooltey (commented):
These are not really related, but do we need to add RTL support for them?

Image
Image

Williamrai (approved):

dbrant:
> These are not really related, but do we need to add RTL support for them?

No to the second one (since it is a purely user-interface element, and not content), but definitely yes to the first one! (Done.)

PR #6642, opened 2026-05-27
[Explore Feed] For you content end of feed

Williamrai:
### What does this do?
- adds end of feed module to For You Content 
- adds circular scrolling

**Phabricator:**
https://phabricator.wikimedia.org/T419757


app/src/main/java/org/wikipedia/feed/ForYouContentTab.kt
```diff
@@ -224,13 +212,103 @@ fun ForYouContentTab(
...
+                            )
+                        }
                     }
                 }
             }
         }
     }
 }
 
+private fun LazyListScope.ForYouModuleItem(
```
cooltey: The function name should start with a lowercase letter (suggested by the IDE).

app/src/main/java/org/wikipedia/feed/ForYouContentTab.kt
```diff
@@ -224,13 +212,103 @@ fun ForYouContentTab(
...
+    module: ForYouModule,
+    index: Int,
+    viewPortHeight: Dp,
+    wikiSite: WikiSite,
+    isDummy: Boolean = false,
```
cooltey: Is this `isDummy` necessary in this case? Will the same key affect anything when scrolling through the feed?

app/src/main/java/org/wikipedia/feed/ForYouContentTab.kt
```diff
@@ -323,6 +401,73 @@ fun EmptyStateActionRow(
     }
 }
 
+@Composable
+fun ForYouEndOfFeedView(
```
cooltey: Just curious, since the UI is pretty much the same as the `ForYouFeedEmptyView`, can we have parameters for the text/icons instead of having two similar functions?

cooltey (changes requested):

app/src/main/java/org/wikipedia/feed/ForYouContentTab.kt
Williamrai: merged into a single function called **ForYouFeedMessageView**

app/src/main/java/org/wikipedia/feed/ForYouContentTab.kt
Williamrai: yes not required its index will make it unique.

app/src/main/java/org/wikipedia/feed/ForYouContentTab.kt
Williamrai: done.

dbrant (changes requested):
Just a couple of design-related comments:

* The title "That's it for today" should be serif font. (same as article titles in the other cards)
* The contents should be aligned to the bottom, not centered vertically.

Williamrai:
> Just a couple of design-related comments: 
> * The contents should be aligned to the bottom, not centered vertically.

I discussed this with the Tais before making it centered vertically and confirmed with her the content should centered vertically rather than bottom-aligned. On taller devices, placing the content at the bottom would leave too much empty space, so centering provides a better visual balance.
image

PR #6643, opened 2026-05-28
Fix: validate the login status for logged-in-required screens

cooltey:
### What does this do?
This prevents the user from seeing the errors in the Activity tab, Year-in-Review, or Watchlist, as the cookie may have expired, but the app still believes the user is logged in. 

This PR creates an `MwCoroutineExceptionHandler` in `ThrowableUtil`, and replaces the general `CoroutineExceptionHandler` to catch the `assertuserfailed` or `notloggedin` error code. If it catches one of the error codes, the app will log out the user and prompt the login dialog to do relogin. 

**Phabricator:**
https://phabricator.wikimedia.org/T427439


app/src/main/java/org/wikipedia/dataclient/Service.kt
```diff
@@ -359,10 +359,10 @@ interface Service {
     @GET(MW_API_PREFIX + "action=query&meta=authmanagerinfo&amirequestsfor=login")
     suspend fun getAuthManagerForLogin(): MwQueryResponse
 
-    @GET(MW_API_PREFIX + "action=query&meta=userinfo&uiprop=groups|blockinfo|editcount|latestcontrib|hasmsg|options")
+    @GET(MW_API_PREFIX + "action=query&meta=userinfo&uiprop=groups|blockinfo|editcount|latestcontrib|hasmsg|options&assert=user")
     suspend fun getUserInfo(): MwQueryResponse
```
dbrant, line 371: We'll need to audit all the other usages of `getUserInfo()` and evaluate the behavior that we want in those cases. (there's not many of them, but still)

In certain cases, we might not want this call to throw, such as the usage in `getBlockMessageHtml()`, and so on.

app/src/main/java/org/wikipedia/activitytab/ActivityTabViewModel.kt
```diff
@@ -206,7 +207,7 @@ class ActivityTabViewModel : ViewModel() {
         if (!AccountUtil.isLoggedIn) {
             return
         }
-        viewModelScope.launch(CoroutineExceptionHandler { _, throwable ->
+        viewModelScope.launch(ThrowableUtil.MwCoroutineExceptionHandler { _, throwable ->
```
dbrant: In addition to handling this error properly, we may also want to change the behavior slightly.

For example, when the user goes to Activity Tab, and we get this error, we may want to explicitly "refresh" the Activity Tab so that it shows the logged-out state, instead of the error state. (We don't need to show an error, because we're already showing an AlertDialog that the user has been logged out.)

app/src/main/java/org/wikipedia/watchlist/WatchlistViewModel.kt
```diff
@@ -24,12 +23,13 @@ import org.wikipedia.util.FeedbackUtil
...
 import org.wikipedia.util.Resource
 import org.wikipedia.util.StringUtil
+import org.wikipedia.util.ThrowableUtil
 import java.io.IOException
 import java.util.Calendar
 
 class WatchlistViewModel : ViewModel() {
 
-    private val handler = CoroutineExceptionHandler { _, throwable ->
+    private val handler = ThrowableUtil.MwCoroutineExceptionHandler { _, throwable ->
```
dbrant: Similarly, if we get this error in the Watchlist activity, we may just want to dump the user out of the activity altogether, because there's no point keeping this activity open with a "logged out" error being shown.

dbrant (changes requested):
This is the correct path, but will need some more behavioral updates.

app/src/main/java/org/wikipedia/dataclient/Service.kt
cooltey: Done, except for the `LoginClient`.

app/src/main/java/org/wikipedia/watchlist/WatchlistViewModel.kt
cooltey: Done!

app/src/main/java/org/wikipedia/activitytab/ActivityTabViewModel.kt
cooltey: Sure, done!

Williamrai (commented):
I created a test account then logged in in the app and desktop, and then I changed the password on the desktop. I tested it this way because it invalidates the session, so the saved cookie gets rejected by the server. I observed the following while testing:

**Case 1**
> **crash on re-entering Activity Tab.** When I opened the Activity tab after password reset, I was redirected to HomeFragment and a logout dialog appeared. I canceled the dialog, then tried to open the Activity tab again, and the app crashed. [Demo](https://github.com/user-attachments/assets/843e2f2a-5c0c-4596-8133-5ef3d60d3abd) The error was: **"FragmentManager is already executing transactions"** [Stack trace Gist](https://gist.github.com/Williamrai/acc288c61af87f50e07235996c063cbf)

**Case 2**
> I changed the password again and opened the activity tab, but this time the Activity tab displayed an "Invalid username" message, while Year in Review showed the logout dialog instead. [Demo](https://github.com/user-attachments/assets/832e1b9d-eb2f-4154-9d3b-36e579fca471)

dbrant:
> **crash on re-entering Activity Tab.**

Whoops, fixed!

> I changed the password again and opened the activity tab, but this time the Activity tab displayed an "Invalid username" message, while Year in Review showed the logout dialog instead.

Should also be fixed now.

dbrant (approved):

PR #6648, opened 2026-05-29
[Explore Feed] Places of Interest Module

Williamrai:
### What does this do?
- adds places of interest module 

**Phabricator:**
https://phabricator.wikimedia.org/T419698


app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
```diff
@@ -560,6 +600,56 @@ class HomeViewModel : ViewModel() {
...
+    private suspend fun buildPlacesModule(): ForYouModule.PlacesOfInterest? {
+        if (!hasLocationPermission()) {
+            return ForYouModule.PlacesOfInterest(age = 0, index = 0, cards = emptyList(), hasLocationPermission = false)
+        }
+        val location = Prefs.placesLastLocationAndZoomLevel?.first
+            ?: return ForYouModule.PlacesOfInterest(age = 0, index = 0, cards = emptyList(), hasLocationPermission = false)
+
+        val cards = getPlacesCards(location)
+        if (cards.isEmpty()) {
+            return null
```
Williamrai, line 640: I reverted the caching logic after realizing I had misread the [requirement](https://phabricator.wikimedia.org/T419698#:~:text=them%20to%20%22Places%22-,If%20location%20is%20granted%3A,-the%20module%20shows). There’s no maximum limit on the number of articles we can show. Instead, I reverted to the previous Home Feed approach, which fetches 10 articles and displays them all which should still provide good coverage of the nearest articles.

app/src/main/java/org/wikipedia/feed/places/PlacesOfInterestModule.kt
```diff
@@ -0,0 +1,232 @@
...
+        Spacer(modifier = Modifier.height(16.dp))
+
+        PlaceholderImageGrid(modifier = Modifier.weight(1f))
+
+        Spacer(modifier = Modifier.height(24.dp))
+
+        Text(
+            modifier = Modifier.fillMaxWidth(),
+            text = context.getString(wikiSite.languageCode, R.string.home_feed_places_of_interest_cta_title),
+            style = MaterialTheme.typography.headlineSmall,
```
dbrant: This should be serif font (like article titles)

app/src/main/java/org/wikipedia/feed/places/PlacesOfInterestModule.kt
```diff
@@ -0,0 +1,232 @@
...
+
+        Spacer(modifier = Modifier.height(24.dp))
+
+        OutlinedButton(
+            modifier = Modifier.align(Alignment.CenterHorizontally),
+            onClick = onGoToPlacesClick,
+            border = BorderStroke(1.dp, WikipediaTheme.colors.primaryColor)
+        ) {
+            Icon(
+                painter = painterResource(R.drawable.ic_location_on_24dp),
```
dbrant: This icon should be the "filled-in" pin, not an outlined pin.

app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
```diff
@@ -170,17 +194,34 @@ class HomeViewModel : ViewModel() {
         state.copy(cards = visibleItems, emptyState = emptyState)
     }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(MAX_STOP_TIMEOUT_MILLIS), CommunityContentState())
 
+    @OptIn(ExperimentalCoroutinesApi::class)
+    private val placesModule: StateFlow<ForYouModule.PlacesOfInterest?> = Prefs.placesLastLocationFlow
```
dbrant, line 205: It looks like the Places module is loaded in a fundamentally different way than all the other modules. This isn't necessarily good or bad, and might even be better in the long term; But because it's so different, can we have some comments that describe how it works, and why it needs to work differently from other modules?

dbrant (changes requested):
just some minor stuff!

app/src/main/java/org/wikipedia/feed/places/PlacesOfInterestModule.kt
Williamrai: done.

app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
Williamrai: added.

app/src/main/java/org/wikipedia/feed/places/PlacesOfInterestModule.kt
Williamrai: done.

dbrant (approved):

PR #6658, opened 2026-06-03
[Explore Feed] reactive hidden card using data store 

Williamrai:
### What does this do?
- Moves the list of hidden feed cards out of `SharedPreferences` and into `DataStore`, and makes both the `Community and For You feeds` **hide/restore** cards reactively.

### Why is this needed?
Hidden cards were stored in the non-observable `Prefs.hiddenCards`, so removing and restoring a card relied on indexing logic and card equality to find it in the list. The For You tab can have multiple independent module lists (not just the interests list), so each would need its own logic to hide and restore cards. This PR simplifies that by moving the preference to DataStore and filtering the hidden cards in combine. 

This also creates a single source of truth for hidden state, and feeds just derives from it.


app/src/main/java/org/wikipedia/feed/news/NewsCard.kt
```diff
@@ -17,7 +17,7 @@ class NewsCard(
     }
 
     override fun dismissHashCode(): Int {
-        return TimeUnit.MILLISECONDS.toDays(date().baseCalendar.time.time).toInt() + wikiSite().hashCode()
+        return TimeUnit.MILLISECONDS.toDays(date().baseCalendar.time.time).toInt() + wikiSite().hashCode() + moduleKey().hashCode()
     }
```
Williamrai, line 21: added `moduleKey().hashCode() `for unique key, without this it was colliding with other card key.

dbrant:
this is the way.

PR #6659, opened 2026-06-03
Possible fix for T395597

dbrant:
Multiple users have reported an issue where the lead image from a _previous_ article remains visible, even after a new article is loaded. Even though we aren't able to reproduce this behavior, I have a suspicion that it has to do with the Fragment lifecycle state and all of the complex sequencing that happens in PageFragmentLoadState. See inline comments for more details.

https://phabricator.wikimedia.org/T395597

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
+++ b/app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
@@ -3,6 +3,7 @@ package org.wikipedia.page
 import android.widget.Toast
 import androidx.lifecycle.lifecycleScope
 import kotlinx.coroutines.CoroutineExceptionHandler
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.async
 import kotlinx.coroutines.launch
@@ -40,6 +41,7 @@ class PageFragmentLoadState(private var model: PageViewModel,
                             private var bridge: CommunicationBridge,
                             private var leadImagesHandler: LeadImagesHandler,
                             private var currentTab: Tab) {
+    private var pageLoadJob: Job? = null
 
     fun load(pushBackStack: Boolean) {
         if (pushBackStack && model.title != null && model.curEntry != null) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
```diff
@@ -117,93 +119,100 @@ class PageFragmentLoadState(private var model: PageViewModel,
...
-                    L.e("Page details network error: ", throwable)
-                    commonSectionFetchOnCatch(throwable)
-                }) {
-                    model.readingListPage = AppDatabase.instance.readingListPageDao().findPageInAnyList(title)
+        pageLoadJob?.let {
+            if (it.isActive) {
+                it.cancel()
+            }
+        }
+        pageLoadJob = fragment.lifecycleScope.launch(CoroutineExceptionHandler { _, throwable ->
```
dbrant, line 127: First, let's keep track of the coroutine as a `Job`, and cancel any previous Job if a new one is being loaded.
This is probably the most likely cause of the actual issue. If a previous Job is taking an abnormally long time, and a new Job begins, they can conflict and race against each other.

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
```diff
@@ -117,93 +119,100 @@ class PageFragmentLoadState(private var model: PageViewModel,
...
+            L.e("Page details network error: ", throwable)
+            commonSectionFetchOnCatch(throwable)
+        }) {
+            if (model.title == null) {
+                return@launch
+            }
```
dbrant, line 133: Invert the check for nullness of `model.title` first, to reduce indentation of the rest of this block.

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
```diff
@@ -234,7 +243,7 @@ class PageFragmentLoadState(private var model: PageViewModel,
     private fun createPageModel(response: Response<PageSummary>,
                                 isWatched: Boolean,
                                 hasWatchlistExpiry: Boolean) {
-        if (!fragment.isAdded || response.body() == null) {
```
dbrant, line 237: This could also be contributing to the issue, and at the very least is complicating the logic unnecessarily. When handling a new `Intent`, it may be possible that the Fragment is not yet resumed (but still alive) when this call happens, and this check disables the rest of the function entirely, preventing the rest of the logic from happening.
We actually don't need this check anymore, because all of the logic actually happens inside a coroutine that is bound to the lifecycleScope of the fragment.

PR #6663, opened 2026-06-05
Home Feed: Random article

dbrant:
* This introduces a couple of slight refactors of the base `ForYouCard` that will now accommodate an arbitrary footer composable, and also dynamically adjust bottom padding to accommodate the presence or absence of the pager dots.

https://phabricator.wikimedia.org/T419755

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/random/RandomModule.kt
+++ b/app/src/main/java/org/wikipedia/feed/random/RandomModule.kt
@@ -1,0 +1,175 @@
+package org.wikipedia.feed.random
+
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.material3.Button
+import androidx.compose.material3.ButtonDefaults
+import androidx.compose.material3.Icon
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.tooling.preview.Preview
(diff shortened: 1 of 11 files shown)


app/src/main/java/org/wikipedia/feed/random/RandomModule.kt
```diff
@@ -0,0 +1,175 @@
...
+        module = module,
+        onCardInView = onCardInView
+    ) { page ->
+        val card = (module.cards[page] as RandomCard)
+        val historyEntry = HistoryEntry(card.title, HistoryEntry.SOURCE_FEED_RANDOM)
+        ForYouCardContent(
+            wikiSite = wikiSite,
+            title = card.title,
+            variation = CardVariation.entries[page % CardVariation.entries.size],
+            backgroundColorIndex = backgroundColorIndex + page,
```
Williamrai: Do we need the `CardVariation.entries[page % CardVariation.entries.size]` and `backgroundColorIndex = backgroundColorIndex + page` since its only one card per day?

app/src/main/java/org/wikipedia/feed/random/RandomModule.kt
```diff
@@ -0,0 +1,175 @@
...
+            module = module,
+            card = module.cards[page],
+            footerText = context.getString(wikiSite.languageCode, R.string.view_random_article_card_title),
+            footerContent = {
+                Button(
+                    onClick = {
+                        context.instrument?.submitInteraction("click", elementId = "random_card_shuffle_button")
+                        context.startActivity(RandomActivity.newIntent(context, wikiSite, Constants.InvokeSource.FEED))
```
Williamrai: instrument is null here, it could be because we have not set it yet in MainActivity. I think we can just pass this click event back up.

app/src/main/java/org/wikipedia/feed/random/RandomModule.kt
dbrant: Done!

app/src/main/java/org/wikipedia/feed/random/RandomModule.kt
dbrant: whoops, Correct!

PR #6668, opened 2026-06-08
Improve haptic feedback

Isira-Seneviratne:
### What does this do?
This updates the vibration effects used in the random and suggested edit fragments to use the [`CONFIRM`](https://developer.android.com/reference/kotlin/androidx/core/view/HapticFeedbackConstantsCompat#CONFIRM()) feedback effect.

### Why is this needed?
The official haptics design principles [recommend](https://developer.android.com/develop/ui/views/haptics/haptics-principles#clear_haptics) using the predefined feedback constants for a consistent effect. The compat class also provides a fallback effect if an effect is unsupported.

**Phabricator:**


dbrant:
The `CONFIRM` feedback effect is barely perceptible on my test devices, as are most of the other predefined constants.
We will be using a custom effect for these cases.

PR #6672, opened 2026-06-10
Home feed: Design updates for Did You Know

dbrant:
* This introduces the ability to _long-press_ links inside a `HtmlText` composable.
* Wire up long-pressing in both `DidYouKnowModule` and `DidYouKnowActivity` (with the same PageOverflowMenu logic as the rest of the feed). There is a bit of code duplication in `DidYouKnowActivity`, but this can be cleaned up and refactored as we go.
* Introduce a tiny function to strip away bold `<b>` tags from html strings, when necessary.

https://phabricator.wikimedia.org/T418898

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/CommunityContentTab.kt
+++ b/app/src/main/java/org/wikipedia/feed/CommunityContentTab.kt
@@ -270,13 +270,12 @@ fun CommunityContentTab(
                                         },
                                         onFooterClick = { onCardFooterClick(card) },
                                         onCardImpression = { onCardImpression(card, cardIndex) },
-                                        pageOverflowContent = { index, offset ->
+                                        pageOverflowContent = { index ->
                                             PageOverflowMenu(
                                                 menuKey = "dyk-${card.date}-$index",
                                                 overflowMenuState = overflowMenuState,
                                                 onDismiss = onPageOverflowDismiss,
-                                                items = overflowMenuState?.items.orEmpty(),
-                                                offset = offset
+                                                items = overflowMenuState?.items.orEmpty()
                                             )
                                         },
                                         onPageOverflowClick = { pageSummary, index ->
--- a/app/src/main/java/org/wikipedia/feed/CommunityContentTab.kt
+++ b/app/src/main/java/org/wikipedia/feed/CommunityContentTab.kt
@@ -270,12 +270,13 @@ fun CommunityContentTab(
                                         },
                                         onFooterClick = { onCardFooterClick(card) },
                                         onCardImpression = { onCardImpression(card, cardIndex) },
-                                        pageOverflowContent = { index ->
+                                        pageOverflowContent = { index, offset ->
                                             PageOverflowMenu(
                                                 menuKey = "dyk-${card.date}-$index",
                                                 overflowMenuState = overflowMenuState,
                                                 onDismiss = onPageOverflowDismiss,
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/wikipedia/feed/didyouknow/DidYouKnowActivity.kt
```diff
@@ -44,6 +59,52 @@ class DidYouKnowActivity : BaseActivity() {
...
+                            context = this,
+                            wikiSite = wikiSite,
+                            pageSummary = pageSummary,
+                            source = source,
+                            menuKey = menuKey,
+                            onOpenPage = { entry ->
+                                startActivity(PageActivity.newIntentForCurrentTab(this, entry, entry.title))
+                            },
+                            onOpenInNewTab = { entry ->
+                                startActivity(PageActivity.newIntentForNewTab(this, entry, entry.title))
```
Williamrai, line 78: Do we need instrumentation for the overflow actions?

app/src/test/java/org/wikipedia/util/StringUtilTest.kt
```diff
@@ -92,6 +92,10 @@ class StringUtilTest {
         assertEquals("te_st", StringUtil.removeHTMLTags("<tag>te_st</tag>"))
     }
 
+    fun testRemoveBoldTags() {
+        assertEquals("Lorem <i>ipsum</i>", StringUtil.removeStyleTags("Lorem <b data=\"123\"><i>ipsum</i></b>"))
```
Williamrai: The test is missing `@Test` annotation and it is also using incorrect function.

dbrant: whoops, done!

app/src/main/java/org/wikipedia/feed/didyouknow/DidYouKnowActivity.kt
dbrant: No, not for this screen. And anyway, the resulting actions are all "implicitly" instrumented. (e.g. opening an article will produce a `link_preview_interaction` event with the appropriate HistoryEntry source.)

Williamrai: okay, got it.

app/src/main/java/org/wikipedia/feed/CommunityContentTab.kt
```diff
@@ -269,7 +269,18 @@ fun CommunityContentTab(
...
+                                            PageOverflowMenu(
+                                                menuKey = "dyk-${card.date}-$index",
+                                                overflowMenuState = overflowMenuState,
+                                                onDismiss = onPageOverflowDismiss,
+                                                items = overflowMenuState?.items.orEmpty()
+                                            )
+                                        },
+                                        onPageOverflowClick = { pageSummary, index ->
+                                            onPageOverflowClick(card, pageSummary, HistoryEntry.SOURCE_FEED_MOST_READ, "dyk-${card.date}-$index")
+                                        }
```
Williamrai, line 283: this should be **HistoryEntry.SOURCE_FEED_DID_YOU_KNOW**

dbrant: yes!! good catch.

PR #6673, opened 2026-06-10
Home feed: Add tooltip explaining recent updates.

dbrant:
This shows a tooltip that tells the user about our new additions to the feed, but _only_ if the user has already been using the feed (`exploreFeedVisitCount > 0`).

https://phabricator.wikimedia.org/T428681


Williamrai:
Just confirming my understanding: the logic relies on `Prefs.exploreFeedVisitCount`, which is already > 0 for existing users since the `old FeedFragment` incremented it. So those users will see the tooltip, while new users will have count 0 and they will never see it. Is that right?

dbrant:
> Just confirming my understanding: the logic relies on `Prefs.exploreFeedVisitCount`, which is already > 0 for existing users since the `old FeedFragment` incremented it. So those users will see the tooltip, while new users will have count 0 and they will never see it. Is that right?

Yes, exactly.

PR #6678, opened 2026-06-14
Fix stale lead image / wrong-refresh after a superseded page load T395597

motiz88:
Claude-authored attempt to fix [T395597](https://phabricator.wikimedia.org/T395597) building on #6659. Claude's description follows.

---

#6659 cancels the previous page-load Job, but Kotlin cancellation is cooperative and only takes effect at a suspension point. When a superseded load's requests were served from cache -- e.g. the article that was open before the user followed an external link to a new one -- its coroutine resumes past the already-complete await()s without ever observing the cancellation and runs createPageModel() anyway. Because createPageModel() builds the page as Page(title = model.title, pageProperties = thisResponse), the stale response's lead image is fused with the current article's title: the new article's text ends up under the previous article's lead image, and pull-to-refresh reloads the old article (it reads the reverted model.title).

Guard the commit so a superseded load is a no-op: after the awaits, bail out if the job is no longer active or model.title is no longer the title this load captured. Also clear the lead image at the start of loadPage() so a previous article's image can't linger under the new article while its metadata loads.

Add PageLoadSupersessionTest, which reproduces the interleaving deterministically (via a manually-pumped dispatcher, no new test dependencies): without the guard the model ends torn, with it the current article stays intact.

### Why is this needed?

**Phabricator:**
[T395597](https://phabricator.wikimedia.org/T395597)

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
+++ b/app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
@@ -6,6 +6,7 @@ import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.async
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import org.wikipedia.R
 import org.wikipedia.WikipediaApp
@@ -188,6 +189,20 @@ class PageFragmentLoadState(private var model: PageViewModel,
             val pageSummaryResponse = pageSummaryRequest.await()
             val watchedResponse = watchedRequest.await()
             val categoriesResponse = categoriesRequest.await()
+
+            // A newer page load may have superseded this one while we were awaiting the
+            // network. Cancellation is cooperative, so a load whose requests were served
+            // from cache (e.g. the article that was already open before a new one was
+            // launched via an external link) can resume past these await()s without ever
+            // observing the cancel() in pageLoad(). If we commit anyway, createPageModel()
+            // fuses this stale response's lead image with the current model.title, leaving
+            // the new article's text under the previous article's lead image (and a refresh
+            // that reloads the wrong page). Re-check that this load is still the current one
+            // before mutating any shared state. (T395597)
+            if (!isActive || model.title !== title) {
+                return@launch
+            }
+
             val isWatched = watchedResponse.query?.firstPage()?.watched == true
             val hasWatchlistExpiry = watchedResponse.query?.firstPage()?.hasWatchlistExpiry() == true
             if (pageSummaryResponse.body() == null) {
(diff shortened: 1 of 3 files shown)


app/src/test/java/org/wikipedia/page/PageLoadSupersessionTest.kt
```diff
@@ -0,0 +1,139 @@
...
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+import org.junit.Assert.assertSame
+import org.junit.Test
+import java.util.ArrayDeque
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Regression test for T395597 ("text of a newly-opened article rendered under the lead
+ * image of a previously-open article; pull-to-refresh reloads the wrong article").
```
dbrant, line 16: This test is a very inadequate representation of the actual logic in `PageFragmentLoadState`.

Retrofit wraps all calls in a `suspendCancellableCoroutine` block, which will respond to cancellation whether it was served from cache or from the real network.

And, crucially, this test does not `.cancel()` the previous Job before launching the new Job, which is exactly the fix originally made in #6659.

If `summaryA` and `summaryB` were both cancellable blocks instead of a plain `CompletableDeferred`, the existing logic would work as expected, without any changes needed.

app/src/main/java/org/wikipedia/page/PageFragmentLoadState.kt
```diff
@@ -188,6 +189,20 @@ class PageFragmentLoadState(private var model: PageViewModel,
             val pageSummaryResponse = pageSummaryRequest.await()
             val watchedResponse = watchedRequest.await()
             val categoriesResponse = categoriesRequest.await()
+
+            // A newer page load may have superseded this one while we were awaiting the
+            // network. Cancellation is cooperative, so a load whose requests were served
+            // from cache (e.g. the article that was already open before a new one was
+            // launched via an external link) can resume past these await()s without ever
```
dbrant, line 196: I'm afraid this doesn't make sense. Retrofit enqueues its calls inside a `suspendCancellableCoroutine` block, which means it will respond to cancellation regardless of whether it returned the content from cache or from the real network. If there was a _chance_ that a Retrofit call ignores cancellation, then we would need to check for `isActive` after every network call, which would be extremely inconvenient.

dbrant (changes requested):
Thanks @motiz88, this feels like a swing-and-a-miss for Claude.

To be very clear: Are you still able to reproduce the issue even after the fix made in #6659? If so, could you share the precise steps to reproduce it?

See comments for additional notes.

motiz88:
Yeah fair enough - I'm happy to try to pin down repro steps if it helps. (Also just realised that the last Play Store publish was in May so #6659 isn't included in it.)

dbrant:
With cautious optimism, all evidence suggests that the issue is resolved.

PR #6679, opened 2026-06-15
Use content extension

Isira-Seneviratne:
### What does this do?
This updates fragments using Jetpack Compose to use the [`content`](https://developer.android.com/reference/kotlin/androidx/fragment/compose/package-summary#(androidx.fragment.app.Fragment).content(kotlin.Function0)) extension.

### Why is this needed?
The `content` extension removes the need to manually create a `ComposeView`, and automatically handles setting up a fragment-appropriate configuration.

**Phabricator:**


dbrant:
Closing as unnecessary churn.

PR #6687, opened 2026-06-17
Home feed: Default view preference.

dbrant:
https://phabricator.wikimedia.org/T426876

### DIFF
--- a/app/src/main/java/org/wikipedia/settings/homefeed/HomeFeedDefaultViewScreen.kt
+++ b/app/src/main/java/org/wikipedia/settings/homefeed/HomeFeedDefaultViewScreen.kt
@@ -14,26 +14,30 @@ import androidx.compose.material3.RadioButtonDefaults
 import androidx.compose.material3.Scaffold
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.Alignment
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/wikipedia/settings/homefeed/HomeFeedDefaultViewScreen.kt
```diff
@@ -0,0 +1,144 @@
...
+                        style = MaterialTheme.typography.bodyLarge,
+                        color = WikipediaTheme.colors.primaryColor
+                    )
+                    Text(
+                        modifier = Modifier.padding(top = 4.dp),
+                        text = stringResource(R.string.explore_feed_preference_personalized_content_title),
+                        style = MaterialTheme.typography.bodyMedium,
+                        color = WikipediaTheme.colors.primaryColor
+                    )
+                }
```
Williamrai, line 120: Text color for both description 
```
text = stringResource(R.string.explore_feed_preference_community_content_title)
and
text = stringResource(R.string.explore_feed_preference_personalized_content_title),
```
should be secondaryColor.

dbrant: done!

PR #6695, opened 2026-06-21
feat: assessments in app

Stevekk11:
### What does this do?
This feature add the assessment label to articles which have it.

### Why is this needed?
For now, you have to open the browser to see the page's assessment, with this you can see it directly in the app.

**Phabricator:**


dbrant:
Thanks--
This is discussed at length in T75299 and T303650. Article assessments are currently not shown in mobile web (Minerva skin) and the mobile apps as a design choice.
If we will display article assessments in the future, it will be done through our [content service](https://gerrit.wikimedia.org/r/admin/repos/mediawiki/services/mobileapps,general), not through client-side API calls.

PR #6698, opened 2026-06-22
Games teaser module

Williamrai:
### What does this do?
- adds Games Teaser Module to For You feed

**Phabricator:**
https://phabricator.wikimedia.org/T419756


app/src/main/java/org/wikipedia/feed/model/WikiGameCard.kt
```diff
@@ -0,0 +1,11 @@
+package org.wikipedia.feed.model
+
+import kotlinx.serialization.Serializable
+import org.wikipedia.feed.wikigames.WikiGame
+@Serializable
+class WikiGameCard(
+    val wikiGame: WikiGame,
+    val date: String
+) : ForYouCard() {
+    override fun dismissHashCode(): Int = 31 * wikiGame.game.ordinal + date.hashCode()
```
dbrant: This is not quite hashy enough: The `ordinal` here is zero, so the entire hash will become `date.hashCode()`.
Maybe we could use `wikiGame.game.name.hashCode` instead of the ordinal?

app/src/main/java/org/wikipedia/feed/wikigames/GamesModule.kt
```diff
@@ -0,0 +1,291 @@
...
+    onCustomizeInterestsClick: (card: Card) -> Unit
+) {
+    val context = LocalContext.current
+
+    ForYouModulePager(
+        modifier = modifier,
+        module = module,
+        onCardInView = onCardInView
+    ) { page ->
+        // space for the floating pager dots
```
dbrant: A good rule to follow is: If you need a comment to explain the purpose of a variable, then why not give the variable a name that explains itself?
Something like `spacingForPagerDots`?

app/src/main/java/org/wikipedia/feed/wikigames/GamesModule.kt
```diff
@@ -0,0 +1,291 @@
...
+        val bottomSpacing = if (module.cards.size > 1) 40.dp else 16.dp
+        when (val card = module.cards[page]) {
+            is WikiGameCard -> when (val game = card.wikiGame) {
+                is WikiGame.OnThisDayGame -> OnThisDayGameModuleCard(
```
dbrant, line 92: When side-scrolling these cards, they look a little weird next to each other:
They have no spacing between each other, but also they are cut-off by padding on the left and right, which feels very different from how other modules scroll.

Image

dbrant (changes requested):

app/src/main/java/org/wikipedia/feed/model/WikiGameCard.kt
Williamrai: done.

app/src/main/java/org/wikipedia/feed/wikigames/GamesModule.kt
Williamrai: done.

app/src/main/java/org/wikipedia/feed/wikigames/GamesModule.kt
Williamrai: done.

PR #6704, opened 2026-06-27
PageTitle: Properly encode special characters in page URLs

Zaidusyy:
PageTitle: Properly encode special characters in page URLs

Question marks in file titles were not escaped when constructing page URLs, causing links to Commons file pages to fail.

Replace URLEncoder with Uri.encode() when building page URLs and add a regression test for file titles containing question marks.

Bug: T430056
Change-Id: I55cff23e5c70af7b06384233f61f779257a61f65

### What does this do?
This commit replaces the usage of `java.net.URLEncoder.encode()` with `android.net.Uri.encode(..., ":/")` when constructing page URLs in `PageTitle.java`. It also adds a regression test to `PageTitleTest.java` to ensure correct encoding of filenames containing question marks.

### Why is this needed?
When constructing MediaWiki URLs, `URLEncoder` aggressively escapes structural characters like the namespace colon (`:`) into `%3A`. This forces the MediaWiki server to perform a 301 Redirect to the canonical, unescaped path. 

During this redirect, actual query delimiters like question marks (`?`) inside filenames are incorrectly unescaped, causing the link to break because the browser treats everything after the `?` as a query parameter rather than part of the filename. 

By using `Uri.encode(..., ":/")`, the colons and slashes are preserved, bypassing the redirect loop entirely while successfully encoding the question mark to `%3F`.

**Phabricator:**
https://phabricator.wikimedia.org/T430056

### DIFF
--- a/app/src/main/java/org/wikipedia/page/PageTitle.java
+++ b/app/src/main/java/org/wikipedia/page/PageTitle.java
@@ -325,17 +325,13 @@ public class PageTitle implements Parcelable {
     }
 
     private String getUriForDomain(String domain) {
-        try {
-            return String.format(
-                    "%1$s://%2$s/wiki/%3$s%4$s",
-                    getWikiSite().scheme(),
-                    domain,
-                    URLEncoder.encode(getPrefixedText(), "utf-8"),
-                    (this.fragment != null && this.fragment.length() > 0) ? ("#" + this.fragment) : ""
-            );
-        } catch (UnsupportedEncodingException e) {
-            throw new RuntimeException(e);
-        }
+        return String.format(
+                "%1$s://%2$s/wiki/%3$s%4$s",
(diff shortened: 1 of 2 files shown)


dbrant:
Woah, I don't really understand how this pull request is physically possible. We no longer have any `.java` files anywhere in our repo; everything was converted to Kotlin years ago. Are you (or your AI agent) a time traveler?

Zaidusyy:
oh wow, I am so embarrassed 🤦‍♂️ 
I recently saw this bug and started hacking on it using an extremely old local clone of the repo I had sitting on my machine I completely forgot to fetch the latest upstream changes before diving in so I was literally working in the past

Let me sync up with the modern Kotlin codebase closing this PR and submitting a fresh one. Sorry for the confusion

PR #6706, opened 2026-06-29
Home Feed: make Bookmarks properly reactive.

dbrant:
This introduces a proper Flow that enables the feed to recompose when articles are added/removed from reading lists, and update the state appropriately.


app/src/main/java/org/wikipedia/feed/ForYouContentTab.kt
Williamrai, line 1: `savedInReadingListTitles` is only passed in **BasedOnInterestMoudle, ContinueReadingModule and BecauseYouReadModule**, its not passed to `PlacesOfInterestArticlesModule, DiscoverArticlesModule and RandomModule.`

app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
```diff
@@ -320,6 +321,25 @@ class HomeViewModel : ViewModel() {
...
+            else AppDatabase.instance.readingListPageDao()
+                .observeSavedApiTitles(
+                    wikiSite.value.languageCode,
+                    titles,
+                    ReadingListPage.STATUS_QUEUE_FOR_DELETE
+                )
+                .map { it.toSet() }
+        }
+        .stateIn(viewModelScope, SharingStarted.WhileSubscribed(MAX_STOP_TIMEOUT_MILLIS), emptySet())
+
```
Williamrai, line 403: I think we should also add the `.catch` like other modules to handle any unexpected errors.

Williamrai (commented):
Refreshing on Community Tab removes the bookmark UI state. [Demo](https://github.com/user-attachments/assets/127c804d-73c6-45a3-9178-18697ff83c57)

PR #6708, opened 2026-06-30
T427507: Standardize feed requests to use UTC dates.

bosankus:
### What does this do?
* Update `HomePreferenceRepository`, `HomeViewModel`, and `OnThisDayViewModel` to use `DateUtil.getUtcRequestDateFor()` for feed content fetching.
* Simplify API parameter formatting by using `UtcDate` properties for year, month, and day.
* Add a `toString()` implementation to the `UtcDate` class for easier debugging.

### Why is this needed?
Standardizing feed requests to use UTC dates ensures consistent and predictable date handling across the app, regardless of device timezone settings. This prevents bugs related to timezone-dependent formatting and makes the codebase more maintainable by centralizing date logic through DateUtil.getUtcRequestDateFor(). By using the UtcDate properties (year, month, day), we simplify API parameter formatting and reduce the chance of date-related bugs in feed content fetching.

**Phabricator:**
https://phabricator.wikimedia.org/T427507

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
+++ b/app/src/main/java/org/wikipedia/feed/HomeViewModel.kt
@@ -423,7 +423,6 @@ class HomeViewModel : ViewModel() {
             val content = ServiceFactory.getRest(wikiSite.value)
                 .getFeedFeatured(date.year, date.month, date.day, wikiSite.value.languageCode)
 
-
             // Construct Card objects based on the day's content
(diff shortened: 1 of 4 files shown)


dbrant:
This fixes one issue, but introduces a different one. This changes the Home feed to use the current UTC date/time, which will make users see the wrong date's contents for a fraction of their day, depending on their UTC offset.
The proper root cause fix for this issue is on the server side, to allow the service to accept a UTC offset as a parameter

PR #6709, opened 2026-06-30
Follow-up to Games module, and some cleanup.

dbrant:
See inline notes.

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/HomeFragment.kt
+++ b/app/src/main/java/org/wikipedia/feed/HomeFragment.kt
@@ -319,7 +319,7 @@ class HomeFragment : Fragment() {
                                         is OnThisDayCardGameState.InProgress -> "continue_click"
                                         is OnThisDayCardGameState.Completed -> "review_click"
                                     }
-                                    instrument.submitInteraction("click", actionSource = wikiGame.cardName, elementId = elementId, actionContext = mapOf("game" to wikiGame.game.name))
+                                    instrument.submitInteraction("click", actionSource = WikiGameCard::class.java.simpleName, elementId = elementId, actionContext = mapOf("game" to wikiGame.game.name))
                                     requireActivity().startActivity(OnThisDayGameActivity.newIntent(
                                         context = requireContext(),
                                         invokeSource = InvokeSource.FEED,
@@ -401,10 +401,9 @@ class HomeFragment : Fragment() {
 
     private fun onCardImpression(card: Card, index: Int) {
         if (cardImpressions.add(card.hideKey)) {
-            val actionSource = if (card is WikiGameCard) card.wikiGame.cardName else card.javaClass.simpleName
             instrument.submitInteraction(
                 "impression",
-                actionSource = actionSource,
+                actionSource = card.javaClass.simpleName,
                 actionContext = mapOf("card_index" to index)
             )
         }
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/wikipedia/feed/discover/DiscoverModule.kt
```diff
@@ -52,13 +52,13 @@ import org.wikipedia.history.HistoryEntry
 import org.wikipedia.theme.Theme
 
 private val discoverPromptImageUrls = listOf(
-    "https://upload.wikimedia.org/wikipedia/commons/thumb/b/be/Red_eyed_tree_frog_edit2.jpg/960px-Red_eyed_tree_frog_edit2.jpg",
```
dbrant, line 55: These image URLs are wayyy too high-resolution for what we need here. We can easily reduce these to 500px, and still have no loss of perceivable quality, and save a lot of bandwidth.

app/src/main/java/org/wikipedia/feed/wikigames/GamesModule.kt
```diff
@@ -60,10 +60,10 @@ import org.wikipedia.theme.Theme
 import org.wikipedia.views.imageservice.ImageService
 
 private val promptImageUrls = listOf(
-    "https://upload.wikimedia.org/wikipedia/commons/thumb/b/be/Red_eyed_tree_frog_edit2.jpg/960px-Red_eyed_tree_frog_edit2.jpg",
```
dbrant, line 63: Same, way too high-res images.

app/src/main/java/org/wikipedia/feed/wikigames/WikiGame.kt
```diff
@@ -1,20 +1,14 @@
 package org.wikipedia.feed.wikigames
 
-import kotlinx.serialization.Serializable
 import org.wikipedia.feed.onthisday.OnThisDay
 import org.wikipedia.games.WikiGames
 
 // Represents all game types that can appear in the Explore feed and GamesHub.
 // This can be extended whenever a new Wiki game is introduced.
-@Serializable
```
dbrant, line 9: None of these things actually need to be Serializable, because of the way this module is loaded. (this module does not get cached in Prefs like other modules do.)

app/src/main/java/org/wikipedia/feed/wikigames/WikiGame.kt
```diff
@@ -1,20 +1,14 @@
...
 sealed class WikiGame {
     // Stable identity of the game type, used for hide keys
     abstract val game: WikiGames
 
-    // per game analytics source name, e.g. "OnThisDayGameCard"
-    val cardName: String get() = "${this.javaClass.simpleName}Card"
```
dbrant, line 15: To keep things simple, let's not have analytics-specific fields in these classes yet. Maybe if/when we have more than one game, we can revisit this.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -319,7 +319,7 @@ class HomeFragment : Fragment() {
                                         is OnThisDayCardGameState.InProgress -> "continue_click"
                                         is OnThisDayCardGameState.Completed -> "review_click"
                                     }
-                                    instrument.submitInteraction("click", actionSource = wikiGame.cardName, elementId = elementId, actionContext = mapOf("game" to wikiGame.game.name))
```
dbrant, line 322: This can be the `simpleName` of the Card, just like all other interactions, since we encode the game name in `actionContext`.

Williamrai (commented):
Thank you for the follow up PR. Would you also be able to update PlacesOfInterestModule to use 500px images?

dbrant:
> Thank you for the follow up PR. Would you also be able to update PlacesOfInterestModule to use 500px images?

ah yes, done!

PR #6710, opened 2026-06-30
Add snackbar when selecting Default Feed View.

dbrant:
Per Slack conversation, we would like to add a confirmation Snackbar when the user changes their Default Feed View selection.


app/src/main/java/org/wikipedia/settings/homefeed/HomeFeedDefaultViewScreen.kt
Williamrai, line 1: Just confirming, we'll only show this snackbar once, right? Not every time the user selects it.

dbrant: yep, just once.

PR #6715, opened 2026-07-01
Survey post explore feed revamp

Williamrai:
### What does this do?
- adds a feedback survey for home feed

[Instrumentation Deck](https://docs.google.com/presentation/d/1OSd5skTwzF5ZYp6-ZNTE50C6LKhNgqd8IsdCxxAZ-ak/edit?slide=id.g3e86f03052a_0_34#slide=id.g3e86f03052a_0_34)

**Phabricator:**
https://phabricator.wikimedia.org/T420939


app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -408,4 +411,51 @@ class HomeFragment : Fragment() {
...
+                )
+            },
+            onSubmit = { feedbackOption, feedbackText ->
+                instrument.submitInteraction(
+                    action = "click",
+                    actionSource = "home_feed_feedback",
+                    elementId = "feedback_submit",
+                    actionContext = buildMap {
+                        feedbackOption?.let { put("feedback_select", it) }
+                        put("feedback_text", feedbackText)
```
dbrant: Since we're really tight on space in `actionContext`, can we establish a convention of having very short names of parameters here? For example, `feedback_select` could be called `score`, and `feedback_text` could be `text`.

Williamrai: Yes, that makes sense. I'll make the change and notify shay about this change as well.

PR #6717, opened 2026-07-02
Transition to Random widget 

Williamrai:
### What does this do?
- adds logic, UI and update code to transition to random widget at the end of reading challenge 

**Phabricator:**
https://phabricator.wikimedia.org/T430859


app/src/main/java/org/wikipedia/widgets/readingchallenge/ReadingChallengeAnalyticsHelper.kt
```diff
@@ -27,4 +28,11 @@ object ReadingChallengeAnalyticsHelper {
             actionContext = if (includeStreak) mapOf("streak_count" to Prefs.readingChallengeStreak) else null
         )
     }
+
+    private fun logRandomArticleHeartbeat() {
```
dbrant, line 32: Do we really need this heartbeat event? When this widget becomes a purely static Random Article widget, it will no longer be updated at regular intervals (right?) so heartbeat events will never fire anyway.

Williamrai: I added this because we [will still have a scheduler](https://phabricator.wikimedia.org/T429850#:~:text=Nice%20to%20have%3A-,Keep%20the%20daily%20updating%20of%20background%20color%20%26%20baby%20globe,-%5Btext%20does%20not) which runs and that updates just the widget colors. But yes we can remove it.

PR #6718, opened 2026-07-06
Support hCaptcha for DiscussionTools talk page edits

kostajh:
Why:
- Talk replies and new topics via discussiontoolsedit fail with a false success (revid 0) when the server requires an hCaptcha challenge

What:
- Parse the captcha nested in the discussiontoolsedit response
- Obtain a token with HCaptchaHelper and resubmit it via the captchaword parameter
- Mint the token against the server-returned site key (captcha.key), the always-challenge key when a challenge is forced (e.g. AbuseFilter)
- Echo wgConfirmEditForceShowCaptcha on the resubmit of a forced challenge
- Error on unhandled captcha types instead of reporting a false success
- Show the hcaptcha-privacy-policy disclaimer on the edit preview, mirroring account creation

Assisted-by: Claude Opus 4.8
Bug: T427936


app/src/main/java/org/wikipedia/captcha/HCaptchaHelper.kt
```diff
@@ -36,14 +36,15 @@ class HCaptchaHelper(
...
+    fun show(siteKey: String? = null) {
         if (hCaptcha == null) {
             val config = RemoteConfig.config.androidv1?.hCaptcha ?: configDefault
             hCaptcha = HCaptcha.getClient(activity)
             hCaptcha?.setup(
                 HCaptchaConfig.builder()
                     .theme(if (WikipediaApp.instance.currentTheme.isDark) HCaptchaTheme.DARK else HCaptchaTheme.LIGHT)
-                    .siteKey(config.siteKey)
+                    // A token is minted against a specific site key, so prefer the server's.
+                    .siteKey(siteKey?.ifEmpty { null } ?: config.siteKey)
```
dbrant, line 46: If we are preferring the sitekey given by the server in this case, should we also be preferring the sitekey given by the server in the case of account creation, too?

app/src/main/java/org/wikipedia/dataclient/discussiontools/DiscussionToolsEditResponse.kt
```diff
@@ -13,6 +13,24 @@ class DiscussionToolsEditResponse : MwResponse() {
             val result: String = "",
             val content: String = "",
             @SerialName("newrevid") val newRevId: Long = 0,
-            val watched: Boolean = false
-    )
+            val watched: Boolean = false,
+            val edit: EditFailure? = null
```
cooltey, line 17: It looks like the `EditFailure` does not only contain `Captcha` but also `result` with its status (`Failure` in the test `json file`).

Would it be possible to change `EditFailure` to a more precise class name?

app/src/main/java/org/wikipedia/edit/preview/EditPreviewFragment.kt
```diff
@@ -170,6 +172,15 @@ class EditPreviewFragment : Fragment(), CommunicationBridgeListener, ReferenceDi
         requireActivity().invalidateOptionsMenu()
     }
 
+    fun showHCaptchaDisclaimer(text: CharSequence) {
+        val binding = _binding ?: return
```
cooltey, line 176: This can be removed.

app/src/main/java/org/wikipedia/talk/TalkReplyViewModel.kt
```diff
@@ -69,20 +92,50 @@ class TalkReplyViewModel(savedStateHandle: SavedStateHandle) : ViewModel() {
...
+        return try {
+            val message = "hcaptcha-privacy-policy"
+            val content = ServiceFactory.get(pageTitle.wikiSite).getMessages(message, null)
+                .query?.allmessages?.find { it.name == message }?.content.orEmpty()
+            StringUtil.parseWikitextExternalLinks(content)
+        } catch (e: Exception) {
+            L.e(e)
+            ""
+        }
+    }
```
cooltey, line 135: It looks like we already have a similar process in `CreateAccountActivityViewModel`. Wondering if we can consolidate them and put it to `HCaptchaHelper`.

PR #6722, opened 2026-07-06
fix: getStatusBarInsets return type Insets? - Insets; simplify call site

tushar743-ui:
### What does this do?
Changes the return type of `getStatusBarInsets()` in MainActivity from Insets? to Insets, and simplifies the call site in `HomeScreen.kt` by removing the unnecessary null-safe operator and fallback (?.top ?: 0 -> .top).

### Why is this needed?

The backing field `statusBarInsets` is initialized to Insets.NONE and only ever assigned from `insetsCompat.getInsets(...)`, both of which are guaranteed non-null. The nullable return type was incorrect and forced callers to handle a null case that can never occur.

**Phabricator:**

### DIFF
--- a/app/src/main/java/org/wikipedia/feed/HomeScreen.kt
+++ b/app/src/main/java/org/wikipedia/feed/HomeScreen.kt
@@ -114,7 +114,7 @@ fun HomeScreen(
 ) {
     val context = LocalContext.current
     val topInset = if (context is MainActivity) {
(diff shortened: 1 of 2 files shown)


dbrant:
Sorry, but this pull request doesn't fix any issues or bugs (even though the title has the word "fix" in it). It removes a single lint warning at best, which is not substantial enough for a PR.

PR #6734, opened 2026-07-10
[Reading lists] migration to compose

Williamrai:
### What does this do?
- migrates ReadingListsFragment to Jetpack Compose
- introduces reactive pipeline by adding ROOM flow and combine 

**Phabricator:**
https://phabricator.wikimedia.org/T431575
https://phabricator.wikimedia.org/T430376

**Test file**
[reading-lists-95.json](https://github.com/user-attachments/files/30064405/reading-lists-95.json)


dbrant:
....I have almost zero notes! Except one thing:
When sorting All Articles by "Date added (newest / oldest)" it seems to be backwards (newest items should be on top). Can we flip it?

PR #6750, opened 2026-07-21
Add banner url params

voyagerfan:
### What does this do?
This PR:

- Updates the CampaignCollection object to replace `appVersion` and  `appInstallId` values with the appropriate data.
- Updates the CampaignCollection testing to ensure these values are replaced in the URL string.

### Why is this needed?
The AppsCampaignConfig.json file will be updated to include key/value pairs for the the app's install ID and version. The app will need to replace these values.

**Phabricator:**
https://phabricator.wikimedia.org/T415791
https://phabricator.wikimedia.org/T432388


noorizanbintiabdulmalik85 (changes requested):
Personal it's private and Not shares

PR #6751, opened 2026-07-22
Wire up revisionId to EditAttemptStep instrumentation.

dbrant:
In our EditAttempStep `saveSuccess` events, we need to log the new revision ID of the edit.
Fortunately the revision ID is readily available in all the places this call is made, so it just needs to be plugged in.

https://phabricator.wikimedia.org/T432664


Williamrai (commented):
This looks good. While testing I also tested for **logSaveFailure** but that was rejected and I got this error. 
`"errorsText":"'.event.revision_id' should be integer"}`
**Full error stack trace:** https://gist.github.com/Williamrai/e4faaf7aa3ae8bd8447f97545e105bde

dbrant:
> .event.revision_id' should be integer

whoops, fixed!

PR #6756, opened 2026-07-27
Allow choice to edit using VisualEditor or Source editor.

dbrant:
https://phabricator.wikimedia.org/T431812
https://phabricator.wikimedia.org/T432409
https://phabricator.wikimedia.org/T422153


app/src/main/java/org/wikipedia/edit/EditorChoiceDialog.kt
```diff
@@ -0,0 +1,261 @@
...
+                    Text(
+                        text = title,
+                        style = MaterialTheme.typography.bodyLarge,
+                        color = WikipediaTheme.colors.primaryColor
+                    )
+
+                    IconButton(
+                        onClick = { }, // TODO: wire up this click
+                        modifier = Modifier
+                            .size(15.dp)
```
cooltey: The size of this icon should be `20` based on the design.

app/src/main/java/org/wikipedia/edit/EditorChoiceDialog.kt
```diff
@@ -0,0 +1,261 @@
...
+                    ) {
+                        Icon(
+                            painter = painterResource(R.drawable.ic_open_in_new_black_24px),
+                            contentDescription = "icon",
+                        )
```
cooltey: The icon color should be in `primaryColor`.
```
tint = WikipediaTheme.colors.primaryColor
```

And, if you are giving the `contentDescription`, please give a meaningful description like "open in an external browser". Since this might be a non-clickable icon, we can just set `null` in this case.

cooltey (changes requested):
Hi @voyagerfan, not sure if this PR is ready for review or not; some code review comments below to keep things going. 

Another thing I noticed: if you have a `Firefox` browser installed and set as the default browser, the editing flow that opens the browser will not work on `Firefox`, and it will redirect back to the app immediately.

app/src/main/java/org/wikipedia/edit/EditorChoiceDialog.kt
voyagerfan: @cooltey nice catch. Not sure where I got 15 from but I've updated it. Also, I removed the IconButton for now. Reference: [ce1d0e0](https://github.com/wikimedia/apps-android-wikipedia/pull/6756/commits/ce1d0e06b758fc386d34bf12a36093cd0666a493)

app/src/main/java/org/wikipedia/edit/EditorChoiceDialog.kt
voyagerfan: Updated the tint and content description. Also since it might be non-clickable I removed the IconButton wrapper.
Reference:  [ce1d0e0](https://github.com/wikimedia/apps-android-wikipedia/pull/6756/commits/ce1d0e06b758fc386d34bf12a36093cd0666a493) and [39579c1](https://github.com/wikimedia/apps-android-wikipedia/pull/6756/commits/39579c11c2c21de30c8e4ef9a1fc2c4d40169ddd)

voyagerfan:
@cooltey I've the feature branch and it looks good.

cooltey (approved):

PR #6768, opened 2026-08-04
Harden authority checking when buiding cookie list.

dbrant:
https://phabricator.wikimedia.org/T433832


app/src/main/java/org/wikipedia/dataclient/SharedPreferenceCookieManager.kt
```diff
@@ -100,7 +100,7 @@ class SharedPreferenceCookieManager(
             val cookiesForDomainSpec = cookieJar[domainSpec]!!
             if (domain.endsWith(domainSpec)) {
                 buildCookieList(cookieList, cookiesForDomainSpec, null)
-            } else if (Service.isWikimediaAuthority(domainSpec)) {
+            } else if (Service.isWikimediaAuthority(domainSpec) && Service.isWikimediaAuthority(domain)) {
```
cooltey, line 103: Do we need to create a test case for this logic?

dbrant: Sure!

PR #6769, opened 2026-08-04
Add taxon IDs to static article topics.

dbrant:
Our list of Topics needs to include each topic's taxon IDs, because these taxons are the thing that is returned by the model classifier, and will now be included in the `mobile-html` payload. We will then use the taxon to identify the `topicId` to which it belongs.
(The topic ids, as defined [here](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/WikimediaMessages/+/refs/heads/master/includes/ArticleTopicFiltersRegistry.php), map onto the taxons, as defined [here](https://github.com/wikimedia/wikitax/tree/master/taxonomies/wikiproject/halfak_20191202/taxons).)

### DIFF
--- a/app/src/main/java/org/wikipedia/topics/ArticleTopics.kt
+++ b/app/src/main/java/org/wikipedia/topics/ArticleTopics.kt
@@ -5,7 +5,8 @@ import org.wikipedia.R
 data class ArticleTopic(
     val topicId: String,
     val msgKey: Int,
-    val queryTopicId: String
+    val queryTopicId: String,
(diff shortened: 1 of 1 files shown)


cooltey (approved):
LGTM. Wondering if we can write a script to update it automatically.

dbrant:
> LGTM. Wondering if we can write a script to update it automatically.

These topics are hardcoded across the ecosystem, and will likely not be updated for a very long time.

PR #6774, opened 2026-08-06
Adding action=`abort` events to editattemptstep for certain edit actions

cooltey:
### What does this do?
1. Adds `abort` action to `EditAttemptStepEvent`
2. Sends `EditAttemptStepEvent.logAbort(title)`  when dismiss `PopupMenu` manually.
3. Replaces `logSaveFailure` with `logAbort` in `EditSectionActivity` as requested.

**Phabricator:**
https://phabricator.wikimedia.org/T433909


app/src/main/java/org/wikipedia/edit/EditSectionActivity.kt
```diff
@@ -388,7 +388,7 @@ class EditSectionActivity : BaseActivity(), ThemeChooserDialog.Callback, EditPre
             binding.editSectionCaptchaContainer.visibility = View.VISIBLE
             captchaHandler.handleCaptcha(null, result)
         } else {
-            EditAttemptStepEvent.logSaveFailure(viewModel.pageTitle)
+            EditAttemptStepEvent.logAbort(viewModel.pageTitle)
```
voyagerfan: Hey @cooltey, I was wondering if it would make sense to move the logAbort call into `onEditFailure`? The `when` block at line 257 calls `onEditFailure` in the nested success case as well as the error case (line 280). I tried testing in airplane mode to simulate no connection, and no logAbort is occurring.

voyagerfan (commented):
quick question on logAbort placement

app/src/main/java/org/wikipedia/edit/EditSectionActivity.kt
cooltey: Nice catch! I revisited the deck and noticed slide `19` has been updated to use the original `logSaveFailure`, that means I'll need to keep the original place where we send `logSaveFailure` and add an additional `logAbort` for other failure messages.

voyagerfan: Oops...After reviewing slide 18 a little more carefully, it looks like the instrumentation needs to occur when we land on edit (after clicking the pencil icon). It looks like we should be logging in the fetching block near line 225.

Also this means we'd need to update for slide 19 when publishing.

cooltey: Sure and done.

PR #6795, opened 2026-08-14
EditAttemptStep updates for edit_count and wiki params 

voyagerfan:
### What does this do?
Updates wiki and edit_count for editattemptstep

### Why is this needed?
To instrument edit interactions

**Phabricator:**
https://phabricator.wikimedia.org/T433000


app/src/main/java/org/wikipedia/descriptions/DescriptionEditViewModel.kt
```diff
@@ -76,6 +78,17 @@ class DescriptionEditViewModel(savedStateHandle: SavedStateHandle) : ViewModel()
         }
     }
 
+    private fun loadUserTotalEdits(): Deferred<Int> = viewModelScope.async(CoroutineExceptionHandler { _, throwable ->
```
voyagerfan: This ViewModel was handled differently than the rest. I noticed we are already make a call to grab `globalUserInfo` so I decoupled it from `fun requestSuggestion` and wrapped it in a function to return a deferred object which can be accessed within the ViewModel and in the fragment.

cooltey: The app sends `EditAttemptStepEvent.logInit` in `onCreate` at `DescriptionEditFragment`, which is wrapped with a `lifecycleScope`, which is not ideal.

My suggestion is:
1. Create a `MutableStateFlow` in the ViewModel, and call the total edits function in `init {}`.
2. Have the state in the  `onViewCreated`, and send the log there. We can send it even if it sends an `Error`.

cooltey (changes requested):

app/src/main/java/org/wikipedia/descriptions/DescriptionEditViewModel.kt
voyagerfan: Done. [7996769](https://github.com/wikimedia/apps-android-wikipedia/pull/6795/commits/7996769474ba161d6a91f21376b4f0ec81c9019d)

app/src/main/java/org/wikipedia/descriptions/DescriptionEditFragment.kt
```diff
@@ -107,7 +107,6 @@ class DescriptionEditFragment : Fragment() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        EditAttemptStepEvent.logInit(viewModel.pageTitle, EditAttemptStepEvent.INTERFACE_OTHER)
     }
```
cooltey, line 110: This can be removed.

app/src/main/java/org/wikipedia/descriptions/DescriptionEditViewModel.kt
```diff
@@ -54,6 +54,12 @@ class DescriptionEditViewModel(savedStateHandle: SavedStateHandle) : ViewModel()
 
     private val _waitForRevisionState = MutableStateFlow(Resource<Boolean>())
     val waitForRevisionState = _waitForRevisionState.asStateFlow()
+    private val _editCount = MutableStateFlow(-1)
+    val editCount = _editCount.asStateFlow()
```
cooltey: For proper architecture here, please wrap the value with `Resource`:
`_editCount = MutableStateFlow(Resource<Int>())`

and add `viewModel.editCount .collect { }` into the `lifecycleScope` before actually sending the log.

app/src/main/java/org/wikipedia/page/PageFragment.kt
```diff
@@ -970,6 +972,12 @@ class PageFragment : Fragment(), BackPressedHandler, CommunicationBridge.Communi
             webView.visibility = View.VISIBLE
         }
 
+        lifecycleScope.launch(CoroutineExceptionHandler { _, t ->
+            L.e(t)
+        }) {
+            editCount = ServiceFactory.get(WikipediaApp.instance.wikiSite).getUserInfo().query?.userInfo?.editCount ?: 0
+        }
```
cooltey: The proper way of doing this is to update `pageLoad()` and check if any of the existing API requests is able to add `userinfo` and grab that `userInfo?.editCount` and send to `onPageMetadataLoaded`. Does this sound good to you?

cooltey (changes requested):

app/src/main/java/org/wikipedia/descriptions/DescriptionEditViewModel.kt
voyagerfan: Sounds good. I updated the DescriptionEditViewModel as well as the corresponding  .collect{} and lifeCycleScope changes here [1a96277](https://github.com/wikimedia/apps-android-wikipedia/pull/6795/commits/1a9627769e46f72cc9934408f8c240c540db2661)

app/src/main/java/org/wikipedia/page/PageFragment.kt
voyagerfan: Sounds good to me. 

- leveraged `pageLoad` here -> [a95b112](https://github.com/wikimedia/apps-android-wikipedia/pull/6795/commits/a95b112497065191a30a3ab812b73cb2288fa530)
- hooked into `onPageMetaDataLoaded -> [7fa9f92](https://github.com/wikimedia/apps-android-wikipedia/pull/6795/commits/7fa9f92cae81edb7ab6079be743d0f7850e6ab05)

cooltey:
Merge this for now. Will have a follow-up PR for this.

cooltey (approved):

PR #6797, opened 2026-08-14
TestKitchen: underpinnings for experiment exposure events.

dbrant:
https://phabricator.wikimedia.org/T434941
https://phabricator.wikimedia.org/T431034

### DIFF
--- a/app/src/main/java/org/wikipedia/analytics/testkitchen/TestKitchenAdapter.kt
+++ b/app/src/main/java/org/wikipedia/analytics/testkitchen/TestKitchenAdapter.kt
@@ -86,8 +86,7 @@ object TestKitchenAdapter : ClientDataCallback, EventSender {
     }
 
     fun submitExperimentExposure(abTest: ABTest) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/analytics/testkitchen/TestKitchenAdapter.kt
```diff
@@ -85,6 +85,11 @@ object TestKitchenAdapter : ClientDataCallback, EventSender {
         return ExperimentImpl(abTest.name, abTest.getGroupName(), subjectId = Prefs.appInstallId, isLoggable = { abTest.shouldInstrument() })
     }
 
+    fun submitExperimentExposure(abTest: ABTest) {
+        val experiment = ExperimentImpl(abTest.name, abTest.getGroupName(), subjectId = Prefs.appInstallId, isLoggable = { abTest.shouldInstrument() })
```
cooltey: This is identical to `getExperiment()`. Is it possible to reuse it?

dbrant: yes!!

PR #6798, opened 2026-08-14
Visual Editor: Instrumentation

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T434241


app/src/main/java/org/wikipedia/edit/EditorChoiceDialog.kt
```diff
@@ -88,8 +100,40 @@ fun showEditorChoiceDialog(
...
+                },
                 onConfirm = { editorChoice, dontShowAgain ->
+                    if (!isSettingsScreen) {
+                        instrument.submitInteraction(
+                            action = "click",
+                            actionSource = "edit_choice_select",
+                            elementId = "edit_choice_submit",
+                            actionContext = mapOf(
+                                "edit_choice" to if (editorChoice == EDITOR_CHOICE_VE) "visual" else "source",
+                                "is_default" to dontShowAgain
```
voyagerfan, line 127: @cooltey minor question. are we sure `is_default` maps to the `dontShowAgain` boolean?

cooltey: From the deck, you can see the checkbox is circled, and the message there is "This default can be changed later in Settings" -- I guess that's why it's named as `is_default`

voyagerfan: I've got your answer. thank you!

PR #6818, opened 2026-08-21
Update maybe later screen for Donation Reminder v3

voyagerfan:
### What does this do?
This PR updates the "maybe later" screen in the donation reminder flow.

### Why is this needed?
The current implementation of the "maybe later" screen does not align with the v3 designs and needs to be updated.

**Phabricator:**
https://phabricator.wikimedia.org/T435345


app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
```diff
@@ -332,131 +343,61 @@ fun DonationReminderContent(
...
+
+            DonationRemindersSwitch(
+                modifier = Modifier
+                    .noRippleClickable {
+                        viewModel.toggleDonationReminders(!isDonationReminderEnabled)
+                    }
+                    .padding(top = 24.dp),
+                isDonationRemindersEnabled = isDonationReminderEnabled,
+                onCheckedChange = { viewModel.toggleDonationReminders(it) }
+            )
```
cooltey: We should only show this toggle when coming from the Settings screen. Please keep the existing logic.

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
```diff
@@ -621,90 +638,79 @@ fun <T : Number> OptionSelector(
...
                         modifier = Modifier,
                         plainTooltipText = stringResource(R.string.donation_reminders_settings_tooltip_info_label)
                     )
                 }
+            }
+            Spacer(modifier = Modifier.width(16.dp))
+            Row(
+                verticalAlignment = Alignment.CenterVertically,
+                horizontalArrangement = Arrangement.spacedBy(16.dp)
+            ) {
```
cooltey: Is this still necessary?

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
```diff
@@ -621,90 +638,79 @@ fun <T : Number> OptionSelector(
...
-                                    onOptionSelected(option)
-                                    isDropdownExpanded = false
-                                }
+                            Button(
+                                colors = ButtonDefaults.buttonColors(
+                                    containerColor = if (isSelected) WikipediaTheme.colors.progressiveColor
+                                    else WikipediaTheme.colors.backgroundColor
+                                ),
+                                onClick = { onOptionSelected(currentOption, SelectedSource.Preset(index)) },
+                                modifier = Modifier.width(if (showArticleLabel) 75.dp else 100.dp)
```
cooltey: Maybe we can use `weight(1f)` with `Row` instead of `LazyRow` to have a better UI presentation, but we will lose the ability to scroll.

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderViewModel.kt
```diff
@@ -137,23 +146,26 @@ class DonationReminderViewModel(savedStateHandle: SavedStateHandle) : ViewModel(
             }
         }
 
-        val context = WikipediaApp.instance
         val presets = (donationConfig?.currencyAmountPresets[currencyCode]?.take(maxPresetItemsInDropdown) ?: listOf(minimumAmount)).toMutableSet()
         // TODO: remove this when experiment is completed in November 2026.
         DonationReminderHelper.updateDonationPresets(presets)
-        val options = presets.map {
+        val sortedPresets = presets.sorted().take(maxPresetItemsInDropdown)
+        val options = sortedPresets.map {
```
cooltey, line 157: I tested, but still not sure why we'll need to sort again here. Would you be able to check this again to see if it is still necessary?

cooltey (changes requested):
1. The Custom amount field should not show the error message immediately after tapping on it, maybe we can grab the logic from `CustomInputDialog`?
2. If coming from Settings and tapping the Custom amount but leaving it empty, we will save `0` as the amount, which is incorrect. We should keep the old amount in this situation. 
3. Should be `Whenever I read` instead of `When I read`
4. The icons beside `Whenever I read` and `Remind me to donate` should be in `inactive_color`
5. `CustomInputDialog` and `CustomInputDialogPreview` can be removed.

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderViewModel.kt
voyagerfan: I tried testing again too. This is what Im seeing.

With .sorted()
Screenshot 2026-08-27 at 8 46 08 AM


Without .sorted()
Screenshot 2026-08-27 at 8 46 25 AM

cooltey: I see. If that was the case, please remove the `sorted` from the `DonationReminderHelper.updateDonationPresets()`, and sort the presets in the viewModel.

```
val presets = (donationConfig?.currencyAmountPresets[currencyCode] ?: listOf(minimumAmount)).toMutableSet()
// TODO: remove this when experiment is completed in November 2026.
DonationReminderHelper.updateDonationPresets(presets)
val sortedPresets = presets.sorted().take(maxPresetItemsInDropdown)
```

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
voyagerfan: Added the guard back [901e3bb](https://github.com/wikimedia/apps-android-wikipedia/pull/6818/commits/901e3bbcaf538af8b6bc9ceb5c30ada7cfbe06ac)

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
voyagerfan: removed the lazy row [c7a0998](https://github.com/wikimedia/apps-android-wikipedia/pull/6818/commits/c7a09986118aa37c3e053618d29787725fe444fd)

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderViewModel.kt
voyagerfan: Done [fe271e6](https://github.com/wikimedia/apps-android-wikipedia/pull/6818/commits/fe271e67f1418623ff4172a3facc31998e53dd30)

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
voyagerfan: Good Catch...its definitely not needed. Done [ad99281](https://github.com/wikimedia/apps-android-wikipedia/pull/6818/commits/ad99281ce14ccc2c871cb655902f089e41c62e7b)

voyagerfan:
> 1. The Custom amount field should not show the error message immediately after tapping on it, maybe we can grab the logic from `CustomInputDialog`?

Ive added a variable `hasUserEditedCustomAmount` for this. We can use the logic from CustomInputDialog, but its different because the state wont survive screen rotation.

> 2. If coming from Settings and tapping the Custom amount but leaving it empty, we will save `0` as the amount, which is incorrect. We should keep the old amount in this situation. 

update -> `if (option.displayText.isBlank() && viewModel.isFromSettings) { }` line 435 in DonationReminderContent

> 3. Should be `Whenever I read` instead of `When I read`

Done [13da72d](https://github.com/wikimedia/apps-android-wikipedia/pull/6818/commits/13da72d95c50e7f424845dd68b350be4529e2148)

> 4. The icons beside `Whenever I read` and `Remind me to donate` should be in `inactive_color`

Done [135b814](https://github.com/wikimedia/apps-android-wikipedia/pull/6818/commits/135b814e60b2a9f9d614618fab3a73fd9d2c12f1)

> 5. `CustomInputDialog` and `CustomInputDialogPreview` can be removed.

forgot to do this but was using CustomDialog as a reference. Can remove in another push

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderViewModel.kt
```diff
@@ -20,7 +18,7 @@ import org.wikipedia.settings.Prefs
 import org.wikipedia.util.log.L
 
 class DonationReminderViewModel(savedStateHandle: SavedStateHandle) : ViewModel() {
-    private val preSelectedArticleFrequency = 15
+    private val preSelectedArticleFrequency = 10
```
cooltey: Can this be removed?

cooltey (commented):
I am still seeing incorrect behavior when I input an invalid amount - the toast message in Settings will show up with the invalid amount. 

Will create a follow-up PR for some final tuning.

PR #6821, opened 2026-08-21
[Instrumented Test] Espresso tests cleanup and refactor

Williamrai:
### What does this do?
- espresso old tests cleanup
- reworks the instrumented test setup
- adds initial Onboarding test 
- adds fixtures and live api tests for the personalization screen
  - fixtures test covers the screen: three onboarding pages, loading interests,  topic selection, home preference. 
  - in fixtures responses come from captured json, so a failure means something wrong with our code
  - live api test covers search only. Search is the one path where the test can enter its own input so it can assert on a specific article
  - the rest of the screen is random articles and the daily feed, so a live test there could only assert that something rendered which the fixtures test proves that already 
- adds unit tests for HomePreferenceRepository, InterestSelectionRepository and
  PersonalizationViewModel
- splits TestSuite into SmokeTestSuite and FixturesTestSuite, a workflow each,
  both manual for now

**Phabricator:**
https://phabricator.wikimedia.org/T434012


dbrant:
✔️  merging to keep things moving

PR #6822, opened 2026-08-21
DonationReminderV3: follow up reminder UI updates

cooltey:
### What does this do?
Minor design changes. We can change the string items after we make the final decision.

**Phabricator:**
https://phabricator.wikimedia.org/T435347


app/src/main/res/layout/view_donation_reminder_card.xml
```diff
@@ -40,17 +39,16 @@
                 android:id="@+id/positiveButton"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
-                style="@style/App.Button.Text"
+                style="@style/App.Button"
                 android:textSize="14sp"
```
voyagerfan: I think this is supposed to be 16.sp?

app/src/main/res/layout/view_donation_reminder_card.xml
```diff
@@ -40,17 +39,16 @@
...
                 tools:text="Lorem ipsum"/>
 
             <com.google.android.material.button.MaterialButton
                 android:id="@+id/negativeButton"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 style="@style/App.Button.Text"
-                android:textColor="?attr/primary_color"
+                android:textColor="?attr/placeholder_color"
                 android:textSize="14sp"
```
voyagerfan: I think this is also supposed to be 16.sp?

voyagerfan (commented):
Hi @cooltey, in addition to the text sizes, I think the text for the positive button is supposed be "Yes, donate now" instead of "Donate Now"

Reference: https://github.com/wikimedia/apps-android-wikipedia/blob/c1dd12d342374ea09aba4f51b3ea297a0ad167bc/app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.kt#L154

app/src/main/res/layout/view_donation_reminder_card.xml
cooltey: Yes! done.

app/src/main/res/layout/view_donation_reminder_card.xml
cooltey: Yes! done.

cooltey:
> Hi @cooltey, in addition to the text sizes, I think the text for the positive button is supposed be "Yes, donate now" instead of "Donate Now"
> 
> Reference:
> 
> https://github.com/wikimedia/apps-android-wikipedia/blob/c1dd12d342374ea09aba4f51b3ea297a0ad167bc/app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.kt#L154

Yes, I was waiting for the final decision from the PM about the title, and now they are updated.

app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.kt
```diff
@@ -150,8 +150,8 @@ class PageHeaderView(context: Context, attrs: AttributeSet? = null) : LinearLayo
             }
 
             val dateText = DateUtil.getMMMMdYYYY(Date(config.setupTimestamp))
-            val messageText = context.getString(R.string.donation_reminders_prompt_message, dateText, articleText, donationAmount)
-            val positiveButtonText = context.getString(R.string.donation_reminders_prompt_positive_button)
+            val messageText = context.getString(R.string.donation_reminders_prompt_message_v2, dateText, articleText, donationAmount)
```
voyagerfan, line 153: It looks like the word "articles" is duplicated in the output string. I think you can remove "articles" from `R.string.donation_reminders_prompt_message_v2` and let `articleText` on line 142 handle the plurals.

voyagerfan (commented):
last minor fix

app/src/main/java/org/wikipedia/page/leadimages/PageHeaderView.kt
cooltey: Done!

PR #6828, opened 2026-08-24
DonationReminderV3: update presets for the experiment

cooltey:
### What does this do?
We pull the donation presets from `MediaWiki:AppsDonationConfig.json`, but for this experiment, we will need to show smaller amounts to users in the test groups.

This PR hardcoded the list and replaced the presets with the desired ones.

### DIFF
--- a/app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderViewModel.kt
+++ b/app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderViewModel.kt
@@ -138,7 +138,9 @@ class DonationReminderViewModel(savedStateHandle: SavedStateHandle) : ViewModel(
         }
 
         val context = WikipediaApp.instance
-        val presets = donationConfig?.currencyAmountPresets[currencyCode]?.take(maxPresetItemsInDropdown) ?: listOf(minimumAmount)
+        val presets = (donationConfig?.currencyAmountPresets[currencyCode]?.take(maxPresetItemsInDropdown) ?: listOf(minimumAmount)).toMutableSet()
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderViewModel.kt
```diff
@@ -138,7 +138,9 @@ class DonationReminderViewModel(savedStateHandle: SavedStateHandle) : ViewModel(
         }
 
         val context = WikipediaApp.instance
-        val presets = donationConfig?.currencyAmountPresets[currencyCode]?.take(maxPresetItemsInDropdown) ?: listOf(minimumAmount)
+        val presets = (donationConfig?.currencyAmountPresets[currencyCode]?.take(maxPresetItemsInDropdown) ?: listOf(minimumAmount)).toMutableSet()
+        // TODO: remove this when experiment is completed in November 2026.
+        DonationReminderHelper.updateDonationPresets(presets)
         val options = presets.map {
```
voyagerfan, line 144: Should we consider sorting the `presets`? Since `updateDonationPresets()` removes and adds to a set, I dont think order is guaranteed anymore?

example - > `val options = presets.sorted().map { }`

voyagerfan (commented):
Minor question.

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderViewModel.kt
cooltey: That makes sense to me, updated!

PR #6830, opened 2026-08-25
DonationReminderV3: fixing glitch when dismiss the reminder card

cooltey:
### What does this do?
When dismissing the reminder card, the app reloads the article view and causes a glitch. This PR handles the layout height in a different way to avoid showing the glitch to users.

**Phabricator:**
https://phabricator.wikimedia.org/T410443


Williamrai (approved):

voyagerfan (commented):
looks good.

PR #6835, opened 2026-08-28
Support logging in with OAuth2.

dbrant:
A minimal prototype of the app logging in via OAuth2 (using our testing OAuth2 [consumer](https://meta.wikimedia.org/w/index.php?title=Special%3AOAuthListConsumers%2Fview%2F50ad79ffa34f64853c96b729e4aa5d8c) on metawiki).

https://phabricator.wikimedia.org/T395459


app/src/main/java/org/wikipedia/auth/OAuthClient.kt
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    companion object {
+        const val CLIENT_ID = "50ad79ffa34f64853c96b729e4aa5d8c"
+        const val REDIRECT_URI = "wikipedia://oauth/callback"
+        const val OAUTH_WIKI = "https://meta.wikimedia.org" // TODO: use current language wiki?
+        const val AUTHORIZATION_ENDPOINT = "/w/rest.php/oauth2/authorize"
+        const val TOKEN_ENDPOINT = "/w/rest.php/oauth2/access_token"
+        const val PROFILE_ENDPOINT = "/w/rest.php/oauth2/resource/profile"
+        const val LOGOUT_ENDPOINT = "/w/rest.php/oauth2/logout" // <--TODO?
```
tgr, line 173: `/w/api.php?action=logout&global=1` (we can add a REST API equivalent if you feel strongly about it)

app/src/main/java/org/wikipedia/auth/OAuthClient.kt
```diff
@@ -0,0 +1,175 @@
...
+        Prefs.tempAccountWelcomeShown = false
+        Prefs.tempAccountCreateDay = 0L
+        ReadingListSyncAdapter.manualSyncWithForce()
+        PollNotificationWorker.schedulePollNotificationJob(WikipediaApp.instance)
+        WikipediaFirebaseMessagingService.updateSubscription()
+    }
+
+    companion object {
+        const val CLIENT_ID = "50ad79ffa34f64853c96b729e4aa5d8c"
+        const val REDIRECT_URI = "wikipedia://oauth/callback"
```
tgr, line 168: Note that you need to use a HTTPS URI to be able to make use if [T423609](https://phabricator.wikimedia.org/T423609).

app/src/main/java/org/wikipedia/auth/OAuthClient.kt
```diff
@@ -0,0 +1,175 @@
...
+        const val OAUTH_WIKI = "https://meta.wikimedia.org" // TODO: use current language wiki?
```
tgr, line 169: Probably best to use the current language wiki. As long as you set the `uselang` parameter, it won't make a lot of difference, but there are some things in the login/signup forms (even in their minimal, no-skin view) that the wiki community can customize, e.g. the enwiki signup form has an extra link to the username policy.

app/src/main/java/org/wikipedia/auth/OAuthClient.kt
```diff
@@ -0,0 +1,175 @@
...
+                    val profile = ServiceFactory.getCoreRest(WikiSite(OAUTH_WIKI)).getOAuthProfile()
+                    finishLogin(profile)
+                    callback.onComplete(null)
+                }
+            }
+        }
+    }
+
+    // TODO: figure out how and when to orchestrate refreshing of tokens.
+    // Currently blocked on https://phabricator.wikimedia.org/T323855
```
tgr, line 138: That has been fixed. You should either keep track of token expiry and use a refresh token if it's expired, or listen for the response header signalling expiry (per [RFC 6750](https://www.rfc-editor.org/rfc/rfc6750.html#section-3.1)) and retry failed requests. The first approach is probably less hassle, and also not at risk of server-side bugs (with HAProxy, the API Gateway and MediaWiki each doing some form of bearer token parsing, there is a good chance something somewhere won't fail the correct way for an expired token).

PR #6837, opened 2026-08-28
Follow-up: Donation Reminder V3 settings screen

cooltey:
### What does this do?
1. Fix the bug: when an invalid amount is entered, we should save the last valid amount from the custom amount field. 
2. Remove old code and rearrange parameters. 
3. Rename string items.


voyagerfan (approved):
reviewed. Thanks for follow up!

PR #6844, opened 2026-08-31
DonationReminderV3: missing translation

cooltey:
### What does this do?
One missing string item from the Variant B screen 😓


cooltey:
self merge this missing string.

PR #6846, opened 2026-09-01
[SDK Update] min sdk to 24

Williamrai:
### What does this do?
- updates min sdk to 24
- removes codes required for sdk 23
This update is required to merge https://github.com/wikimedia/apps-android-wikipedia/pull/6842


app/build.gradle
```diff
@@ -37,7 +37,7 @@ android {
 
     defaultConfig {
         applicationId 'org.wikipedia'
-        minSdk = 23
+        minSdk = 24
```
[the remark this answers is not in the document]
Williamrai: This workflow is not running actively and will be removed in this PR https://github.com/wikimedia/apps-android-wikipedia/pull/6821

app/src/main/java/org/wikipedia/dataclient/okhttp/OkHttpConnectionFactory.kt
```diff
@@ -40,16 +35,6 @@ object OkHttpConnectionFactory {
                 .addInterceptor(TitleEncodeInterceptor())
                 .addInterceptor(HttpLoggingInterceptor().setLevel(Prefs.retrofitLogLevel))
 
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N) {
-            val certFactory = CertificateFactory.getInstance("X.509")
-            val certificates = HandshakeCertificates.Builder()
-                .addPlatformTrustedCertificates()
-                .addTrustedCertificate(certFactory.generateCertificate(WikipediaApp.instance.resources.openRawResource(R.raw.isrg_root_x1)) as X509Certificate)
```
dbrant, line 47: Excellent - We can now also remove the `raw/isrg_root...` files that we were bundling.

Williamrai: these are also used in `android:networkSecurityConfig="@xml/network_security_config"`. is it safe to remove them?

dbrant: ehhhh ok then, let's keep these a while longer until we're sure they're no longer necessary

dbrant:
Reminder for next release: keep the previous app bundle as "retained", so it remains available for older devices.

PR #6849, opened 2026-09-02
Donation reminder v3 maybe later punchlist

voyagerfan:
### What does this do?
This PR addresses UI punch list created from design review of the maybe later screen

- updates to m3 label large for pill headers
- updates type token for pills
- color matching for placeholder text and currency symbol in the custom amount field
- adds spacing below the `beta` tag in the loading state
- aligns icons/tooltip with the text & and makes sizing dynamic with font size changes
- prevents line breaks inside the pill, adds flowrow to allow pills to stack.

### Why is this needed?
There are several UI issues that need to be fixed

**Phabricator:**
https://phabricator.wikimedia.org/T435345

### DIFF
--- a/app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
+++ b/app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
@@ -7,6 +7,7 @@ import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.FlowRow
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.WindowInsets
@@ -693,8 +694,8 @@ fun <T : Number> OptionSelector(
             }
         }
         Spacer(modifier = Modifier.width(16.dp))
-        Row(
-            verticalAlignment = Alignment.CenterVertically,
+        FlowRow(
+            itemVerticalAlignment = Alignment.CenterVertically,
             horizontalArrangement = Arrangement.spacedBy(16.dp)
         ) {
             option.options.forEachIndexed { index, currentOption ->
@@ -712,6 +713,7 @@ fun <T : Number> OptionSelector(
                     ) {
                         Text(
                             text = currentOption.displayText,
+                            maxLines = 1,
                             color = if (isSelected) WikipediaTheme.colors.paperColor
                             else WikipediaTheme.colors.primaryColor,
                             style = MaterialTheme.typography.titleMedium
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
```diff
@@ -657,32 +665,39 @@ fun <T : Number> OptionSelector(
...
+                    iconSize = iconSize
                 )
             }
         }
         Spacer(modifier = Modifier.width(16.dp))
-        Row(
-            verticalAlignment = Alignment.CenterVertically,
+        FlowRow(
+            itemVerticalAlignment = Alignment.CenterVertically,
             horizontalArrangement = Arrangement.spacedBy(16.dp)
```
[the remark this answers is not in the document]
voyagerfan: No negative affects were noticed with this implementation.

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
```diff
@@ -657,32 +665,39 @@ fun <T : Number> OptionSelector(
...
         ) {
```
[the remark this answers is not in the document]
voyagerfan: Nice catch. Will add in the next commit

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
```diff
@@ -206,6 +210,9 @@ fun DonationReminderAppBar(
     menuItems: List<DonationReminderDropDownMenuItem> = emptyList()
 ) {
     var expanded by remember { mutableStateOf(false) }
+    val iconSize = with(LocalDensity.current) {
+        16.sp.toDp()
+    }
```
cooltey: I don't think you'll need this if you have `verticalAlignment = Alignment.CenterVertically` in line 291.

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
```diff
@@ -760,7 +777,8 @@ private fun DonationRemindersSwitch(
 @Composable
 fun InfoTooltip(
     modifier: Modifier = Modifier,
-    plainTooltipText: String
+    plainTooltipText: String,
+    iconSize: Dp? = null
```
cooltey: I don't think we need this.

cooltey (changes requested):

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
voyagerfan: ~~`Alignment.CenterVertically` handles the alignment, but I confirmed that the all icons should resize as the text size increases so I think we may actually need to keep this. Im open to a different resizing implementation if you'd like to go a different direction~~

Removing the icon resizing after double confirming

app/src/main/java/org/wikipedia/donate/donationreminder/DonationReminderScreen.kt
voyagerfan: ~~Confirmed that we need to resize icons as text increase. Im open to other ideas if we want to go another direction.~~

Removing the icon resizing after double confirming

PR #6857, opened 2026-09-03
New Articles AB Test: set up for second round.

dbrant:
This sets the stage for the second round of the New Articles A/B test, once it's ready to go (via remote configuration).

This also introduces an event (sent within the same `instrument` as the Feed) that reports the total latency when the For You section is fetched from the network, so that we can monitor changes in latency more directly. (already discussed and approved by Shay.)

https://phabricator.wikimedia.org/T435671


app/src/main/java/org/wikipedia/feed/interests/NewWithinInterestABTest.kt
```diff
@@ -3,7 +3,7 @@ package org.wikipedia.feed.interests
 import org.wikipedia.analytics.ABTest
 import org.wikipedia.settings.RemoteConfig
 
-class NewWithinInterestABTest : ABTest("new-within-interest", GROUP_SIZE_2) {
+class NewWithinInterestABTest : ABTest("new-within-interest-redux", GROUP_SIZE_2) {
```
dbrant, line 6: ^ Changing the "name" of the experiment will automatically cause existing users to be re-bucketed, so there won't be any confusion with the previous experiment.

app/src/main/java/org/wikipedia/feed/HomeFragment.kt
```diff
@@ -173,6 +189,7 @@ class HomeFragment : Fragment(), LinkPreviewDialog.LoadPageCallback {
         (requireActivity() as? MainActivity)?.onTabChanged(NavTab.HOME)
         viewModel.updateTabCount()
         viewModel.updateSelectedLanguageIfNeeded()
+        instrument.stopFunnel()
```
dbrant, line 192: ^ Let's stop and re-start the funnel here, because Shay was seeing `impression` events with missing funnel, which means that some impressions were being sent after `onPause`, so we still need the funnel to be active at that point.

PR #6859, opened 2026-09-03
DonationReminderV3: move bottom buttons to bottomBar in Scaffold

cooltey:
### What does this do?
Currently, we put the bottom buttons inside the `DonationReminderContent`, but sometimes this introduces an issue where there is more space at the top on different devices. 

This PR moves the bottom buttons to a separate function and puts them in the `bottomBar`.

**Before:**

Screenshot_20260903_160618
Screenshot_20260903_160541

**After:**

Screenshot_20260903_160658
Screenshot_20260903_160803
Screenshot_20260903_160447
Screenshot_20260903_160405
Screenshot_20260903_160246


voyagerfan (approved):
Awesome job with the state hoisting and bottom bar + api24 fix!

PR #6866, opened 2026-09-04
DonationReminderV3: add the missing string for "Update reminder" action

cooltey:
### What does this do?
For the V3 version, we will keep showing the button if coming from Settings.


voyagerfan (approved):
Looks good

PR #6876, opened 2026-09-09
Wire up instrumentation for Notifications

cooltey:
**Phabricator:**
https://phabricator.wikimedia.org/T430965


app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSettingsViewModel.kt
```diff
@@ -54,6 +55,21 @@ class RecommendedReadingListSettingsViewModel : ViewModel() {
...
+            RecommendedReadingListUpdateFrequency.MONTHLY -> "monthly"
+        }
+        TestKitchenAdapter.client.getInstrument("apps-notifications")
+            .submitInteraction(
+                action = "submit_click",
+                actionSource = "discover",
+                actionSubtype = "settings",
+                actionContext = mapOf(
+                    "article_count" to _uiState.value.articlesNumber,
+                    "frequency" to frequencyForEvent
```
voyagerfan, line 70: I think you can substitute `frequencyForEvent` with `_uiState.value.updateFrequency.name.lowercase()`. That way, you can remove the `when` block parsing.

cooltey: In that way, we will send localized strings to the server if the device language is not English.

app/src/main/java/org/wikipedia/readinglist/recommended/RecommendedReadingListSettingsViewModel.kt
```diff
@@ -54,6 +55,21 @@ class RecommendedReadingListSettingsViewModel : ViewModel() {
...
+                )
+            )
```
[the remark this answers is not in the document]
cooltey: It's expected to send the `submit_click` if it is disabled.

app/src/main/java/org/wikipedia/readinglist/ReadingListFragment.kt
```diff
@@ -1026,6 +1031,8 @@ class ReadingListFragment : Fragment(), MenuProvider, ReadingListItemActionsDial
                 Prefs.isRecommendedReadingListNotificationEnabled = true
                 requestPermissionAndScheduleRecommendedReadingNotification()
                 update()
+                TestKitchenAdapter.client.getInstrument("apps-notifications")
+                    .submitInteraction(action = "click", actionSource = "discover", actionSubtype = "discover_home_modal", elementId = "notification_bell_on")
```
voyagerfan: I believe the actionSubtype is supposed to be `discover_home` instead of `discover_home_modal`. (slide 10 of instrumentation deck)

cooltey: Yes! 😓 Fixed.

PR #6896, opened 2026-09-15
[Evergreen] shows plural string only when count  is greater than 1

Williamrai:
### What does this do?
- shows plural string only when coun is greater than 1


app/src/main/java/org/wikipedia/createaccount/CreateAccountEncourageScreen.kt
```diff
@@ -234,7 +234,7 @@ private fun encourageCards(uiState: CreateAccountEncourageViewModel.UiState): Li
 
 @Composable
 private fun countedTitle(count: Int, @PluralsRes titleRes: Int, @StringRes zeroTitleRes: Int): String {
-    return if (count > 0) pluralStringResource(titleRes, count, count) else stringResource(zeroTitleRes)
+    return if (count > 1) pluralStringResource(titleRes, count, count) else stringResource(zeroTitleRes)
```
dbrant: This does not feel correct. A `plural` resource handles the `1` case automatically, and this if/else statement was specifically intended to handle the `0` case differently from the nonzero case.
Is there an issue that this is meant to solve?

dbrant (changes requested):

app/src/main/java/org/wikipedia/createaccount/CreateAccountEncourageScreen.kt
Williamrai: Haley requested that we omit the number and use the generic title when the count is 1 same as 0. The german and french translations also uses incorrect plural wording for 1, so I reused the existing generic strings as a quick fix.

dbrant: In that case... could we just plug in the generic string into the `<one>` case of the plurals?

If we leave the code like this, it will be caught as a bug by the next human or agent.

dbrant: Alternatively, let's just leave a detailed comment explaining our reasoning, as you laid out above.

Williamrai: i think your solution is good. I pushed that solution.

PR #6897, opened 2026-09-15
Set Visual Editor visibility to always to `true`

cooltey:
### What does this do?
A temporary PR for testing purposes.


cooltey:
The remote config is `on`, closed.

PR #6904, opened 2026-09-18
Follow up: Entry point card for semantic search v2

cooltey:
### What does this do?
1. Update the title font weight to `Bold`.
2. Remove the switch items' listeners.


app/src/main/java/org/wikipedia/settings/dev/DeveloperSettingsPreferenceLoader.kt
```diff
@@ -323,38 +322,6 @@ internal class DeveloperSettingsPreferenceLoader(fragment: PreferenceFragmentCom
             }
         }
 
-        (findPreference(R.string.preference_developer_semantic_search_override_supported_language) as TwoStatePreference).apply {
```
voyagerfan, line 326: Since we're removing this, we'll need to swap the keys out in `developer_preferences.xml`. They currently point to strings in `dev_settings_strings.xml`

cooltey: Updated.

voyagerfan (approved):
Looks good!

